PR #212, opened 2019-01-28
Hackathon rotational autocorrelation

vyasr:
This pull request adds the ability to compute the correlation between a set of orientations in two states. The output produced is a scalar quantity measuring the total correlation. The primary use-case is as an order parameter comparing the initial state of a system to some future state.


cpp/order/RotationalAutocorrelation.cc
```diff
@@ -0,0 +1,118 @@
+// Copyright (c) 2010-2018 The Regents of the University of Michigan
```
bdice: Update copyrights in this PR to 2019

cpp/order/RotationalAutocorrelation.cc
```diff
@@ -0,0 +1,118 @@
...
+        {
+        if (l + k - a - b >= 0)
+            {
+            sum_tracker += cpow(std::conj(xi),k) * cpow(zeta, b-k) *
+                              cpow(std::conj(zeta), a-k) * cpow(-xi, l+k-a-b) /
+                              factorial(k) / factorial(l+k-a-b) /
+                              factorial(a-k) / factorial(b-k);
+            }
+        }
+    sum_tracker *= std::sqrt(factorial(a) * factorial(l-a) * factorial(b) * factorial(l-b) / (float(l)+1));
```
bdice, line 68: I'm wondering if there is a more numerically stable way to do this... Lots of (large?) factorial computations with `float` precision.

cpp/order/RotationalAutocorrelation.h
```diff
@@ -0,0 +1,127 @@
...
+ *  harmonics from the 2-sphere to the 3-sphere. For details, see Harmonic
+ *  functions and matrix elements for hyperspherical quantum field models
+ *  (https://doi.org/10.1063/1.526210).
+ */
+std::complex<float> hypersphere_harmonic(const std::complex<float> xi, std::complex<float> zeta,
+                                          const int l, const int m1, const int m2);
+
+//! Compute the total rotational autocorrelation for a set of orientations.
+/*! The desired autocorrelation function is the rotational analog of the
+ *  dynamic structure factor, which provides information on the dynamcs of
```
bdice: dynamcs ➡️  dynamics

cpp/order/RotationalAutocorrelation.h
```diff
@@ -0,0 +1,127 @@
...
+            return m_l;
+            }
+
+        //! Get the number of orientations used in the last call to compute.
+        unsigned int getN()
+            {
+              return m_N;
+            }
+
+        //! Get a reference to the last computed global angle array.
```
bdice: Is this comment correct? "global angle"?

cpp/order/RotationalAutocorrelation.h
```diff
@@ -0,0 +1,127 @@
...
+        //! Get a reference to the last computed value of the rotational autocorrelation.
+        float getRotationalAutocorrelation()
+            {
+            return m_Ft;
+            }
+
+        //! Compute the rotational autocorrelation.
+        /*! \param ref_ors Quaternions in initial frame.
+         *  \param ors Quaternions in current frame.
+         *  \param Np The number of orientations.
```
bdice: Change `Np` to `N` here and in the function, to match the implementation.

cpp/order/RotationalAutocorrelation.h
```diff
@@ -0,0 +1,127 @@
...
+class RotationalAutocorrelation
+    {
+    public:
+        //! Explicit default constructor for Cython.
+        RotationalAutocorrelation() {}
+
+        //! Constructor
+        /*! \param l The order of the spherical harmonic.
+         */
+        RotationalAutocorrelation(int l) : m_l(l), m_N(0), m_Ft(0) {}
```
bdice, line 73: Should `l` be an `unsigned int`?

freud/order.pyx
```diff
@@ -1347,3 +1347,102 @@ cdef class SolLiqNear(SolLiq):
...
+    colloidal plastic crystals of hard polyhedra – phase behavior and
+    directional entropic forces" by Karas et al. (currently in preparation).
+    The output is not a correlation function, but rather a scalar value that
+    measures total system orientational correlation with an initial state. As
+    such, the output can be treated as an order parameter measuring degrees of
+    rotational (de)correlation. For analysis of a trajectory, the compute call
+    needs to be done at each trajectory frame.
+
+    .. moduleauthor:: Andrew Karas
+    .. moduleauthor:: Vyas Ramasubramani
```
bdice: Add emails, like other modules in freud (I think this is consistent elsewhere?)

freud/order.pyx
```diff
@@ -1347,3 +1347,102 @@ cdef class SolLiqNear(SolLiq):
...
+            Order of the hyperspherical harmonic. Must be a positive, even
+            integer.
+
+    Attributes:
+        num_orientations (unsigned int):
+            The number of orientations used in computing the last set.
+        azimuthal (int):
+            The azimuthal quantum number, which defines the order of the
+            hyperspherical harmonic. Must be a positive, even integer.
+        ra_array (:class:`numpy.ndarray`):
```
bdice: I don't like the name but `rotational_autocorrelation` is so long... Also do we need to provide a shape for the array?

freud/order.pyx
```diff
@@ -1347,3 +1347,102 @@ cdef class SolLiqNear(SolLiq):
...
+
+    def __dealloc__(self):
+        del self.thisptr
+
+    def compute(self, ref_ors, ors):
+        """Calculates the rotational autocorrelation function for a single frame.
+
+        Args:
+            ref_ors ((:math:`N_{orientations}`, 4) :class:`numpy.ndarray`):
+                Reference orientations for time t=0.
```
bdice: I think "for the initial frame" is better than "for time t=0."

tests/test_order_RotationalAutocorrelation.py
```diff
@@ -0,0 +1,74 @@
...
+            orientations = data['orientations']
+
+            ra2 = freud.order.RotationalAutocorrelation(2)
+            l2 = []
+            for i in range(orientations.shape[0]):
+                ra2.compute(orientations[0, :, :], orientations[i, :, :])
+                l2.append(ra2.autocorrelation)
+            npt.assert_array_almost_equal(l2, data['l2auto'])
+
+            ra6 = freud.order.RotationalAutocorrelation(2)
```
bdice: Seems this test is using wrong info - use `l=6` and the corresponding `data['l6auto']` below.

bdice (changes requested):
Thanks for fixing this up, @vyasr! Please review the comments below and check the coverage to be sure your tests hit >80% of it.

bdice:
We may need to switch to `using namespace std;` to fix problems with Appveyor builds.

edit: It's the issue with `std::min` and `std::max`, see `cpp/locality/AABB.h` for the `#undef` needed to fix this.

cpp/order/RotationalAutocorrelation.cc
vyasr: That's a fair question. The factorial argument is bounded by l, which I assume will typically be in the single digits. I'm not sure there's a good way around this without trying to rework the underlying math, can you think of anything?

bdice: I don't think there is a clear way to mathematically simplify this expression. I considered proposing `lgamma` but it's not thread-safe and probably doesn't matter much if `l` is a small integer. We could consider switching to `double` precision for this summation, or leaving it as-is. Just wanted to think about it a bit.
https://en.cppreference.com/w/cpp/numeric/math/lgamma

cpp/order/RotationalAutocorrelation.h
vyasr: Technically yes, but I think that it's easier to leave it signed so that we don't run into issues when we do things like compare it with signed `m1` and `m2` values. We could change it if you have a strong preference though.

bdice: Ignore this - it's better as it is for the reason you mentioned. I missed the checks for "even and positive" in the Cython code during my first read-through.

bdice (approved):

PR #221, opened 2019-02-07
Bugfixes in LocalQl, LocalQlNear, LocalWl, LocalWlNear

vyasr:
1) Local*Near classes weren't using NearestNeighbor
2) Removed some poor typing of wigner arrays in LocalWl
3) Add args and kwargs to LocalQl to ensure that positional and keyword args are treated correctly


bdice (changes requested):
@vyasr All these changes look correct to me, after our discussion yesterday. It would be good to include tests that validate the correct behavior for the `kwargs` like `kn` and the `Near` variants' compute methods (ideally tests that would have failed before this PR).

bdice (approved):

PR #226, opened 2019-02-15
Apply clang-format

bdice:
## Description
This PR sets a clang-format standard for the C++ code in freud.

## Motivation and Context
Our C++ codebase is sometimes hard to read and inconsistently styled. This PR will create a uniform standard for C++ code going forward.

This style could be applied gradually as we work on individual modules or all-at-once (ideally after merging spatial data and before the hackathon, TBD on both).

This website is very helpful for visualizing the output of clang-format: https://zed0.co.uk/clang-format-configurator/

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,27 @@
+﻿---
+BasedOnStyle: WebKit
+AlignAfterOpenBracket: Align
+AlignTrailingComments: 'true'
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: 'false'
+AllowShortLoopsOnASingleLine: 'false'
+BreakBeforeBraces: Attach
+BreakConstructorInitializers: BeforeColon
+ColumnLimit: '110'
+CompactNamespaces: 'true'
+Cpp11BracedListStyle: 'true'
+FixNamespaceComments: 'true'
+IndentWidth: '4'
+Language: Cpp
+NamespaceIndentation: None
+PointerAlignment: Left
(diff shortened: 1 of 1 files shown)


bdice:
@vyasr Would you check out this branch and offer your thoughts on the `.clang-format` settings I chose? See the website linked above for an interactive test of these and other options.

vyasr:
This mostly looks fine to me. The two keys I'm not completely sure about are AllowShortFunctionsOnASingleLine and BreakBeforeBraces. I don't think we should ever allow single-line functions, but I'm not sure that Empty and None have the same result here. I always go back and forth on when I like breaks before braces; part of me really likes the alignment of opening and closing braces. It's visually easy to follow that, but I'm definitely inconsistent and I think I mostly prefer that for longer functions(/classes/namespaces/loops). I'm fine with just choosing something for now and making edits later if we want to.

Should we just add this as a recommend pre-commit hook? That way code will just get formatted as we continue to make changes.

bdice:
@vyasr I'm open to changes on those settings - can you make changes to those two options so that you like the result? I'm not sure what values to put in there to make it look like you want.

Yes, I'm going to recommend making it a pre-commit hook and add it to our CI style checks after we update the entire codebase.

vyasr:
The website you provided is currently down, I'm just waiting for it to come back up so that I can test and finalize, then I'm happy to merge.

vyasr (approved):
We're just going to go with this style for now, and make potential changes if we see style we really don't like.

PR #229, opened 2019-02-21
Fix/pypi deployment

vyasr:
This fixes some of the PyPI deployment issues we experienced with 1.0. The Linux issue is fixed by manually checking out the code instead of using CircleCI's checkout script, which forces cloning over ssh on this VM that's too old to handle it. The Mac deployment had a bug that was causing it to just retest the same Python version's installation of freud multiple times, which I've fixed, but the underlying cause of the deployment failure is still present. The problem there appears to be in the most recent release of SciPy. I've opened issue scipy/scipy#9858, we can see if they're able to solve that problem.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -192,6 +192,7 @@ references:
 
             # Install from and test all wheels
             for VERSION in ${PY_VERSIONS[@]}; do
+                pyenv global ${VERSION}
                 pip install freud_analysis --no-index -f ~/ci/freud/wheelhouse
                 cd ~/ci/freud/tests
                 python -m unittest discover . -v
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,17 @@ references:
   load_code: &load_code
     checkout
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -24,6 +24,17 @@ references:
...
     checkout
 
+  load_code_centos: &load_code_centos
+    # For use with old OS that doesn't support clone over ssh
+    run:
+      name: Checkout repository without ssh
+      command: |
+        cd ~/ci/freud
+        git clone ${CIRCLE_REPOSITORY_URL} .
+        export CIRCLE_TAG=v1.0.0  # Hard code for testing
```
bdice: @vyasr You need to remove this, right?

bdice:
@vyasr Can you fix whatever appears to be the diff in the `fsph` submodule? I think you might have committed that by accident but it's behind `master`.

bdice (changes requested):

vyasr:
@bdice fixed

bdice (approved):

PR #235, opened 2019-03-07
Enable ParticleBuffer to have different buffers in x, y, z.

bdice:
## Description
This PR adds the ability to pass a three-element list or array to the `buffer` argument of `ParticleBuffer.compute()`, as well as correcting a bug from numerical imprecision issues when using `images=True`.

## Motivation and Context
I commonly use the `images=True` flag when replicating periodic systems but there is not an easy way to replicate with a different number of images in each direction. This is helpful if the goal of replication is to achieve a desired number of particles, or if the box is very long in one dimension.

## How Has This Been Tested?
I added new tests for this additional behavior. Existing tests for the single value of buffer still pass.

I also found and corrected a bug in the `ParticleBuffer` which produced too many particles due to numerical imprecision in cases where particles were exactly on the boundary of a box with an unusual box length (e.g. it failed for `L=2*sqrt(0.5)`). A test has been added for this specific case, which initially failed and now passes with the corrected algorithm.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/box/ParticleBuffer.cc
+++ b/cpp/box/ParticleBuffer.cc
@@ -46,13 +46,17 @@ void ParticleBuffer::compute(const vec3<float> *points,
 
     if (use_images)
         {
-        images = vec3<int>(1 + ceil(buff.x), 1 + ceil(buff.y), 1 + ceil(buff.z));
-        m_buffer_box = Box(images.x * L.x, images.y * L.y, images.z * L.z, xy, xz, yz, is2D);
+        images = vec3<int>(ceil(buff.x), ceil(buff.y), ceil(buff.z));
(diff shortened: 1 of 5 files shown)


bdice:
Status: I'm having some difficulty where particles near the edge of the box are being replicated or left out when they shouldn't be. I'm not sure what's wrong yet. It's possibly floating point precision errors...

vyasr (approved):
I'm OK merging this for now since it helps your research, but I would prefer to split the images/non-images code paths on the C++ side at some point. It would require a bit of code duplication (the for loops and the image calculation in the loop), but it would make some of the code less confusing and more efficient. The logic behind the ternary operator isn't immediately obvious, and you have to check the images flag in the loop every time.

PR #241, opened 2019-03-16
Rewrite neighbor finding in freud

vyasr:
## Description
This pull request defines a new API for lazily finding and iterating over neighbors by separating the building of the acceleration structure from the querying of neighbors. This API supports efficiently querying the same data structure an arbitrary number of times for new neighbors. Additionally, neighbors can be retrieved as lists or using a lazy generator.

## Motivation and Context
This change makes it possible to reuse data structures that accelerate neighbor queries for a given set of data. Reusing these data structures will allow sequences of analyses on the same data to proceed much more efficiently. The ability to iteratively extract neighbors of a point without building a NeighborList will also provide performance gains since it avoids the overhead of looping twice to build the structure and then query it.

## How Has This Been Tested?
Extensive unit tests have been added in the `test_locality_NeighborQuery.py` module. These tests mirror the existing tests for the LinkCell class but are modified to use the new API for the LinkCell class (and for the AABBQuery class, which does not support the old `compute`-based API).

## Types of changes

## Checklist:


cpp/locality/LinkCell.cc
```diff
@@ -17,64 +18,56 @@ using namespace tbb;
...
-    if (cell_width != m_cell_width)
+    if (cell_width != m_cell_width || box != m_box)
         {
-        vec3<float> L = m_box.getNearestPlaneDistance();
-        vec3<unsigned int> celldim  = computeDimensions(m_box, cell_width);
+        vec3<unsigned int> celldim = computeDimensions(box, cell_width);
         // Check if box is too small!
-        bool too_wide =  cell_width > L.x/2.0 || cell_width > L.y/2.0;
-        if (!m_box.is2D())
+        // will only check if the box is not null
```
bdice: The order of these comments is confusing. Perhaps write `// Check if the box is not null` here, and `// Check if the box is too small` above the `if` statement below.

freud/_locality.pxd
```diff
@@ -8,6 +8,35 @@ from libcpp.memory cimport shared_ptr
 from libcpp.vector cimport vector
 cimport freud._box
 
+cdef extern from "NeighborQuery.h" namespace "freud::locality":
+    cdef cppclass NeighborPoint:
+        unsigned int id
+        unsigned int ref_id
+        float distance
+        bool operator==(NeighborPoint)
```
bdice, line 16: Should this interface also expose the distance-based comparator?

freud/locality.pyx
```diff
@@ -14,16 +14,273 @@ import warnings
...
+    .. moduleauthor:: Vyas Ramasubramani <vramasub@umich.edu>
+
+    .. versionadded:: 1.1.0
+    """
+
+    def __iter__(self):
+        cdef freud._locality.NeighborPoint npoint
+        cdef float[:, ::1] l_points = self.points
+
+        if self.query_type == 'nn':
```
bdice: Would it be possible to refactor this so that `self.iterator` is returned by a function or property that does all this? It would reduce duplication between `__iter__` and `toNList` as well as duplication with `AABBQueryResult` since that would be able to just define its corresponding function differently and use the parent class's methods for everything else.

freud/locality.pyx
```diff
@@ -14,16 +14,273 @@ import warnings
...
+    def __cinit__(self):
+        if type(self) is NeighborQuery:
+            raise RuntimeError(
+                "The NeighborQuery class is abstract, and should not be "
+                "directly instantiated"
+            )
+
+    @property
+    def box(self):
+        return self.box
```
bdice: How does this know to return the C++ box...? Is this duplicated name too confusing?

freud/locality.pyx
```diff
@@ -589,17 +826,22 @@ cdef class IteratorLinkCell:
...
     .. moduleauthor:: Joshua Anderson <joaander@umich.edu>
+    .. moduleauthor:: Vyas Ramasubramani <vramasub@umich.edu>
 
     Args:
         box (:class:`freud.box.Box`):
             Simulation box.
         cell_width (float):
             Maximum distance to find particles within.
+        points (:class:`np.ndarray`):
+            The points associated with this class, if used in the new mode
```
bdice: "in the new mode" -> "as a NeighborQuery object"?

bdice (changes requested):
I have a few minor comments/suggestions to resolve. Overall, this is great work, @vyasr. Thanks so much!

freud/locality.pyx
vyasr: Good call, that cleans this up quite a bit.

freud/locality.pyx
vyasr: Yeah, that's a good call. Note that we don't want to return the C++ box, we want to return a Python object, but I agree that the conflict between the property name and the member is bad. I'll modify this.

freud/locality.pyx
vyasr: Yup, I'll clarify.

bdice (approved):
👍 great work, you rock!

PR #242, opened 2019-03-17
Steinhardt class: A new single class for the Steinhardt order Parameter

b-butler:
## Description
The goal is to create a class Steinhardt that better emulates the functionality
of other classes within freud.order with the Steinhardt order parameter. This
functionality is currently encapsulated into the LocalQl and LocalWl classes.
This new class would in theory become the new standard API will leaving the
other classes for backwards-compatibility.

Some goals for this branch are ...
+ Provide a single compute method in the python interface
+ Use flags at construction to determine the particular order parameter given by
  compute.
+ Keep current and improve further - when applicable - performance of
  class with an eye towards potenial further parallelization.
+ Improve cpp code readability and better emulate the DRY mantra.

## Motivation and Context
Most other classes within the order module have a single compute method exposed
to the python interface. However, with the Steinhardt order parameters currently
8 exist between 2 classes. Furthermore, the C++ code use to determine the
Steinhardt order parameter requires code repetition. The current python
interface also exposes the user to this complexity. Until recently, the logic
for the compute methods were confusingly split between the Cython and C++
layers. Further improvements to code readability and maintainability are
warrented and this new API would allow for that.

## Pull Request Status
- [x] Create draft Steinhardt.cc. This should initially mostly be copying code
  from LocalQl and LocalWl to file.
- [x] Create working compute method.
  - [x] Impliment logic using member flags to calculate only what is necessary.
  - [x] Abstract array reallocation to function that only reallocates used
    arrays.
  - [x] Potentially move individual computational steps in compute method to
    private functions.
  - [ ] Add parallelization to loops over particle number.
- [x] Clean up code where applicable.
- [x] Add Steinhardt class to \_order.pxd.
- [x] Add Steinhardt class to order.pyx.
- [x] Create tests for Steinhardt class.
- [x] Update documentation

## Checklist:


cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,495 @@
+// Copyright (c) 2010-2019 The Regents of the University of Michigan
```
vyasr, line 1: Can you retab this file? It looks like in a lot of places you end up with 8 spaces instead of 4 for tabs.

vyasr: Also make sure you follow our (admittedly strange) code style with respect to brace indentation, see [here](https://freud.readthedocs.io/en/latest/sourceconventions.html#cpp)

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,495 @@
...
+    {
+    nlist->validate(m_Np, m_Np);
+
+    memset((void*) m_Qlmi.get(), 0, sizeof(complex<float>)*(2*m_l+1)*m_Np);
+    memset((void*) m_Qli.get(), 0, sizeof(float)*m_Np);
+
+    parallel_for(tbb::blocked_range<size_t>(0, m_Np),
+        [=] (const blocked_range<size_t>& r)
+        {
+	// May not be necessary check
```
vyasr: What check isn't necessary? Is there something we can remove here, or is this a copied comment from old code?

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,495 @@
...
+        const float rminsq = m_rmin * m_rmin;
+        const float rmaxsq = m_rmax * m_rmax;
+        const float normalizationfactor = 4*M_PI/(2*m_l+1);
+        const size_t *neighbor_list(nlist->getNeighbors());
+	// --------------------------
```
vyasr: Remove

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,495 @@
...
+
+                if (i == j)
+                    {
+                    continue;
+                    }
+
+                const vec3<float> delta = m_box.wrap(points[j] - ref);
+                const float rsq = dot(delta, delta);
+
+		// Replace conditional with function call above.
```
vyasr: Is this your comment?

cpp/order/Steinhardt.h
```diff
@@ -0,0 +1,230 @@
...
+            }
+
+        //! Get the number of particles used in the last compute
+        unsigned int getNP()
+            {
+            return m_Np;
+            }
+
+		//! Get the last calculated order parameter
+		std::shared_ptr<float> getQl()
```
vyasr: Fix indentation

cpp/order/Steinhardt.h
```diff
@@ -0,0 +1,230 @@
...
+                             const vec3<float> *points,
+                             unsigned int Np);
+
+        //! \internal
+        //! helper function to reduce the thread specific arrays into one array
+        void reduce();
+
+
+
+	private:
```
vyasr: Unindent `private`

vyasr: Also remove a couple of the extra blank lines above

cpp/order/Steinhardt.h
```diff
@@ -0,0 +1,230 @@
...
+        //  vector<complex<float> > with values for m = -l..l.
+        virtual void computeYlm(const float theta, const float phi,
+                                std::vector<std::complex<float> > &Ylm);
+
+		//! Reallocates only the necesary arrays when the number of particles changes
+		// unsigned int Np number of particles
+		void reallocateArrays(unsigned int Np);
+
+		//! Calculates the base Ql order parameter before further modifications
+		//if any.
```
vyasr: Add a space `// if any`

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
...
+    The :data:`average` argument in the constructor provides access to a variant
+    of this parameter that performs a average over the first and second shell
+    combined [Lechner2008]_. To compute this parameter, we perform a second
+    averaging over the first neighbor shell of the particle to implicitly
+    include information about the second neighbor shell. This averaging is
+    performed by replacing the value :math:`\overline{Q}_{lm}(i)` in the
+    original definition by the average value of :math:`\overline{Q}_{lm}(k)`
+    over all the :math:`k` neighbors of particle :math:`i` as well as itself.
+
+    The :data:`norm` constructor arguement provides
```
vyasr: type `argument`

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
...
+    normalized versions of the plain :math:`Ql` or :math:`Wl` or normalized
+    average if the :data:`average` flag is set to true.  where the normalization
+    is performed by dividing by the average :math:`Q_{lm}` values over all
+    particles.
+
+    .. moduleauthor:: Xiyu Du <xiyudu@umich.edu>
+    .. moduleauthor:: Vyas Ramasubramani <vramasub@umich.edu>
```
vyasr, line 533: Add yourself! :)

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
...
+            Simulation box.
+        rmax (float):
+            Cutoff radius for the local order parameter. Values near the first
+            minimum of the RDF are recommended.
+        l (unsigned int):
+            Spherical harmonic quantum number l. Must be a positive number.
+        rmin (float):
+            Can look at only the second shell or some arbitrary RDF region.
+        average (bool):
+            Determines whether to calcuate the averaged Steinhardt order
```
vyasr: typo `calculate`

tests/test_order_Steinhardt.py
```diff
@@ -0,0 +1,90 @@
...
+class TestSteinhardt(unittest.TestCase):
+    def test_shape(self):
+        N = 1000
+
+        box = freud.box.Box.cube(10)
+        np.random.seed(0)
+        positions = np.random.uniform(-box.Lx/2, box.Lx/2,
+                                      size=(N, 3)).astype(np.float32)
+
+        positions.flags['WRITEABLE'] = False
```
vyasr: For consistency with our other tests, I would remove this here and in other tests whether it's still present.

vyasr (changes requested):
There are some style issues, but the content all looks fine to me. There are a few places where it would be nice to combine a little more of the logic, but unfortunately different code paths differ just enough that it's not possible.

cpp/order/Steinhardt.h
```diff
@@ -0,0 +1,230 @@
...
+
+
+
+	private:
+        //! Spherical harmonics calculation for Ylm filling a
+        //  vector<complex<float> > with values for m = -l..l.
+        virtual void computeYlm(const float theta, const float phi,
+                                std::vector<std::complex<float> > &Ylm);
+
+		//! Reallocates only the necesary arrays when the number of particles changes
```
bdice: typo `necessary`

freud/_order.pxd
```diff
@@ -74,6 +74,20 @@ cdef extern from "TransOrderParameter.h" namespace "freud::order":
         unsigned int getNP()
         float getK()
 
+cdef extern from "Steinhardt.h" namespace "freud::order":
+    cdef cppclass Steinhardt:
+        Steinhardt(const freud._box.Box &, float, unsigned int, float,
+                bool, bool, bool) except +
```
bdice: You'll need to fix the code here to address the `flake8` code style issues that are preventing tests from running. See here for info about the `flake8` pre-commit hook, which can help you catch these early: https://freud.readthedocs.io/en/stable/sourceconventions.html#python

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
     def __str__(self):
         return repr(self)
 
+cdef class Steinhardt:
+    R"""Compute the local Steinhardt [Steinhardt1983]_rotationally invariant
+    :math:`Q_l` :math:`Wl` order parameter for a set of points.
+
+    Implements the local rotationally invariant :math:`Q_l` or :math:`Wl` order
```
bdice: `W_l` needs an underscore

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
...
+    combined [Lechner2008]_. To compute this parameter, we perform a second
+    averaging over the first neighbor shell of the particle to implicitly
+    include information about the second neighbor shell. This averaging is
+    performed by replacing the value :math:`\overline{Q}_{lm}(i)` in the
+    original definition by the average value of :math:`\overline{Q}_{lm}(k)`
+    over all the :math:`k` neighbors of particle :math:`i` as well as itself.
+
+    The :data:`norm` constructor arguement provides
+    normalized versions of the plain :math:`Ql` or :math:`Wl` or normalized
+    average if the :data:`average` flag is set to true.  where the normalization
```
bdice: Fix the incomplete sentence here. Maybe write "Normalization is performed by..."

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
...
+    Attributes:
+        box (:class:`freud.box.Box`):
+            Box used in the calculation.
+        num_particles (unsigned int):
+            Number of particles.
+        St (:math:`\left(N_{particles}\right)` :class:`numpy.ndarray`):
+            The last computed selected variant of the Steinhardt order
+            parameter for each particle (filled with NaN for particles with no
+            neighbors).
+    .. todo:: move box to compute, this is old API
```
bdice: We'll want to actually make this change - the `box` shouldn't be in the constructor anymore, only in the `compute` method.

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
...
+        """
+        self.box = box
+
+    @property
+    def num_particles(self):
+        cdef unsigned int np = self.stptr.getNP()
+        return np
+
+    @property
+    def St(self):
```
bdice: I discussed with @vyasr a while back - I think we'd like to use the property name `order` for all order parameters' data going forward. Since this is a new class, let's rename `St` to `order`.

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
...
+        if self.stptr.getUseWl():
```
bdice, line 604: I'd like to see this logic implemented on the C++ level.

freud/order.pyx
```diff
@@ -494,6 +494,173 @@ cdef class TransOrderParameter:
...
+            The last computed selected variant of the Steinhardt order
+            parameter for each particle (filled with NaN for particles with no
+            neighbors).
+    .. todo:: move box to compute, this is old API
+    """  # noqa: E501
+    cdef freud._order.Steinhardt * stptr
+    cdef freud.box.Box m_box
+    cdef rmax
+
+    def __cinit__(self, box, rmax, l, rmin=0, average=False, norm=False,
```
bdice: We'll probably refactor this to use the new neighbor-finding code after #272 has been merged. Then we will handle the neighbor-finding logic (e.g. `rmax`, `rmin`, `k` nearest neighbors) via a `QueryArgs` argument to the `compute` method. That way, we'll also encompass the behavior of the current `LocalQlNear` and `LocalWlNear` classes.

bdice:
@b-butler Thanks for your work on this! It will be a really nice feature improvement. We'll do a few iterations on this to get it right, since this feature is so heavily used. Ping @vyasr and I if you have questions or when you've made a round of changes and are ready for more review.

freud/order.pyx
vyasr: To clarify, you want there to be a single function on the C++ side that returns the correct order parameter depending on whether or not the flag indicates to use Wl or Ql? Keep in mind that the two currently return different types of output, so I'm not sure you can accomplish that.

I also don't think the Wl really needs to be complex from what I recall of the math defining it, but IIRC you do get meaningful imaginary components from the Wl classes. I haven't really used them in years, so I don't know what to make of that, and I would suggest holding off making any changes to the types, etc, for a separate PR where we do more explicit validation checking against e.g. pyboo.

freud/order.pyx
vyasr: Agreed. For now it's fine that this only supports the distance cutoffs.

cpp/order/Steinhardt.cc
b-butler: Will do

cpp/order/Steinhardt.cc
b-butler: Copied from old code

cpp/order/Steinhardt.cc
b-butler: No that I think was left over from the hackathon, but I must have missed it

freud/order.pyx
b-butler: I would have liked to impliment the logic in C++, but vyasr is right. The type difference between Wl and Ql prevents the logic from being on the C++ side.

b-butler:
I addressed all the comments and fixed some other stuff I noticed.

bdice:
@b-butler Good work so far! A few broad tasks for refactoring and simplification:

1. Many of the computations (e.g. normalization) are repeated through the code. For example, I would prefer to see one "normalize" call that accepts shared pointers as arguments and uses the flags of the class to determine how to specialize, instead of `computeNorm`, `computeAveNorm`, `computeNormWl`, `computeAveNormWl`.
2. If the `norm` flag is false, do we need to compute `m_Qlm` at all? From what I can tell, this data is only used if `norm` is true. It would be good to verify that and make that step conditional if possible.
3. There is a loop in `computeAve` that is labeled "Can be parallelized." It would be good to parallelize this.

freud/order.pyx
bdice: Just saw this - editing my last comment about next steps since this isn't possible. Sorry for the confusion.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,510 @@
...
+    // divide by the square root.
+    float normalizationfactor = sqrt(4*M_PI/(2*m_l+1));
+
+    // Get wigner3j coefficients from wigner3j.cc
+    m_wigner3jvalues = getWigner3j(m_l);
+
+    for (unsigned int i = 0; i < m_Np; i++)
+        {
+        // Revert Ql normalization
+        m_Qli.get()[i] /= normalizationfactor;
```
bdice: It doesn't appear like this result is being used - I looked at the original `LocalWl` to confirm. The original code had an option for `enableNormalization` that causes this to be useful, but it doesn't appear that it was ever possible to control this on the Python side. I will remove this in the current changes I am making to the code.

bdice:
I did some clean up and reduced the code repetition for the `Wl` methods. There is still more to be done with reducing repetition, but it may not be exactly what I was originally saying above about `norm` methods. Suggestions 2 and 3 are still valid. Here's the task list I have in mind:
- [ ] Only compute Qlm if needed
- [ ] Parallelize `computeAve` loop
- [ ] Refactor remaining blocks of repeated code into functions
- [ ] Remove unnecessary member variable arrays (e.g. it doesn't make sense to have multiple member variables for `m_WliAveNorm` ... `m_Wli`, etc. if only one of them can be used at a time).

I'm thinking about creating a class that wraps around our shared pointer arrays and plays nicely with Cython, taking the `makeArray` method I just wrote for this PR to the next level. We might combine some nice behaviors of shared pointers and `std::vector` to create a wrapper layer that handles things like using `memset` to zero out the data, easily instantiating an array with a default deleter, making it simpler to iterate over the data, etc. This kind of code is repeated lots of places in freud and it's a prime target for refactoring in my opinion just because it would help simplify our codebase a lot.  Thoughts, @vyasr @jinsoo960?

bdice:
It's strange to me that we return full arrays for the "normalized" order parameters - it's exactly the same value for every particle.

cpp/order/Steinhardt.cc
bdice: ^ done.

bdice:
I finished the refactoring tasks I wanted to do - @b-butler, please take a look and feel free to pick up with the remaining tasks when you have time.

b-butler:
Will do @bdice. Just saw all your commits and recommendations.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+/*! \file Steinhardt.cc
+    \brief Computes variants of Steinhardt order parameters.
+*/
+
+namespace freud { namespace order {
+
+// Calculating Ylm using fsph module
+void Steinhardt::computeYlm(const float theta,
+						    const float phi,
+							std::vector<std::complex<float> > &Ylm)
```
vyasr: Fix spacing here.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+    if (Ylm.size() != 2*m_l+1)
+        {
+        Ylm.resize(2*m_l+1);
+        }
+
+    fsph::PointSPHEvaluator<float> sph_eval(m_l);
+
+    unsigned int j(0);
+    // old definition in compute (theta: 0...pi, phi: 0...2pi)
+    // in fsph, the definition is flipped
```
vyasr: I'd delete this comment, it's referring to a pre-fsph implementation of Steinhardt that must be 3 years old at this point.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+    if (m_Wl)
+        {
+        m_WliOrder = Steinhardt::makeArray<complex<float> >(Np);
+        }
+    }
+
+void Steinhardt::compute(const box::Box& box,
+						 const locality::NeighborList *nlist,
+						 const vec3<float> *points,
+						 unsigned int Np)
```
vyasr: Fix spacing.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+    {
+    // Conditionally reinitialize arrays if size differs from previous call.
+    if (m_Np != Np)
+        {
+        m_Np = Np;
+        Steinhardt::reallocateArrays(Np);
```
vyasr: You shouldn't need this qualification, just `reallocateArrays` is enough.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+        }
+
+    // Computes the base Qlmi required for each specialized order parameter
+    Steinhardt::baseCompute(box, nlist, points);
+
+    if (m_average)
+        {
+        Steinhardt::computeAve(box, nlist, points);
```
vyasr: Same as above.

vyasr: I see this in a number of places, I'd just drop all unnecessary `Steinhardt::`.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+        }
+
+	// Reduce Qlm
+	Steinhardt::reduce();
+
+    if (m_Wl)
+        {
+        memset((void*) m_WliOrder.get(), 0, sizeof(complex<float>)*m_Np);
```
vyasr: Shouldn't this already have been zeroed by `reallocateArrays`?

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+        if (m_average)
+            {
+            Steinhardt::aggregateWl(m_WliOrder, m_QlmiAve);
+			m_NormWl = Steinhardt::normalizeWl();
```
vyasr: Fix spacing.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+            }
+        else
+            {
+            Steinhardt::aggregateWl(m_WliOrder, m_Qlmi);
+			m_NormWl = Steinhardt::normalizeWl();
```
vyasr: Fix spacing.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+        }
+        {
+        if (m_average)
+            {
+            m_Norm = Steinhardt::normalize();
+            }
+        else
+            {
+            m_Norm = Steinhardt::normalize();
+            }
```
vyasr: I'm confused as to why this conditional is here.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+        }
+    }
+
+void Steinhardt::baseCompute(const box::Box& box,
+							 const locality::NeighborList *nlist,
+							 const vec3<float> *points)
```
vyasr: Fix spacing.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+    {
+    nlist->validate(m_Np, m_Np);
+
+    memset((void*) m_Qlmi.get(), 0, sizeof(complex<float>)*(2*m_l+1)*m_Np);
+    memset((void*) m_Qli.get(), 0, sizeof(float)*m_Np);
```
vyasr: These should be set by `reallocateArrays`.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+		if (!m_average)
+			{
+			bool Qlm_exists;
+			m_Qlm_local.local(Qlm_exists);
+			if (!Qlm_exists)
+				{
+				m_Qlm_local.local() = new complex<float> [2*m_l+1];
+				memset((void*) m_Qlm_local.local(), 0, sizeof(complex<float>) * (2*m_l+1));
+				}
+			}
```
vyasr: Fix the indentation of this block, and I think you can also get rid of the memset since `reallocateArrays` does it.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+                for (unsigned int k = 0; k < (2*m_l+1); ++k)
+                    {
+                    const unsigned int index = (2*m_l+1) * i + k;
+                    m_Qlmi.get()[index] /= neighborcount;
+                    // Add the norm, which is the (complex) squared magnitude
+                    m_Qli.get()[i] += norm(m_Qlmi.get()[index]);
+					if (!m_average)
+						{
+						m_Qlm_local.local()[k] += m_Qlmi.get()[index] / float(m_Np);
+						}
```
vyasr: Fix indentation.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+                    }
+            m_Qli.get()[i] *= normalizationfactor;
+            m_Qli.get()[i] = sqrt(m_Qli.get()[i]);
+            } // Ends loop over particles i for Qlmi calcs
+        });
+    }
+
+void Steinhardt::computeAve(const box::Box& box,
+							const locality::NeighborList *nlist,
+							const vec3<float> *points)
```
vyasr: Indentation

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+    {
+    const size_t *neighbor_list(nlist->getNeighbors());
+
+    memset((void*) m_QlmiAve.get(), 0, sizeof(complex<float>)*(2*m_l+1)*m_Np);
+    memset((void*) m_QliAve.get(), 0, sizeof(float)*m_Np);
```
vyasr: Dealt with in reallocateArrays

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,391 @@
...
+                    source.get()[particle_index + u3];
+                counter++;
+                }
+            } // Ends loop over Wigner 3j coefficients
+        } // Ends loop over particles
+   }
+
+void Steinhardt::reduce()
+	{
+	memset((void*) m_Qlm.get(), 0, sizeof(complex<float>)*(2*m_l+1));
```
vyasr: Remove in favor of reallocateArrays

cpp/order/Steinhardt.h
```diff
@@ -0,0 +1,201 @@
...
+// This file is from the freud project, released under the BSD 3-Clause License.
+
+#ifndef STEINHARDT_H
+#define STEINHARDT_H
+
+#include <complex>
+#include <cstring>
+#include <memory>
+#include <stdexcept>
+#include <tbb/tbb.h>
```
vyasr, line 9: I think you can remove some of these (cstring, stdexcept).

vyasr (changes requested):
This is nice, much shorter than the older methods and will significantly clean up that whole chunk of code. I still have some change requests, but it's getting there!

The formatting is still all over the place. Since you first started working on this PR, we've started using clang-format as our C++ style requirement. Can you merge in master and then try to apply that clang-format to your code (let Bradley or me know if you need help doing that)? That should help fix a lot of the spacing issues.

cpp/order/Steinhardt.cc
```diff
@@ -30,151 +28,154 @@ void Steinhardt::computeYlm(const float theta,
...
-            }
+        {
+            m_Norm = normalize();
+        }
         else
-            {
-            m_Norm = Steinhardt::normalize();
-            }
+        {
+            m_Norm = normalize();
```
bdice: Seems like this `if` is redundant?

cpp/order/Steinhardt.cc
```diff
@@ -30,151 +28,154 @@ void Steinhardt::computeYlm(const float theta,
...
-                    {
+                {
                     continue;
-                    }
+                }
 
                 const vec3<float> delta = box.wrap(points[j] - ref);
                 const float rsq = dot(delta, delta);
 
                 if (rsq < rmaxsq && rsq > rminsq)
```
bdice: See if tests still pass if you remove this `if` statement. We have a problem with this check, see #297. We may need to change the preprocessing we do to the NeighborList to correct the behavior after removing this line.

cpp/order/Steinhardt.cc
```diff
@@ -184,208 +185,196 @@ void Steinhardt::baseCompute(const box::Box& box,
...
+            {
+                const unsigned int n(neighbor_list[2 * bond + 1]);
+
+                if (n == i)
+                {
+                    continue;
+                }
+
+                const vec3<float> rn = points[n];
+                // rin = rn - ri, from i pointing to j.
```
bdice: Fix comment, `from i pointing to neighbor n`.

cpp/order/Steinhardt.cc
```diff
@@ -184,208 +185,196 @@ void Steinhardt::baseCompute(const box::Box& box,
...
+                const vec3<float> rin = box.wrap(rn - ri);
+                const float rinsq = dot(rin, rin);
+
+                if (rinsq < rmaxsq && rinsq > rminsq)
```
bdice: See above comment about removing this check.

cpp/order/Steinhardt.cc
```diff
@@ -30,151 +28,154 @@ void Steinhardt::computeYlm(const float theta,
...
-			m_Qlm_local.local(Qlm_exists);
-			if (!Qlm_exists)
-				{
-				m_Qlm_local.local() = new complex<float> [2*m_l+1];
-				memset((void*) m_Qlm_local.local(), 0, sizeof(complex<float>) * (2*m_l+1));
-				}
-			}
+        const float normalizationfactor = 4 * M_PI / (2 * m_l + 1);
+        const size_t* neighbor_list(nlist->getNeighbors());
+
```
bdice, line 138: Add comment: `// Initialize thread-local m_Qlm and compute it in this function, if we won't average over neighbors later`

cpp/order/Steinhardt.cc
```diff
@@ -184,208 +185,196 @@ void Steinhardt::baseCompute(const box::Box& box,
...
-				// Add the norm, which is the complex squared magnitude
-				m_QliAve.get()[i] += norm(m_Qlmi.get()[index]);
-				}
-			m_QliAve.get()[i] *= normalizationfactor;
-			m_QliAve.get()[i] = sqrt(m_QliAve.get()[i]);
-			} // Ends loop over particles i for Qlmi calcs
-		}); //End parallel function
-    }
+    parallel_for(tbb::blocked_range<size_t>(0, m_Np), [=](const blocked_range<size_t>& r) {
+        bool QlmAve_exists;
```
bdice: Suggesting a change from `QlmAve_exists` to `Qlm_exists` as above, so it's clear these represent the same thing.

cpp/order/Steinhardt.cc
```diff
@@ -0,0 +1,380 @@
...
+
+    // Reduce Qlm
+    reduce();
+
+    if (m_Wl)
+    {
+        if (m_average)
+        {
+            aggregateWl(m_WliOrder, m_QlmiAve);
+            m_NormWl = normalizeWl();
```
bdice: Move this below the `if` statement (so it's only specified once).

cpp/order/Steinhardt.cc
bdice: Add a similar comment below in the average method, something like `// Initialize thread-local m_Qlm for storing neighbor-averaged QlmiAve, reduced over particles`

cpp/order/Steinhardt.cc
bdice: Specifically we need to check the behavior where `rmin > 0`, when there are neighbors within that inner shell distance.

bdice (changes requested):
This is looking great. A few comments left. I'm really happy with how this is turning out, it's much simpler. 🙌

b-butler:
The tests are currently failing in the github tests, but work in my local environment.

vyasr:
@b-butler the failing tests weren't really your fault. You needed to merge up to current master and make a couple minor changes to your calls to `convert_array`, but also there were issues with the `common` module being moved from pure Python to Cython that were causing problems. Should be fixed now. Remind me to review in the next day or so if I haven't yet.

b-butler:
@vyasr thanks for figuring it out!

vyasr (approved):
@bdice I'm happy with this PR and I think it's ready. If you're happy with this as well, I recommend merging #317 and then making the change here to the enumerable_thread_specific array we use here, then merging in this PR.

PR #215, opened 2019-01-30
[WIP] Parallelize Steinhardt Ql calculation.

bdice:
This PR parallelizes the computation of Steinhardt Ql order parameters and other variants present in `freud`.

I opened this PR as a work-in-progress so that I could track the performance.

## To-Do
- [ ] Refactor variants of `LocalQl.compute` to use a shared parallel compute method
  - [ ] `LocalQl.computeAve`
  - [ ] `LocalQl.computeNorm`
  - [ ] `LocalQl.computeAveNorm`
  - [ ] `LocalWl.compute`
  - [ ] `LocalWl.computeAve`
  - [ ] `LocalWl.computeNorm`
  - [ ] `LocalWl.computeAveNorm`
  - [ ] Review `SolLiq` internals, refactor appropriately
- [ ] Create regression tests

## Benchmarks
Input data: fcc structure, 20x20x20 unit cells (=32k particles) with noise=0.1
System: 16-core Intel Xeon Silver 4110 @ 2.10 GHz.
- Q6: 0.95 seconds before (commit 8bb083f), 0.15 seconds with parallelism (commit e901a97) (6.3x speedup over 16 cores)

PR #247, opened 2019-04-17
Parallelize Steinhardt - Part 1

bdice:
## Description
This PR parallelizes part of the Steinhardt order parameters in freud. PR #215 has additional changes but the changes contained in this PR should be ready to merge.

## Types of changes

## Checklist:


cpp/order/LocalQl.cc
```diff
@@ -50,69 +50,81 @@ void LocalQl::compute(const locality::NeighborList *nlist, const vec3<float> *po
...
+    parallel_for(tbb::blocked_range<size_t>(0, m_Np),
+        [=] (const blocked_range<size_t>& r)
         {
-        const vec3<float> ref(points[i]);
-        unsigned int neighborcount(0);
-
-        for (; bond < nlist->getNumBonds() && neighbor_list[2*bond] == i; ++bond)
+        const float rminsq = m_rmin * m_rmin;
+        const float rmaxsq = m_rmax * m_rmax;
+        const float normalizationfactor = 4*M_PI/(2*m_l+1);
```
vyasr: I'd leave these definitions outside the parallel_for, no need to redefine inside.

cpp/order/LocalQl.cc
```diff
@@ -50,69 +50,81 @@ void LocalQl::compute(const locality::NeighborList *nlist, const vec3<float> *po
...
+        m_Qlm_local.local(Qlm_exists);
+        if (!Qlm_exists)
             {
-            const unsigned int j(neighbor_list[2*bond + 1]);
+            m_Qlm_local.local() = new complex<float> [2*m_l+1];
+            memset((void*) m_Qlm_local.local(), 0, sizeof(complex<float>) * (2*m_l+1));
+            }
 
-            if (i == j)
+        size_t bond(nlist->find_first_index(r.begin()));
```
vyasr, line 70: Can we put this inside the for loop initializer instead?

cpp/order/LocalQl.cc
```diff
@@ -50,69 +50,81 @@ void LocalQl::compute(const locality::NeighborList *nlist, const vec3<float> *po
...
+        const size_t *neighbor_list(nlist->getNeighbors());
+
+        bool Qlm_exists;
+        m_Qlm_local.local(Qlm_exists);
+        if (!Qlm_exists)
             {
-            const unsigned int j(neighbor_list[2*bond + 1]);
+            m_Qlm_local.local() = new complex<float> [2*m_l+1];
+            memset((void*) m_Qlm_local.local(), 0, sizeof(complex<float>) * (2*m_l+1));
+            }
```
vyasr, line 68: The memory management here seems very dubious. Should we convert the `m_Ql*_local` objects into `tbb::enumerable_thread_specific<std::unique_ptr...>` instead of raw pointers? Otherwise I'm not sure how we don't just orphan a bunch of memory when the LocalQl object goes out of scope.

cpp/order/LocalQl.cc
```diff
@@ -50,69 +50,81 @@ void LocalQl::compute(const locality::NeighborList *nlist, const vec3<float> *po
...
+                    {
+                    // phi is usually in range 0..2Pi, but
+                    // it only appears in Ylm as exp(im\phi),
+                    // so range -Pi..Pi will give same results.
+                    float phi = atan2(delta.y, delta.x);     // -Pi..Pi
+                    float theta = acos(delta.z / sqrt(rsq)); // 0..Pi
+
+                    // If the points are directly on top of each other for whatever reason,
+                    // theta should be zero instead of nan.
+                    if (rsq == float(0))
```
vyasr: Can we get rid of the cast here? I would expect it to be done implicitly.

cpp/order/LocalQl.cc
vyasr: Have a look at [this code](https://github.com/glotzerlab/freud/blob/1756bb3b7eaa9cd3b2e4894b77d0c374148ad78c/cpp/density/RDF.cc#L64) showing how to free the relevant memory.

vyasr (changes requested):
There is some cleanup to do, as well as fixing a memory management problem. Let me know when you've had a chance to go through this again.

vyasr (approved):

PR #249, opened 2019-04-19
Windows min/max fix.

bdice:
## Description
Fixing the spatial data branch for compatibility with Windows. @vyasr We'll merge this if/when Appveyor tests pass.

## Motivation and Context
`std::min` and `std::max` conflict with a Windows header. This attempts to solve the problem globally.

## Types of changes


vyasr:
It looks like the Appveyor tests still fail on this branch, is that just because you didn't rebuild CPP? It's still failing on the std::min/std::max lines in AABB.cc.

vyasr (approved):

vyasr (approved):

PR #250, opened 2019-04-22
Optimize rotational autocorrelation

vyasr:
## Description
This PR cuts the runtime of the rotational autocorrelation function code by removing some work done inside loops that didn't need to be. Additionally, since the pow function is being called frequently I reimplemented a simple integer version that is much faster.

## Motivation and Context
The rotational autocorrelation function as defined can't be accelerated by the usual autocorrelation tricks using FFTs since the calculation for any pair of time points is not as simple as an inner product, so this code needs to be reasonably fast to support calculation on large trajectories.

## How Has This Been Tested?
No new tests were added, but existing tests pass.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:


bdice (approved):
@vyasr This looks great to me! Is it noticeably faster?

vyasr:
Yes, it's roughly 4x faster! It's a constant factor change though. I unfortunately wasn't able to do any clever math to reduce the actual complexity of the run because the sums are too ugly to be amenable to standard series sum simplifications.

PR #251, opened 2019-04-22
Add support for pyfftw in MSD

vyasr:
## Description
This PR adds support for using pyFFTW instead of SciPy or NumPy for fast Fourier transforms.

## Motivation and Context
NumPy's FFT is typically slower than SciPy's, so in the current master branch the code uses `scipy.fftpack.fft` if it's available, otherwise resorting to `numpy.fft.fft`. However, sometimes even the SciPy performance is insufficient, and using a better library like pyFFTW is preferable. Since FFTW is the _de facto_ standard for fast FFTs and the Python package is well supported, that is the natural choice to add support for. There are cases where SciPy is faster, but those are generally cases where performance is not a noticeable bottleneck. On the other hand, there are cases where SciPy (and NumPy) are essentially as slow as a normal DFT, whereas PyFFTW performs significantly better.

Note that I added some logging to catch the most obvious case where this performance difference will be noticeable. However, I'm not 100% sure we should be doing that, so if we don't want to I can remove that check and the additional logging that I introduced.

## How Has This Been Tested?
I just checked that this passes existing tests.

## Types of changes

## Checklist:


bdice (approved):
Looks good! (Reviewed offline)

vyasr:
This PR should not be merged until we have a more careful discussion regarding licensing.

bdice:
There is ongoing work by @peterbell10 to add configurable backend support to `scipy.fftpack`.

Proposal: https://mail.python.org/pipermail/scipy-dev/2019-May/023506.html

Design document: https://docs.google.com/document/d/1hGMc6ZrsmSi_LSixI5oFbo5cZkPUo92sfmnVthg1u6Y/edit#

vyasr:
@bdice I spoke to Karen and gave her the SciPy context, and she agrees that what we're doing should be safe. So feel free to give this PR one more review if you'd like, then I think we can merge it.

Issue #243, opened 2019-03-20
Binary pypi package fails due to numpy linkage on google colab


klarh:
**Describe the bug**
Google colab supports installing packages straight from pypi into a notebook's environment using pip. However, the numpy bundled with their container setup seems to be incompatible with the way freud's binary package was built for pypi. I haven't been paying too much close attention to how we're building binaries for pypi, but if it's not inconvenient, it may be useful to have them support whatever is special about this environment out of the box.

**To Reproduce**
1. Install freud in a colab notebook: `!pip install freud-analysis`
2. Import freud

**Error output**
`ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject`

**System configuration (please complete the following information):**
 - OS: linux
 - Version of Python: 3.6
- Version of freud: 1.0
- Version of numpy: 1.14.6

**Additional context**
An example notebook is located [here](https://colab.research.google.com/drive/1-2E6sVXAKxKnRsrUV_GoPbjrj0CfcKEr)

bdice:
@klarh We're definitely seeing this on our personal systems, but I believe the issue I've seen was with conda-forge builds. @k-l-wang can you comment with a description of your related issue?

I don't know how to solve this at the moment. @vyasr we should try to reproduce this and figure out how to build it better.

klarh:
FWIW, the binary versions seem to work on my local box with (non-conda) python 3.7.2 and both numpy 1.16.1 (from arch repository) and numpy 1.16.2 (binary version available on pypi).

vyasr:
I presume that is due to the wheels being built against NumPy 1.16 during the CI builds that upload to PyPI. The CircleCI script looks at the requirements.txt file to find requirements, but that file does not specify anything about versions, so it probably pulls the most recent NumPy on PyPI and builds against that. If we force it to build against NumPy 1.10 (the oldest NumPy that we claim to support) then this problem should go away.

klywang:
**Describe the bug**
(a) Updating freud < 1.0 to freud 1.0 in a conda environment corrupts the numpy installation. Downgrading freud to a version < 1.0 or reinstalling numpy using `pip install numpy -I` resolves the issue.

[...]

PR #252, opened 2019-04-23
Build PyPI wheels with older NumPy

vyasr:
## Description
Changes the deployment script to use the oldest version of NumPy supported by freud (1.10).

## Motivation and Context
Currently `pip install` from PyPI tries to use wheels, and when it finds appropriate wheels for a given Python version it installs them even if those wheels don't work for the existing NumPy versions due to ABI incompatibilities. This change builds the PyPI wheels with the oldest possible version of NumPy so that users can use PyPI wheels no matter which NumPy version they have installed beforehand.

## How Has This Been Tested?
I downloaded the relevant Docker images and ran these manually to ensure that the build succeeded and tests passed. However, I have not tried actually uploading to PyPI and then installing from there. I also have not tested the OS X wheel building process because we have no way to do that locally; we'll have to wait until this is merged and made live to deal with that problem I think.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,8 +31,15 @@ references:
       command: |
         cd ~/ci/freud
         git clone ${CIRCLE_REPOSITORY_URL} .
-        git fetch --force origin "refs/tags/${CIRCLE_TAG}"
-        git checkout -q "$CIRCLE_TAG"
+        if [ -n "$CIRCLE_TAG" ]
+        then
+          git reset --hard "$CIRCLE_SHA1"
+          git checkout -q "$CIRCLE_TAG"
+        elif [ -n "$CIRCLE_BRANCH" ]
+        then
+          git reset --hard "$CIRCLE_SHA1"
+          git checkout -q -B "$CIRCLE_BRANCH"
+        fi
 
   get_requirements: &get_requirements
     run:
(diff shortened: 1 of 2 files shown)


vyasr:
This fixes issue #243.

.circleci/config.yml
```diff
@@ -122,9 +129,22 @@ references:
...
-                "${PYBIN}/pip" install -r ~/ci/freud/requirements.txt --progress-bar=off
-                "${PYBIN}/pip" wheel ~/ci/freud/ -w ~/wheelhouse/ --no-deps
+              # Split the echo command and the version since python2
+              # --version doesn't return the value, just prints it.
+              echo "Building for "
+              ${PYBIN}/python --version
+
+              "${PYBIN}/python" -m pip install cython --no-deps --ignore-installed -q --progress-bar=off
+              rm -rf numpy-1.10.0rc2
+              curl -sSLO https://github.com/numpy/numpy/archive/v1.10.0rc2.tar.gz
```
bdice: I would prefer to use the last bugfix release of 1.10; 1.10.4 instead of 1.10.0(rc2). https://github.com/numpy/numpy/releases/tag/v1.10.4

You can see the bugfixes between 1.10.0 and 1.10.4 here: https://docs.scipy.org/doc/numpy-1.11.0/release.html

bdice (approved):
Besides the NumPy version question, this looks good. 👍

.circleci/config.yml
vyasr: Good call, for some reason I decided to initially go with the earliest release but the ABI shouldn't change so might as well use the newest bugfix release.

Issue #246, opened 2019-04-13
Segmentation Fault on Power 8 system (Summit)


csadorf:
**Brief description**

Executing unit tests and computing a radial distribution function causes a Segmentation Fault on [INCITE Summit](https://www.olcf.ornl.gov/for-users/system-user-guides/summit).

**To Reproduce**

1. Install freud into an environment on Incite Summit with the following module environment:

```
$ module list
Currently Loaded Modules:
  1) hsi/5.0.2.p5   3) lsf-tools/2.0           5) DefApps     7) spectrum-mpi/10.2.0.11-20190201   9) cmake/3.13.4  11) netlib-lapack/3.8.0           13) hdf5/1.10.3
  2) xalt/1.1.3     4) darshan-runtime/3.1.7   6) gcc/6.4.0   8) cuda/9.2.148                     10) git/2.13.0    12) python/3.6.6-anaconda3-5.3.0
```
and this command:

```
python setup.py install
```
2. Then execute:

```
cd tests
python -m unittest discover -v
```

**Error output**

```
(env-summit) [csadorf@login1.summit tests]$ python -m unittest discover . -v
test_2_dimensional (test_box_Box.TestBox) ... ok
test_box_volume (test_box_Box.TestBox) ... ok
test_construct (test_box_Box.TestBox)
Test correct behavior for various constructor signatures ... ok
test_coordinates (test_box_Box.TestBox) ... ERROR
test_cube (test_box_Box.TestBox) ... ok
test_equal (test_box_Box.TestBox) ... ok
test_fraction (test_box_Box.TestBox) ... ERROR
test_from_box (test_box_Box.TestBox)
Test various methods of initializing a box ... ok
test_get_length (test_box_Box.TestBox) ... ok
test_get_tilt_factor (test_box_Box.TestBox) ... ok
test_images (test_box_Box.TestBox) ... ERROR
test_matrix (test_box_Box.TestBox) ... ok
test_multiply (test_box_Box.TestBox) ... ok
test_periodic (test_box_Box.TestBox) ... ok
test_set_length (test_box_Box.TestBox) ... ok
test_square (test_box_Box.TestBox) ... ok
test_str (test_box_Box.TestBox) ... ok
test_to_dict (test_box_Box.TestBox)
Test converting box to dict ... ok
test_tuple (test_box_Box.TestBox) ... ok
test_unwrap (test_box_Box.TestBox) ... ok
test_vectors (test_box_Box.TestBox)
[...]

PR #254, opened 2019-04-23
Fix/test precision

bdice:
## Description
Many tests failed on Summit's POWER9 architecture because numerical checks in the tests were expecting higher precision. Tests have failed for similar reasons in the past on x86_64 architectures, but testing on a new architecture brought several more instances to light. This addresses part of the problems in #246.

## Types of changes

## Checklist:


bdice:
It appears that the tests for Nematic are still failing, and it's not clear why. @jglaser Can you reproduce this on a POWER9 architecture (e.g. Summit)? Build this branch and try to run `test_order_Nematic.py`.

```
======================================================================
FAIL: test_imperfect (__main__.TestNematicOrder)
Test imperfectly aligned systems.
----------------------------------------------------------------------
  File "tests/test_order_Nematic.py", line 88, in test_imperfect
    npt.assert_equal(op.director, np.array([0, 1, 0]))
  File "/home/ubuntu/.local/lib/python3.6/site-packages/numpy/testing/_private/utils.py", line 348, in assert_equal
    return assert_array_equal(actual, desired, err_msg, verbose)
  File "/home/ubuntu/.local/lib/python3.6/site-packages/numpy/testing/_private/utils.py", line 896, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "/home/ubuntu/.local/lib/python3.6/site-packages/numpy/testing/_private/utils.py", line 819, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Arrays are not equal

Mismatch: 66.7%
Max absolute difference: 0.18725421
Max relative difference: nan
 x: array([0.187254, 0.982311, 0.      ], dtype=float32)
 y: array([0, 1, 0])
```

vyasr (commented):
I'm still not quite sure why the numerical precision is different on this architecture. Could we maybe get some feedback from @joaander on why this might be the case before proceeding? It does seem to be the case that the problem is primarily one of math being done differently on different architectures, but I want to make sure that this doesn't indicate a problem in the way our computations are done internally. For multithreaded computation I would expect the nonassociativity of floating point arithmetic to matter, but if we run these tests single-threaded then it should at least be consistent; do we observe the same output when run multiple times?

joaander:
I recommend `numpy.testing.all_close` with `rtol=1e-6` for comparing non-zero single-precision values `atol` when comparing values to 0. I do *not* recommend using `assert_almost_equal`. Machine epsilon for single precision is 1.19e-07, so 1e-6 relative precision is the most one can reasonably expect regardless of architecture.

PR #256, opened 2019-04-24
Use constant memoryviews to prevent errors with read-only inputs.

### LINKED ISSUE #244: Buffer source array is read-only
**Describe the bug**
Tried to use `pythia` and an error occurred in freud because the numpy array was read-only. It should be possible to compute analyses on read-only particle arrays.

**To Reproduce**
I tried to reproduce this bug with @bdice but we couldn't figure out what caused it. We know that switching the two lines here from `cdef float[:, ::1]` to `cdef const float[:, ::1]` fixes the problem.

https://github.com/glotzerlab/freud/blob/9f89a2211e18d192fdedd4121e9e65c62b92ce64/freud/environment.pyx#L478-L480

We expected this script to reproduce the problem but it doesn't show the error as expected:
```python
import freud
import numpy as np

points = np.eye(3)
points.flags.writeable = False
orientations = np.array([[1, 0, 0, 0]] * 3)
orientations.flags.writeable = False
box = freud.box.Box.cube(L=6)
nlist = freud.locality.NearestNeighbors(1, 1).compute(box, points, points).nlist
ld = freud.environment.LocalDescriptors(1, 4, 2)
ld.compute(box, 1, points, points, orientations, nlist=nlist)
print(ld.sph)
```

**Error output**
```
 Encountered error during program execution: 'buffer source array is read-only'
Execute with '--show-traceback' or '--debug' to get more information.
  File "./project.py", line 371, in <module>
    PathwayProject(config=source_project.config).main()
  File "/nfs/turbo/bi-vislab/bdice/code/signac-flow/flow/project.py", line 2737, in main
    _exit_or_raise()
  File "/nfs/turbo/bi-vislab/bdice/code/signac-flow/flow/project.py", line 2705, in main
    args.func(args)
  File "/nfs/turbo/bi-vislab/bdice/code/signac-flow/flow/project.py", line 2425, in _main_run
    run()
  File "/nfs/turbo/bi-vislab/bdice/code/signac-flow/flow/legacy.py", line 193, in wrapper
    return func(self, jobs=jobs, names=names, *args, **kwargs)
  File "/nfs/turbo/bi-vislab/bdice/code/signac-flow/flow/project.py", line 1608, in run
    np=np, timeout=timeout, progress=progress)
  File "/nfs/turbo/bi-vislab/bdice/code/signac-flow/flow/project.py", line 1408, in run_operations
    self._fork(operation, timeout)
  File "/nfs/turbo/bi-vislab/bdice/code/signac-flow/flow/project.py", line 1478, in _fork
    self._operation_functions[operation.name](operation.job)
  File "./project.py", line 342, in analyze_pathways
    train_inputs, train_outputs = makeDataset(mode, job.fn(dump_fname), frame_slice, frames)
  File "./project.py", line 121, in makeDataset
    descriptors = list(globalDescriptors(structure_fname, mode, frame_slice, limit=800)) #100 vs. 200, 300
  File "./project.py", line 114, in globalDescriptors
    yield computeParticleDescriptors(frame.box, frame.positions, mode, limit=limit)
  File "./project.py", line 82, in computeParticleDescriptors
    descriptors = pythia.spherical_harmonics.abs_neighbor_average(box, positions, **kwargs)
  File "/nfs/turbo/bi-vislab/bdice/code/pythia/pythia/spherical_harmonics.py", line 122, in abs_neighbor_average
    nlist))
  File "/nfs/turbo/bi-vislab/bdice/code/pythia/pythia/spherical_harmonics.py", line 88, in neighbor_average
    comp.compute(freud_box, nNeigh, positions, positions, orientations, nlist=nlist)
  File "freud/environment.pyx", line 480, in freud.environment.LocalDescriptors.compute
  File "stringsource", line 654, in View.MemoryView.memoryview_cwrapper
  File "stringsource", line 349, in View.MemoryView.memoryview.__cinit__
ValueError: buffer source array is read-only
```

 - OS: Arch Linux 5.0.2
 - Version of Python: 3.6.7
 - Version of freud: 1.0

**Additional context**
The `const` feature was added to Cython to solve this exact problem: https://github.com/cython/cython/issues/1605

bdice:
## Description
I added `const` to memoryviews in all `.pyx` files.

## Motivation and Context
See #244 for a description of the problem. This resolves #244.

## How Has This Been Tested?
The existing tests use writable arrays, so this won't appear any different in our testing. Maybe we need to run some tests with non-writable arrays?

## Types of changes

## Checklist:


freud/box.pyx
```diff
@@ -779,7 +779,7 @@ cdef class ParticleBuffer:
             dereference(self.thisptr.getBufferParticles().get()).size()
         if not buffer_size:
             return np.array([[]], dtype=np.uint32)
-        cdef unsigned int[::1] buffer_ids = \
+        cdef const unsigned int[::1] buffer_ids = \
```
bdice, line 782: I'm not sure if it's correct to use `const` in cases like this. The tests still pass, but this is kind of an output array...?

vyasr:
I think we probably need to add two kinds of tests here:

1. Create a read-only array and show that it fails to work in `freud` before these changes, and that these changes fix it.
2. For one of the functions where you've changed the returned object (or any internal object that isn't a user parameter) into a `const` memoryview, run the function and then try and modify the output and see if the behavior is as expected. This request is specifically related to the comment you made on `box.pyx`; I'm not sure that we should be making `const` memoryviews anywhere except explicitly for user input parameters.

vyasr:
One other note, in the docs section on installing freud, we should add a minimum Cython version that we require if you want to build from source. Currently we omit that since Cython isn't a hard requirement, but we should include that in our documentation. This information is especially important now, because our code won't compile anymore without at least Cython 0.28.

Thanks for taking care of this!

bdice:
@vyasr What do you think the writable behavior of output arrays should be? In general, I don't think that users should be allowed to modify the contents of freud's output arrays (e.g. `rdf.R[5] = 100`). However, copies are writable (see below). I started adding tests and I'll re-request review when it's ready for your attention.

## Copies of read-only arrays are writable.

```python
>>> import numpy as np
>>> a = np.eye(3)
>>> a
array([[1., 0., 0.],
       [0., 1., 0.],
       [0., 0., 1.]])
>>> a.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
>>> a.flags['WRITEABLE'] = False
>>> a[0, 1] = 3
  File "<stdin>", line 1, in <module>
ValueError: assignment destination is read-only
>>> b = a.copy()
>>> b
array([[1., 0., 0.],
       [0., 1., 0.],
       [0., 0., 1.]])
>>> b.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
>>> b[0, 1] = 3
>>> b
array([[1., 3., 0.],
       [0., 1., 0.],
       [0., 0., 1.]])
>>>
```

vyasr:
So that I understand, currently you've added tests where you manually set the WRITABLE flag to False and then checked that everything still works. To support this, you've added the `const` keyword to all memoryviews onto input arrays. The open question is what to do with output arrays, right?

I agree that output arrays should not be writeable, but it's fine if copies are; this is consistent with the fact that write now output arrays are not even guaranteed to be valid if the original object is destroyed, and users are responsible for copying them if they want to keep working with them.

bdice:
@vyasr Currently setting the output memory views to `const` does nothing once they're returned as NumPy arrays.

For clarification, an input NumPy array with `flags['WRITEABLE'] == False` is only usable via a `const` memory view, but the opposite is not true: a `const` memory view does not create a NumPy array with `flags['WRITEABLE'] == False` when using `np.asarray(memoryview)`. I think it's fine to leave the `const` declarations on everything (including output arrays), since it provides a bit of no-write safety on memory views that shouldn't be writable by anything except for the C++ side.

We can revisit whether/how we want to set the `WRITEABLE` flag to `False` on all outputs, but that isn't necessary to resolve this bug.

vyasr:
Right, my question is whether we should be setting the writeable flag to false on the `np.asarray` outputs that we return. It seems like that's what we want, and what you suggested before, right?

> What do you think the writable behavior of output arrays should be? In general, I don't think that users should be allowed to modify the contents of freud's output arrays (e.g. `rdf.R[5] = 100`).

bdice:
Yeah, but let's save that (setting writeable to False) for another PR.

vyasr (approved):

PR #240, opened 2019-03-16
Issue210: Add __repr__ and __str__ methods for pretty representations of objects

gaofy95:
## Description

Add __repr__ and __str__ methods to box module, cluster module, density module and environment module.

## Motivation and Context

The objects created by freud lack "pretty representations". Adding __repr__ and __str__ methods can help users quickly debug code in the terminal or in a Jupyter notebook.

## How Has This Been Tested?

Create an object of each class, and call repr and str on each object to check the output.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #257, opened 2019-04-25
__repr__ and __str__ methods for pretty representations, Part 2

### LINKED ISSUE #210: Add __repr__ and __str__ methods for pretty representations of objects.
The objects created by freud lack "pretty representations," which can help users quickly debug code in the terminal or in a Jupyter notebook.

Currently, the behavior looks like this:
```python
>>> import freud
>>> box = freud.box.Box.cube(L=3)
>>> box  # this output is not helpful, I think this calls __repr__
<freud.box.Box object at 0x7f3bae62ced0>
>>> str(box)  # this output is more helpful, but doesn't follow the desired format below
'Box(Lx=3.0, Ly=3.0, Lz=3.0, xy=0.0, xz=0.0, yz=0.0, dimensions=3)'
```

See the behavior of `__repr__` and `__str__` in fresnel for an example of the desired behavior:

```python
class Camera(object):
    ...(stuff)...

    def __repr__(self):
        return "fresnel.camera.orthographic(position={0}, look_at={1}, up={2}, height={3})".format(self.position, self.look_at, self.up, self.height)

    def __str__(self):
        return "<Camera object with position {0}>".format(self.position);
```

https://github.com/glotzerlab/fresnel/blob/master/fresnel/camera.py#L96

bdice:
## Description
I added `__repr__` and `__str__` methods to most of the remaining objects in `freud`, excepting the `locality` module since I didn't know what kind of representation to use for objects without simple constructors.

## Motivation and Context
Resolves #210, continuation of #240.

## How Has This Been Tested?
I added tests that ensure that `str(obj) == str(eval(repr(obj)))`

## Types of changes

## Checklist:


freud/environment.pyx
```diff
@@ -544,13 +544,13 @@ cdef class LocalDescriptors:
         return self.thisptr.getLMax()
 
     def __repr__(self):
-        return ("freud.environment.{cls}(num_neigh={num_neigh}, " +
+        return ("freud.environment.{cls}(num_neighbors={num_neigh}, " +
                 "lmax={lmax}, rmax={rmax}, " +
```
vyasr: For consistency, let's remove the plus signs here.

freud/order.pyx
```diff
@@ -82,6 +82,11 @@ cdef class CubaticOrderParameter:
             orientation.
     """  # noqa: E501
     cdef freud._order.CubaticOrderParameter * thisptr
+    cdef t_initial
+    cdef t_final
+    cdef scale
+    cdef n_replicates
+    cdef seed
```
vyasr, line 79: Why do we need these? We already have properties for all of these that pull directly from the C++ class, is there a reason not to just use those?

vyasr: We would need to add properties for the `seed` and `n_replicates` as well; if you prefer we could eliminate the properties entirely and just make these attributes public. However, since we've in general avoided making `cdef`'ed attributes public, we'd probably have to introduce properties again to access the values, so it seems like a change that gains us nothing.

vyasr (changes requested):
There are a couple of unnecessary things introduced, otherwise LGTM.

Bigger picture, we should probably think a bit about how we want attributes to be stored and accessed in freud. Right now things are very scattered between C++ and Python, so there are C++ getter functions in almost all classes but they are used inconsistently because the properties in Python sometimes use these and sometimes use Python-level class attributes.

freud/order.pyx
bdice: I removed the attributes that were duplicated as properties, my bad. I kept `seed` and `n_replicates` for now.

vyasr: That's fine for now, we can do a more thorough revision at some point.

vyasr (approved):

PR #272, opened 2019-05-10
Feature/generic query

### LINKED ISSUE #261: Error in _getIterator not handled, gives segfaults
**Describe the bug**
When using an RDF with `rmax > L/2`, an error is thrown in the C++ code. However, this error is not being handled appropriately and gives a segfault for the script below, as of commit 1361df24309d8d1fe384fe22a02d31aecc797c82. It may be appropriate to add an `assertRaises` test for the exception.

**To Reproduce**
```python
import freud
rdf = freud.density.RDF(rmax=5, dr=0.1)
box = freud.box.Box(Lx=5, Ly=5, is2D=True)
points = [[0, 0, 0], [1, 1, 0], [1, -1, 0]]

# This should fail because rmax > L/2, but it doesn't catch
# the exception and segfaults instead
rdf.compute(box, points, points)
```

**Error output**
```python
RuntimeError: The AABBQuery rcut is too large for this box.
Exception ignored in: 'freud.locality.NeighborQueryResult._getIterator'
RuntimeError: The AABBQuery rcut is too large for this box.
/bin/bash: line 1:  8134
Segmentation fault      (core dumped) python '/home/bdice/freud-segfault.py'

shell returned 139
```

 - OS: Bash on Windows
 - Version of Python: 3.6.7
 - Version of freud: 1361df24309d8d1fe384fe22a02d31aecc797c82 (current master)

vyasr:
This branch adds the ability to query based on a set of arguments that determines what type of query to use.

## Description

The PR adds a `queryWithArgs` method in C++ that takes a `QueryArgs` object that contains parameters with which the type of query to perform may be determined. These features have been exposed to Python for testing, and some of them will probably have to stay on the Python side to support the more generic compute API going forward, but for now I have left them as private variables so that we have some flexibility.

This pull request has also turned into an extensive set of bug fixes and minor changes for the behavior of the dynamic querying functions. Here is a list of problems that were fixed:

1. The `NeighborQueryIterator` was returning results in the `point, ref_point` order, but for backwards compatibility and consistency with the `NeighborList` generation we need to swap that (until 2.0).
2. The `toNeighborList` function was finding one neighbor fewer than it should when performing a k-nearest neighbors query when self-neighbors were excluded because it was based on ball queries. I fixed this by temporarily increment the number of neighbors to find so that it would find enough.
3. In order to support the argument-based querying, I had to move the self-exclusion logic into C++, where it was originally in Python.
4. The counting logic for the k-nearest neighbors query in AABBQuery was incorrect and did not ensure that enough neighbors were found.
5. AABBQuery::query is based on growing ball queries by some scale, but if that scale goes beyond the max permitted bounds, it stops searching. If that scale factor is too large, it can easily miss neighbors that are within the permitted region, so I had to manually make sure it searched that radius.
6. Even with the previous fix, AABBQuery::query could not guarantee finding k neighbors if some of them were outside the maximum safe distance for AABB ball queries. To ensure that enough neighbors are found, I made it possible to perform AABBQuery::queryBall with larger radii, with the understanding that the caller would have to filter out duplicates after the fact. This function is only an internal API, and is now used by AABBQuery::query.
7. Fixed behavior for 2 dimensions.
8. Ensured that errors thrown in C++ when AABB rcuts are too large are correctly propagated to Python (closes #261).

I also added tests corresponding to all of these bugs.

## Motivation and Context

This generic querying feature will enable the generic compute class that we are planning, which will encapsulate the key logic for neighbor-based calculations.

## How Has This Been Tested?

I added extra tests for the specific new query function, as well as additional tests to catch the `NeighborList` bugs that I fixed.

## Types of changes

## Checklist:


cpp/locality/AABBQuery.h
```diff
@@ -40,23 +41,75 @@ class AABBQuery : public NeighborQuery
...
+                if (args.nn == -1)
+                    throw std::runtime_error("You must set nn in the query arguments.");
+                if (args.scale == -1)
+                    {
+                    args.scale = 1.1;
+                    }
+                if (args.rmax == -1)
+                    {
+                    vec3<float> L = this->getBox().getL();
+                    args.rmax = 0.1*std::min(std::min(L.x, L.y), L.z);
```
bdice: What if it's 2D?

cpp/locality/NeighborQuery.h
```diff
@@ -37,16 +37,42 @@ struct NeighborPoint {
...
+     *  values for parameters.
+     */
+    QueryArgs() : nn(-1), rmax(-1), scale(-1), exclude_ii(false) {}
+
+    enum QueryType {
+        ball,         //! Query based on distance cutoff.
+        nearest       //! Query based on number of requested neighbors.
+    };
+
+    QueryType mode;     //! The number of nearest neighbors to find.
```
bdice: Fix comment.

cpp/locality/NeighborQuery.h
```diff
@@ -37,16 +37,42 @@ struct NeighborPoint {
...
+    unsigned int nn;    //! The number of nearest neighbors to find.
```
bdice: Please don't make it unsigned and initialize it as `-1`. Use an `int` for sanity.

freud/locality.pyx
```diff
@@ -26,6 +25,103 @@ cimport numpy as np
...
+        * mode: The query mode, either 'ball' or 'nearest'.
+        * rmax: The cutoff distance for a ball query.
+        * nn: The number of nearest neighbors to find.
+        * exclude_ii: Whether or not to include self-neighbors.
+
+    Args:
+        query_args (dict):
+            A dictionary of query arguments.
+
+    Returns
```
bdice: Missing colon?

freud/locality.pyx
```diff
@@ -26,6 +25,103 @@ cimport numpy as np
...
+    cdef dict invalid_args = {}
+    for key, val in query_args.items():
+        if hasattr(qa, key):
+            setattr(qa, key, val)
+        else:
+            invalid_args[key] = val
+    if invalid_args:
+        raise ValueError(
+            "The following invalid query arguments were provided: "
+            ", ".format("{} = {}".format(key, val)))
```
bdice: Probably meant `join`.

cpp/locality/NeighborQuery.h
bdice: This code also throws a warning.

cpp/locality/NeighborQuery.h
```diff
@@ -81,13 +107,39 @@ class NeighborQuery
...
+                }
+            else if (args.mode == QueryArgs::nearest)
+                {
+                return this->query(points, N, args.nn, args.exclude_ii);
+                }
+            else
+                {
+                assert("Invalid query mode provided to generic query function.");
+                }
+            }
```
bdice, line 134: Throws a warning:
```
cpp/locality/NeighborQuery.h:134:13: warning: control may reach end of non-void function [-Wreturn-type]
            }
            ^
```

cpp/locality/NeighborQuery.h
```diff
@@ -118,6 +170,20 @@ class NeighborQuery
...
+        virtual void validateQueryArgs(QueryArgs& args)
+            {
+            if (args.mode == QueryArgs::ball)
+                {
+                if (args.rmax == -1)
+                    throw std::runtime_error("You must set rmax in the query arguments.");
+                }
+            else if (args.mode == QueryArgs::nearest)
+                {
+                if (args.nn == -1)
```
bdice, line 182: Throws a warning:
```
cpp/locality/NeighborQuery.h:182:29: warning: comparison of integers of different signs: 'unsigned int' and 'int' [-Wsign-compare]
                if (args.nn == -1)
                    ~~~~~~~ ^  ~~
```

cpp/locality/AABBQuery.h
```diff
@@ -40,23 +41,75 @@ class AABBQuery : public NeighborQuery
...
+                }
+            else if (args.mode == QueryArgs::nearest)
+                {
+                return query(points, N, args.nn, args.rmax, args.scale, args.exclude_ii);
+                }
+            else
+                {
+                assert("Invalid query mode provided to generic query function.");
+                }
+            }
```
bdice, line 70: Similar problem to previous warning:
```
cpp/locality/AABBQuery.h:69:13: warning: control may reach end of non-void function [-Wreturn-type]
            }
            ^
```

cpp/locality/AABBQuery.h
```diff
@@ -40,23 +41,75 @@ class AABBQuery : public NeighborQuery
...
+        virtual void validateQueryArgs(QueryArgs& args)
+            {
+            if (args.mode == QueryArgs::ball)
+                {
+                if (args.rmax == -1)
+                    throw std::runtime_error("You must set rmax in the query arguments.");
+                }
+            else if (args.mode == QueryArgs::nearest)
+                {
+                if (args.nn == -1)
```
bdice, line 110: Similar to previous warning:
```
cpp/locality/AABBQuery.h:99:29: warning: comparison of integers of different signs: 'unsigned int' and 'int' [-Wsign-compare]
                if (args.nn == -1)
                    ~~~~~~~ ^  ~~
```

cpp/locality/AABBQuery.cc
```diff
@@ -208,44 +223,94 @@ NeighborPoint AABBQueryIterator::next()
...
+                    break;
+                    }
+                else if ((m_r_cur >= max_plane_distance) || (m_all_distances.size() >= m_k))
                     {
-                    // Note that we use reverse iterators to sort in descending
-                    // order so that we can use pop_back to remove the item from
-                    // the vector before returning it.
-                    std::sort(m_current_neighbors.rbegin(), m_current_neighbors.rend());
+                    // Once this condition is reached, either we found enough
+                    // neighbors beyond the normal min_plance_distance
```
bdice: `min_plane_distance`

cpp/locality/NeighborQuery.h
```diff
@@ -245,8 +314,52 @@ class NeighborQueryIterator {
...
+         * All it needs to do is increase the counter of points to find, find
+         * them, and then reset.
+         */
+        virtual NeighborList *toNeighborList()
+            {
+            if (m_exclude_ii)
+                m_k += 1;
+            NeighborList *nlist = NeighborQueryIterator::toNeighborList();
+            if (m_exclude_ii)
+                m_k -= 1;
```
bdice: I can't remember how this works in C++ but a `try...finally` sort of construct would be nice, so that `m_k` ends up being decremented again even if the `toNeighborList` fails.

freud/locality.pyx
```diff
@@ -181,6 +280,37 @@ cdef class NeighborQuery:
     def points(self):
         return np.asarray(self.points)
 
+    def _queryGeneric(self, points, query_args):
+        # This function is temporarily included for testing and WILL be
+        # removed in future releases.
```
bdice, line 285: Can you help me understand what your plan is for this? Should it _ever_ be released if you're planning to remove it soon?

freud/locality.pyx
```diff
@@ -745,15 +898,12 @@ cdef class AABBQuery(NeighborQuery):
...
 
-        # Ensure that enough neighbors are found when excluding
-        if exclude_ii:
-            k += 1
-
         # Default guess value
         if r == 0:
-            r = 0.1*min(self._box.L)
+            r = 0.1*min(
+                [length for length in self._box.L if length != 0])
```
bdice: If this is meant to address 2D behavior, I'd like to check the `is2D` property and make that logic explicit, rather than rely on `Lz=0` as this seems to do.

bdice (changes requested):
Awesome work, @vyasr. I have some smaller things to address, but overall this looks great.

cpp/locality/AABBQuery.h
vyasr: Good catch, I knew I would miss that somewhere.

cpp/locality/NeighborQuery.h
vyasr: What would you suggest? I left this warning because it's safe, but we can get rid of it by rewriting the code. One option would be to just assume that the function gets a valid argument. Another would be to rewrite the `else if` as just an `else`, and put in a nested `if` statement that checks that the value is the expected one and fails if that's the case. I don't want to throw an exception here since this function isn't in the public API, and any failure would be due to a developer error.

cpp/locality/NeighborQuery.h
vyasr: Good idea.

freud/locality.pyx
vyasr: We can't test this functionality without exposing it to Python. My original thought was that we would remove it once we refactor the rest of freud to use NeighborQuery objects, since those would implicitly test the function. On the other hand, we could just leave this as a private function that exists only for unit tests. Now that I've written so many tests, I would be open to the latter. Thoughts?

cpp/locality/NeighborQuery.h
vyasr: I'm a little confused at why this is an unsigned int here, I changed it locally already. I assume it will fix itself when I push all the other changes though.

cpp/locality/NeighborQuery.h
bdice: @vyasr Maybe this warning would go away if you throw a runtime error in the `else` block?

bdice: @vyasr I would still throw an exception. "Developer error" is a strange reason to reject this in my opinion - it's still an error that the user should know about so it can be debugged.

Test it and see if the warning goes away. If not, then we should find an alternate approach.

freud/locality.pyx
bdice: I don't see a reason why we should remove it. It might even be useful in the public API someday. I would take off this comment about it being removed in the future.

bdice (changes requested):
Requesting a couple minor changes.

freud/locality.pyx
vyasr: After discussing offline, we clarified that enabling a public API with this method requires adding substantial logic to the Cython side that is currently all in C++. As a result, we will not plan to implement this now, but may do so in the future.

cpp/locality/NeighborQuery.h
vyasr: After discussing offline, we came to the consensus that while this is an appropriate usage of assert, we want to move towards making it easier to eventually use freud as a pure C++ library. In that light, it makes sense to add more explicit error checking to C++ functions, so we can convert this to an exception.

bdice (approved):
Remaining comments were addressed offline. Thanks @vyasr!

PR #279, opened 2019-05-16
Local/benchmark test

jinsooihm:
## Description

Added benchmark feature to `circle.ci`. 
When pushed to git, print and save benchmark data in` circle.ci`.
Exit with `1` if the new commit is slower than the origin/master by 0.7 times.

## Further work
The input sizes to run benchmarks on should be calibrated in a way that `circle.ci` should have enough resource to run. 
Some of the functions can be improved in organization.
Currently, benchmark takes very long on `circle.ci`.
More documentation can be added.
Only some of `freud` module benchmarks (`locality`, `density`) have been added yet.

## Types of changes

## Checklist:


.circleci/config.yml
```diff
@@ -84,6 +84,31 @@ references:
       path: test-reports
       destination: test-reports
 
+  benchmark: &benchmark
+    run:
+      name: Run benchmarks
+      command: |
+          pip install --user gitpython
+          echo 'export PYTHONPATH=$PYTHONPATH:.' >> $BASH_ENV
+          source $BASH_ENV
```
bdice, line 95: Is there a reason why you don't just call `export PYTHONPATH=$PYTHONPATH:.` directly?

.circleci/config.yml
```diff
@@ -84,6 +84,31 @@ references:
...
+          BENCHSCR="benchmarks/benchmarker.py"
+          echo "Running benchmark on current HEAD"
+          python "$BENCHSCR" run -c
+          echo "Comparing with orign/master"
```
bdice: `origin`

.circleci/config.yml
```diff
@@ -84,6 +84,31 @@ references:
...
+          python${PYVER} setup.py build_ext --inplace --COVERAGE --ENABLE-CYTHON
+          git checkout "${CIRCLE_SHA1}" -- benchmarks/
+          echo "Running benchmark on origin/master"
+          python "$BENCHSCR" run -c
+          python "$BENCHSCR" compare origin/master "${CIRCLE_SHA1}"
+
+  store_benchmarks: &store_benchmarks
+    store_artifacts:
+      path: benchmarks/reports
+      destination: benchmarks-reports
```
bdice, line 121: The current output on CircleCI for "Uploading artifacts" says:
```
Uploading /home/ci/ci/freud/benchmarks/reports to benchmarks-reports
No artifact files found at /home/ci/ci/freud/benchmarks/reports
```
I think `path` has to point to a specific file - please refer to the CircleCI docs and figure out how to upload all the result files generated by the benchmarks.

benchmarks/benchmark.py
```diff
@@ -25,43 +26,44 @@
...
+    # Override this method to run the benchmark.
+    def bench_run(self, N):
         pass
 
+    # Returns timer instance with overridden setup and run functions.
+    def setup_timer(self, N):
+        test_wrapper = self
+        setup = "test_wrapper.bench_setup(N)"
+        stmt = "test_wrapper.bench_run(N)"
+        varmapping = {"test_wrapper": test_wrapper, "N": N}
```
bdice: Just put `self` in for `test_wrapper` and remove line 43.

benchmarks/benchmark.py
```diff
@@ -25,43 +26,44 @@
...
     # \param number Number of times to call run()
     # \param print_stats Print stats to stdout
     # \returns The average time for each call to run()
     #
-    def run_benchmark(self, N=None, number=100, print_stats=False):
-        # setup
-        self.setup(N)
-        self.run(N)
+    def run_benchmark(self, N=None, number=100, print_stats=False, repeat=1):
+        # initilize timer
```
bdice: initialize

benchmarks/benchmarker.py
```diff
@@ -0,0 +1,329 @@
...
+        description="Execute performance tests in various categories for "
+                    "specific data space sizes (N).")
+    parser_run.add_argument(
+        '-o', '--output', nargs='?', default='benchmark.json',
+        help="Specify which collection file to store results "
+             "to or '-' for None, "
+             "default='benchmark.json'.")
+    parser_run.add_argument(
+        '-c', '--circleci', action='store_true',
+        help="Flag for running on circle.ci to fix thread number")
```
bdice: I'd like to remove the thread count from the signature of every `run` method, too -- we could use the environment variable `BENCHMARK_NPROC` just like we're already using `BENCHMARK_NPROC_INCREMENT`, to make the implementation of all this cleaner.

Alternatively we could refactor this to take an integer and limit to that number of threads, instead of making it specific to CircleCI. The CircleCI config can specify 2 threads via the command line with something like `-j 2` or `--threads 2` instead of `-c`.

Ask @vyasr for his opinion.

benchmarks/benchmark.py
```diff
@@ -138,7 +139,8 @@ def run_size_scaling_benchmark(self, N_list, number=1000,
     #       problem size (down to a minimum of 1).
     #
     def run_thread_scaling_benchmark(self, N_list, number=1000,
-                                     print_stats=True):
+                                     print_stats=True, repeat=1,
+                                     on_circleci=False):
```
bdice: See my other comment below - I'd like to use something like `limit_threads=P` or an environment variable `BENCHMARK_NPROC` instead of `on_circleci=False`.

benchmarks/benchmarker.py
```diff
@@ -0,0 +1,329 @@
...
+# param name: Name of benchmark run to print
+# param Ns: List containing N values to run benchmark on
+# param number: Number of times to run time
+# param classobj: Benchmark class object to benchmark on
+# param print_stats: Print stats if true
+# param on_circleci: Limit thread number if ran on circle.ci
+# param kwargs: Initializer variables for classobj
+# return: Dictionary containing benchmark information
+def do_some_benchmarks(name, Ns, number, classobj, print_stats,
+                       on_circleci, **kwargs):
```
bdice: In `freud`, we use Google-style docstrings (put the function documentation below the `def` and use the format below).
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

bdice: I'd like to rename this function to something like `run_benchmarks`.

benchmarks/benchmarker.py
```diff
@@ -0,0 +1,329 @@
...
+import git
+import json
+import os
+import argparse
+import sys
+import unittest
+import importlib
+
+
+# To add a new benchmark,
```
bdice: This description of adding benchmarks should end up in the freud documentation, right below the Unit Tests in the developer guide. https://freud.readthedocs.io/en/stable/design.html#unit-tests

benchmarks/benchmarker.py
```diff
@@ -0,0 +1,329 @@
...
+def do_some_benchmarks(name, Ns, number, classobj, print_stats,
+                       on_circleci, **kwargs):
+    if print_stats:
+        print(benchmark_desc(name, kwargs))
+
+    # initialize classobj instance
+    try:
+        b = classobj(**kwargs)
+    except TypeError:
+        print("Wrong set of initizliation keyword \
```
bdice: initialization

bdice (changes requested):
Nice work, @jinsoo960. I have requested some changes in the PR.

benchmarks/benchmarker.py
```diff
@@ -0,0 +1,329 @@
...
+    parser_compare.add_argument(
+        'rev_other', default='master', nargs='?',
+        help="The git revision to compare against. "
+             "Valid arguments are  for example "
+             "a branch name, a tag, a specific commit id, "
+             "or 'HEAD', defaults to 'master'.")
+    parser_compare.add_argument(
+        'rev_this', default='HEAD', nargs='?',
+        help="The git revision that is benchmarked. "
+             "Valid arguments are  for example "
```
bdice, line 408: extra space

benchmarks/benchmarker.py
```diff
@@ -0,0 +1,329 @@
...
+                        print("Threads: {}, N: {}, "
+                              "ratio: {:0.2f}".format(str(i), N, ratio))
+
+                        info = {"name": this_res["name"],
+                                "params": this_res["params"],
+                                "threads": i,
+                                "N": N,
+                                "ratio": ratio}
+
+                        if ratio < 1:
```
bdice: Is there a way you could write this and the printing logic above as a single function?

benchmarks/benchmarker.py
```diff
@@ -0,0 +1,329 @@
...
+                print("Something is wrong with {}".format(m))
+
+    save_benchmark_result(results, args.output)
+
+
+# compare runtime of two commits and save the comparison
+# result
+# exit 1 if the runtime of rev_this is slower than
+# the runtime of rev_other by more than the threshold ratio
+# STRUCUTRE CAN BE IMPROVED the logic is simple
```
bdice: Add this as a function docstring in complete sentences.

.circleci/config.yml
```diff
@@ -84,6 +84,31 @@ references:
...
+          python "$BENCHSCR" run -c
+          echo "Comparing with orign/master"
+          git reset --hard origin/master
+          echo "PYVER=${PYVER}"
+          rm freud/*.cpp
+          python${PYVER} setup.py build_ext --inplace --COVERAGE --ENABLE-CYTHON
+          git checkout "${CIRCLE_SHA1}" -- benchmarks/
+          echo "Running benchmark on origin/master"
+          python "$BENCHSCR" run -c
+          python "$BENCHSCR" compare origin/master "${CIRCLE_SHA1}"
```
bdice, line 116: I'd really like to put the comparison into its own task so that it appears as a separate section in the CircleCI output. Something called `Compare benchmarks`, right below `Run benchmarks`.

image

bdice: This would also clarify that the benchmarks succeeded but their _comparison_ failed (e.g. by being slow).

.circleci/config.yml
jinsooihm: https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-shell-command seems to recommend doing so.

benchmarks/benchmark.py
```diff
@@ -25,43 +26,48 @@
 # Like the unit test module, it is designed to be an overridden class.
 # Users must override run() to run their benchmark.
 # setup() is called once at the start of every benchmark for one time setup.
-
-
-class benchmark(object):
-
+class Benchmark:
```
vyasr: Can we keep the explicit `(object)` inheritance for now? As long as we're supporting Python2.7 I don't want to run into any unexpected MRO surprises.

benchmarks/benchmark_box_ParticleBuffer.py
```diff
@@ -0,0 +1,42 @@
...
+
+def run():
+    Ns = [1000, 5000, 10000]
+    print_stats = True
+    number = 100
+    name = 'freud.box.ParticleBuffer'
+
+    # L = 10
+    # buf = L*0.5
+    # images = False
```
vyasr: Can we remove?

benchmarks/benchmarker.py
```diff
@@ -0,0 +1,424 @@
...
+    Returns:
+        str: String with name and parameters.
+
+    """
+    s = name + ": \n\t"
+    s += ", ".join("{} = {}".format(str(k), str(v)) for k, v in params.items())
+    return s
+
+
+def run_benchmarks(name, Ns, number, classobj, print_stats,
```
vyasr: Can we change the signature to `print_stats=True` and save ourselves the duplication above?

benchmarks/benchmark_environment_BondOrder.py
```diff
@@ -0,0 +1,51 @@
...
+class BenchmarkEnvironmentBondOrder(Benchmark):
+    def __init__(self, rmax, k, num_neighbors, n_bins_t, n_bins_p):
+        self.rmax = rmax
+        self.k = k
+        self.num_neighbors = num_neighbors
+        self.n_bins_t = n_bins_t
+        self.n_bins_p = n_bins_p
+
+    def bench_setup(self, N):
+        # n = 4
```
vyasr: Remove

benchmarks/benchmarker.py
vyasr: I'm also happy with adding default arguments to anything else (like `number`) if they're pretty standard across our benchmarks, but `print_stats` is the obvious one.

benchmarks/benchmark_order_SolLiq.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
+import freud
+from benchmark import Benchmark
+from benchmarker import run_benchmarks
+
+
+class BenchmarkOrderSolLiq(Benchmark):
+    def __init__(self, L, rmax, Qthreshold, Sthreshold, _I):
```
vyasr: Is there a reason the last argument is "hidden" (with an underscore)?

benchmarks/util.py
```diff
@@ -0,0 +1,73 @@
+import numpy as np
+import freud
+
+
+def quatRandom():
```
vyasr: If you're already using rowan in a benchmark, you can probably use it here as well. I would probably remove this function entirely.

doc/source/design.rst
```diff
@@ -51,6 +51,54 @@ behavior of the module. These tests should be simple and short, testing
...
 
+Benchmarks
+==========
+
+Modules can be benchmarked in the following way. 
+
+.. code-block:: python
+    
+    # benchmarks/benchmark_example_Example.py
+    import Benchmark from benchmark
```
vyasr: Need to swap these

vyasr (commented):
Almost there! Some minor cleanup to do then this is ready to go I think.

jinsooihm:
Reply to 

> Is there a reason the last argument is "hidden" (with an underscore)?

The reason is that `flake8` does not like having `I` as a variable name.

vyasr:
Got it, `flake8` is pretty strict about that. If you can't come up with a better name, then the better solution to renaming the variable `_I` is just tell `flake8` to ignore that line with `#noqa`. For an example see [order.py](https://github.com/glotzerlab/freud/blob/master/freud/order.pyx#L1552).

.circleci/config.yml
```diff
@@ -102,6 +138,26 @@ references:
       - *test_cov
       - *store
 
+  build_and_benchmark: &build_and_benchmark
```
bdice, line 141: @jinsoo960 Sorry, I was unclear earlier. It looks like we're running the benchmarks twice (once to see if the benchmarks pass and again with comparison). You already did what I actually wanted when you split the `benchmark` step into `benchmark` and `comparison`.

We're running more CI than we really need to in this case. Just remove the `comparisons` job and add `- *comparison` as a step to this job.

benchmarks/benchmark.py
```diff
@@ -25,43 +26,48 @@
...
+        self.bench_run(N)
+
+    # Returns timer instance with overridden setup and run functions.
+    def setup_timer(self, N, num_threads):
+        setup = "self.bench_setup(N)"
+        stmt = "self.bench_run_parallel(N, num_threads)"
+        varmapping = {"self": self, "N": N, "num_threads": num_threads}
+        return timeit.Timer(stmt=stmt, setup=setup, globals=varmapping)
+
     # Perform the benchmark
```
bdice: Please convert these comments to a Google docstring. (and so on, below)

benchmarks/benchmark.py
```diff
@@ -1,16 +1,17 @@
...
 import pstats
 import sys
 import multiprocessing
 import numpy
+import timeit
 
 from freud import parallel
 
+
 # Run a given method and print benchmark results
```
bdice: Please convert these comments to a Google docstring.

benchmarks/benchmark_density_LocalDensity.py
```diff
@@ -0,0 +1,43 @@
...
+class BenchmarkDensityLocalDensity(Benchmark):
+    def __init__(self, nu, rcut):
+        self.nu = nu
+        self.rcut = rcut
+
+    def bench_setup(self, N):
+        box_size = math.sqrt(N*self.nu)
+        seed = 0
+        np.random.seed(seed)
+        self.pos = np.random.random_sample((N, 3)).astype(np.float32) \
```
bdice, line 20: Optional: We could create a helper function `random_points(box, N)` in the `util` file for all our benchmarks, that generates random points between `[0, 1]` in each dimension (accounting for the `box` dimensionality, with z=0 for 2D boxes), and then uses `box.makeCoordinates`.

bdice: I would prefer to initialize the box in the setup function, not the run function.

benchmarks/benchmark_order_LocalQlNear.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+import freud
+from benchmark import Benchmark
+from benchmarker import run_benchmarks
+
+
+class BenchmarkOrderLocalQlNear(Benchmark):
+    def __init__(self, L, rmax, _I, kn):
```
bdice: The `_I` parameter here is actually not an `I` "eye" at all - it's supposed to be a lowercase `l` "ell." flake8 will also be mad if you use a lowercase `l`, since it's easy to mix up with a capital `I` (as we see). Please rename `_I` to `sph_l` for clarity, that's the convention we've adopted in a few places.

(same for other Steinhardt-variant classes, RotationalAutocorrelation, etc.)

doc/source/design.rst
```diff
@@ -51,6 +51,54 @@ behavior of the module. These tests should be simple and short, testing
...
 
+Benchmarks
+==========
+
+Modules can be benchmarked in the following way. 
+
+.. code-block:: python
+    
+    # benchmarks/benchmark_example_Example.py
+    from benchmark import Benchmark
```
bdice: Missing something like `from benchmarker import run_benchmarks`

Please check this to be sure this example file matches the implemented design. To make it simpler, we could just include a real benchmark with [the `.. literalinclude::` directive](https://www.sphinx-doc.org/en/1.5/markup/code.html#directive-literalinclude) and say "The following code is an example benchmark for the `freud.density.RDF` module," etc.

doc/source/design.rst
```diff
@@ -51,6 +51,54 @@ behavior of the module. These tests should be simple and short, testing
...
+                              rmax=rmax, dr=dr)
+
+    if __name__ == '__main__':
+        run()
+
+in a file :code:`benchmark_example_Example` in the :code:`benchmarks` directory.
+More concrete examples can be found in the :code:`benchmarks` directory.
+The runtime of :code:`BenchmarkExample.bench_run` will be timed for :code:`number`
+of times for input sizes of :code:`Ns`. Its runtime with respect to the number of threads
+will also be measured. When a commit is pushed, circle.ci will
```
bdice: The latter sentence might be clearer if you said, "Benchmarks are run as a part of continuous integration, with performance comparisons between the current commit and the `master` branch."

bdice (changes requested):
Some comments for improvement, then this should be ready to go.

bdice:
@jinsoo960 One more thing I forgot to mention - there's a folder of benchmarks you labeled "to be removed." If that code can be superseded by proper benchmarks for each of those modules, just remove those files completely.

vyasr:
@bdice I think all your comments have been addressed, but can you check to be sure?

bdice:
@vyasr @jinsoo960 I made some changes, I approve it in the current state after all tests pass (I don't think I broke anything but I didn't test it locally so let's read the CI benchmark reports one more time before merging).

bdice (approved):

vyasr:
New changes look fine, and all tests pass.

PR #281, opened 2019-05-20
Fix/parallel numthreads

vyasr:
## Description
The Python variable tracking the number of threads in use was not being set correctly because we weren't correctly setting a global variable.

## Motivation and Context
Since the variable is private, this bug should have had limited effects on the user. The main exception is that if a user called `setNumThreads(N)` and then used the context manager `with NumThreads(M)`, upon exiting the context manager the number of threads would revert to 0 rather than the N that the user originally set.

## How Has This Been Tested?
I added explicit tests of the private module variable to make sure its value was being appropriately stored.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 * Corrected type of `y_max` argument to PMFTXY2D from int to float.
 * Reduce logging verbosity about array conversion.
+* Fixed number of threads set upon exiting the NumThreads context manager.
 
 ## v1.0.0 - 2019-02-08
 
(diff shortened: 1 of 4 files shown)


bdice (approved):
Looks good to me.

bdice:
@vyasr Is there a reason we have `_numThreads` as private? I was trying to check how many threads were active recently, and I couldn't find a way to check that. Perhaps we can expose it as a renamed public variable, `freud.parallel.num_threads`? We can do that in this PR or another PR.

vyasr:
No, I don't think there is except that nobody asked for anything else. I'd be fine with making that variable public. However, I'd also want to make it settable, to make for a nicer API, e.g. `freud.parallel.num_threads = 5`. Since that will require creating a lightweight class capable of handling this logic, I'd say let's push that to a separate PR.

PR #284, opened 2019-05-22
LinkCell computes cell neighbors on need basis

jinsooihm:
## Motivation and Context

LinkCell used to compute cell neighbors for all cells, which was an overhead for sparse inputs, since there are some cells that never get used, if the cell does not contain any particle.
For example, suppose we generate random points within `-L/2` and `L/2`. 
If we use box size of `L` for `LinkCell.compute` method, the old implementation and the new implementation run essentially at same speed, if the particles are densely distributed.
However, if we increase the box size, for example `3*L`, the particles within the box are concentrated in the center, and there are no particles in the boundary.
The old implementation still computes cell neighbors on those outer cells that do not ever get used, and gives us essentially `L/rcut`^3 complexity, which can be observed from explosion in the runtime.
The new implementation does not compute cell neighbors on those outer cells, and thus it runs at essentially same speed as the box of size `L` case. 

## How Has This Been Tested?

It passed the tests on circle.ci, and the runtime behavior has been measured.

comparison_rcut_dev0 5
The new implementation.
comparison_rcut_stable0 5
The stable implementation.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/LinkCell.cc
+++ b/cpp/locality/LinkCell.cc
@@ -83,7 +83,6 @@ void LinkCell::updateInternal(const box::Box& box, float cell_width)
                 throw runtime_error("At least one cell must be present.");
                 }
             m_celldim = celldim;
-            // computeCellNeighbors();
             }
         m_cell_width = cell_width;
         }
@@ -273,9 +272,9 @@ const std::vector<unsigned int>& LinkCell::computeCellNeighbors(unsigned int cur
     {
     std::vector<unsigned int> neighbor_cells;
     vec3<unsigned int> l_idx = m_cell_index(cur_cell);
-    unsigned int i = l_idx.x;
-    unsigned int j = l_idx.y;
-    unsigned int k = l_idx.z;
+    int i = (int) l_idx.x;
+    int j = (int) l_idx.y;
+    int k = (int) l_idx.z;
 
     // loop over the neighbor cells
     int starti, startj, startk;
--- a/cpp/locality/LinkCell.cc
+++ b/cpp/locality/LinkCell.cc
@@ -272,61 +272,61 @@ const std::vector<unsigned int>& LinkCell::computeCellNeighbors(unsigned int cur
     {
     std::vector<unsigned int> neighbor_cells;
     vec3<unsigned int> l_idx = m_cell_index(cur_cell);
-    int i = (int) l_idx.x;
-    int j = (int) l_idx.y;
-    int k = (int) l_idx.z;
+    const int i = (int) l_idx.x;
+    const int j = (int) l_idx.y;
+    const int k = (int) l_idx.z;
 
     // loop over the neighbor cells
     int starti, startj, startk;
     int endi, endj, endk;
     if (m_celldim.x < 3)
         {
-        starti = (int)i;
+        starti = i;
         }
     else
         {
-        starti = (int)i - 1;
+        starti = i - 1;
         }
     if (m_celldim.y < 3)
         {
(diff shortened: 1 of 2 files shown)


cpp/locality/LinkCell.cc
```diff
@@ -269,93 +269,91 @@ void LinkCell::compute(const box::Box& box,
...
-                    {
-                    startk = (int)k;
-                    }
-                else
-                    {
-                    startk = (int)k - 1;
-                    }
+    std::vector<unsigned int> neighbor_cells;
+    vec3<unsigned int> l_idx = m_cell_index(cur_cell);
+    unsigned int i = l_idx.x;
```
bdice: I'd like to cast to `int` here, instead of casting `i`, `j`, `k` to `int` later.

bdice (commented):
@jinsoo960 Can you remove the commented code? Also I'd like to see how the new LinkCell code performs vs. AABBQuery with various numbers of threads, perhaps for even smaller rcut like 0.1 (for L=10).

cpp/locality/LinkCell.cc
```diff
@@ -83,7 +83,7 @@ void LinkCell::updateInternal(const box::Box& box, float cell_width)
                 throw runtime_error("At least one cell must be present.");
                 }
             m_celldim = celldim;
-            computeCellNeighbors();
+            // computeCellNeighbors();
```
vyasr: Remove

cpp/locality/LinkCell.h
```diff
@@ -423,9 +424,23 @@ class LinkCell : public NeighborQuery
...
+                CellNeighbors::const_accessor a;
+                if(m_cell_neighbors.find(a, cell)) 
+                    {
+                        return a->second;
+                    }
+                else 
+                    {
+                        // computeCellNeighbors(cell);
+                        // m_cell_neighbors.find(a, cell);
+                        // return a->second;
```
vyasr: Remove these as well.

vyasr (approved):
Just saw Bradley's comment; I concur, otherwise this looks good. Didn't require too extensive of a change.

cpp/locality/LinkCell.h
```diff
@@ -467,7 +482,9 @@ class LinkCell : public NeighborQuery
         vec3<unsigned int> m_celldim;  //!< Cell dimensions
 
         std::shared_ptr<unsigned int> m_cell_list;                 //!< The cell list last computed
-        std::vector< std::vector<unsigned int> > m_cell_neighbors; //!< List of cell neighbors to each cell
+        // std::vector< std::vector<unsigned int> > m_cell_neighbors; //!< List of cell neighbors to each cell
+        typedef tbb::concurrent_hash_map<unsigned int, std::vector<unsigned int> >  CellNeighbors;
```
bdice: Remove the line above and move its comment to this line: `//!< Hash map of cell neighbors for each cell`

cpp/locality/LinkCell.h
```diff
@@ -467,7 +482,9 @@ class LinkCell : public NeighborQuery
...
+        CellNeighbors m_cell_neighbors;
```
bdice: Remove the commented line above and move its description to this line: `//!< Hash map of cell neighbors for each cell`

bdice (commented):
I think this is ready to merge after comments are addressed. I would still like to see the scaling with threads.

bdice:
I'm going to update this branch right now, addressing the review comments, so it can go out with the 1.1 release.

PR #288, opened 2019-05-24
Use NumPy arrays for more return types in Box.

### LINKED ISSUE #280: box.L should return an array instead of a tuple
Almost all outputs in `freud` are NumPy arrays. The property `box.L` returns a tuple.

This might have been intentional since tuples are immutable, but returning a tuple makes it tougher to use with some other NumPy behaviors (e.g. the ability to multiply or divide by a constant) without calling `np.asarray(box.L)`.

bdice:
## Motivation and Context
This PR resolves #280 and related places where return types are not NumPy arrays.

## How Has This Been Tested?
Tests were updated to support NumPy arrays.

## Types of changes

## Checklist:

### DIFF
--- a/freud/box.pyx
+++ b/freud/box.pyx
@@ -298,13 +298,12 @@ cdef class Box:
 
         vecs = freud.common.convert_array(
             vecs, vecs.ndim, dtype=np.float32, contiguous=True)
-        images = np.empty_like(vecs, dtype=int)
 
         if vecs.ndim == 1:
-            images[:] = self._getImage(vecs)
+            images = np.asarray(self._getImage(vecs), dtype=int)
         elif vecs.ndim == 2:
-            for i, vec in enumerate(vecs):
-                images[i] = self._getImage(vec)
+            images = np.asarray([self.getImage(vec) for vec in vecs],
+                                dtype=int)
         return images
 
     def _getImage(self, vec):
--- a/freud/box.pyx
+++ b/freud/box.pyx
@@ -72,7 +72,7 @@ cdef class Box:
             The xz tilt factor.
         yz (float):
             The yz tilt factor.
-        L (tuple, settable):
+        L (:math:`\left(3\right)` :class:`numpy.ndarray`, settable):
             The box lengths along x, y, and z.
         Lx (float, settable):
             The x-dimension length.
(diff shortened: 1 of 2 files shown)


freud/box.pyx
```diff
@@ -298,13 +298,14 @@ cdef class Box:
...
 
         if vecs.ndim == 1:
-            vecs[:] = self._getImage(vecs)
+            images[:] = self._getImage(vecs)
         elif vecs.ndim == 2:
             for i, vec in enumerate(vecs):
-                vecs[i] = self._getImage(vec)
-        return vecs
+                images[i] = self._getImage(vec)
+        return images
```
vyasr, line 307: This function might be a bit more readable if you just made a list `images` and then returned `np.asarray(images)`. Then you could do the 2d case as just a listcomp and the 1d case wouldn't need the colon. But I'll leave it to you whether or not you care enough to make the change.

tests/test_box_Box.py
```diff
@@ -193,7 +193,7 @@ def test_vectors(self):
         b_list = [1, 2, 3, 0.1, 0.2, 0.3]
         Lx, Ly, Lz, xy, xz, yz = b_list
         box = freud.box.Box.from_box(b_list)
-        self.assertEqual(
+        npt.assert_allclose(
```
vyasr, line 196: Does this actually fail as an equality check from precision issues?

tests/test_box_Box.py
```diff
@@ -208,21 +208,21 @@ def test_vectors(self):
 
     def test_periodic(self):
         box = freud.box.Box(1, 2, 3, 0, 0, 0)
-        false = [False, False, False]
-        true = [True, True, True]
-        self.assertEqual(box.periodic, true)
+        falses = [False, False, False]
+        trues = [True, True, True]
```
vyasr: I would prefer these written either using `np.[zeros|ones](3, dtype=bool)` or `[False|True]*3`, but again I'll leave it to your discretion if you actually want to make the change.

vyasr: Actually, I take it back. Can you remove these arrays entirely and just change all the tests to `npt.assert_array_equal(box.periodic, True)`? That looks much better.

vyasr (changes requested):
I have one requested change to the test, a couple of other optional ones. Then this is good to go.

tests/test_box_Box.py
bdice: @vyasr I changed it so it would match the other tests in the same section, which do have to use `assert_allclose` for precision reasons. We've been getting by on this one, because there isn't any math to throw off the precision of this specific lattice vector, but I think it's better to make the checks act the same.

vyasr (approved):

PR #290, opened 2019-05-28
Refactored convert array

jinsooihm:
## Description

Made the default `dtype` of `freud.common.convert_array` to be `numpy.float32` and removed the parameter `array_name`. 
Relevant changes were made in other modules that use the `freud.common.convert_array` method.
Also, removed the parameter `contiguous=True` since it is always true.
Added default argument for `dimensions` to be `None`.
If `dimensions=None`, no dimensionality check is done. 

## Motivation and Context

The code looks shorter and simpler.

## How Has This Been Tested?

Passed all the current tests.

## Types of changes

## Checklist:

### DIFF
--- a/freud/box.pyx
+++ b/freud/box.pyx
@@ -228,8 +228,7 @@ cdef class Box:
                 "Invalid dimensions for fractions given to makeCoordinates. "
                 "Valid input is an array of shape (3,) or (N,3).")
 
-        fractions = freud.common.convert_array(
-            fractions, fractions.ndim, dtype=np.float32, contiguous=True)
+        fractions = freud.common.convert_array(fractions, fractions.ndim)
 
         if fractions.ndim == 1:
             fractions[:] = self._makeCoordinates(fractions)
@@ -261,8 +260,7 @@ cdef class Box:
                 "Invalid dimensions for vecs given to makeFraction. "
                 "Valid input is an array of shape (3,) or (N,3).")
 
-        vecs = freud.common.convert_array(
-            vecs, vecs.ndim, dtype=np.float32, contiguous=True)
+        vecs = freud.common.convert_array(vecs, vecs.ndim)
 
         if vecs.ndim == 1:
             vecs[:] = self._makeFraction(vecs)
@@ -296,8 +294,7 @@ cdef class Box:
                 "Invalid dimensions for vecs given to getImage. "
                 "Valid input is an array of shape (3,) or (N,3).")
 
-        vecs = freud.common.convert_array(
-            vecs, vecs.ndim, dtype=np.float32, contiguous=True)
+        vecs = freud.common.convert_array(vecs, vecs.ndim)
 
         if vecs.ndim == 1:
             vecs[:] = self._getImage(vecs)
@@ -350,8 +347,7 @@ cdef class Box:
                 "Invalid dimensions for vecs given to wrap. "
                 "Valid input is an array of shape (3,) or (N,3).")
 
-        vecs = freud.common.convert_array(
-            vecs, vecs.ndim, dtype=np.float32, contiguous=True)
+        vecs = freud.common.convert_array(vecs, vecs.ndim)
 
         if vecs.ndim == 1:
             # only one vector to wrap
@@ -393,10 +389,8 @@ cdef class Box:
                 "Invalid dimensions for vecs given to unwrap. "
                 "Valid input is an array of shape (3,) or (N,3).")
 
-        vecs = freud.common.convert_array(
-            vecs, vecs.ndim, dtype=np.float32, contiguous=True)
-        imgs = freud.common.convert_array(
-            imgs, vecs.ndim, dtype=np.int32, contiguous=True)
+        vecs = freud.common.convert_array(vecs, vecs.ndim)
+        imgs = freud.common.convert_array(imgs, vecs.ndim, dtype=np.int32)
 
         if vecs.ndim == 1:
             # only one vector to unwrap
@@ -740,8 +734,7 @@ cdef class ParticleBuffer:
                 each side, meaning that one image doubles the box side lengths,
                 two images triples the box side lengths, and so on.
         """
-        points = freud.common.convert_array(
-            points, 2, dtype=np.float32, contiguous=True, array_name='points')
+        points = freud.common.convert_array(points, 2, array_name='points')
 
         if points.shape[1] != 3:
             raise RuntimeError(
--- a/freud/box.pyx
+++ b/freud/box.pyx
@@ -228,7 +228,7 @@ cdef class Box:
                 "Invalid dimensions for fractions given to makeCoordinates. "
                 "Valid input is an array of shape (3,) or (N,3).")
 
-        fractions = freud.common.convert_array(fractions, fractions.ndim)
+        fractions = freud.common.convert_array(fractions)
 
         if fractions.ndim == 1:
             fractions[:] = self._makeCoordinates(fractions)
@@ -260,7 +260,7 @@ cdef class Box:
                 "Invalid dimensions for vecs given to makeFraction. "
                 "Valid input is an array of shape (3,) or (N,3).")
 
-        vecs = freud.common.convert_array(vecs, vecs.ndim)
+        vecs = freud.common.convert_array(vecs)
 
         if vecs.ndim == 1:
             vecs[:] = self._makeFraction(vecs)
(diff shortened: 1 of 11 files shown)


freud/cluster.pyx
```diff
@@ -99,7 +99,7 @@ cdef class Cluster:
                 Simulation box (Default value = None).
         """
         points = freud.common.convert_array(
-            points, 2, dtype=np.float32, contiguous=True)
+            points, 2, dtype=np.float32)
```
bdice: Remove dtype (it defaults to float32). Also please combine lines 101-102 (and similarly elsewhere in this PR, if the function can be written on one line without hitting flake8 line length limits).

freud/cluster.pyx
```diff
@@ -247,13 +247,12 @@ cdef class ClusterProperties:
         else:
             b = freud.common.convert_box(box)
 
-        points = freud.common.convert_array(
-            points, 2, dtype=np.float32, contiguous=True)
+        points = freud.common.convert_array(points, 2, dtype=np.float32)
```
bdice: Remove dtype.

freud/common.py
```diff
@@ -37,7 +35,7 @@ def convert_array(array, dimensions, dtype=None,
 
     if array.ndim != dimensions:
         raise TypeError("{}.ndim = {}; expected ndim = {}".format(
-            array_name or "array", array.ndim, dimensions))
+            "array", array.ndim, dimensions))
```
bdice: Replace this format string with the literal string `array`:
```python
raise TypeError("array.ndim = {}; expected ndim = {}".format(array.ndim, dimensions))
```

freud/msd.pyx
```diff
@@ -142,19 +142,15 @@ cdef class MSD:
...
             raise TypeError(
                 'positions should be a 3-dimensional array of shape'
                 '(N_frames, N_particles, 3)'
             )
 
         if images is not None:
-            images = freud.common.convert_array(
-                images, 3, dtype=np.float32, contiguous=True,
-                array_name="images")
+            images = freud.common.convert_array(images, 3)
```
bdice: @vyasr Why do we treat `images` as floating values?

freud/box.pyx
```diff
@@ -228,8 +228,7 @@ cdef class Box:
                 "Invalid dimensions for fractions given to makeCoordinates. "
                 "Valid input is an array of shape (3,) or (N,3).")
 
-        fractions = freud.common.convert_array(
-            fractions, fractions.ndim, dtype=np.float32, contiguous=True)
+        fractions = freud.common.convert_array(fractions, fractions.ndim)
```
bdice: There are a couple cases like this one (and `face_orientations` in the PMFT module) where we're not actually doing anything with the number of dimensions. I'd like to see `dimensions=None` by default, and remove this parameter from calls to `convert_array` where we don't enforce a specific array dimensionality. This would require that we change the logic in `convert_array` to `if dimensions is not None and array.ndim != dimensions:`.

freud/common.py
```diff
@@ -10,8 +10,8 @@
 logger = logging.getLogger(__name__)
 
 
-def convert_array(array, dimensions, dtype=None,
-                  contiguous=True, array_name=None):
+def convert_array(array, dimensions, dtype=np.float32,
+                  contiguous=True):
```
bdice: I'm inclined to require `contiguous=True` in all cases, and remove this option from the function signature. We cannot use arrays in C++ unless they're contiguous, and all previous calls to `convert_array` set this to `True`.

bdice (changes requested):
Good work! I suggested a few changes.

freud/common.py
```diff
@@ -20,26 +19,22 @@ def convert_array(array, dimensions, dtype=None,
...
-        dimensions (int): Expected dimensions of the array.
+        dimensions (int): Expected dimensions of the array. If 'None',
+            no dimensionality check will be done (Default value = 'None').
         dtype: code:`dtype` to convert the array to if :code:`array.dtype`
             is different. If `None`, :code:`dtype` will not be changed.
-            (Default value = None).
-        contiguous (bool): Whether to cast the array to a contiguous (Default
-            value = True).
+            (Default value = `numpy.float32`).
         array. Default behavior casts to a contiguous array.
```
bdice: I think this line is left over from a previous commit and needs to be removed?

freud/common.py
```diff
@@ -20,26 +19,22 @@ def convert_array(array, dimensions, dtype=None,
...
-        array_name (str): Name of the array, used for errors (Default value =
-            None).
 
     Returns:
         py:class:`numpy.ndarray`: Array.
```
bdice: Probably needs to be `:class:` without the `py:`. Typically this is written as `:py:class:` (note that this line is missing a colon on the left side anyway) but we have left off the `:py` elsewhere in the docs, since it's implied as the default Sphinx domain for Python packages.

tests/test_common.py
```diff
@@ -20,12 +20,12 @@ def test_convert_array(self):
         npt.assert_equal(z.dtype, np.float32)
         # now make contiguous
         npt.assert_equal(y.flags.contiguous, False)
-        z = common.convert_array(y, 2, contiguous=True)
+        z = common.convert_array(y, 2)
         npt.assert_equal(z.flags.contiguous, True)
         # test the dim_message
```
bdice: There isn't a `dim_message` anymore - please remove this portion of the test.

bdice: We should explicitly test two additional cases: passing in a list of lists, and converting array data of type `np.float64`.

bdice (changes requested):
A few more edits requested, then I think this will be ready.

freud/msd.pyx
vyasr: There is no good reason for that. @jinsoo960 please change that to `np.int32`. Good catch.

tests/test_common.py
```diff
@@ -20,16 +20,30 @@ def test_convert_array(self):
...
+        z = common.convert_array(y, dtype=np.float64)
+        npt.assert_equal(z.dtype, np.float64)
+
+        # test for list of list input
+        yl = [list(r) for r in y]
+        zl = common.convert_array(yl, 2)
+        z = common.convert_array(y, 2)
+        npt.assert_equal(z, zl)
+
+        # test for dimensions defualt argument
```
vyasr: Typo default

tests/test_common.py
```diff
@@ -20,16 +20,30 @@ def test_convert_array(self):
         npt.assert_equal(z.dtype, np.float32)
         # now make contiguous
         npt.assert_equal(y.flags.contiguous, False)
-        z = common.convert_array(y, 2, contiguous=True)
+        z = common.convert_array(y, 2)
         npt.assert_equal(z.flags.contiguous, True)
         # test the dim_message
         try:
```
vyasr: Can we change this test to use `with self.assertRaises(TypeError)`?

vyasr (changes requested):
While we are making these changes, there are a couple of other standardizations that I would like to consider adding to `convert_array`.

1) Many methods perform some sort of shape check after calling `convert_array`; could we implement a shape check into this function? One method to accomplish this would be to accept tuples where any entry that is `None` is not checked for length, e.g. `points = convert_array(points, 2, (None, 3))`. That would also allow us to subsume the dimensionality check into a check on the length of the shape tuple. 

2) A related point, there are a number of methods that do either `convert_array(array.squeeze(), ...)` or `convert_array(np.atleast_2d(array), ...)`. If we changed to the shape-based check, we could perform the appropriate resizing in `convert_array`, then perform the shape check to verify that the resizing gave the expected output.

@jinsoo960 @bdice what are your thoughts on either of these proposals? I think that would help clean things up a bit more.

jinsooihm:
@vyasr 
1. If we were to incorporate the shape check into `convert_array`, then would we want to also take runtime error messages as an input?

bdice (approved):
Looks good! Thanks @jinsoo960. 🎉

bdice:
> While we are making these changes, there are a couple of other standardizations that I would like to consider adding to `convert_array`.
> 
> 1. Many methods perform some sort of shape check after calling `convert_array`; could we implement a shape check into this function? One method to accomplish this would be to accept tuples where any entry that is `None` is not checked for length, e.g. `points = convert_array(points, 2, (None, 3))`. That would also allow us to subsume the dimensionality check into a check on the length of the shape tuple.
> 2. A related point, there are a number of methods that do either `convert_array(array.squeeze(), ...)` or `convert_array(np.atleast_2d(array), ...)`. If we changed to the shape-based check, we could perform the appropriate resizing in `convert_array`, then perform the shape check to verify that the resizing gave the expected output.
> 
> @jinsoo960 @bdice what are your thoughts on either of these proposals? I think that would help clean things up a bit more.

I think these belong in a separate PR. @vyasr I'd like to merge this as-is and file an issue for the changes you suggested, if you're alright with that.

vyasr:
I'm fine with making those changes in a separate PR. I'll merge this and open an issue for that.

vyasr (approved):

PR #291, opened 2019-05-29
Feature/voronoi typing

yzjin:
## Description
Changed the  data structure of neighbor list from a ```list of list``` to a ```dictionary``` in ```computeNeighbors```.
Built a well-sorted neighbor list using sparse matrix and removed unnecessary array copying.
Removed redundant code in ```nlist``` and made it a pure getter.

## Motivation and Context
Speed up voronoi calculation.

## How Has This Been Tested?
```test_voronoi.py``` was updated to support well-sorted neighbor list right now.
Passed other relevant tests.

## Types of changes

## Checklist:

### CODE REVIEW

File: freud/voronoi.pyx [hunk 1]
```diff
@@ -322,41 +313,56 @@ class Voronoi(object):
...
+        max_index = np.max(nlist[:]+1)
+        # convert neighbor list to a csr matrix and set weights = 1
+        sparse_neighbors = csr_matrix(
+            (np.ones(len(nlist)), (nlist.index_i, nlist.index_j)),
+            shape=(max_index, max_index))
+        # take a numShell power of the matrix
+        # sum over all shell neighbors, and convert to a linked list
+        sparse_neighbors = sum(
+            [sparse_neighbors**k for k in range(1, numShells+1)])
+        lil_neighbors = sparse_neighbors.tolil(copy=False)
```
bdice: Can you remove the `copy=False`? I think we actually want a copy in most cases, since we want to convert it to the list-of-lists arrangement in memory. Then we can index `lil_neighbors` efficiently.

bdice (changes requested):
@yzjin I requested one change, otherwise this looks good.

freud/voronoi.pyx
bdice: Also the tests failed on Python 3.5, perhaps because older versions of SciPy don't have the `copy` argument. Another reason to drop it.

bdice (approved):
Looks good to me.

vyasr (approved):

PR #295, opened 2019-05-31
Added functional API to loop over neighbors

jinsooihm:
## Description

To support taking in `NeighborQuery` object as input and not build unnecessary `NeighborList`, a wrapper function, `freud::locality::loopOverNeighbors`, has been added.
`freud::locality::loopOverNeighbors` has a logic for deciding what to do depending on the input, and iterates over the neighbors to do some computation.

Implemented for
  - `Cluster`
  - `CorrelationFunction`
  - `RDF`
  - `PMFT`
  - `BondOrder`

We need to first change the order of `ref_point` and `point` in NeighborList for this framework to work nicely for:
  - `Steinhardt`
  - `HexOrderParameter`
  - `TransOrderParameter`
  - `LocalDensity`

These do not use `NeighborList`:
  - `ClusterProperties`
  - `GaussianDensity`
  - `CubaticOrderParameter`
  - `NematicOrderParameter`
  - `RotationalAutocorrelation`

These require the knowing the `NeighborList`:
  - `AngularSeparation`
  - `LocalBondProjection`
  - `LocalDescriptors`
  - `SolLiq`

These use special data structure that may not be thread safe. There is a support for non-parallel loop logic now:
  - `MatchEnv`

## Motivation and Context

Since this new implementation does not require building a `NeighborList`, the runtime is faster than the stable implementation, in which `freud.locality.make_default_nlist` is called to build a `NeighborList`.
Also, at least for `freud.cluster.Cluster.computeClusters`, there is no noticeable overhead from additional function calls resulting from using `freud::locality::loop_over_NeighborList` compared to the stable implementation.

## How Has This Been Tested?

New tests have been added and passes all the new and already existing tests.

## TODO
`cpp/cluster/dset.h` has to be attributed to [this repositry](https://github.com/wjakob/dset).

## Types of changes

- [x] API breaking change (this branch is branched off from #328)

## Checklist:


cpp/cluster/Cluster.cc
```diff
@@ -76,56 +79,55 @@ Cluster::Cluster(float rcut)
...
+
+    freud::locality::loop_over_NeighborList(nq, points, Np, qargs, nlist, 
+        [&rmaxsq, &dj, &box, points] (size_t i, size_t j) 
+            { 
+            vec3<float> p = points[i];
+            if (i != j)
                 {
-                if (i != j)
+                // compute r between the two particles
+                vec3<float> delta = p - points[j];
```
bdice: Write as one expression, `vec3<float> delta = box.wrap(p - points[j]);`

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+                size_t j(neighbor_list[2*bond + 1]);
+                cf(i, j);
+                }
+        });
+    }
+
+// Body should be an object taking in 
+// with operator(size_t begin, size_t end)
+template<typename Body>
+void for_loop_wrapper(bool parallel, size_t begin, size_t end, const Body& body)
```
bdice: Does this belong in `util/`? It seems more generally useful.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+                                  QueryArgs qargs, const NeighborList* nlist, const ComputePairType& cf)
+    {
+    // check if nlist exists
+    if(nlist != NULL)
+        {
+
+        // if nlist exists, loop over it parallely
+        const size_t *neighbor_list(nlist->getNeighbors());
+        size_t n_bonds = nlist->getNumBonds();
+        parallel_for(tbb::blocked_range<size_t>(0, n_bonds),
```
bdice: Should this use the parallel wrapper?

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+// ComputePairType should be a void function that takes (ref_point, point) as input.
+template<typename ComputePairType>
+void loop_over_NeighborList(const NeighborQuery *ref_points, const vec3<float> *points, unsigned int Np,
+                                  QueryArgs qargs, const NeighborList* nlist, const ComputePairType& cf)
+    {
+    // check if nlist exists
+    if(nlist != NULL)
+        {
+
+        // if nlist exists, loop over it parallely
```
bdice: "in parallel" ;)

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+
+        // dummy implementation for pure virtual function in the parent class
+        virtual std::shared_ptr<NeighborQueryIterator> queryBall(const vec3<float> *points, unsigned int N, float r, bool exclude_ii=false) const
+        {
+            return nullptr;
+        }
+
+    };
+
+// ComputePairType should be a void function that takes (ref_point, point) as input.
```
bdice: Clarify that the points are actually indices, not `vec3<float>` coordinates.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+                    cf(i, j);
+                    }
+            });
+        }
+    else
+        {
+        // if nlist does not exist, check if ref_points is an actual NeighborQuery
+        std::shared_ptr<NeighborQueryIterator> iter;
+        AABBQuery* abq = nullptr;
+        if(const RawPoints* rp = dynamic_cast<const RawPoints*>(ref_points))
```
bdice: I didn't know how this worked - a comment would be useful. "If a dynamic_cast fails, the result of the conversion will be a null pointer." (source: https://www.learncpp.com/cpp-tutorial/12-9-dynamic-casting/)

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+            // if ref_points is RawPoints, build a NeighborQuery
+            abq = new AABBQuery(ref_points->getBox(), ref_points->getRefPoints(), ref_points->getNRef());
+            iter = abq->queryWithArgs(points, Np, qargs);
+            }
+        else
+            {
+            iter = ref_points->queryWithArgs(points, Np, qargs);
+            }
+
+        // iterate over the query object parallely
```
bdice: "in parallel"

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+        tbb::parallel_for(tbb::blocked_range<size_t>(0, Np),
```
bdice: Use parallel wrapper?

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+#include <tbb/tbb.h>
+
+#include "NeighborQuery.h"
+#include "AABBQuery.h"
+#include "NeighborList.h"
+#include "Index1D.h"
+
+namespace freud { namespace locality {
+
+class RawPoints : public NeighborQuery
```
bdice: Should this go in another file, `NeighborQuery.h` or `RawPoints.h`?

freud/locality.pyx
```diff
@@ -719,6 +723,28 @@ cdef class NeighborList:
...
+        ref_points (:class:`freud.locality.AABBQuery`,
+            :class:`freud.locality.LinkCell`, or :class:`numpy.ndarray`):
+            NeighborQuery object or numpy array to build :class:`RawPoints` out of.
+
+    Returns:
+        :class:`freud.locality.NeighborQuery`
+            The same :class:`NeighborQuery` object if one is given or :class:`RawPoints`
+            built from :code:`box` and :code:`ref_points`.
+    """  # noqa: E501
+    if isinstance(ref_points, AABBQuery) or isinstance(ref_points, LinkCell):
```
bdice: Should this just check if it's an instance of the parent class `NeighborQuery`?

freud/locality.pyx
```diff
@@ -719,6 +723,28 @@ cdef class NeighborList:
...
 
+def make_default_nq(box, ref_points):
+    R"""Helper function to return a NeighborQuery object.
+
+    Args:
+        box (:class:`freud.box.Box`):
+            Simulation box.
+        ref_points (:class:`freud.locality.AABBQuery`,
+            :class:`freud.locality.LinkCell`, or :class:`numpy.ndarray`):
+            NeighborQuery object or numpy array to build :class:`RawPoints` out of.
```
bdice: ```NeighborQuery object or NumPy array used to build :class:`RawPoints`.```

freud/locality.pyx
```diff
@@ -806,6 +832,46 @@ def make_default_nlist_nn(box, ref_points, points, n_neigh, nlist=None,
     return nn.nlist, nn
 
 
+cdef class RawPoints(NeighborQuery):
+    R"""Use an AABB tree to find neighbors.
```
bdice: Update docstrings, set yourself as author, update versionadded, etc.

bdice:
Overall this looks really great! Todo:
- [x] Merge in `master`
- [x] Apply clang-format
- [x] Remove old commented blocks
- [ ] Add attribution / source / license of DisjointSet
- [x] Move `cluster/dset.h` to `util/DisjointSet.h`
- [x] Refactor and remove previous DisjointSet implementation if possible (also check MatchEnv)
- [ ] Review suggested PR changes

Are the `LocalDensity` changes ready for review? It looked unfinished (or untested?) at first glance.

Longer term, I'd like to consider removing the `rsq < rmaxsq` check in C++ (here and in other compute modules) and just build clusters using the network of neighbors found via the NeighborQuery/NeighborList. I'd like users to be able to pass a `NeighborList` directly and use it, but that would make it unclear with the current constructor argument `rcut` (see also #297). If we decided to do that, we would need to remove the `rcut` constructor argument from Cluster and instead require an `r` query argument to produce the current behavior.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,136 @@
...
+        const size_t *neighbor_list(nlist->getNeighbors());
+        size_t n_bonds = nlist->getNumBonds();
+        parallel_for(tbb::blocked_range<size_t>(0, n_bonds),
+            [=] (const tbb::blocked_range<size_t>& r)
+            {
+                for(size_t bond = r.begin(); bond !=r.end(); ++bond)
+                    {
+                    size_t i(neighbor_list[2*bond]);
+                    size_t j(neighbor_list[2*bond + 1]);
+                    cf(i, j);
```
bdice: Is it possible to include neighbor distances and weights here? `cf(i, j, dist, weight);`?

Something like
```c++
vec3<float> rij(ref_points->getBox().wrap(ref_points->getRefPoints()[i] - points[j]));
float dist(sqrt(dot(rij, rij)));
```
(make sure I put `i` and `j` in the right order, etc.)

Adding `distance`s will make it much easier to write the actual compute functions.

We don't currently use `weight`s to my knowledge but I want to run some analyses that incorporate weights so this is a good time to add that generality. Weights would need to be added to the definition of a neighbor bond in the NeighborQuery code. For both `query` and `queryBall`, the weights should always be `1`.

cpp/locality/NeighborComputeFunctional.h
jinsooihm: This was just an idea, I am not sure where it should belong.

cpp/locality/NeighborComputeFunctional.h
jinsooihm: I think this would be better done with making NeighborList to contain distance also. 

Which I did for now.

cpp/locality/NeighborComputeFunctional.h
jinsooihm: I think it would be sufficient to put in `NeighborQuery.h`.

cpp/locality/NeighborComputeFunctional.h
jinsooihm: Could you elaborate more on adding weights to NeighborQuery?

cpp/locality/LinkCell.h
```diff
@@ -312,8 +312,8 @@ class IteratorCellShell
     bool m_is2D;     //!< true if the cell list is 2D
 };
 
-bool compareFirstNeighborPairs(const std::vector<std::tuple<size_t, size_t, float>>& left,
-                               const std::vector<std::tuple<size_t, size_t, float>>& right);
+bool compareFirstNeighborPairs(const std::vector<std::tuple<size_t, size_t, float, float>>& left,
```
vyasr: Can we move this function to NeighborList.h? It doesn't really fit here.

cpp/locality/NeighborComputeFunctional.cc
```diff
@@ -0,0 +1,5 @@
+#include "NeighborComputeFunctional.h"
```
vyasr: Do we need this file?

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,101 @@
...
+#include "AABBQuery.h"
+#include "Index1D.h"
+#include "NeighborList.h"
+#include "NeighborQuery.h"
+
+namespace freud { namespace locality {
+
+// Body should be an object taking in
+// with operator(size_t begin, size_t end)
+template<typename Body> void for_loop_wrapper(bool parallel, size_t begin, size_t end, const Body& body)
```
vyasr: Could you change this to be a more formal docstring? Something to indicate the meaning of the other arguments in addition to body. You could follow [doxygen](http://www.doxygen.nl/manual/docblocks.html), but anything is fine for now.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,101 @@
...
+    {
+        // if nlist exists, loop over it in parallel.
+        loop_over_NeighborList_parallel(nlist, cf);
+    }
+    else
+    {
+        // if nlist does not exist, check if ref_points is an actual NeighborQuery
+        std::shared_ptr<NeighborQueryIterator> iter;
+        std::shared_ptr<AABBQuery> abq;
+        if (const RawPoints* rp = dynamic_cast<const RawPoints*>(ref_points))
```
vyasr: Perhaps add a comment to specify that this returns NULL, making the conditional false, when the cast fails. I suspect this syntax will be unfamiliar to most future freud developers.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,101 @@
...
+                                              ref_points->getNRef());
+            iter = abq.get()->queryWithArgs(points, Np, qargs);
+        }
+        else
+        {
+            iter = ref_points->queryWithArgs(points, Np, qargs);
+        }
+
+        // iterate over the query object in parallel
+        for_loop_wrapper(true, 0, Np, [iter, qargs, &cf](size_t begin, size_t end) {
```
vyasr: Can we capture `iter` by reference? Capturing `qargs` by reference should definitely be safe.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,101 @@
...
+            for (size_t i = begin; i != end; ++i)
+            {
+                std::shared_ptr<NeighborQueryIterator> it = iter->query(i);
+                np = it->next();
+                while (!it->end())
+                {
+                    if (!qargs.exclude_ii || i != np.ref_id)
+                    {
+                        // TODO
+                        // weight set to 1 for now
```
vyasr: I would add a note that changing this will happen when we bring Voronoi into the NeighborQuery infrastructure.

cpp/locality/NeighborComputeFunctional.h
vyasr: Same goes for the other functions in this file.

cpp/locality/NeighborQuery.h
```diff
@@ -257,7 +258,7 @@ class NeighborQueryIterator
      */
     virtual NeighborList* toNeighborList()
     {
-        typedef tbb::enumerable_thread_specific<std::vector<std::pair<size_t, size_t>>> BondVector;
```
vyasr, line 260: Can we add a `typedef <std::tuple<size_t, size_t, float> > Bond`? We need the tuple again later, and this will make it more readable.

vyasr: Also it will make it easier to change if we need to make weights a more first-class object.

cpp/locality/NeighborQuery.h
```diff
@@ -370,6 +375,34 @@ class NeighborQueryQueryIterator : virtual public NeighborQueryIterator
...
+        : NeighborQuery(box, ref_points, Nref)
+    {}
+
+    ~RawPoints() {}
+
+    // dummy implementation for pure virtual function in the parent class
+    virtual std::shared_ptr<NeighborQueryIterator> query(const vec3<float>* points, unsigned int N,
+                                                         unsigned int k, bool exclude_ii = false) const
+    {
+        return nullptr;
```
vyasr: I would have these query methods throw exceptions to make debugging easier if someone tries to use these.

cpp/util/DisjointSet.h
```diff
@@ -0,0 +1,162 @@
+#if !defined(__UNIONFIND_H)
```
vyasr: So we'll want to add the [entire repository](https://github.com/wjakob/dset) as a git submodule in the `extern/` directory. Our `setup.py` file should already add everything in there to the include path, so you shouldn't have to change any of the #includes. We need to make sure that the license is included, though, which will happen if you add the repository as a submodule.

freud/cluster.pyx
```diff
@@ -98,15 +98,18 @@ cdef class Cluster:
             box (:class:`freud.box.Box`, optional):
                 Simulation box (Default value = None).
         """
+        cdef freud.locality.NeighborQuery nq = \
+            freud.locality.make_default_nq(self.m_box, points)
+        cdef freud._locality.NeighborList * nlistptr \
+            = freud.locality.make_nlistptr(nlist)
+
+        points = nq.points
```
vyasr, line 109: Is this actually necessary? The NeighborQuery object's points should just be a pointer back to the same array anyway, right? 
In either case, I think the `convert_array` logic below should happen before you make the default object, and afterwards you can refer to `nq.points` if needed.

freud/locality.pyx
```diff
@@ -370,6 +370,10 @@ cdef class NeighborQuery:
         return NeighborQueryResult.init(
             self.nqptr, points, exclude_ii, r=r, k=0)
 
+    cdef freud._locality.NeighborQuery * get_ptr(self) nogil:
+        R"""Returns a pointer to the raw C++ object we are wrapping."""
+        pass
```
vyasr: Is this right?

vyasr: I see why you did this. Since you're doing a `dynamic_cast` in the underlying C++, does it work to just return `self.nqptr` here and have that get passed into the C++ functions? Or are there potential problems that arise if the C++ code tries to use the object without correctly casting it?

setup.py
```diff
@@ -340,14 +340,22 @@ def find_tbb(tbb_root=None, tbb_include=None, tbb_link=None):
...
 # Source files required for all modules.
-sources_in_all = []
+sources_in_all = [
+    os.path.join("cpp", "locality", "NeighborQuery.cc"),
+    os.path.join("cpp", "locality", "AABBQuery.cc"),
+    os.path.join("cpp", "locality", "LinkCell.cc"),
+    os.path.join("cpp", "locality", "NearestNeighbors.cc"),
+    os.path.join("cpp", "locality", "NeighborList.cc"),
+    os.path.join("cpp", "box", "Box.cc")
+]
```
vyasr, line 346: Do you actually need to add all of these back in? I would assume that at least LinkCell and NearestNeighbors could be removed, and I don't think Box.cc does anything (we should just remove it anyway).

cpp/locality/LinkCell.cc
```diff
@@ -128,8 +128,8 @@ const vec3<unsigned int> LinkCell::computeDimensions(const box::Box& box, float
     return dim;
 }
 
-bool compareFirstNeighborPairs(const std::vector<std::tuple<size_t, size_t, float>>& left,
-                               const std::vector<std::tuple<size_t, size_t, float>>& right)
+bool compareFirstNeighborPairs(const std::vector<std::tuple<size_t, size_t, float, float>>& left,
+                               const std::vector<std::tuple<size_t, size_t, float, float>>& right)
```
vyasr: I have a comment below about adding a `typedef` for bonds. It might actually be best to move that typedef to the `NeighborList`, and maybe have two separate typedefs, one with weights and one without (e.g. `typedef  std::tuple<size_t, size_t, float>` Bond; typedef std::tuple<size_t, size_t, float, float>` WeightedBond;`). I assume those will get unified at some point down the line.

vyasr (changes requested):
I've requested some changes regarding cleaning up the code, but the structure all looks right. Two bigger picture questions come to mind. First, I think we'd want to isolate as much of the switching logic to C++ as possible. So far this PR largely follows the original spec I had envisioned, and I don't know if we can reduce the logic on the Python side any further (I don't see any way around exporting the `RawPoints` class to Python and creating instances of it), but it would be worth spending a little time thinking about that to make sure we've avoided splitting the logic between Python and C++ as much as possible. The second issue that pops up in a few places is related to the NeighborList weights. Right now they're kind of an afterthought tacked on to support the current mode of operation of the Voronoi module. I had some comments below about ways to clean it up a little, but @bdice might have more ideas on how we should give it more first-class support. There may be other weighting schemes that make sense as well, so it's worth planning a little bit while we're revamping so much of the code that touches this logic.

cpp/locality/NeighborComputeFunctional.cc
jinsooihm: `Box` has an empty `cc` file so I thought we might need it. Is it not necessary?

cpp/locality/NeighborComputeFunctional.h
jinsooihm:
```
// check if ref_points is a pointer to a RawPoints object
// dynamic_cast will fail if ref_points is not actually pointing to RawPoints
// and return a null pointer. Then, the assignment operator will return
// a null pointer, making the condition in the if statement to be false.
// This is a typical C++ way of checking the type of a polymorphic class
// using pointers and casting.
```
Would this be good enough?

cpp/locality/NeighborQuery.h
jinsooihm: I actually did that when testing, and was wondering why it was not typedef-ed.

freud/cluster.pyx
jinsooihm: Should convert_array be done inside `make_deafault_nq` since it is there where we check for the type of `ref_points`?

freud/locality.pyx
jinsooihm: I did this to make it consistent with other classes that use `get_ptr()` to get the underlying pointer.

cpp/locality/NeighborComputeFunctional.h
bdice: @jinsoo960 The neighbors found by NeighborQuery classes currently include two indices and a distance. I would like to add a `float weight` field to these. For ball queries and k-nearest neighbor queries, the weights would always default to `1.0`.

cpp/locality/NeighborComputeFunctional.cc
bdice: @jinsoo960 Try removing it for now, I think the `Box.cc` may exist for other reasons that Cython demands. Not sure.

cpp/locality/NeighborComputeFunctional.h
bdice: Perfect.

freud/cluster.pyx
bdice: @jinsoo960 I recommend that we hold off on this until the convert array PR is merged -- it's better to avoid the merge conflicts and clean this up as we transition to NeighborQueries.

bdice: However, I agree that the data being passed into the NeighborQuery has to go through `convert_array` first - either before the `nq` is created or in the definition of `make_default_nq`.

cpp/cluster/Cluster.cc
```diff
@@ -16,114 +18,48 @@ using namespace std;
...
+                                                    if (dist < m_rcut)
+                                                    {
+                                                        // merge the two sets using the disjoint set
+                                                        if (!dj.same(i, j))
+                                                        {
+                                                            dj.unite(i, j);
+                                                        }
+                                                    }
+                                                }
+                                            });
```
vyasr: Is this the indentation requested by clang? I would prefer that everything in the body of the lambda for our `loopOverNeighbors` function be indented just one level up from the indentation level of the function call. This is generally our practice for `tbb::parallel_for`.

cpp/cluster/Cluster.cc
```diff
@@ -16,114 +18,48 @@ using namespace std;
...
-    }
+    DisjointSets dj(m_num_particles);
+
+    locality::QueryArgs qargs;
+    qargs.mode = locality::QueryArgs::QueryType::ball;
+    qargs.rmax = m_rcut;
+
+    freud::locality::loopOverNeighbors(nq, points, Np, qargs, nlist,
+                                            [this, &dj, &box, points](size_t i, size_t j, float dist, float weight) {
+                                                if (i != j)
```
vyasr: Since the changes in #328 are incorporated into this one, I think we should get rid of this `i != j` check here, right?

cpp/cluster/Cluster.h
```diff
@@ -66,8 +48,9 @@ class Cluster
     Cluster(float rcut);
 
     //! Compute the point clusters
-    void computeClusters(const box::Box& box, const freud::locality::NeighborList* nlist,
-                         const vec3<float>* points, unsigned int Np);
+    void computeClusters(const freud::locality::NeighborQuery* nq, const box::Box& box,
```
vyasr: Can we just use the `NeighborQuery`'s box and get rid of the `box` argument here?

cpp/density/CorrelationFunction.cc
```diff
@@ -9,7 +9,10 @@
 #include <emmintrin.h>
 #endif
 
+#include <iostream>
+
```
vyasr: Let's remove these now that debugging is done.

cpp/density/CorrelationFunction.cc
```diff
@@ -108,63 +111,38 @@ template<typename T> void CorrelationFunction<T>::reset()
 
 template<typename T>
 void CorrelationFunction<T>::accumulate(const box::Box& box, const freud::locality::NeighborList* nlist,
```
vyasr: Same as above, can we remove the box?

cpp/cluster/Cluster.cc
```diff
@@ -16,114 +18,48 @@ using namespace std;
...
-                    }
-                }
-            }
-        }
-    }
+    DisjointSets dj(m_num_particles);
+
+    locality::QueryArgs qargs;
+    qargs.mode = locality::QueryArgs::QueryType::ball;
+    qargs.rmax = m_rcut;
```
vyasr: This makes sense for now, but we should make a follow-up pull request that just sets up default query arguments for each compute class on the python side so that the C++ can always just operate directly on the query, rather than hardcoding them in C++.

cpp/density/CorrelationFunction.cc
```diff
@@ -108,63 +111,38 @@ template<typename T> void CorrelationFunction<T>::reset()
...
-        for (size_t i = r.begin(); i != r.end(); i++)
+    float dr_inv = 1.0f / m_dr;
+    freud::locality::loopOverNeighbors(nq, points, Np, qargs, nlist,
+    [=](size_t i, size_t j, float dist, float weight)
         {
-            // get the cell the point is in
-            vec3<float> ref = ref_points[i];
-            for (; bond < nlist->getNumBonds() && neighbor_list[2 * bond] == i; ++bond)
+            // check that the particle is not checking itself, if it is the same list
+            if(dist < m_rmax)
```
vyasr: Remove this logic.

cpp/density/CorrelationFunction.cc
```diff
@@ -108,63 +111,38 @@ template<typename T> void CorrelationFunction<T>::reset()
...
             {
-                const size_t j(neighbor_list[2 * bond + 1]);
+                // float r = sqrtf(rsq);
+                float r = dist;
```
vyasr: Remove the commented line, and we can also just start using `dist` in place of `r` to avoid creating an extra intermediate variable (or rename `dist` to `r` in the function signature).

cpp/cluster/Cluster.cc
vyasr: Actually, in this PR can we just construct the QueryArgs object in Python and pass it through? That will make the future change I proposed much easier.

vyasr:
We need to think a little bit about what should be stored by compute classes. In particular, I think we should just start depending on the box in the NeighborQuery object, but if you (@jinsoo960) think that's too big a change maybe we should do that in a separate pull request? What are your thoughts.

cpp/density/GaussianDensity.cc
```diff
@@ -27,8 +27,7 @@ GaussianDensity::GaussianDensity(unsigned int width, float r_cut, float sigma)
 
 GaussianDensity::GaussianDensity(unsigned int width_x, unsigned int width_y, unsigned int width_z,
                                  float r_cut, float sigma)
-    : m_box(box::Box()), m_width_x(width_x), m_width_y(width_y), m_width_z(width_z), m_rcut(r_cut),
-      m_sigma(sigma)
+    : m_box(box::Box()), m_width_x(width_x), m_width_y(width_y), m_width_z(width_z), m_rcut(r_cut)
```
vyasr: This change seems like a typo?

cpp/density/GaussianDensity.h
```diff
@@ -7,9 +7,11 @@
 #include <memory>
 
 #include "Box.h"
+#include "ETSWrapper.h"
 #include "Index1D.h"
 #include "ThreadStorage.h"
 #include "VectorMath.h"
+#include "NeighborComputeFunctional.h"
```
vyasr: Since GaussianDensity hasn't actually been changed to use either the ETSWrapper or the `loopOverNeighbors` function, we should remove these includes for now.

cpp/density/LocalDensity.cc
```diff
@@ -38,56 +39,55 @@ void LocalDensity::compute(const box::Box& box, const freud::locality::NeighborL
...
+                  num_neighbors += 1.0f;
+              }
+              else if (r < (m_rcut + m_diameter/2.0f))
+              {
+                  // partially count particles that intersect the rcut sphere
+                  // this is not particularly accurate for a single particle, but works well on average for
+                  // lots of them. It smooths out the neighbor count distributions and avoids noisy spikes
+                  // that obscure data
+                  num_neighbors += 1.0f + (m_rcut - (r + m_diameter/2.0f)) / m_diameter;
+              }
```
vyasr, line 77: This is fine for now, but we should probably resolve this logic by just adapting the query arguments on the python side to build a neighbor list with the slightly expanded rcut, then in C++ it would just be two conditional branches (`if (r_cut > m_crut - m_diameter/2)... else...`).

cpp/density/LocalDensity.cc
```diff
@@ -38,56 +39,55 @@ void LocalDensity::compute(const box::Box& box, const freud::locality::NeighborL
...
+          if (m_box.is2D())
+              {
+              // local density is area of particles divided by the area of the circle
+              m_density_array.get()[i] = (m_volume * m_num_neighbors_array.get()[i]) / (M_PI * m_rcut * m_rcut);
+              }
+          else
+              {
+              // local density is volume of particles divided by the volume of the sphere
+              m_density_array.get()[i] = (m_volume * m_num_neighbors_array.get()[i]) / (4.0f/3.0f * M_PI * m_rcut * m_rcut * m_rcut);
+              }
```
vyasr, line 90: A minor optimization would be to compute the area/volume outside the for loop, then inside the loop you just divide by whichever you get.

cpp/environment/BondOrder.cc
```diff
@@ -135,106 +136,172 @@ void BondOrder::reset()
...
-    // compute the order parameter
-    parallel_for(blocked_range<size_t>(0, n_ref), [=](const blocked_range<size_t>& br) {
-        float dt_inv = 1.0f / m_dt;
-        float dp_inv = 1.0f / m_dp;
-        Index2D sa_i = Index2D(m_nbins_t, m_nbins_p);
+    freud::locality::loopOverNeighbors(ref_points, points, n_p, qargs, nlist, 
+    [=] (size_t i, size_t j, float dist, float weight)
+    {
+        vec3<float> ref_pos = ref_points->getRefPoints()[i];
+        quat<float> ref_q(ref_orientations[i]);
```
vyasr: We could make these references to avoid any copying.

vyasr: Also `q` below.

cpp/environment/BondOrder.cc
```diff
@@ -135,106 +136,172 @@ void BondOrder::reset()
...
+        vec3<float> delta = m_box.wrap(points[j] - ref_pos);
```
vyasr: Just name this `v` and get rid of `v` below.

cpp/environment/BondOrder.cc
```diff
@@ -135,106 +136,172 @@ void BondOrder::reset()
...
+// #endif
+
+//                 // increment the bin
+//                 if ((ibint < m_nbins_t) && (ibinp < m_nbins_p))
+//                 {
+//                     ++m_local_bin_counts.local()[sa_i(ibint, ibinp)];
+//                 }
+//             }
+//         }
+//     });
```
vyasr: Delete this block.

vyasr:
Related, should we be storing the NeighborQuery object? Or perhaps we don't even really need to be storing the box? The fact that we have a box from the NeighborQuery object complicates some of this, we should think through the implications in terms of what information a class really needs. It might be that some logic should just be moved from constructors into `compute` calls, perhaps with conditionals to check if the box has changed.

cpp/environment/MatchEnv.cc
```diff
@@ -691,23 +691,20 @@ void MatchEnv::cluster(const freud::locality::NeighborList* env_nlist,
...
-                        unsigned int a = dj.find(i);
-                        unsigned int b = dj.find(j);
-                        if (a != b)
-                            dj.merge(i, j, vec_map, rotation);
-                    }
+                    // merge the two sets using the disjoint set
+                    unsigned int a = dj.find(i);
+                    unsigned int b = dj.find(j);
+                    if (a != b)
+                        dj.merge(i, j, vec_map, rotation);
```
vyasr, line 707: For future reference, we should probably change this to use the new parallel-safe disjoint set class as much as possible.

cpp/locality/NeighborList.h
```diff
@@ -88,8 +94,15 @@ class NeighborList
     std::shared_ptr<size_t> m_neighbors;
     //! Neighbor list per-bond weight array
     std::shared_ptr<float> m_weights;
+    //! Neighbor list per-bond distance array
+    std::shared_ptr<float> m_distances;
 };
 
+typedef std::tuple<size_t, size_t, float> Bond;
+typedef std::tuple<size_t, size_t, float, float> WeightedBond;
```
vyasr: Is there a reason not to just generalize the `NeighborPoint` structure in NeighborQuery.h to also contain weights?

cpp/locality/LinkCell.cc
```diff
@@ -181,7 +172,7 @@ void LinkCell::compute(const box::Box& box, const vec3<float>* ref_points, unsig
     // for quick access later (not ref_points)
     computeCellList(box, points, Np);
 
-    typedef std::vector<std::tuple<size_t, size_t, float>> BondVector;
+    typedef std::vector<WeightedBond> BondVector;
```
vyasr: I think we should change everything using the `WeightedBond`, see my comments on `NeighborList.h`. We can discuss there and make changes if you agree that it's appropriate.

cpp/locality/NearestNeighbors.cc
```diff
@@ -53,7 +53,7 @@ void NearestNeighbors::compute(const box::Box& box, const vec3<float>* ref_pos,
     m_box = box;
     m_neighbor_list.resize(num_ref * m_num_neighbors);
 
-    typedef std::vector<std::tuple<size_t, size_t, float>> BondVector;
+    typedef std::vector<WeightedBond> BondVector;
```
vyasr: Same comment as LinkCell regarding WeightedBond.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
+#ifndef NEIGHBOR_COMPUTE_FUNCTIONAL_H
+#define NEIGHBOR_COMPUTE_FUNCTIONAL_H
+
+#include <iostream>
```
vyasr: Remove.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+#include "Index1D.h"
+#include "NeighborList.h"
+#include "NeighborQuery.h"
+
+namespace freud { namespace locality {
+//! Wrapper for for-loop
+/*! \param parallel If true, run body in parallel.
+    \param begin Beginning index.
+    \param end Ending index.
+    \param body Body should be an object taking in
```
vyasr: Remove "taking in"

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+           with operator(size_t begin, size_t end).
+*/
+template<typename Body> void forLoopWrapper(bool parallel, size_t begin, size_t end, const Body& body)
```
vyasr: Can we change the signature order to have the `parallel` bool after `end`? That feels more natural.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+#include <memory>
+#include <tbb/tbb.h>
+
+#include "AABBQuery.h"
+#include "Index1D.h"
+#include "NeighborList.h"
+#include "NeighborQuery.h"
+
+namespace freud { namespace locality {
+//! Wrapper for for-loop
```
vyasr: Can you add a little more detail on what justifies this wrapper (e.g. "wrapper that simplifies generation of parallel code on demand").

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+
+//! Wrapper iterating looping over NeighborQuery or NeighborList
+/*! \param ref_points NeighborQuery object to iterate over
+    \param points Points
+    \param Np Number of points
+    \param qargs Query arguments
+    \param nlist Neighbor List. If not NULL, loop over it. Otherwise, use ref_points
+           appropriately with given qargs.
+    \param cf A void function that takes
+           (ref_point_index, point_index, distance, weight) as input.
```
vyasr: We should probably be consistent with the `forLoopWrapper` docstring; in either case, a function or a functor will work, right?

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+*/
+template<typename ComputePairType>
+void loopOverNeighbors(const NeighborQuery* ref_points, const vec3<float>* points, unsigned int Np,
+                       QueryArgs qargs, const NeighborList* nlist, const ComputePairType& cf,
+                       bool parallel = true)
```
vyasr, line 47: The argument order and the default value for parallelism being inconsistent with `forLoopWrapper` will be a little confusing to developers. I would expect both functions to have similar behavior. I assume your rationale was that at present there are multiple places `forLoopWrapper` is used that are not actually parallel-safe? I would prefer to make the two signatures align unless there's a compelling reason not to.

cpp/locality/NeighborComputeFunctional.h
vyasr: Actually, let's hold off on this change until other PR changes are more finalized. After looking at the other loop functions below, I'm not quite sure what the most natural choice of APIs is given the number of functions we could be dispatching to.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+    // dynamic_cast will fail if ref_points is not actually pointing to RawPoints
+    // and return a null pointer. Then, the assignment operator will return
+    // a null pointer, making the condition in the if statement to be false.
+    // This is a typical C++ way of checking the type of a polymorphic class
+    // using pointers and casting.
+    if (const RawPoints* rp = dynamic_cast<const RawPoints*>(ref_points))
+    {
+        // if ref_points is RawPoints, build a NeighborQuery
+        abq = std::make_shared<AABBQuery>(ref_points->getBox(), ref_points->getRefPoints(),
+                                          ref_points->getNRef());
```
vyasr, line 110: At some point we should probably benchmark to see if defaulting to AABBQuery or LinkCell is a better choice. Not urgent right now though, especially since the user is free to make a smart choice themselves.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+        iter = abq.get()->queryWithArgs(points, Np, qargs);
```
vyasr: You can directly use the arrow operator on shared pointers, no need to get the underlying pointer.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+            np = it->next();
+            while (!it->end())
+            {
+                //! Warning! If qargs.exclude_ii is true, NeighborPoint with same indices
+                // will not be considered regardless of ref_points and points
+                // being same set of points
+                if (!qargs.exclude_ii || i != np.ref_id)
+                {
+                    // TODO when Voronoi gets incorporated in NeighborQuery infrastructure
+                    // weight set to 1 for now
```
vyasr: This should probably read that weight will need to get passed through. But if (as per one of my previous comments) we update the `NeighborPoint` struct to also include weights, then this becomes simpler to deal with.

vyasr: Also, we might not be planning on making Voronoi a NeighborQuery object anymore, instead relying on the NeighborList side of logic to just have Voronoi make a NeighborList. In either case, though, I think that it makes sense to support weights within the NeighborPoint object.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+                    cf(np.ref_id, i, np.distance, 1);
+                }
+                np = it->next();
+            }
+        }
+    });
+}
+
+// This function does not work for now since ref_point point orders are different
```
vyasr, line 140: I haven't really reviewed the next three functions since I understand the limitation that's holding it up, but can you summarize what the eventual purpose of these functions will be? Will they replace the looping functions above?

cpp/locality/NeighborList.h
vyasr: Related: renaming that to `NeighborBond` or so probably makes sense. I'm sure I had a reason for calling it `NeighborPoint`, but it really represents a bond.

cpp/locality/NeighborQuery.h
```diff
@@ -291,12 +292,13 @@ class NeighborQueryIterator
...
         size_t* neighbor_array(nl->getNeighbors());
         float* neighbor_weights(nl->getWeights());
+        float* neighbor_distance(nl->getDistances());
 
         parallel_for(tbb::blocked_range<size_t>(0, num_bonds), [&](const tbb::blocked_range<size_t>& r) {
             for (size_t bond(r.begin()); bond < r.end(); ++bond)
             {
-                neighbor_array[2 * bond] = linear_bonds[bond].first;
-                neighbor_array[2 * bond + 1] = linear_bonds[bond].second;
+                std::tie(neighbor_array[2 * bond], neighbor_array[2 * bond + 1], neighbor_distance[bond])
```
vyasr: This line becomes a little more verbose if the bond is a struct instead of a tuple since it needs to be split, but it will be equally fast and probably more readable than using `std::tie`.

cpp/order/SolLiq.cc
```diff
@@ -7,6 +7,7 @@
 #include <tbb/tbb.h>
```
vyasr, line 7: As far as I can tell, the only changes you've made to this file are switching the DisjointSet class being used and removing all of the `exclude_ii` logic. The parallelization of parts of this file could cause merge conflicts, so beware of that when finally merging `next` back into this branch. They shouldn't actually be conflicts, but git may have some trouble finding them.

cpp/order/SolLiq.h
```diff
@@ -17,6 +17,7 @@
 #include "Cluster.h"
 #include "ThreadStorage.h"
 #include "VectorMath.h"
+#include "NeighborList.h"
```
vyasr, line 19: I don't think you actually need this, do you?

cpp/order/SolLiq.h
```diff
@@ -17,6 +17,7 @@
 #include "Cluster.h"
```
vyasr: Do we still need cluster, now that we're using a different `DisjointSet`? I assume we just need to include the new class (once it's been moved to `extern`).

cpp/util/DisjointSet.h
vyasr: @jinsoo960 it looks like this still needs to be done, let me know if you need help with it.

cpp/util/ETSWrapper.h
```diff
@@ -0,0 +1,106 @@
+#ifndef ETSWRAPPER_H
```
vyasr: I've merged `master` into `next`, you'll need to merge in `next` for this file to disappear (since it's not actually new in this PR).

cpp/util/NdHistogram.h
```diff
@@ -5,7 +5,9 @@
 #include <tbb/tbb.h>
```
vyasr: Same comment as with ETSWrapper.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -0,0 +1,262 @@
...
+    // if nlist does not exist, check if ref_points is an actual NeighborQuery
+    std::shared_ptr<NeighborQueryIterator> iter;
+    std::shared_ptr<AABBQuery> abq;
+    // check if ref_points is a pointer to a RawPoints object
+    // dynamic_cast will fail if ref_points is not actually pointing to RawPoints
+    // and return a null pointer. Then, the assignment operator will return
+    // a null pointer, making the condition in the if statement to be false.
+    // This is a typical C++ way of checking the type of a polymorphic class
+    // using pointers and casting.
+    if (const RawPoints* rp = dynamic_cast<const RawPoints*>(ref_points))
```
vyasr, line 106: We can probably avoid having this in C++ entirely by being a bit smarter about combining the `make_default_[nlist|nq]` logic in Python. If we combine that into one function that does (with the proper type definitions, etc):
```
if nlist is None:
    nlist = NULL
    if isinstance(ref_points, np.ndarray):
        ref_points = RawPoints(ref_points)
elif isinstance(ref_points, np.ndarray):
    ref_points = AABBQuery(ref_points)
```

That should simplify logic in both Python and C++, but let me know if I missed something. Also, if you prefer we can wait to make that change in a future PR.

freud/cluster.pyx
vyasr: The `convert_array` logic is all merged, so we should be able to move forward on this.

freud/density.pyx
```diff
@@ -132,20 +143,20 @@ cdef class FloatCF(Compute):
         else:
             l_values = values
 
-        defaulted_nlist = freud.locality.make_default_nlist(
-            b, ref_points, points, self.rmax, nlist, None)
-        cdef freud.locality.NeighborList nlist_ = defaulted_nlist[0]
+        # defaulted_nlist = freud.locality.make_default_nlist(
+        #     b, ref_points, points, self.rmax, nlist, exclude_ii)
+        # cdef freud.locality.NeighborList nlist_ = defaulted_nlist[0]
```
vyasr: Remove.

freud/density.pyx
```diff
@@ -350,21 +374,21 @@ cdef class ComplexCF(Compute):
...
```
vyasr: Remove.

freud/density.pyx
```diff
@@ -713,8 +737,8 @@ cdef class LocalDensity(Compute):
         # local density of each particle includes itself (cutoff
         # distance is r_cut + diam/2 because of smoothing)
         defaulted_nlist = freud.locality.make_default_nlist(
-            b, ref_points, points, self.r_cut + 0.5*self.diameter, nlist,
-            exclude_ii=False)
+            b, ref_points, points,
+            self.r_cut + 0.5*self.diameter, nlist, False)
```
vyasr, line 733: I believe I commented in the C++ that we could do this, so looks like we're thinking the same thing. Since you've already made this change, I would recommend removing the additional check that's on the C++ side for points beyond the far cutoff (unless I missed something).

freud/density.pyx
```diff
@@ -836,27 +860,34 @@ cdef class RDF(Compute):
...
         cdef const float[:, ::1] l_points = points
-        cdef unsigned int n_ref = l_ref_points.shape[0]
         cdef unsigned int n_p = l_points.shape[0]
 
-        defaulted_nlist = freud.locality.make_default_nlist(
-            b, ref_points, points, self.rmax, nlist)
-        cdef freud.locality.NeighborList nlist_ = defaulted_nlist[0]
+        # defaulted_nlist = freud.locality.make_default_nlist(
+        #     b, ref_points, points, self.rmax, nlist, exclude_ii)
+        # cdef freud.locality.NeighborList nlist_ = defaulted_nlist[0]
```
vyasr: Remove.

freud/environment.pyx
```diff
@@ -197,27 +208,25 @@ cdef class BondOrder(Compute):
                 ('Unknown BOD mode: {}. Options are:'
                     'bod, lbod, obcd, oocd.').format(mode))
 
-        defaulted_nlist = freud.locality.make_default_nlist_nn(
-            b, ref_points, points, self.num_neigh, nlist, None, self.rmax)
-        cdef freud.locality.NeighborList nlist_ = defaulted_nlist[0]
+        # defaulted_nlist = freud.locality.make_default_nlist_nn(
+        #     b, ref_points, points, self.num_neigh,
+        #     nlist, exclude_ii, self.rmax)
+        # cdef freud.locality.NeighborList nlist_ = defaulted_nlist[0]
```
vyasr: Remove.

freud/locality.pyx
vyasr: Right, I guess what I'm asking is whether you could just have this function return self.nqptr and get rid of the corresponding functions in the subclasses. Will the dynamic cast get the right type of object? I would expect that Cython generates the right C++ code for this, but can't be 100% sure.

freud/locality.pyx
```diff
@@ -782,6 +810,45 @@ def make_default_nlist_nn(box, ref_points, points, n_neigh, nlist=None,
...
+
+    .. versionadded:: 1.1.0
+
+    """  # noqa: E501
+
+    def __cinit__(self, box, points):
+        cdef const float[:, ::1] l_points
+        if type(self) is RawPoints:
+            # Assume valid set of arguments is passed
+            self.queryable = True
```
vyasr: I don't think this attribute is necessary, I only added to LinkCell because it still supports the old API.

tests/test_density_ComplexCF.py
```diff
@@ -83,10 +83,10 @@ def test_random_points(self):
         absolute_tolerance = 0.1
         # first bin is bad
         ocf.accumulate(freud.box.Box.square(box_size), points, comp,
-                       points, np.conj(comp))
+                       None, np.conj(comp))
```
vyasr: Maybe we should change the argument order so that users don't have to explicitly pass `None` in order to get the expected `ref_points` == `points` behavior? I guess it's complicated by the fact that for ComplexCF the typical use is passing the conjugate values in for the same points. @bdice what do you think make sense as defaults here?

tests/test_density_FloatCF.py
```diff
@@ -82,10 +82,10 @@ def test_random_points(self):
         # first bin is bad
         ocf.accumulate(box, points, ang)
         npt.assert_allclose(ocf.RDF, correct, atol=absolute_tolerance)
-        ocf.compute(box, points, ang, points, ang)
+        ocf.compute(box, points, ang, None, ang)
```
vyasr: The FloatCF case is probably easier to solve than the ComplexCF case, so maybe we should start thinking about this here. I think that, at minimum, users should be able to skip passing in both `points` and `values` and get an autocorrelation out. I'm OK with users having to pass in points a second time in order to do a correlation function with the same spatial points but with different values, but it does raise questions about self-exclusion. I'm inclined to just make the docs clear that if `points == ref_points` but `ref_values != values`, you need to explicitly pass `exclude_ii=True` to get the behavior that you probably want; would you agree @jinsoo960 and @bdice? As long as the docs are very clear, I think that's the best API choice. Probably same goes for ComplexCF?

tests/test_environment_AngularSeparation.py
```diff
@@ -70,7 +70,7 @@ def test_getNP(self):
         with self.assertRaises(AttributeError):
             ang.n_ref
 
-        ang.computeNeighbor(box, ors, ors, points, points, equiv_quats)
+        ang.computeNeighbor(box, ors, ors, points, None, equiv_quats)
```
vyasr: Since you changed the underlying function to have a default value for `equiv_quats`, just remove the last two arguments everywhere in this file.

tests/test_density_FloatCF.py
vyasr: The user could also pass `values` as a keyword argument, bypassing the problem and allowing freud to correctly autodetect `exclude_ii = True`.

tests/test_environment_LocalBondProjection.py
```diff
@@ -59,7 +59,7 @@ def test_num_particles(self):
         proj_vecs = np.asarray([[0, 0, 1]])
 
         ang = freud.environment.LocalBondProjection(rmax, num_neigh)
-        ang.compute(box, proj_vecs, points, ors, points, equiv_quats)
+        ang.compute(box, proj_vecs, points, ors, None, equiv_quats)
```
vyasr: I would make the same change as for `AngularSeparation`: give `equiv_quats` a default value of `(1, 0, 0, 0)` and simplify all of these calls.

tests/test_pmft.py
```diff
@@ -747,7 +747,7 @@ def test_shift_two_particles_dead_pixel(self):
 
         for pm in [noshift, shift]:
             pm.compute(freud.box.Box.cube(3), points, orientations,
-                       points, orientations, face_orientations=None)
+                       face_orientations=None)
```
vyasr: I think you can remove the last argument, it's `None` by default.

vyasr (changes requested):
@jinsoo960 thanks again for the great work here! Sorry it's taken me so long to review. I had a lot of comments, but most of them should be pretty quick to address. Once you've gone through the obvious, feel free to ping me with any questions on the others. I'll be sure to respond quickly to keep the process moving along.

jinsooihm:
> We need to think a little bit about what should be stored by compute classes. In particular, I think we should just start depending on the box in the NeighborQuery object, but if you (@jinsoo960) think that's too big a change maybe we should do that in a separate pull request? What are your thoughts.

I think removing box from the function signature will be nicer since it would look cleaner. It would not be too much change.

jinsooihm:
> Related, should we be storing the NeighborQuery object? Or perhaps we don't even really need to be storing the box? The fact that we have a box from the NeighborQuery object complicates some of this, we should think through the implications in terms of what information a class really needs. It might be that some logic should just be moved from constructors into `compute` calls, perhaps with conditionals to check if the box has changed.

I think we will need to do another survey of the classes to see whether some things are truly needed in constructor (perhaps it makes sense to initialize many things in the constructor if the class has `accumulate`), or just as an initialization call inside `compute`.

cpp/cluster/Cluster.cc
jinsooihm: This was the behavior before I think before adding `dist` and `weight` as parameters to the lambda, but now it gives a result that makes sense.

jinsooihm: I have not run clang-format for while, but I think it would be better to do it after everything is done.

cpp/cluster/Cluster.cc
jinsooihm: Yeah, I forgot to change it. Thanks.

cpp/cluster/Cluster.cc
jinsooihm: Seems like I forgot to change things in Cluster to match what I am doing in other places. In most other methods, I think I am building QueryArgs in python.

cpp/density/GaussianDensity.cc
jinsooihm: I have no idea how this happened.. (and seems like the tests are not catching this).

cpp/locality/NeighborComputeFunctional.h
jinsooihm: Anything that has `operator()` defined for `size_t, size_t, float, float` will in theory "work".

cpp/locality/NeighborComputeFunctional.h
jinsooihm: It is inconsistent mainly because I added the parallel option here at the last stage, and did not want to go and change everything...

cpp/locality/NeighborComputeFunctional.h
jinsooihm: Maybe we can write on the documentation some recommendations like 
"if the distribution of your particle is dense then use ..."

cpp/locality/NeighborComputeFunctional.h
jinsooihm: These are for some classes that have normalizing or some other preprocess and postprocess logic for each `ref_point`.

cpp/order/SolLiq.h
jinsooihm: I think this is necessary?

cpp/order/SolLiq.h
jinsooihm: It is not necessary anymore.

cpp/util/NdHistogram.h
jinsooihm: What do you mean?

cpp/locality/NeighborComputeFunctional.h
jinsooihm: The reason that I made them into separate functions is because of Cython typing issues, so I should figure out how this will work in a single function perhaps returning a tuple. 
Are you saying that we want to impose a restriction that if someone were to call `loopOverNeighborQuery`, the person should call with actual query object, not `RawPoints`?

cpp/locality/NeighborList.h
jinsooihm: Would it be better to get rid of the tuples `Bond` and `WeightedBond` and instead use the `NeighborBond` struct and define a comparator for it?

bdice: @jinsoo960 Yes!

vyasr:
How would you suggest removing the box entirely? Would it be an optional parameter if users don't pass in a NeighborQuery, but just a numpy array? Or would you just require a NeighborQuery object?

One possibility I was considering was that we could change the Python API to take `ref_points` as either a NeighborQuery object or a tuple `(freud.box.Box, np.ndarray)`. That way users would only need to provide a box if not providing a NeighborQuery object. I'm undecided as to whether that's clunky, I might need to try it out to convince myself for sure one way or another. But I feel like users could probably get used to that.

vyasr:
@jinsoo960 please re-request review when this PR is ready for another round of review, just to make sure I don't miss it.

jinsooihm:
> How would you suggest removing the box entirely? Would it be an optional parameter if users don't pass in a NeighborQuery, but just a numpy array? Or would you just require a NeighborQuery object?
> 
> One possibility I was considering was that we could change the Python API to take `ref_points` as either a NeighborQuery object or a tuple `(freud.box.Box, np.ndarray)`. That way users would only need to provide a box if not providing a NeighborQuery object. I'm undecided as to whether that's clunky, I might need to try it out to convince myself for sure one way or another. But I feel like users could probably get used to that.

I was saying that `box` can be removed from the input to C++ since we are putting a `NeighborQuery` object. 
For the user side, I think it might be awkward to use the "tuple or NeighborQuery" method for something like `GaussianDensity` since it only takes in `box` and `points` and does not do computation based on neighbors.

vyasr:
I agree that it could be odd for something like `GaussianDensity`. I would only do this for something that actually does neighbor finding. Such methods would be identified by also taking a `query_args` `dict` as an argument. I think we might also want to differentiate terminology: for example, `ref_points` and `points` should be restricted to things that require neighbor finding, and other classes would use `positions`. I don't really like that latter name (maybe we go with `tree` and `points` for neighbor-based classes, then reuse `points` for non-querying computes), but I think that kind of differentiation is necessary and would help alleviate further confusion.

tests/test_density_FloatCF.py
jinsooihm: What would be the default behavior the users expect when the user wants to use same set of points for `points` and `ref_points` but want to use different values?

freud/cluster.pyx
vyasr: Given our decision to be much stricter on when the automatic exclude_ii is done, I think this change is no longer necessary in any case. Further changes to the `make_default_*` functions are now done in the downstream PR, right?

freud/cluster.pyx
```diff
@@ -100,11 +100,18 @@ cdef class Cluster(Compute):
...
+            mode="ball", rmax=self.rmax, exclude_ii=True)
+        points = nq.points
 
-        defaulted_nlist = freud.locality.make_default_nlist(
-            self.m_box, points, points, self.rmax, nlist, True)
-        cdef freud.locality.NeighborList nlist_ = defaulted_nlist[0]
+        points = freud.common.convert_array(points, (None, 3))
+        if points.shape[1] != 3:
+            raise RuntimeError(
+                'Need a list of 3D points for computeClusters()')
```
vyasr: We can remove this extra `if` statement, right?

freud/density.pyx
```diff
@@ -108,10 +108,26 @@ cdef class FloatCF(Compute):
                 NeighborList to use to find bonds (Default value =
                 :code:`None`).
         """
+        exclude_ii = points is None
+
+        if (points is None) != (values is None):
+            raise ValueError("Either both points and values have to be"
+                             "provided or both points and values have to"
+                             "be None")
```
vyasr: Actually, I think users should be able to pass in values as a kwarg to take advantage of automatic `exclude_ii`; the check should only be enforcing that they don't pass in a second set of points without values, which I think is more nonsensical.

freud/density.pyx
```diff
@@ -326,10 +339,26 @@ cdef class ComplexCF(Compute):
...
```
vyasr: Same as previous comment in FloatCF.

freud/environment.pyx
```diff
@@ -671,33 +676,34 @@ cdef class MatchEnv(Compute):
...
-        R"""Test if the motif provided by refPoints1 is similar to the motif
-        provided by refPoints2.
+    def isSimilar(self, ref_points1, ref_points2,
+                  threshold, registration=False):
+        R"""Test if the motif provided by ref_points1 is similar to the motif
+        provided by ref_points2.
 
         Args:
-            refPoints1 ((:math:`N_{particles}`, 3) :class:`numpy.ndarray`):
+            ref_points1 ((:math:`N_{particles}`, 3) :class:`numpy.ndarray`):
```
vyasr: To bring this in line with the rest of freud, can we just rename these to points and ref_points, rather than ref_points[1|2]?

tests/test_density_ComplexCF.py
```diff
@@ -83,10 +83,10 @@ def test_random_points(self):
         absolute_tolerance = 0.1
         # first bin is bad
         ocf.accumulate(freud.box.Box.square(box_size), points, comp,
-                       points, np.conj(comp))
+                       points, np.conj(comp), qargs={"exclude_ii": True})
```
vyasr, line 86: Given my other suggested change, perhaps it's also worth testing that the `(box, points, comp, values=np.conj(comp))` API also gives the same result.

tests/test_density_FloatCF.py
```diff
@@ -82,10 +82,11 @@ def test_random_points(self):
         # first bin is bad
         ocf.accumulate(box, points, ang)
         npt.assert_allclose(ocf.RDF, correct, atol=absolute_tolerance)
-        ocf.compute(box, points, ang, points, ang)
+        ocf.compute(box, points, ang)
         npt.assert_allclose(ocf.RDF, correct, atol=absolute_tolerance)
         ocf.reset()
-        ocf.accumulate(box, points, ang, points, ang)
+        ocf.accumulate(box, points, ang, points,
+                       ang, qargs={'exclude_ii': True})
```
vyasr, line 89: Same comment as ComplexCF test regarding passing `values` as a kwarg.

vyasr (approved):
@jinsoo960 there are a few final comments, but they're mostly cosmetic. I also made some typo fixes where it was easy to do through the GitHub API. Once those are done, I'm happy to merge this. 

@bdice since this is such a fundamental rewrite in some ways it's probably worth your time to review as well, but I know you're crunched for time so if you don't think you have time just let me know and we can go ahead and move forward. I've gone through three pretty thorough rounds of review, so I'm reasonably confident that this PR is in good shape.

freud/cluster.pyx
jinsooihm: Since `points` comes from `nq.points`, it is guaranteed that `convert_array` was called on it with dimension `(None, 3)`.
So, I removed unnecessary `convert_array` from all places.

vyasr: Got it.

cpp/util/DisjointSet.h
vyasr: This is done now.

PR #296, opened 2019-05-31
Added clang-format check in circleci

jinsooihm:
## Description

Circleci runs style check for C++ files also using `clang-format`.
If a file in `cpp` directory does not meet the style specified by `.clang-format` file, circleci fails and will not run any further functionality tests.
Also, `flake8` check is now separated from functionality check.
Due to some difficulty and version issues of `clang-format`, the style check is done in OS X image.

## Motivation and Context

Style enforcement is now done on C++ codes also.
The style is not exactly same as current convention, but very close to it.

## Types of changes

## Checklist:


bdice:
@jinsoo960 I would prefer to find another option instead of using OS X images if possible (e.g. installing clang format into a Linux distro and making a docker image, editing our existing docker images, etc.) for a few reasons.
1. We are limited to one OS X build concurrently.
2. I believe our total OS X minutes per month are limited.
3. The latency of launching a style check should be minimal, and I would guess the OS X images are less optimized.
4. The style check can’t be run locally with the CircleCI infrastructure if it uses the OS X image.

jinsooihm:
@bdice We can use `clang-format-4.0` on Ubuntu 16.04 image, but some of the options for `clang-format` are not available in that version.

bdice:
@jinsoo960 We can install recent versions of clang from a separate source, like https://apt.llvm.org/ for Ubuntu, and make that part of our Docker image build script. @vyasr manages those images right now, and it would be helpful to update one of those images to use Python 3.7 anyway.

jinsooihm:
@bdice I believe the following script should do it.
```
apt-get update
apt-get -y install wget
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key| apt-key add - # Fingerprint: 6084 F3CF 814B 57C1 CF12 EFD5 15CF 4D18 AF4F 7421
apt-get -y install software-properties-common
apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main"
apt-get update
apt-get -y install clang-format-8
```
I tried on a local docker image of `vramasub/freud_py35_ubuntu` as root.

bdice:
@jinsoo960 If clang-format 6 is new enough, then I would change to the `ubuntu:18.04` image (which should happen anyway) and install that from the default source.

jinsooihm:
@bdice I checked and `clang-format-6.0` has all the necessary options.

bdice (approved):
Thanks @jinsoo960!

vyasr (approved):
Made a couple small changes, it's ready.

Issue #283, opened 2019-05-20
Error in parallel.setNumThreads


jinsooihm:
**Describe the bug**
`freud.parallel.setNumThreads` does not seem to change the number of threads correctly in the following situation.

**To Reproduce**
```
#script1.py
import freud
import numpy as np
import time

def main():
    rcut = 1.0
    L = 10
    N = 20000

    fbox = freud.box.Box.cube(L)
    seed = 0
    np.random.seed(seed)
    points = np.random.uniform(-L/2, L/2, (N, 3))

    freud.parallel.setNumThreads(0)
    start = time.time()
    lc = freud.locality.LinkCell(fbox, rcut)
    lc.compute(fbox, points, points, exclude_ii=True)
    end = time.time()
    print(end - start)

    freud.parallel.setNumThreads(5)
    start = time.time()
    lc = freud.locality.LinkCell(fbox, rcut)
    lc.compute(fbox, points, points, exclude_ii=True)
    end = time.time()
    print(end - start)

    freud.parallel.setNumThreads(3)
    start = time.time()
    lc = freud.locality.LinkCell(fbox, rcut)
    lc.compute(fbox, points, points, exclude_ii=True)
    end = time.time()
    print(end - start)

    freud.parallel.setNumThreads(1)
    start = time.time()
    lc = freud.locality.LinkCell(fbox, rcut)
    lc.compute(fbox, points, points, exclude_ii=True)
    end = time.time()
    print(end - start)

if __name__=="__main__":
    main()
```

```
#script2.py
import freud
import numpy as np
import time

def main():
    rcut = 1.0
    L = 10
    N = 20000

    fbox = freud.box.Box.cube(L)
    seed = 0
    np.random.seed(seed)
    points = np.random.uniform(-L/2, L/2, (N, 3))

    start = time.time()
    lc = freud.locality.LinkCell(fbox, rcut)
    lc.compute(fbox, points, points, exclude_ii=True)
    end = time.time()
    print(end - start)

    freud.parallel.setNumThreads(5)
    start = time.time()
    lc = freud.locality.LinkCell(fbox, rcut)
    lc.compute(fbox, points, points, exclude_ii=True)
    end = time.time()
    print(end - start)

    freud.parallel.setNumThreads(3)
    start = time.time()
    lc = freud.locality.LinkCell(fbox, rcut)
    lc.compute(fbox, points, points, exclude_ii=True)
    end = time.time()
    print(end - start)

[...]

PR #298, opened 2019-06-03
Always call setNumThreads at import time

bdice:
## Description
We must call `setNumThreads` at import time, to ensure that the thread count can be altered at runtime.

I added a function `getNumThreads`. I also removed the code that is specific to setting 1 thread for nyx/flux.

## Motivation
Currently parallelism isn't working as expected in freud. See #283. 

## How Has This Been Tested?
@jinsoo960 will test this. 😄
- [x] Write a test that ensures get/setNumThreads work as expected.

## Types of changes

## Checklist:

### DIFF
--- a/freud/parallel.pyx
+++ b/freud/parallel.pyx
@@ -4,16 +4,12 @@
 R"""
 The :class:`freud.parallel` module controls the parallelization behavior of
 freud, determining how many threads the TBB-enabled parts of freud will use.
-By default, freud tries to use all available threads for parallelization unless
-directed otherwise.
+freud uses all available threads for parallelization unless directed otherwise.
 """
 
 cimport freud._parallel
 
-# Override TBB's default autoselection. This is necessary because once the
-# automatic selection runs, the user cannot change it.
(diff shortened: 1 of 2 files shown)


vyasr:
Looks good to me. I'll hold off doing anything until Jin Soo adds tests.

jinsooihm:
It seems to behave correctly now, but how should this be formerly (quantitatively) tested?
#283 was kind of a qualitative observation.

bdice:
@jinsoo960 Just write tests that ensure `set` and `get` behave as expected. Just one test that does something like:

```python
# Initially zero
self.assertEqual(get(), 0)

# Make sure set and get work
set(1)
self.assertEqual(get(), 1)

# Make sure you can change it
set(2)
self.assertEqual(get(), 2)

# Test the context manager
with NumThreads(3):
    self.assertEqual(get(), 3)
self.assertEqual(get(), 2)
```

jinsooihm:
@bdice Tests are added.

vyasr (approved):

PR #299, opened 2019-06-04
Fix wheel building process

vyasr:
Use correct Python binary in Mac and update pip to support progress bars. Force linux wheels to clone over http instead of ssh (default changed in CircleCI v2). Remove unnecessary requirement six


bdice (approved):
I applied some minor edits - looks good to me!

PR #300, opened 2019-06-04
Reduced code duplication in PMFT and RDF

jinsooihm:
## Description

Added `accumulateGeneral`, `resetGeneral` and other various convenience functions in `NdHistogram` class to capture repeated logic in other `PMFT` subclasses and `RDF`.
One caveat is that the structure of `NdHistogram` is very specific to `PMFT` and `RDF` that it would not really help someone who is trying to make a new histogram class. 
Also, it might actually just make more sense to make it as a friend class that just provides useful functionality instead of a parent.

## Motivation and Context

Modules that are histogram based have a lot of repeated logic.
`CorrelationFunction` can be included in the framework, if we make `NdHistogram` to be templated, and perhaps make #295 more generalized to use in `accumulate`. 
Reduction is sufficiently different that it seems it should be implemented separately.

Possible ideas.
1. CorrelationFunction has two different arrays that it computes in parallel, while all others have one array. We can make the initialization check for each threads to be a member function or lambda. (Does that work correctly with `tbb::enumerable_thread_specific.local()`?)
2. Make #295 be more general, i.e. able to do some checks between `parallel_for` and actual loop between neighbors and use that in `accumulate`. But this might make things just a bit more convoluted.
    

## How Has This Been Tested?

Passes all the current tests.

## Types of changes

## Checklist:

   (Not `clang-format`ed yet)


cpp/pmft/PMFTXY2D.cc
```diff
@@ -41,177 +41,91 @@ PMFTXY2D::PMFTXY2D(float x_max, float y_max, unsigned int n_x, unsigned int n_y)
...
+    Index2D b_i = Index2D(m_n_x, m_n_y);
+
+    accumulateGeneral(box, n_ref, nlist, n_p, m_n_x * m_n_y, [=](size_t i, size_t j) {
+        vec3<float> ref = ref_points[i];
+        vec3<float> delta = this->m_box.wrap(points[j] - ref);
+        float rsq = dot(delta, delta);
+
+        // check that the particle is not checking itself
+        // 1e-6 is an arbitrary value that could be set differently if needed
+        if (rsq < 1e-6)
```
bdice, line 101: Oh gosh. This needs to be handled differently with `exclude_ii`-type logic.

cpp/pmft/PMFTXYT.cc
```diff
@@ -47,189 +46,93 @@ PMFTXYT::PMFTXYT(float x_max, float y_max, unsigned int n_x, unsigned int n_y, u
...
-        [=] (const blocked_range<size_t>& r)
-            {
-            for (size_t i = r.begin(); i != r.end(); i++)
-                {
-                m_pcf_array.get()[i] = (float)m_bin_counts.get()[i] * norm_factor * inv_jacobian * inv_num_dens;
-                }
-            });
-    }
+{
+    float jocob_factor = (float) 1.0 / m_jacobian;
```
bdice: Rename this: `jacobian_factor`

cpp/util/NdHistogram.h
```diff
@@ -0,0 +1,121 @@
...
+#include <memory>
+#include <tbb/tbb.h>
+
+#include "Box.h"
+#include "NeighborList.h"
+
+namespace freud { namespace util {
+
+template<typename T>
+std::shared_ptr<T> returnEmptyArray(unsigned int size)
```
bdice: Rename to `makeEmptyArray`?

cpp/util/NdHistogram.h
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    protected:
+        box::Box m_box;
+        unsigned int m_frame_counter; //!< Number of frames calculated
+        unsigned int m_n_ref;         //!< The number of reference points
+        unsigned int m_n_p;           //!< The number of points
+        bool m_reduce;                //!< Whether or not the PCF needs to be reduced
+
+        std::shared_ptr<float> m_pcf_array;         //!< Array of PCF/RDF computed
```
bdice: Specify that PCF stands for Pair Correlation Function in the comment. `//!< Array of computed pair correlation function`

cpp/density/RDF.cc
```diff
@@ -28,262 +27,187 @@ RDF::RDF(float rmax, float dr, float rmin)
...
 //! CumulativeCount class to perform a parallel reduce to get the cumulative count for each histogram bin
 class CumulativeCount
+{
+private:
+    float m_sum;
+    float* m_N_r_array;
+    float* m_avg_counts;
+
+public:
+    CumulativeCount(float* N_r_array, float* avg_counts)
```
bdice: This entire `CumulativeCount` class is a holdover from TBB before lambdas. This should definitely be rewritten.
https://www.threadingbuildingblocks.org/docs/help/reference/algorithms/parallel_scan_func.html

bdice (commented):
Some comments below. More review is needed.

cpp/pmft/PMFTXY2D.cc
jinsooihm: Is it guaranteed that if the indices of two points are same, then the two points are same points?

cpp/density/RDF.cc
jinsooihm: For some reason, it seems that tbb on Ubuntu cannot find the template for
```
template<typename Range, typename Value, typename Scan, typename Combine>
Value parallel_scan( const Range& range, const Value& identity,
                     const Scan& scan, const Combine& combine
                     [, partitioner] );
```
The implementation using this version fails to compile on Ubuntu containers on circleci, but compiles on my computer (Mac) and arch container.

jinsooihm: The reason is that this `parallel_scan` was added for the 2018 update 1 for `tbb` but the default on Ubuntu is using the 2017 version.

benchmarks/benchmark_PMFT_PMFTR12.py
```diff
@@ -24,6 +24,7 @@ def bench_setup(self, N):
 
     def bench_run(self, N):
         self.pmft.compute(self.box, self.points, self.orientations)
+        self.pmft.bin_counts
```
vyasr, line 27: We can remove this, right? The computation isn't lazy so we don't need to benchmark the access

vyasr: Remove below as well, unless I'm missing something.

cpp/density/RDF.cc
```diff
@@ -28,262 +27,192 @@ RDF::RDF(float rmax, float dr, float rmin)
...
-            m_sum = b.m_sum;
-            }
-    };
-
+        m_sum = temp;
+    }
+    CumulativeCount( CumulativeCount& b, split )
+        : m_sum(0), m_N_r_array(b.m_N_r_array), m_avg_counts(b.m_avg_counts)
+    {
+    }
```
vyasr, line 92: Does our clang style suggest always splitting this line? Just wondering.

vyasr (changes requested):
@jinsoo960 could you fix merge conflicts and merge in master, then I'll review again? I'm having a bit of a hard time separating out which lines are actually changed and which are due to clang-formatting, and I'm worried that I'll miss something important as a result.

benchmarks/benchmark_PMFT_PMFTR12.py
jinsooihm: I believe `reduce` does not get called in `compute`, but only when you try to access `bin_counts`.

cpp/density/RDF.cc
jinsooihm: Seems like it does not split it when I tried right now.

cpp/pmft/PMFTXY2D.cc
bdice: We are fixing this in #328.

bdice:
@jinsoo960 wants to remove 4 lines of code before merging this.

edit: nevermind.

bdice (approved):

PR #302, opened 2019-06-07
Defined _repr_png_ for some modules

jinsooihm:
## Description

`_repr_png_` method for integration with `IPython` was added for some modules that have obvious representation as 2D diagrams.

- `cluster.Cluster`
- `density.ComplexCF`
- `density.FloatCF`
- `density.RDF`
- `environment.MatchEnv`
- `msd.MSD`
- `pmft.PMFTXY2D`
- `voronoi.Voronoi`

If there are more that need to be added, please comment.

## Motivation and Context

It looks nice on Jupyter

## How Has This Been Tested?

Tests have been added to see if the `_repr_png_` fails.
Also, if the input is 3D, some tests have been added to check if `_repr_png_` returns `None`.

## Types of changes

## Checklist:


freud/cluster.pyx
```diff
@@ -169,6 +169,24 @@ cdef class Cluster:
     def __str__(self):
         return repr(self)
 
+    def _repr_png_(self):
+        try:
+            count = np.unique(self.cluster_idx, return_counts=True)
+        except ValueError:
+            print("computeClusters should be called before to show")
```
bdice: Just remove this print statement - since `_repr_png_` is called implicitly by IPython, I think this may be confusing to the user.

freud/cluster.pyx
```diff
@@ -169,6 +169,24 @@ cdef class Cluster:
...
+            return None
+        if not self.box.is2D():
+            return None
+        else:
+            count_sorted = sorted((freq, keys)
+                                  for keys, freq in zip(count[0], count[1]))
+            freqs = [i[0] for i in count_sorted[-10:]]
+            keys = [i[1] for i in count_sorted[-10:]]
```
bdice: Is this going to return the clusters ordered like 📶 instead of largest to smallest? I'd prefer to use `sorted(..., key=lambda x: -x[0])` and `count_sorted[:10]` to show largest to smallest.

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+
+    Args:
+        x (list): x values of the bar graph.
+        height (list): Height values corresponding to :code:`x`.
+        title (str): Title of the graph. (Default value = :code:`None`).
+        xlabel (str): Label of x axis. (Default value = :code:`None`).
+        ylabel (str): Label of y axis. (Default value = :code:`None`).
+
+    Returns:
+        bytes: Byte representation of the graph in png file if import
```
bdice: `if matplotlib is available` is clearer than `if import succeeds`

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+        fig = Figure()
+        ax = fig.subplots()
+        ax.bar(x=x, height=height)
+        ax.set_title(title)
+        ax.set_xlabel(xlabel)
+        ax.set_ylabel(ylabel)
+        ax.set_xticks(x)
+        ax.set_xticklabels(x)
+        f = io.BytesIO()
+        canvas = FigureCanvasAgg(fig) # noqa F841
```
bdice: What happens if you just write `FigureCanvasAgg(fig)` without the `canvas =`? Also please add a comment like `# Sets an Agg backend so this figure can be rendered` (or whatever you think makes the most sense).

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+        ylabel (str): Label of y axis. (Default value = :code:`None`).
+
+    Returns:
+        bytes: Byte representation of the graph in png file if import
+            succeeds. Otherwise :code:`None`.
+    """
+    try:
+        from matplotlib.figure import Figure
+        from matplotlib.backends.backend_agg import FigureCanvasAgg
+        import io
```
bdice: We could put these module imports at the top.

```python
# Top of the file
import io
try:
    from matplotlib.figure import Figure
    from matplotlib.backends.backend_agg import FigureCanvasAgg
    MATPLOTLIB = True
except ImportError:
    MATPLOTLIB = False
```
...
```python
# in function definitions
if not MATPLOTLIB:
    return None
fig = Figure()
ax = fig.subplots()
```
...

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+        ax.set_title(title)
+        ax.set_xlabel(xlabel)
+        ax.set_ylabel(ylabel)
+        f = io.BytesIO()
+        canvas = FigureCanvasAgg(fig) # noqa F841
+        fig.savefig(f, format='png')
+        return f.getvalue()
+
+
+def make_polygon(sides, radius=1):
```
bdice: I would remove this function since it's not used anywhere.

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+                         for theta in thetas])
+    return vertices
+
+
+def pmft_plot(pmft):
+    """ Helper function to draw 2D PMFT diagram.
+
+    Args:
+        pmft (:class:`freud.pmft.PMFTXY2D`):
+            PMFT2DXY instance.
```
bdice: Typo: PMFTXY2D

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+        import io
+        from scipy.ndimage.filters import gaussian_filter
+    except ImportError:
+        return None
+    else:
+        pmft_arr = np.copy(pmft.PMFT)
+
+        # Do some simple post-processing for plotting purposes
+        pmft_arr[np.isinf(pmft_arr)] = np.nan
+        dx = (2.0 * 3.0) / pmft.n_bins_X
```
bdice: All this math used to set the bounding box is unfortunately not generalizable to all PMFTs. We should use the values in `pmft.X` and `pmft.Y`, because this example code is hard-coded for a specific shape size.

https://freud.readthedocs.io/en/stable/pmft.html#freud.pmft.PMFTXY2D

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+            x = -3.0 + dx * i
+            for j in range(pmft.n_bins_Y):
+                y = -3.0 + dy * j
+                if ((x*x + y*y < 1.5) and (np.isnan(pmft_arr[j, i]))):
+                    pmft_arr[j, i] = 10.0
+        w = int(2.0 * pmft.n_bins_X / (2.0 * 3.0))
+        center = int(pmft.n_bins_X / 2)
+
+        # Get the center of the histogram bins
+        pmft_smooth = gaussian_filter(pmft_arr, 1)
```
bdice: I'm not sure if smoothing is appropriate for displaying the raw data. Taking this out would also remove a dependency.

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+        reduced_y = y[center-w:center+w]
+
+        # Plot figures
+        fig = Figure(figsize=(12, 5), facecolor='white')
+        values = [-2, -1, 0, 2]
+        norm = Normalize(vmin=-2.5, vmax=3.0)
+        n_values = [norm(i) for i in values]
+        colors = viridis(n_values)
+        colors = colors[:, :3]
+        # verts = make_polygon(sides=6, radius=0.6204)
```
bdice: Remove.

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+        x = pmft.X
+        y = pmft.Y
+        reduced_x = x[center-w:center+w]
+        reduced_y = y[center-w:center+w]
+
+        # Plot figures
+        fig = Figure(figsize=(12, 5), facecolor='white')
+        values = [-2, -1, 0, 2]
+        norm = Normalize(vmin=-2.5, vmax=3.0)
+        n_values = [norm(i) for i in values]
```
bdice: We shouldn't hard-code these values in the previous few lines.

freud/common.py
```diff
@@ -56,3 +56,246 @@ def convert_box(box):
...
+        fig.subplots_adjust(right=0.85)
+        cbar_ax = fig.add_axes([0.88, 0.1, 0.02, 0.8])
+        fig.colorbar(im, cax=cbar_ax)
+        f = io.BytesIO()
+        canvas = FigureCanvasAgg(fig) # noqa F841
+        fig.savefig(f, format='png')
+        return f.getvalue()
+
+
+def draw_voronoi(box, cells, color_by_sides=False):
```
bdice: @klarh I was thinking about the related Voronoi-drawing code in `plato` - anything we should consider when creating this IPython representation of `freud` Voronoi objects?

freud/density.pyx
```diff
@@ -211,6 +211,11 @@ cdef class FloatCF:
     def __str__(self):
         return repr(self)
 
+    def _repr_png_(self):
+        return freud.common.line_plot(self.R, self.RDF,
+                                      title="Float CF",
```
bdice: `title="Correlation Function"`

freud/density.pyx
```diff
@@ -211,6 +211,11 @@ cdef class FloatCF:
...
+                                      xlabel="r", ylabel="p*q(r)")
```
bdice: `ylabel="C(r)"`

freud/density.pyx
```diff
@@ -402,6 +407,11 @@ cdef class ComplexCF:
     def __str__(self):
         return repr(self)
 
+    def _repr_png_(self):
+        return freud.common.line_plot(self.R, np.real(self.RDF),
+                                      title="Complex CF",
```
bdice: `title="Correlation Function"`

freud/density.pyx
```diff
@@ -402,6 +407,11 @@ cdef class ComplexCF:
...
+                                      xlabel="r", ylabel="C(r)")
```
bdice: Try this: `ylabel="\Re(C(r))"` - we want to indicate it's the Real part.

freud/environment.pyx
```diff
@@ -869,6 +870,23 @@ cdef class MatchEnv:
     def __str__(self):
         return repr(self)
 
+    def _repr_png_(self):
+        try:
+            count = np.unique(self.clusters, return_counts=True)
+        except ValueError:
+            print("clusters should be called before to show")
```
bdice: Remove as above.

freud/environment.pyx
```diff
@@ -869,6 +870,23 @@ cdef class MatchEnv:
...
+            return None
+        if not self.m_box.is2D():
+            return None
+        count_sorted = sorted((freq, keys)
+                              for keys, freq in zip(count[0], count[1]))
+        freqs = [i[0] for i in count_sorted[-10:]]
+        keys = [i[1] for i in count_sorted[-10:]]
```
bdice: Re-sort as in `Cluster`.

tests/test_density_ComplexCF.py
```diff
@@ -127,6 +127,23 @@ def test_repr(self):
...
+        num_points = 100
+        box_size = rmax*3.1
+        np.random.seed(0)
+        points = np.random.random_sample((num_points, 3)).astype(np.float32) \
+            * box_size - box_size/2
+        ang = np.random.random_sample((num_points)).astype(np.float64) \
+            * 2.0 * np.pi
+        comp = np.exp(1j*ang)
+        ocf = freud.density.ComplexCF(rmax, dr)
+        # first bin is bad
```
bdice: Is this comment just copy-pasted? I'm actually not sure what this refers to...

bdice (changes requested):
Change requests below. Overall looks good. Can you post screenshots of the example plots? Just trying to make sure it looks realistic / useful.

freud/cluster.pyx
jinsooihm: Seems like you also need to make `keys` to be a list of string to make the bar plot be ordered from the largest cluster to the smallest.

tests/test_density_ComplexCF.py
jinsooihm: It is copy-pasted.

freud/density.pyx
jinsooihm: Do people usually use this \Re for real part? I feel like ones I have seen just put normal text Re.

jinsooihm:
Some screenshots:
- RDF
  Screen Shot 2019-06-10 at 1 34 26 PM
- Cluster / MatchEnv
  Screen Shot 2019-06-10 at 1 32 16 PM
- Complex CF
  Screen Shot 2019-06-10 at 1 41 07 PM
- Float CF
  Screen Shot 2019-06-10 at 1 42 07 PM
- Voronoi
  Screen Shot 2019-06-10 at 1 31 10 PM
- PMFTXY2D
  Screen Shot 2019-06-10 at 1 31 02 PM

jinsooihm:
More screenshots:
- MSD window mode
  Screen Shot 2019-06-10 at 2 09 29 PM
- MSD direct mode
  Screen Shot 2019-06-10 at 2 09 25 PM

freud/common.py
```diff
@@ -56,3 +63,266 @@ def convert_box(box):
...
+    Returns:
+        bytes: Byte representation of the graph in png file if matplotlib is
+        available. Otherwise :code:`None`.
+    """
+    count_sorted = sorted([(freq, key)
+                          for key, freq in zip(keys, freqs)],
+                          key=lambda x: -x[0])
+    freqs = [i[0] for i in count_sorted[:num_cluster_to_plot]]
+    keys = [str(i[1]) for i in count_sorted[:num_cluster_to_plot]]
+    return freud.common.bar_plot(keys, freqs,
```
bdice: We should make it clear if there are more clusters than are being shown. Maybe add `(total clusters: N)` as a text annotation, or in the xlabel or title?

freud/density.pyx
```diff
@@ -832,3 +844,7 @@ cdef class RDF:
 
     def __str__(self):
         return repr(self)
+
+    def _repr_png_(self):
+        return freud.common.line_plot(self.R, self.RDF,
```
bdice: Add `xlabel="$r$", ylabel="$g(r)$"`

freud/density.pyx
```diff
@@ -402,6 +408,12 @@ cdef class ComplexCF:
     def __str__(self):
         return repr(self)
 
+    def _repr_png_(self):
+        return freud.common.line_plot(self.R, np.real(self.RDF),
+                                      title="Correlation Function",
+                                      xlabel=r"$r$",
+                                      ylabel=r"$\Re(C(r))$")
```
bdice: Change to normal text Re

freud/density.pyx
```diff
@@ -211,6 +211,12 @@ cdef class FloatCF:
     def __str__(self):
         return repr(self)
 
+    def _repr_png_(self):
+        return freud.common.line_plot(self.R, self.RDF,
+                                      title="Correlation Function",
+                                      xlabel=r"$r$",
+                                      ylabel=r"$\langle p \ast q \rangle (r)$")
```
bdice: Use `C(r)` instead of `p \ast q`

freud/common.py
jinsooihm: I made it to print `Keys of {} largest clusters (total clusters: {})`

jinsooihm: Screen Shot 2019-06-10 at 3 02 54 PM
Screen Shot 2019-06-10 at 3 03 03 PM

freud/plot.py
```diff
@@ -0,0 +1,272 @@
...
+                    [-2, -1, 0, 2], colors=colors)
+
+        fig.subplots_adjust(right=0.85)
+        cbar_ax = fig.add_axes([0.88, 0.1, 0.02, 0.8])
+        fig.colorbar(im, cax=cbar_ax)
+        f = io.BytesIO()
+        # Sets an Agg backend so this figure can be rendered
+        FigureCanvasAgg(fig)
+        fig.savefig(f, format='png')
+        return f.getvalue()
```
vyasr: A lot of this plotting logic is specific to the specific data in the PMFT example. I would simplify this to the bare minimum required to generate a simple PMFT plot without the bells and whistles to make the plot especially pretty. My inclination is to just do the simple heat map and remove the extra logic for the contour as well. @bdice may disagree with some of these recommendations though.

vyasr (commented):
This looks really nice! I can't think of any other modules that should have png reprs added, @bdice can you?

freud/plot.py
bdice: Yes, I agree we should simplify this to the minimum.

jinsooihm: Screen Shot 2019-06-13 at 9 40 19 PM

Something like this?
Or is this still specific to the data?

freud/environment.pyx
```diff
@@ -869,6 +870,17 @@ cdef class MatchEnv:
     def __str__(self):
         return repr(self)
 
+    def _repr_png_(self):
+        import plot
+        try:
+            count = np.unique(self.clusters, return_counts=True)
+        except ValueError:
+            return None
+        if not self.m_box.is2D():
```
bdice: Why do we limit to 2D if we're plotting `MatchEnv` cluster sizes?

freud/density.pyx
```diff
@@ -402,6 +409,13 @@ cdef class ComplexCF:
     def __str__(self):
         return repr(self)
 
+    def _repr_png_(self):
+        import plot
+        return plot.line_plot(self.R, np.real(self.RDF),
+                              title="Correlation Function",
+                              xlabel=r"$r$",
+                              ylabel=r"$Re(C(r))$")
```
bdice: Does this put `Re` in plain text or math font? I'm looking for this: 
image

freud/plot.py
bdice: @jinsoo960 The code is doing a bunch of scaling and trickery that is specific to this hexagon analysis - the colorbar needs to find its min/max bounds from the min/max of the PMFT, and the x/y limits of the plot should correspond to the PMFT's actual bin locations. Last I checked, it was scaling the axes by 2/3 and only showing a cropped version of the actual data, for the purposes of this plot.

freud/density.pyx
jinsooihm: It seems like `matplotlib` does not have all the math fonts, but I think `operatorname{Re}` gets close to it.

freud/plot.py
jinsooihm: Screen Shot 2019-06-13 at 10 46 06 PM
Screen Shot 2019-06-13 at 10 46 16 PM

This is an almost bare minimum version, with smoothing an all other logic removed.
Only processing done is changing `inf` to `nan`.

bdice: Close- bring back the part that labels the `x` and `y` ticks by the values of `pmft.X` and `pmft.Y`, instead of labeling them by the index of the bin.

bdice: Also change the title of the plot to `2D Potential of Mean Force`.

bdice:
## To-do:
- [x] Take a look at pandas' [DataFrame.plot](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.html) - it takes `ax` as an argument and returns an `ax` object. We should refactor the `_repr_png_` methods into `plot` methods, so we can use this plotting code in our examples. There should be a separate `_repr_png_` method still, which will call `import plot; return plot.to_png_bytes(self.plot())` or something similar.
- [x] Add $k_B T$ to the colorbar of the PMFTXY2D, like this: 
image

bdice (approved):
Looks good to me!

PR #304, opened 2019-06-10
Refactor/voronoi

yzjin:
## Description
Voronoi is incorporated in locality module and port to C++. The cython counterparts are updated to coordinate.
Merged `compute()` and `computeNeighbors()` into one function to reduce code redundancy. 
Refactored the computation of `compute()` in C++
Stored bonds using `NeighborList` class instead of python `dict`.
`locality.Voronoi` constructor now doesn't require `box` and `buff`.
`compute()` now takes in `(box, position, buffer, image)` arguments.

## Motivation and Context
The Voronoi computation is slow for certain systems. The current implementation is in cython using `SciPy.spatial.voronoi` package and contains duplicate codes. Some bottlenecks of the code includes inefficient implementation of the neighbor list and repeated code in several functions(e.g. `compute()` and `computeNeighbors()`). We would like to improve the performance by using C++ to replace cython in implementation. API for compute functions are more compatible to their functionality. 

## How Has This Been Tested?
Updated tests to support the code. Passes existing voronoi tests except `test_nlist_symmetric`.
### Bug description
This version will lead to duplicate (i, j) pairs in neighbor lists. 

### Reproduce
```
def test_nlist_symmetric(self):
        # Test that the voronoi neighborlist is symmetric
        L = 10  # Box length
        rbuf = 3  # Cutoff radius
        N = 40  # Number of particles

        box = freud.box.Box.cube(L)  # Initialize box
        np.random.seed(0)
        points = np.random.uniform(-L/2, L/2, (N, 3)).astype(np.float32)
        vor = freud.locality.Voronoi()
        vor.compute(
            box=box, positions=points, buffer=rbuf, images=False)
        nlist = vor.nlist

        ijs = set(zip(nlist.index_i, nlist.index_j))
        jis = set(zip(nlist.index_j, nlist.index_i))

        # we shouldn't have duplicate (i, j) pairs in the
        # resulting neighbor list
        npt.assert_equal(len(ijs), len(nlist))
        npt.assert_equal(len(ijs), len(jis))

        # every (i, j) pair should have a corresponding (j, i) pair
        self.assertTrue(all((j, i) in jis for (i, j) in ijs))
```

### Error Output
```
  File "test_locality_voronoi.py", line 150, in test_nlist_symmetric
    npt.assert_equal(len(ijs), len(nlist))
AssertionError:
Items are not equal:
 ACTUAL: 591
 DESIRED: 593
```
## Types of changes

## Checklist:


cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,195 @@
...
+
+#include <iostream>
+#include <algorithm>
+#include <sstream>
+#include <stdexcept>
+#include <tbb/tbb.h>
+#include <tuple>
+#include <cmath>
+#include <memory>
+#include <vector>
```
vyasr, line 8: Can you check these includes? At the very least, you can remove iostream, sstream, and stdexcept, but I'm also not sure that you need memory (I don't see you making any shared pointers), algorithm (you might need this one for std::min and std::max), or cmath (maybe you need it for sqrt?). I would try removing those and see which cause problems.

cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,195 @@
...
+
+#include "Box.h"
+#include "NeighborList.h"
+#include "Voronoi.h"
+#include "NeighborQuery.h"
+
+using namespace std;
+using namespace tbb;
```
vyasr: For the most part, you look like you're using fully qualified names of everything below, so can you just remove these two `using` declarations and fix any cases where you aren't fully specifying what you're calling?

cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,195 @@
...
+    bool operator()(const NeighborBond &n1, NeighborBond &n2) const {
+        if (n1.index_i != n2.index_i) {
+            return n1.index_i < n2.index_i;
+        }
+        if (n1.index_j != n2.index_j) {
+            return n1.index_j < n2.index_j;
+        }
+        return n1.weight < n2.weight;
+    }
+};
```
vyasr: Can we just make this a function? It's a little confusing to see a stateless functor, and this might as well follow the `compareFirstNeighborPairs` example in LinkCell.cc.

cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,195 @@
...
+                unsigned int j = ridge_points[2*ridge+1];
+                float weight = 0;
+
+                // Reject bonds between two image particles
+                if (i >= N && j >= N)
+                    continue;
+
+                bool exclude_ii = true;
+                // We DO allow bonds from a particle to its own image
+                if (exclude_ii && i == j)
```
vyasr: No need to define `exclude_ii`, just check if `i==j`.

cpp/locality/Voronoi.h
```diff
@@ -0,0 +1,69 @@
...
+     */
+    bool operator< (const NeighborBond &n) const
+    {
+        if (index_i < n.index_i) {
+            return true;
+        } else if (index_j < n.index_j) {
+            return true;
+        }
+        return weight < n.weight;
+    }
```
vyasr: Is this actually what you want? I would have expected that you only check weights if the indexes are equal, but if `index_i > n.index_i` but `weight < n.weight`, this returns true. If you intend to prioritize sorting by weight over index-based sorting, then I would put that first in the sequence of conditionals.

cpp/locality/Voronoi.h
```diff
@@ -0,0 +1,69 @@
...
+};
+
+
+class Voronoi
+    {
+    public:
+        // default constructor
+        Voronoi();
+
+        // void compute(const vec3<double>* vertices, const std::vector<int>* ridge_points, const std::vector<int>* ridge_vertices);
```
vyasr: Delete

freud/locality.pyx
```diff
@@ -1374,3 +1392,239 @@ cdef class NearestNeighbors:
...
+
+cdef class Voronoi:
+    R"""Compute the Voronoi tessellation of a 2D or 3D system using qhull.
+    This uses :class:`scipy.spatial.Voronoi`, accounting for periodic
+    boundary conditions.
+
+    .. moduleauthor:: Benjamin Schultz <baschult@umich.edu>
+    .. moduleauthor:: Yina Geng <yinageng@umich.edu>
+    .. moduleauthor:: Mayank Agrawal <amayank@umich.edu>
+    .. moduleauthor:: Bradley Dice <bdice@bradleydice.com>
```
vyasr, line 1389: Add yourself @yzjin ! :)

tests/test_voronoi.py
```diff
@@ -3,6 +3,7 @@
 import freud
 import unittest
 import util
+from scipy.sparse import csr_matrix
```
vyasr: Remove this, I think this is a holdover from when you tried to Cythonize the method and is no longer relevant.

freud/locality.pyx
```diff
@@ -5,22 +5,40 @@ R"""
...
 cimport freud._locality
 cimport freud.box
 cimport numpy as np
 
+logger = logging.getLogger(__name__)
+
+try:
+    from scipy.spatial import Voronoi as qvoronoi
+    from scipy.spatial import ConvexHull
+    from scipy.sparse import csr_matrix
```
vyasr: Remove this.

cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,195 @@
...
+                    continue;
+
+                // Bonds where a ridge goes to infinity (index -1) have weight 0
+                bool weighted_bond = true;
+                vector<int> current_ridge_vertex_ids;
```
vyasr: If you change this to `vector<vec3<float> > current_ridge_vertices` and just push_back the vertex directly, it saves you some logic in the length/area calculation.

cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,195 @@
...
+                        current_ridge_vertex_ids.push_back(ridge_vertices[ridge_vert_id]);
+                    }
+                }
+
+                if (weighted_bond) {
+                    if (box.is2D()) {
+                        // 2D weight is the length of the ridge edge
+                        int v1ind = current_ridge_vertex_ids[0];
+                        int v2ind = current_ridge_vertex_ids[1];
+                        vec3<double> v1 = vertices[v1ind];
```
vyasr: Can you add a comment near the top of this file indicating that you have to use double precision to get meaningful results? Otherwise this just looks wrong in freud, and it's also not obvious why in some cases you have double and others you have float.

cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,195 @@
...
+                            c0 = 1;
+                        }
+
+                        double projected_area = 0;
+                        int n_verts = vertex_coords.size();
+                        for (int step = 0; step < n_verts; step++)
+                        {
+                            int n1 = step % n_verts;
+                            int n2 = (step + 1) % n_verts;
+                            int n3 = (step - 1 + n_verts) % n_verts;
```
vyasr: Is there any reason you can't use unsigned ints for all of these (including `step`)? If there are compiler warnings associated with the `step-1` being potentially negative, you might be able to trick it by rewriting this as ((step + n_verts) - 1) or something like that. If not, then leave it.

cpp/locality/Voronoi.h
```diff
@@ -0,0 +1,69 @@
...
+        void compute(const box::Box &box, const vec3<double>* vertices,
+            const int* ridge_points, const int* ridge_vertices,
+            unsigned int n_ridges, unsigned int N, const int* expanded_ids,
+            const int* ridge_vertex_indices);
+         NeighborList *getNeighborList()
```
vyasr: Add a space above this line.

cpp/locality/Voronoi.h
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2010-2019 The Regents of the University of Michigan
+// This file is from the freud project, released under the BSD 3-Clause License.
+
+#ifndef VORONOI_H
+#define VORONOI_H
+
+#include "VectorMath.h"
+#include <vector>
+
+using namespace std;
```
vyasr: Remove this.

cpp/locality/Voronoi.h
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2010-2019 The Regents of the University of Michigan
+// This file is from the freud project, released under the BSD 3-Clause License.
+
+#ifndef VORONOI_H
+#define VORONOI_H
+
+#include "VectorMath.h"
+#include <vector>
```
vyasr: It looks like this is working right now because you have it in your CPP file, but you should #include NeighborList directly in this file.

vyasr (changes requested):
Thanks for the PR Yezhi! This will be really nice to have in freud, and moves us towards being able to use it with the new NeighborQuery infrastructure.

cpp/locality/Voronoi.h
yzjin: Actually the current implementation doesn't need this comparing function. I include this because I'm not sure if it will be needed in the future. Would you still like me to change per your suggestions or I can delete it?

bdice: @yzjin I would remove this logic if it's not being used, just to prevent confusion.

bdice:
@yzjin In files where we use `std::max` and `std::min`, we have to add a special bit of code that prevents problems when compiling on Windows. You can see the Appveyor tests failed - please include this snippet at the top of `Voronoi.cc` to fix those. (See `environment/MatchEnv.cc` and `locality/LinkCell.cc` for examples.)

```c++
#if defined _WIN32
#undef min // std::min clashes with a Windows header
#undef max // std::max clashes with a Windows header
#endif
```

cpp/locality/Voronoi.h
yzjin: Updated

bdice:
## To-do:
- [ ] Add `exclude_ii` option to `compute`
- [x] Investigate the case where `i` is bonded to an image of `j` and `j` is bonded to an image of `i` - currently there's no way to prevent this from creating duplicate bonds. See below.
- [x] We could force users to always use a buffer of 2 images (remove the argument to `compute`), so that they're guaranteed the correct results (with periodicity) at the cost of potentially lower speed. Personally, I always use enough buffer to guarantee the correct results. Would this be okay with you, @vyasr?

If we force users to be safe (by always using 2 images), then we may be able to resolve the duplicate real-image bond problem by only adding bonds where `real_i < real_j` -- and doing this symmetrically (real-to-image and image-to-real). If `exclude_ii` is false, then when `real_i == real_j` we can add bonds asymmetrically (only real-to-image, not image-to-real).

@yzjin We should construct some test cases and ensure this strategy works. Happy to explain this in person on a whiteboard, but I think this solves the issue.

cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,203 @@
...
+    \brief Build a cell list from a set of points.
+*/
+
+namespace freud { namespace locality {
+
+// Default constructor
+Voronoi::Voronoi()
+{}
+
+// A compare funtion feeding to tbb::parallel_sort to sort NeighborBonds
```
bdice: typo: `function`

Change to `A comparison function used to sort NeighborBonds.`

cpp/locality/Voronoi.h
```diff
@@ -0,0 +1,48 @@
...
+struct NeighborBond
+{
+    NeighborBond() : index_i(0), index_j(0), weight(0) {}
+
+    NeighborBond(unsigned int index_i, unsigned int index_j, float w) :
+        index_i(index_i), index_j(index_j), weight(w) {}
+
+    unsigned int index_i;     //! The point id.
+    unsigned int index_j;     //! The reference point id.
+    float weight;             //! The weight of this bond.
```
bdice, line 22: Add a `float distance` field and compute the distance between the points in the `compute` method (to ensure that `i-i` bonds aren't treated as zero-distance later).

bdice: To compute distances, you'll probably need to adapt the signature of `compute` to pass in `vec3<float> expanded_points`, or just pass in the entire `ParticleBuffer` object in the same way that we pass in `Box` objects to various compute methods and then query the `ParticleBuffer` for its data.

cpp/locality/Voronoi.cc
```diff
@@ -0,0 +1,203 @@
...
+        m_box = box;
+
+        // iterate over ridges in parallel
+        BondVector bonds;
+        tbb::parallel_for(tbb::blocked_range<size_t>(0, n_ridges), [&] (const tbb::blocked_range<size_t> &r) {
+             BondVector::reference local_bonds(bonds.local());
+            for (size_t ridge(r.begin()); ridge != r.end(); ++ridge) {
+                unsigned int i = ridge_points[2*ridge];
+                unsigned int j = ridge_points[2*ridge+1];
+                float weight = 0;
```
bdice, line 78: Distance computation can be done right here. I think you'll get the wrong answers if you call `box.wrap` on `rij`, so don't do that.
```c++
vec3<float> rij(expanded_points[j] - expanded_points[i]);
float distance = sqrt(dot(rij, rij));
```

cpp/locality/Voronoi.h
bdice: After #295 is merged, I believe we won't need this `struct` at all. Voronoi's `struct` should be the same struct used by other `NeighborQuery` classes when that happens.

doc/source/locality.rst
```diff
@@ -50,3 +51,9 @@ Nearest Neighbors
 
 .. autoclass:: freud.locality.NearestNeighbors(rmax, n_neigh, scale=1.1, strict_cut=False)
    :members: compute, getNeighborList, getNeighbors, getRsq
+
+Voronoi
+=================
```
bdice: Remove the extra `===`s

cpp/locality/Voronoi.cc
bdice: This should be superseded by #295 if `Voronoi`'s struct matches the other struct in `NeighborQuery`. Then this comparison function can be removed/replaced.

vyasr:
I'm fine with defaulting to the minimum safe buffer width. I think it should be a user-specified parameter so that if you know that you need less buffer you can do a more efficient calculation, but the default should guarantee correctness.

Please re-request review from me when you think this PR is ready for another round of review after the items in the checklist have been addressed.

bdice (commented):
Oops - I wrote this review a while back and forgot to submit it. Hope these changes are still applicable.

bdice:
@yzjin I discussed this PR with @vyasr. Let's apply the changes I suggested above, then rename this class to `locality._Voronoi`. We will hide the documentation for now, and continue to improve it before releasing it as part of the public API. There are some API-breaking changes that I'd like to make but we need to finish some other freud PRs first -- making it "private" with an underscore would let us continue to use it in our own research work without promising a public API yet.

cpp/locality/Voronoi.cc
vyasr: I agree, let's plan to merge those changes once #295 is ready. We can merge this PR and simplify it later.

vyasr (approved):

PR #306, opened 2019-06-11
Using Eigen for diagonalization

### LINKED ISSUE #293: LocalDescriptors should use Eigen for diagonalization
We currently only use `HOOMDMatrix.cc` / `HOOMDMatrix.h` for one thing: diagonalizing a real symmetric 3x3 matrix in `LocalDescriptors.cc`.
https://github.com/glotzerlab/freud/blob/master/cpp/environment/LocalDescriptors.cc#L112

We should find eigenvalues / eigenvectors with Eigen, instead. https://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html

There is an example of this in `NematicOrderParameter.cc`. https://github.com/glotzerlab/freud/blob/master/cpp/order/NematicOrderParameter.cc#L164-L165

**Note that Eigen returns eigenvalues in [_increasing_ order](https://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html#a3df8721abcc71132f7f02bf9dfe78e41).** NematicOrderParameter is doing too much work to order its eigenvalues, and that should also be fixed.

Then we should remove `HOOMDMatrix` from the includes in `LocalDescriptors`, remove `HOOMDMatrix` from the `environment` module sources in `setup.py`, and delete the `HOOMDMatrix` files from the `cpp/util` directory.

jinsooihm:
## Description

Dependence on `HOOMDMatrix` was removed by adding a function that diagonalizes 3x3 matrix using `Eigen`.
Also, `NematicOrderParameter` and its test has been modified to use the new function.

## Motivation and Context

Resolves #293 

## How Has This Been Tested?

An existing test has been modified.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change) (perhaps?)

## Checklist:


cpp/util/diagonalize.cc
```diff
@@ -0,0 +1,50 @@
+#include "diagonalize.h"
+
+namespace freud { namespace util {
+
+void diagonalize33SymmetricMatrix(float mat[9], float eigen_vals[3], float eigen_vecs[9])
+{
+    Eigen::Matrix3f m;
```
bdice: We might be able to use `Eigen::Map<Eigen::Matrix3f> m(&mat);`

cpp/util/diagonalize.cc
```diff
@@ -0,0 +1,50 @@
...
+    Index2D a_i = Index2D(3);
+    for (unsigned int i = 0; i < 3; ++i)
+    {
+        for (unsigned int j = 0; j < 3; ++j)
+        {
+            m(i, j) = mat[a_i(i, j)];
+        }
+    }
+
+    Eigen::SelfAdjointEigenSolver<Eigen::MatrixXf> es;
```
bdice: Probably use `Matrix3f` here?

cpp/util/diagonalize.cc
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    eigen_vecs[a_i(j, j)] = 0.0;
+                }
+        // set order parameter to zero so it's easily detectable
+        eigen_vals[0] = eigen_vals[1] = eigen_vals[2] = 0.0;
+    }
+    else
+    {
+        // columns are eigenvectors
+        Eigen::Matrix3f eigen_vec = es.eigenvectors();
```
bdice: Might be able to use the `Map` syntax here too, to assign directly into `eigen_vecs` without needing the double-for loop.

setup.py
```diff
@@ -353,12 +353,13 @@ def find_tbb(tbb_root=None, tbb_include=None, tbb_link=None):
         os.path.join("cpp", "locality", "NeighborList.cc"),
     ],
     environment=[
-        os.path.join("cpp", "util", "HOOMDMatrix.cc"),
+        os.path.join("cpp", "util", "diagonalize.cc"),
```
bdice: Swap this and the following line - put `diagonalize` last.

tests/test_order_Nematic.py
```diff
@@ -50,44 +50,46 @@ def test_perfect(self):
...
+        npt.assert_allclose(op_perp.director, u, atol=1e-1)
+        self.assertFalse(np.all(op_perp.director == u))
 
-        npt.assert_allclose(op.nematic_order_parameter, 0.25, atol=1e-6)
-        npt.assert_equal(op.director, np.array([0, 1, 0]))
-        npt.assert_allclose(op.nematic_tensor, np.diag([0.25, 0.25, -0.5]),
-                            rtol=1e-6, atol=1e-6)
+        npt.assert_allclose(
+            op_perp.nematic_tensor, np.diag([-0.5, 1, -0.5]), atol=1e-1)
+        self.assertFalse(np.all(op_perp.nematic_tensor
```
bdice: Break the line after `np.all(` for clarity.

bdice (commented):
Looks great - some minor suggestions, then this is ready to merge.

cpp/util/diagonalize.cc
```diff
@@ -0,0 +1,27 @@
...
+    {
+        // numerical issue, return identity matrix
+        Eigen::Matrix3f id = Eigen::Matrix3f::Identity();
+        Eigen::Map<Eigen::Matrix3f> (eigen_vecs, 3, 3) = id;
+        // set eigenvalues to zero so it's easily detectable
+        eigen_vals[0] = eigen_vals[1] = eigen_vals[2] = 0.0;
+    }
+    else
+    {
+        Eigen::Map<Eigen::Matrix3f> (eigen_vecs, 3, 3) = es.eigenvectors();
```
bdice, line 22: I don't think we need the `3, 3` or the `3` below, because these are fixed-size matrix types.

> A fixed-size read-only vector of integers might be declared as
> `Map<const Vector4i> mi(pi);`
> where `pi` is an `int *`. In this case the size does not have to be passed to the constructor, because it is already specified by the Matrix/Array type.

https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html

bdice (approved):
One suggested edit - I'll do it directly in GitHub and we'll make sure CI passes.

Edit: didn't work, let's merge as-is.

cpp/util/diagonalize.cc
bdice: This didn't work - we'd have to cast the results of `es` to the fixed-size types, too. I just reverted my change and we'll merge as-is.

Issue #199, opened 2019-01-17
Use atleast_2d everywhere


bdice:
We should make sure that 1d inputs (single points) are allowed for analysis methods in freud.

vyasr:
This should be done in conjunction with #268

bdice:
@vyasr I talked with @jinsoo960 about this, we're not sure if we want to make a change for this in `convert_array` after all. We may just keep the call to `atleast_2d` before calling `convert_array` in the cases where we explicitly want this behavior. Otherwise it may confuse users and/or give unexpected answers by accepting inputs of the wrong dimensionality.

vyasr:
For the box functions, here's code for a decorator that should do what we want:
```
def support_1d(func):
    def wrapper(self, arg1, *args, **kwargs):
        if len(arg1.shape) == 1:
            flat = True
            arg1 = np.atleast_2d(arg1)
        else:
            flat = False
        ret = func(arg1, *args, **kwargs)
        if flat:
            return ret.squeeze()
        else:
            return ret
```

PR #307, opened 2019-06-11
Refactoring convert array

### LINKED ISSUE #268: convert_array should standardize more
Many of the arguments to the `convert_array` function essentially never change. We should standardize at least the `dtype` and `contiguous` arguments further. Standardizing `dtype` here would also make it easier to enable double precision support if we choose to do so.

jinsooihm:
## Description

Shape check moved into `convert_array` function.
`squeeze` and `atleast_2d` is not implemented yet.

## Motivation and Context

Related to #199.
Resolves #268.

## How Has This Been Tested?

Few tests have been added and it passes all the existing tests.

## Types of changes

## Checklist:


vyasr (changes requested):
After discussion, we'd like to merge the shape and dimension checks into one check. That should simplify this further.

jinsooihm:
Done! 
One side note:
I was in the branch `feature/property` and there `common` is a `pyx` file and thus produces a `common.cpython-36m-darwin.so` file in the directory `freud`.
When I checked out this branch, that `common.cpython-36m-darwin.so` file still exists, and importing `common` will not import `common.py`, but instead will import the implementation in `common.cpython-36m-darwin.so` file. (This can be checked from just opening a python interpreter and try printing `freud.common.__file__`.)
It really messed things up on my local machine because all the changes to `common.py` does not get reflected, and even running `python setup.py clean` does not clean it.

vyasr:
Thanks for the heads up. Makes sense, that annoyance probably won't go away until we merge that branch, so good to keep it in mind.

freud/box.pyx
```diff
@@ -391,7 +391,8 @@ cdef class Box:
                 "Valid input is an array of shape (3,) or (N,3).")
 
         vecs = freud.common.convert_array(vecs)
-        imgs = freud.common.convert_array(imgs, vecs.ndim, dtype=np.int32)
+        imgs = freud.common.convert_array(imgs, (None, ) * vecs.ndim,
```
vyasr: Can you just make this check `vecs.shape` and then remove the extra check above?

vyasr (approved):
Everything looks good! There were a couple of small things I wanted changed, but they were were all tiny changes so it made more sense for me to just make them myself.

freud/cluster.pyx
```diff
@@ -131,10 +128,8 @@ cdef class Cluster:
             keys((:math:`N_{particles}`) :class:`numpy.ndarray`):
```
bdice: Is this missing a space after `keys`?

freud/cluster.pyx
```diff
@@ -245,16 +240,9 @@ cdef class ClusterProperties:
...
-            raise RuntimeError(
-                'Need a list of 3D points for computeClusterProperties()')
+        points = freud.common.convert_array(points, shape=(None, 3))
         cluster_idx = freud.common.convert_array(
-            cluster_idx, 1, dtype=np.uint32)
-        if cluster_idx.shape[0] != points.shape[0]:
-            raise RuntimeError(
-                ('cluster_idx must be a 1D array of matching length/number'
-                    'of particles to points'))
+            cluster_idx, dtype=np.uint32, shape=(points.shape[0],))
```
bdice: Switch the order of arguments.

freud/common.py
```diff
@@ -10,30 +10,40 @@
...
+    .. moduleauthor:: Jin Soo Ihm <jinihm@umich.edu>
 
     Args:
         array (:class:`numpy.ndarray`): Array to check and convert.
-        dimensions (int): Expected dimensions of the array. If 'None',
-            no dimensionality check will be done (Default value = 'None').
+        shape: (tuple of int and :code:`None`): Expected shape of the array.
+            Only the dimensions that are not :code:`None` are checked.
+            (Default value = :code:'None').
         dtype: code:`dtype` to convert the array to if :code:`array.dtype`
```
bdice: Missing colon before `code:` ➡️ `:code:`

freud/environment.pyx
```diff
@@ -435,13 +422,8 @@ cdef class LocalDescriptors:
...
 
-            orientations = freud.common.convert_array(orientations, 2)
-            if orientations.shape[1] != 4:
-                raise TypeError('orientations should be an Nx4 array')
-
-            if orientations.shape[0] != points_ref.shape[0]:
-                raise ValueError(
-                    "orientations must have the same size as points_ref")
+            orientations = freud.common.convert_array(orientations,
+                shape=(points_ref.shape[0], 4)) # noqa
```
bdice: Not sure why this is `# noqa` - is there a `flake8` complaint? I like to mark `# noqa: E501` or whatever the `flake8` rule is that it breaks.

freud/locality.pyx
```diff
@@ -284,7 +284,8 @@ cdef class NeighborQuery:
         # This function is temporarily included for testing and may be
         # removed in future releases.
         # Can't use this function with old-style NeighborQuery objects
-        points = freud.common.convert_array(np.atleast_2d(points), 2)
+        points = freud.common.convert_array(np.atleast_2d(points),
+                                            (None, None))
```
bdice: Why not `shape=(None, 3)`?

freud/locality.pyx
```diff
@@ -459,16 +458,16 @@ cdef class NeighborList:
                 Array of per-bond weights (if :code:`None` is given, use a
                 value of 1 for each weight) (Default value = :code:`None`).
         """
-        index_i = freud.common.convert_array(index_i, 1, dtype=np.uint64)
-        index_j = freud.common.convert_array(index_j, 1, dtype=np.uint64)
+        index_i = freud.common.convert_array(index_i, (None,), dtype=np.uint64)
+        index_j = freud.common.convert_array(index_j, (None,), dtype=np.uint64)
```
bdice: We could get rid of the shape check below by enforcing `index_j` has the same shape as `index_i`.

freud/locality.pyx
```diff
@@ -459,16 +458,16 @@ cdef class NeighborList:
...
 
         if index_i.shape != index_j.shape:
             raise TypeError('index_i and index_j should be the same size')
 
         if weights is None:
             weights = np.ones(index_i.shape, dtype=np.float32)
         else:
-            weights = freud.common.convert_array(weights, 1)
+            weights = freud.common.convert_array(weights, (None, ))
```
bdice: Check to make sure `weights` is the same shape as `index_i`.

freud/locality.pyx
```diff
@@ -812,7 +807,8 @@ cdef class AABBQuery(NeighborQuery):
             # Assume valid set of arguments is passed
             self.queryable = True
             self._box = freud.common.convert_box(box)
-            self.points = freud.common.convert_array(points, 2).copy()
+            self.points = freud.common.convert_array(
+                points, (None, None)).copy()
```
bdice: As above, can we check for shape `(None, 3)`?

freud/locality.pyx
```diff
@@ -827,7 +823,8 @@ cdef class AABBQuery(NeighborQuery):
         # This function is temporarily included for testing and WILL be
         # removed in future releases.
         # Can't use this function with old-style NeighborQuery objects
-        points = freud.common.convert_array(np.atleast_2d(points), 2)
+        points = freud.common.convert_array(
+            np.atleast_2d(points), (None, None))
```
bdice: `(None, 3)`?

freud/msd.pyx
```diff
@@ -142,24 +142,10 @@ cdef class MSD:
...
-            if images.shape[2] != 3:
-                raise TypeError(
-                    'images should be a 3-dimensional array of shape'
-                    '(N_frames, N_particles, 3)'
-                )
-
-            if not positions.shape == images.shape:
-                raise TypeError(
-                    'The positions and images must have the same shape')
+            images = freud.common.convert_array(images, dtype=np.int32,
```
bdice: Swap dtype and shape.

freud/order.pyx
```diff
@@ -233,7 +231,7 @@ cdef class NematicOrderParameter:
 
         cdef vec3[float] l_u = vec3[float](u[0], u[1], u[2])
         self.thisptr = new freud._order.NematicOrderParameter(l_u)
-        self.u = freud.common.convert_array(u, 1)
+        self.u = freud.common.convert_array(u, (None, ))
```
bdice: Should be shape `(3,)`

bdice:
We're pretty inconsistent throughout about whether we define the shape as a keyword argument or just a plain argument, like `freud.common.convert_array(arr, shape=(None, 3))` vs. `freud.common.convert_array(arr, (None, 3))`. I like defining it as a keyword argument for clarity, but that's a stylistic preference. Consistency would be good either way.

bdice:
Also check for other instances of the `(None, None)` pattern that I mentioned - I stopped leaving comments after the first few. It doesn't seem right to me (should be `(None, 3)` if I understand correctly) but it was repeated a lot so I may have misinterpreted something.

bdice (changes requested):
See comments.

vyasr:
The majority of cases were using keyword args, so I switched to doing that everywhere.

freud/environment.pyx
```diff
@@ -149,14 +149,14 @@ cdef class BondOrder:
         Args:
             box (:class:`freud.box.Box`):
                 Simulation box.
-            ref_points ((:math:`N_{particles}`, 3) :class:`numpy.ndarray`):
+            ref_points ((:math:`N_{ref_points}`, 3) :class:`numpy.ndarray`):
```
bdice: I think the docs will render this wrong unless you escape the underscore. Please check.

bdice (approved):
One doc fix - then good to merge. 👍

PR #309, opened 2019-06-11
Verify values have been computed

### LINKED ISSUE #167: Verify values have been computed
**[Original report](https://bitbucket.org/glotzer/freud/issue/167) by Bradley Dice (Bitbucket: [bdice](https://bitbucket.org/bdice), GitHub: [bdice](https://github.com/bdice)).**

All data retrieval methods that require calling a `compute()` method first should raise an exception if the requested data has not yet been computed.

jinsooihm:
## Description

Added a parent class that contains decorators for compute, property/method, and reset method to set and check for flags whether to allow access to a property/method or not.
Raise an `AttributeError` if the user should not access.
Nice thing is that minimal change is required on already existing modules.
Implemented for:
- Cluster
- Density
- Environment
- Interface
- MSD
- Order
- PMFT
- Voronoi

Also, some documentations were fixed.

## Motivation and Context

Resolves #167.

## How Has This Been Tested?

Tests are added.

Screen Shot 2019-06-11 at 5 48 32 PM

## Types of changes

## Checklist:

   Seems like many parts of the documentation have to be edited to tell the user that which variables require which method to be called before.


freud/cluster.pyx
```diff
@@ -8,11 +8,12 @@ of clusters of points in a system.
 
 import numpy as np
 import warnings
-import freud.common
 import freud.locality
+import freud.common
```
bdice: Revert to keep these sorted alphabetically.

freud/common.pyx
```diff
@@ -0,0 +1,150 @@
...
+            func(self, *args, **kwargs)
+        return wrapper
+
+
+def convert_array(array, dimensions=None, dtype=np.float32):
+    """Function which takes a given array, checks the dimensions,
+    and converts to a supplied dtype and/or makes the array
+    contiguous.
+
+    .. moduleauthor:: Eric Harper <harperic@umich.edu>
```
bdice, line 131: @jinsoo960 add yourself here!

bdice (commented):
This design looks great to me! Now just apply it everywhere. 🖥 👍

freud/common.pyx
```diff
@@ -0,0 +1,179 @@
...
+            Decorator decorating appropriate property method.
+        """
+        if isinstance(key, str):
+            key = (key,)
+
+        def _computed_property_with_key(func):
+            @property
+            @wraps(func)
+            def wrapper(self, *args, **kwargs):
+                if not any(self._called_compute[k] for k in key):
```
bdice: If you change this to `self._called_compute.get(k, False)` then it will automatically return `False` if the key isn't set - this would avoid the need to call `self._set_compute_flag("computeClusters")` and similar methods in the constructors of the non-standard classes. Making `_called_compute` into a [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) is another option.

I'd prefer to remove the `_set_compute_flag` if possible, after that change is made.

freud/msd.pyx
```diff
@@ -194,19 +196,21 @@ cdef class MSD:
 
         return self
 
-    @property
+    @Compute._computed_property()
     def box(self):
-        return self.box
+        return self.m_box
```
bdice: The `box` isn't a computed property for `MSD` - it's a constructor argument, so no need to call `compute` to get the result. This is unusual among `freud` classes, but for a good reason, since the MSD is only meaningful for constant boxes (unlike many other analyses, which can allow box changes across frames). I think this can be reverted to just saving `self.box` in the constructor and having that be directly accessible by the user, as before.

bdice (changes requested):
A couple of requested changes. Overall this is great! Also good tests!

Please merge in `master` and add a line to the Changelog.

bdice:
@jinsoo960
## To-do:
- [x] merge in `master`
- [x] Add a changelog line
- [x] Add decorators to the new plotting features. Make sure the `_repr_png_` returns `None` instead of raising an error if it hasn't been computed yet. `object.plot()` should raise an error.

bdice (approved):
👍

PR #311, opened 2019-06-13
Added atleast and squeeze for box functions / Moved operations on multiple vectors logic to C++

### LINKED ISSUE #199: Use atleast_2d everywhere

jinsooihm:
## Description

Some `Box` method allows the user to use a single vector or list of vectors.
Appropriate converting up and down of input and output of those methods are now done using `atleast_2d` and `squeeze`.
Also, operations on multiple vectors are now done in C++ side in parallel.

## Motivation and Context

Resolves #199 

## How Has This Been Tested?

Passes tests.

## Types of changes

## Checklist:


freud/box.pyx
```diff
@@ -33,6 +34,54 @@ logger = logging.getLogger(__name__)
...
 
+
+def _support_1d(func):
+    R"""Decorator for class methods that accept one dimensional input or
+    two dimensional input for the first argument.
+    """
+    @wraps(func)
+    def wrapper(self, arg1, *args, **kwargs):
+        arg1 = np.asarray(arg1)
+        if len(arg1.shape) == 1:
```
bdice: Use `arg1.ndim` instead of `len(arg1.shape)`

freud/box.pyx
```diff
@@ -33,6 +34,54 @@ logger = logging.getLogger(__name__)
 # _always_ do that, or you will have segfaults
 np.import_array()
 
+
+def _support_1d(func):
+    R"""Decorator for class methods that accept one dimensional input or
+    two dimensional input for the first argument.
+    """
+    @wraps(func)
+    def wrapper(self, arg1, *args, **kwargs):
```
bdice: Does this decorator work if you specify the input array as a keyword argument?

bdice (commented):
One question, one comment.

freud/box.pyx
jinsooihm: Nope.
```
def _support_1d(func):
    R"""Decorator for class methods that accept one dimensional input or
    two dimensional input for the first argument.
    """
    @wraps(func)
    def wrapper(self, *args, **kwargs):
        if len(args):
            arg1 = args[0]
        else:
            arg1 = kwargs[next(iter(kwargs))]
        arg1 = np.asarray(arg1)
        if arg1.ndim == 1:
            flat = True
            arg1 = np.atleast_2d(arg1)
        else:
            flat = False

        ret = func(self, arg1)
        if flat:
            return ret.squeeze()
        else:
            return ret
    return wrapper
```
is the only way I can think of to handle that case.

jinsooihm: For `_support_1d_2args`, it would be more complicated and might not worth making a decorator.

bdice: Seems like the only way to do better would be this method, which would break our Python 2 compatibility (the older Py2-compatible way is considered legacy/deprecated). Even if we didn't care about compatibility, this strikes me as over-engineered.
https://docs.python.org/3/library/inspect.html#inspect.getfullargspec

In the case that we can't support both positional and keyword arguments, I think we have to just write the logic directly (`flat = ...`; `if flat: squeeze...` etc.) into each of the functions... even though it removes the niceness of having a decorator. 😦 It would be silly (and API-breaking) to require either positional or keyword args from a user.

freud/box.pyx
```diff
@@ -223,26 +225,28 @@ cdef class Box:
...
+        fractions = freud.common.convert_array(fractions, shape=(None, 3))
 
-        if fractions.ndim == 1:
-            fractions[:] = self._makeCoordinates(fractions)
-        elif fractions.ndim == 2:
-            for i, f in enumerate(fractions):
-                fractions[i] = self._makeCoordinates(f)
-        return fractions
+        for i, f in enumerate(fractions):
+            fractions[i] = self._makeCoordinates(f)
```
bdice: Use `_makeCoordinates_many`.

freud/box.pyx
```diff
@@ -255,26 +259,26 @@ cdef class Box:
...
-            vecs[:] = self._makeFraction(vecs)
-        elif vecs.ndim == 2:
-            for i, vec in enumerate(vecs):
-                vecs[i] = self._makeFraction(vec)
-        return vecs
+        flatten = vecs.ndim == 1
+        vecs = np.atleast_2d(vecs)
+        vecs = freud.common.convert_array(vecs, shape=(None, 3))
+
+        vecs = self._makeFraction_many(vecs)
```
bdice: Please refactor to remove the `_makeFraction` and `_makeFraction_many` methods, just put that logic directly in here. Same for the other methods.

freud/box.pyx
```diff
@@ -381,26 +386,18 @@ cdef class Box:
...
-        elif vecs.ndim == 2:
-            vecs += imgs[:, [0]]*self.getLatticeVector(0)
-            vecs += imgs[:, [1]]*self.getLatticeVector(1)
-            if self.dimensions == 3:
-                vecs += imgs[:, [2]]*self.getLatticeVector(2)
-        return vecs
+        vecs += imgs[:, [0]]*self.getLatticeVector(0)
+        vecs += imgs[:, [1]]*self.getLatticeVector(1)
+        if self.dimensions == 3:
+            vecs += imgs[:, [2]]*self.getLatticeVector(2)
```
bdice: Does it work to write `imgs[:, 2]` instead? Avoid putting the single index in brackets if the other form works. Same for other indices above.

cpp/box/Box.h
```diff
@@ -200,6 +201,21 @@ class Box
         return v;
     }
 
+    //! Convert fractional coordinates into real coordinates in place
+    /*! \param vecs Vectors of fractional coordinates between 0 and 1 within
+     *         parallelpipedal box
```
bdice: typo: `parallelepipedal`

bdice (changes requested):
@jinsoo960 Please review the comments below. Also change the title and description / other messages of this PR so it matches what was actually done.

freud/box.pyx
```diff
@@ -223,26 +225,28 @@ cdef class Box:
...
     def _makeCoordinates(self, f):
         cdef const float[::1] l_vec = f
         cdef vec3[float] result = self.thisptr.makeCoordinates(
             <const vec3[float]&> l_vec[0])
         return [result.x, result.y, result.z]
 
+    def _makeCoordinates_many(self, vecs):
+        cdef const float[:, ::1] l_points = vecs
+        cdef unsigned int Np = l_points.shape[0]
+        with nogil:
```
bdice, line 232: I _think_ Cython automatically drops the GIL since the method is marked `nogil`. Please verify this in the Cython docs and then remove with `with nogil:` context manager.

freud/box.pyx
vyasr: I concur, see my overall PR comment.

freud/box.pyx
vyasr: Using the extra brackets preserves the 2D-ness of the array, which is necessary in this case to multiply each dimension correctly. It subsets an (N, 3) array to an (N, 1) array to multiply the lattice vector, which is of shape (3, ). Without the bracket it would try to multiple an array of shape (N, ) with an array of shape (3, ), which will fail.

freud/box.pyx
vyasr: I believe marking a function as `nogil` just indicates that it's safe to be called while the gil GIL is released, but does not release the GIL. However, this function is also not currently marked as `nogil`, unless I missed some history?

vyasr (changes requested):
I don't think we need to maintain so many methods, it introduces additional redundancy. I think we should remove all of the old C++ methods that only operated on single vectors, then rename the new ones to the old ones (remove the `_many`), and then reflect that on the Python side. We can also get rid of the helper functions in Python (like `_unwrap`) and just put the code in the main function (e.g. `unwrap`) like @bdice suggested in one of his comments.

freud/box.pyx
jinsooihm: According to Cython,
> Note The nogil function annotation declares that it is safe to call the function without the GIL. It is perfectly allowed to execute it while holding the GIL. The function does not in itself release the GIL if it is held by the caller.

so, I think we need `nogil`.

freud/box.pyx
jinsooihm: I actually just removed these lines since it just does what `unwrap` does in C++ side.

jinsooihm:
`makeCoordinates`, `makeFraction`, `getImage` and `unwrap` are by default operating on multiple vectors, and only these are exposed to Cython.
`makeCoordinatesSingle` and `makeFractionSingle` are in C++ side since they are used in other C++ functions. 
`wrap` is on a single vector, and there is `wrapMany` to operate on multiple vectors since `wrap` is used a lot in C++ right now. Once we implement `NeighborQuery` with distance, I believe some of these can be removed and we can consider making `wrap` to operate on multiple vectors and add `wrapSingle`

freud/box.pyx
```diff
@@ -15,6 +15,7 @@ from __future__ import print_function
 import warnings
 import numpy as np
 from collections import namedtuple
+from functools import wraps
```
vyasr: We're not using this anymore, are we?

freud/box.pyx
```diff
@@ -33,6 +34,7 @@ logger = logging.getLogger(__name__)
 # _always_ do that, or you will have segfaults
 np.import_array()
 
+
```
vyasr: Shouldn't need this, 2 lines is enough for flake8.

vyasr (changes requested):
This is almost there. If possible, I would like to remove the awkward "single" naming scheme by just overloading the functions. That's the more natural C++ solution, and in theory should work with Cython. If, in practice, you find that there are issues with exposing those to Cython, then let me know and we can just leave things as is.

jinsooihm:
> This is almost there. If possible, I would like to remove the awkward "single" naming scheme by just overloading the functions. That's the more natural C++ solution, and in theory should work with Cython. If, in practice, you find that there are issues with exposing those to Cython, then let me know and we can just leave things as is.

Oops, I forgot that such things existed (clearly, I have not been programming for long time). Overloading seems to work properly.

vyasr (approved):

PR #313, opened 2019-06-15
Update formatting of several repo documents and remove unused imports.

bdice:
## Description
A handful of minor fixes:

- Updated the formatting of the Contributor Agreement to match HOOMD-blue
- Clarified flake8 configuration to ignore fewer relevant errors
- Removed unused imports
- Specify that we use OneFlow for `freud` branches
- Standardized `repr` formatting

## Types of changes

## Checklist:

### DIFF
--- a/ContributorAgreement.md
+++ b/ContributorAgreement.md
@@ -1,29 +1,28 @@
-freud Contributor Agreement
+# freud Contributor Agreement
 
 These terms apply to your contribution to the freud Open Source Project ("Project") owned or managed by the Regents of the University of Michigan ("Michigan"), and set out the intellectual property rights you grant to Michigan in the contributed materials. If this contribution is on behalf of a company, the term "you" will also mean the company you identify below. If you agree to be bound by these terms, fill in the information requested below and provide your signature.
 
 1. The term "contribution" means any source code, object code, patch, tool, sample, graphic, specification, manual, documentation, or any other material posted or submitted by you to a project.
-1. With respect to any worldwide copyrights, or copyright applications and registrations, in your contribution:
+2. With respect to any worldwide copyrights, or copyright applications and registrations, in your contribution:
     * you hereby assign to Michigan joint ownership, and to the extent that such assignment is or becomes invalid, ineffective or unenforceable, you hereby grant to Michigan a perpetual, irrevocable, non-exclusive, worldwide, no-charge, royalty-free, unrestricted license to exercise all rights under those copyrights. This includes, at Michigan's option, the right to sublicense these same rights to third parties through multiple levels of sublicensees or other licensing arrangements;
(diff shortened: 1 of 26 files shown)


setup.cfg
```diff
@@ -7,7 +7,9 @@ message = Bump up to version {new_version}.
 [flake8]
 filename = *.py,*.pyx,*.pxi,*.pxd
 exclude = .eggs,*.egg,build,extern,doc/source/examples
-ignore = E901,E225,E226,E227,E999,W503,W504,F401
+ignore = E225,E226,E227,E999,W504
```
vyasr, line 10: Is there a good reason to ignore these? I'm generally in favor of enforcing white space around operators.

ContributorAgreement.md
```diff
@@ -1,29 +1,28 @@
-freud Contributor Agreement
```
vyasr, line 1: The changes to this file are fine, but this combination of changes seems somewhat odd. If we assume that nobody is ever rendering this file as markdown, then there's no reason to mark the first line as a header. If we assume that the lines are rendered, then explicitly numbering below is unnecessary because markdown does it for you, and it's generally better not to have to manually number things to avoid human error. Is there a reason you chose to make these changes (especially since this file is identical to the ones used in our group's other repos)?

vyasr (changes requested):

setup.cfg
bdice: Yes - `flake8` doesn't naturally support Cython syntax. `flake8` reports a lot of false positives in `pyx` and `pxd` files if we don't ignore this subset of errors/warnings. However, there were a few errors/warnings that I had originally ignored for other reasons (I didn't know enough at the time to resolve them) but they're fixed now. I want to **stop ignoring these**:
- `F401`: detect unused imports
- `W503`: Line break occurred before a binary operator
- `E901`: some syntax error? Not sure what this was triggering on.

bdice: (To build this list, I disabled all the flags and re-enabled the flags one at a time to check what each one did. That way I know this list is minimal, while rejecting the false positives that `flake8` finds at its default settings.)

ContributorAgreement.md
bdice: I'm changing this so that it _matches_ other group software. We applied an identical set of changes to HOOMD-blue and fresnel recently.

We _should_ assume this file is rendered as Markdown - it ends in `.md`. The re-numbering, while not strictly necessary for Markdown renderers, makes it more readable for plain text editors.

setup.cfg
vyasr: Ah right, it thinks pointers are multiplication. That makes sense.

vyasr (approved):

PR #317, opened 2019-06-19
Wrapper class for tbb::enumerable_thread_specific

### LINKED ISSUE #315: Gaussian density segmentation fault
**Describe the bug**
A clear and concise description of what the bug is and what you expected to happen.
`GaussianDensity.compute` method can give segmentation fault if it is called first with a 2D box and then called with a 3D box.
This is because the thread local arrays are only initialized newly if it does not already exists.
But the size of the array depends on the input `box` to `compute` method. 

**To Reproduce**
1. Start with a system with a box and particles like ...
2. Create an analysis method for ...
3. Call the function ...
`points` can be any valid points.
```
testBox = freud.box.Box.square(box_size)
diff.compute(testBox, points)

testBox = freud.box.Box.cube(box_size)
diff.compute(testBox, points)
```

**Error output**
If applicable, add terminal output or screenshots to help explain your problem.
```
bus error
```

 - OS: [e.g. macOS]
 - Version of Python [e.g. 3.7]
 - Version of freud [e.g. 1.0]

**Additional context**

### LINKED ISSUE #316: Gaussian Density memory leak
**Describe the bug**
`GaussianDensity` module has memory leak.
This is probably because `GaussianDensity` does not have `__dealloc__` defined.

**To Reproduce**
Running `tests/test_density_GaussianDensity.py` produces memory leak.

**Error output**
Screen Shot 2019-06-17 at 8 00 00 PM

jinsooihm:
## Description

Wrapper class for `tbb::enumerable_thread_specific<T *>` has been implemented with convenience functions, such as `reset`, `updateSize`, constructor and destructor.
The new class `ETSArrayWrapper<T>` still provides same interface as `tbb::enumerable_thread_specific<T *>` for `begin()`, `end()`, and `local()`, which are the only use cases in `freud`.
Also, `__dealloc__` has been added for `GaussianDensity`.
One thing to note is that the `imshow` method of `matplotlib` has memory leak issue.
Every instance of `tbb::enumerable_thread_specific` has been replaced except for Steinhardt classes.

## Motivation and Context

Less code duplication and abstraction of memory management.
Resolves #316. 
Resolves #315.

## How Has This Been Tested?

Added a new test in `test_density_GaussianDensity` when the size of the thread specific array changes. 
Ran tests locally with memory check.

## Types of changes

## Checklist:


cpp/util/ETSWrapper.h
```diff
@@ -0,0 +1,106 @@
...
+#define ETSWRAPPER_H
+
+#include <tbb/tbb.h>
+
+namespace freud { namespace util {
+
+//! Wrapper classfor enumerable_thread_specific<T*>
+/*! It is expected that default value for T is 0.
+ */
+template<typename T> class ETSArrayWrapper
```
bdice: I would prefer a name like `ThreadStorage` that is clearer to new readers. I don't think the abbreviation ETS will make sense unless a reader knows TBB well.

cpp/util/ETSWrapper.h
```diff
@@ -0,0 +1,106 @@
...
+    //! Destructor
+    ~ETSArrayWrapper()
+    {
+        deleteArray();
+    }
+
+    //! Update size of the thread local arrays
+    /*! \param size New size of the thread local arrays
+     */
+    void updateSize(unsigned int size)
```
bdice: Can we rename this to `resize`? It fits better with other C++ methods for doing essentially the same thing.

bdice (changes requested):
Some naming suggestions. @jinsoo960 Can you elaborate on the matplotlib memory leak you mentioned in the PR description?

bdice:
Also is there a reason not to update the Steinhardt classes?

jinsooihm:
@bdice 
Seems like it is an issue that people often have problem with,
`https://stackoverflow.com/questions/3623600/python-matplotlib-memory-not-being-released-when-specifying-figure-size`
Most of the suggestions say call `close()` on figure object, which I am not sure how to do in `_repr_png_()`, since if I call it, the image would not show.
I thought it can be implemented for Steinhardt classes once #242 gets done.

bdice:
@jinsoo960 Yes, we need to try to close figures. `matplotlib` also starts throwing warnings/errors if figures don't get closed - good catch. 😛 You should be able to close the figure after saving to the `BytesIO` object. Try this:
```python
def ax_to_bytes(ax):
    ...
    fig.savefig(f, format='png')
    fig.close()
    return f.getvalue()
```

We're going to have to maintain the existing Steinhardt classes for a while longer, since it's one of the most-used features. For now, it'd be good to keep them up to date with other broad code base changes.

jinsooihm:
@bdice 
Actually, it seems like the problem is not from `imshow`, but from `from mpl_toolkits.axes_grid1.colorbar import colorbar`. 
I actually removed this dependency, since it was giving weird error of empty plot when calling `_repr_png_()` for second time in Jupyter lab (this issue is briefly described in #318).
So, #318 fixes this issue.
In addition, I added there calling `fig.clf()` to clear the figure just in case.

cpp/util/ThreadStorage.h
```diff
@@ -0,0 +1,106 @@
...
+     */
+    T* makeNewEmptyArray(unsigned int size)
+    {
+        T* tmp = new T[size];
+        memset((void*) tmp, 0, sizeof(T) * size);
+        return tmp;
+    }
+
+    unsigned int m_size; //!< size of thread local array
+    tbb::enumerable_thread_specific<T*> array; //!< public to expose all functions
```
vyasr: Not sure what you intended with this comment.

cpp/util/ThreadStorage.h
```diff
@@ -0,0 +1,106 @@
+#ifndef THREADSTORAGE_H
+#define THREADSTORAGE_H
+
+#include <tbb/tbb.h>
+
+namespace freud { namespace util {
+
+//! Wrapper classfor enumerable_thread_specific<T*>
```
vyasr: "class for"

vyasr (changes requested):
There's one comment I didn't understand, so I didn't just fix the typo that I found, but otherwise I'm happy to merge this ASAP.

bdice:
@jinsoo960 Searching for `thread_specific` in the code base, I see a lot of places where we have iterators (some are `typedef`'d). See PMFT, several locality classes, RDF, CorrelationFunction, etc. Are these something that can be replaced with the ThreadStorage iterator?

bdice:
Please add changelog line: `Changed: Refactored C++ handling of thread-local storage.`

cpp/util/ThreadStorage.h
jinsooihm: Oops. It used to be public and I forgot to change the comment.

jinsooihm:
> @jinsoo960 Searching for `thread_specific` in the code base, I see a lot of places where we have iterators (some are `typedef`'d). See PMFT, several locality classes, RDF, CorrelationFunction, etc. Are these something that can be replaced with the ThreadStorage iterator?

I do not think we can or should use this wrapper class in `locality`, since they are not using `enumerable_thread_specific` container to store arrays, but just tuples.
I changed the iterator names in other cases from `tbb::enumerable_thread_specific<T*>::const_iterator` to `util::ThreadStorage<T>::const_iterator`.
I might prefer `auto` since it is a long name.

vyasr (approved):
I agree that this class is not suitable for the locality `enumerable_thread_specific<std::vector<..>>`. Cleaning up the iterators is good, though.

PR #318, opened 2019-06-19
Feature/repr png more

### LINKED ISSUE #314: Add more plottable analyses
There are a couple compute classes whose results can be easily visualized that were not handled in #302. I suggest the following:

- [GaussianDensity in 2D systems](https://freud.readthedocs.io/en/stable/examples/module_intros/Density-GaussianDensity.html) - use real X and Y coordinates, and check what happens if the 2D box is sheared.
- [LocalQl (and its variants) as a histogram](https://freud.readthedocs.io/en/stable/examples/module_intros/Order-Steinhardt.html) - let the X axis be automatically determined by the data

jinsooihm:
## Description

`_repr_png_()` and `plot()` have been added for `GaussianDensity` when 2D and `LocalQl`, `LocalQlNear`, `LocalWl`, and `LocalWlNear`.
Also, an issue of getting additional empty plot when using `PMFTXY2D` in Jupyter Lab has been fixed.

## Motivation and Context

Resolves #314.

## How Has This Been Tested?

Tests have been added and the images are available as screenshots.

Screenshots in #314.

## Types of changes

## Checklist:


freud/order.pyx
```diff
@@ -740,6 +740,31 @@ cdef class LocalQl(Compute):
...
+            ax (:class:`matplotlib.axes`): Axis to plot on. If :code:`None`,
+                make a new figure and axis. (Default value = :code:`None`)
+
+        Returns:
+            (:class:`matplotlib.axes`): Axis with the plot.
+        """
+        import plot
+        return plot.histogram_plot(self.Ql,
+                                   title="Local Ql",
+                                   xlabel=r"$Ql$",
```
bdice: Make the _x_ label like `"$Q_4$"` - insert the actual value of `l`. Same for the title, and for other methods.

freud/plot.py
```diff
@@ -142,20 +174,76 @@ def pmft_plot(pmft, ax=None):
...
-        cb = colorbar(im, cax=cax)
-        cb.set_label_text(r"$k_B T$")
+        cb = Colorbar(cax, im)
+        cb.set_label(r"$k_B T$")
+
         return ax
 
 
+def plot_density(density, box, ax=None):
+    """Helper function to density diagram.
```
bdice: "to _plot_ density diagram."

freud/plot.py
```diff
@@ -142,20 +174,76 @@ def pmft_plot(pmft, ax=None):
...
+
+    Args:
+        density (:class:`numpy.ndarray`):
+            Array containing density.
```
bdice, line 218: Specify array shape and/or dimensionality requirements.

freud/plot.py
```diff
@@ -142,20 +174,76 @@ def pmft_plot(pmft, ax=None):
...
+        box (:class:`freud.box.Box`):
+            Simulation box.
+        color_by_sides (bool):
```
bdice: Remove this.

freud/plot.py
```diff
@@ -142,20 +174,76 @@ def pmft_plot(pmft, ax=None):
...
+            If :code:`True`, color cells by the number of sides.
+            (Default value = :code:`False`)
+
+    Returns:
+        bytes: Byte representation of the diagram in png format if matplotlib
```
bdice: Fix.

bdice (changes requested):
(Sorry, I started reviewing this before it was marked as "ready for review" - my mistake.) Let's install `matplotlib` on CircleCI so that it can try to render the plots, just toss it into this PR. It will make our tests run more of the real code paths that we expect.

freud/order.pyx
```diff
@@ -740,6 +740,31 @@ cdef class LocalQl(Compute):
...
+        """Plot Ql.
+
+        Args:
+            ax (:class:`matplotlib.axes`): Axis to plot on. If :code:`None`,
+                make a new figure and axis. (Default value = :code:`None`)
+
+        Returns:
+            (:class:`matplotlib.axes`): Axis with the plot.
+        """
+        from plot import histogram_plot as hplot
```
bdice: Let's skip the abbreviation here. It's only used once and it's not a standard abbreviation like `np` for `numpy`. One reason that we try to avoid this is that it's hard to search the codebase for specific functions if they're abbreviated in unexpected ways. If you're having difficulty with long line lengths, just wrap to a new line after `return histogram_plot(`.

freud/plot.py
```diff
@@ -30,10 +42,12 @@ def bar_plot(x, height, title=None, xlabel=None, ylabel=None, ax=None):
...
         xlabel (str): Label of x axis. (Default value = :code:`None`).
         ylabel (str): Label of y axis. (Default value = :code:`None`).
+        ax (:class: matplotlib.Axes.axes): axes object to plot on.
+            If :code:`None`, make a new axes and figure object.
+            (Default value = :code:`None`).
 
     Returns:
-        bytes: Byte representation of the graph in png format if matplotlib is
-        available. Otherwise :code:`None`.
+        :class: matplotlib.Axes.axes: axes object with the diagram.
```
bdice: I think the class is `matplotlib.axes.Axes` ([docs](https://matplotlib.org/3.1.0/api/axes_api.html#matplotlib.axes.Axes)). To render this correctly, we need to do two things:
1. Fix the syntax everywhere to
```python
:class:`matplotlib.axes.Axes`
```
2. Add `matplotlib` to our documentation via [intersphinx](http://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html). See the bottom of `doc/source/conf.py`.

freud/plot.py
```diff
@@ -107,22 +125,58 @@ def line_plot(x, y, title=None, xlabel=None, ylabel=None, ax=None):
     return ax
 
 
+def histogram_plot(x, title=None, xlabel=None, ylabel=None, ax=None):
+    """Helper function to draw a histogram graph.
+
+    .. moduleauthor:: Jin Soo Ihm <jinihm@umich.edu>
+
+    Args:
+        x (list): x values of the line graph.
```
bdice: Fix docs.

freud/plot.py
```diff
@@ -142,20 +196,75 @@ def pmft_plot(pmft, ax=None):
         ax.yaxis.set_ticks([i for i in range(int(ylims[0]), int(ylims[1]+1))])
         ax.set_xlabel(r'$x$')
         ax.set_ylabel(r'$y$')
-
         ax.set_title('PMFT')
+
+        ax_divider = make_axes_locatable(ax)
+        cax = ax_divider.append_axes("right", size="7%", pad="10%")
```
bdice, line 200: This is a cool feature, I've struggled with sizing colorbars onto existing figures before.

freud/plot.py
```diff
@@ -142,20 +196,75 @@ def pmft_plot(pmft, ax=None):
...
+    if ax is None:
+        fig = Figure()
+        ax = fig.subplots()
+
+    xlims = (-box.Lx/2, box.Lx/2)
+    ylims = (-box.Ly/2, box.Ly/2)
+
+    ax.set_title('Gaussian Density')
+    ax.set_xlabel(r'$x$')
+    ax.set_xlabel(r'$x$')
```
bdice: ylabel, y

freud/plot.py
```diff
@@ -142,20 +196,75 @@ def pmft_plot(pmft, ax=None):
...
+
+    ax_divider = make_axes_locatable(ax)
+    cax = ax_divider.append_axes("right", size="7%", pad="10%")
+
+    im = ax.imshow(density, extent=[xlims[0], xlims[1], ylims[0], ylims[1]])
+
+    transform = matplotlib.transforms.Affine2D().skew(box.xy, 0)
```
bdice: For now, I don't think we want to transform this. In the future I'd like to fix the behavior of GaussianDensity and then skew it accordingly, see #319. Please just comment out the code that shears the image for now, and document that it's related to issue #319.

bdice (changes requested):
Suggested comments below.

freud/plot.py
jinsooihm: This is pretty weird. So there are two almost same classes, `mpl_toolkits.axes_grid1.colorbar.colorbar` and `mpl_toolkits.axes_grid1.colorbar.Colorbar`. 
The documentation says `colorbar` is a `pyplot` thin wrapper while `Colorbar` is not.
So I believe using `colorbar` as in this [example](https://matplotlib.org/3.1.0/gallery/axes_grid1/demo_colorbar_with_axes_divider.html) causes memory leak since it calls `matplotlib.pyplot`.

jinsooihm: I believe this is also the reason why the `_repr_png_()` for `PMFT2DXY` gave another empty plot, since the empty plot is probably from `matplotlib.pyplot`.

jinsooihm: This is the bug that existed, of having an empty figure below PMFT plot.
Screen Shot 2019-06-19 at 10 27 54 PM

And I think the below example shows that the bug was caused from `matplotlib.pyplot` called from `colorbar` since the `plt.plot()` is actually shown below the PMFT plot, where there used to be an empty figure.
Screen Shot 2019-06-19 at 10 28 01 PM

freud/order.pyx
```diff
@@ -740,6 +745,67 @@ cdef class LocalQl(Compute):
...
+        xlabel = r"{} ${}_{{{}}}$".format(mode_str, name, self.sph_l)
+        return plot.histogram_plot(plot_data,
+                                   title=title,
+                                   xlabel=xlabel,
+                                   ylabel=r"Number of particles",
+                                   ax=ax)
+
+    @Compute._computed_method(("compute", "computeNorm",
+                               "computeAve", "computeAveNorm"))
+    def plot(self, mode, ax=None):
```
bdice: Change the signature to `plot(self, ax=None, mode=None)` (swapped order).
Then
```python
if mode is None:
    mode = self.plot_mode
```
and update the docstring to say it defaults to the most recently computed attribute.

bdice: And similarly for the other classes.

freud/plot.py
```diff
@@ -9,13 +9,26 @@
 
 
 def ax_to_bytes(ax):
+    """Helper function to convert figure to png file.
+
+    .. moduleauthor:: Jin Soo Ihm <jinihm@umich.edu>
+
+    Args:
+        ax (:class: matplotlib.axes.Axes): axes object to plot on.
+            If :code:`None`, make a new axes and figure object.
```
bdice: Fix documentation.
```axes object to plot.```
No default.

File: freud/plot.py [hunk 14]
```diff
@@ -107,22 +126,57 @@ def line_plot(x, y, title=None, xlabel=None, ylabel=None, ax=None):
...
+    """Helper function to draw a histogram graph.
+
+    .. moduleauthor:: Jin Soo Ihm <jinihm@umich.edu>
+
+    Args:
+        values (list): values of the histogram.
+        title (str): Title of the graph. (Default value = :code:`None`).
+        xlabel (str): Label of x axis. (Default value = :code:`None`).
+        ylabel (str): Label of y axis. (Default value = :code:`None`).
+        ax (:class: matplotlib.axes.Axes): axes object to plot on.
```
bdice: Probably need backticks around the special class names, here and in other methods.
```rst
:class:`matplotlib.axes.Axes`
```

freud/plot.py
bdice: @jinsoo960 Good catch. If this is fixed, please resolve this conversation.

freud/plot.py
```diff
@@ -142,20 +196,76 @@ def pmft_plot(pmft, ax=None):
...
+
+    Returns:
+        :class: matplotlib.axes.Axes: axes object with the diagram.
+    """
+    if not MATPLOTLIB:
+        return None
+    try:
+        from mpl_toolkits.axes_grid1.axes_divider import make_axes_locatable
+        from matplotlib.colorbar import Colorbar
+        # import matplotlib.transforms
```
bdice: Let's remove this and the other commented code - I have changed my mind about #319 for now. I'm no longer certain that we'll need this feature. We can always pull it back from the git history if we need it someday.

bdice (changes requested):
Some suggested changes below. Thanks!

bdice (approved):

PR #321, opened 2019-06-21
Feature/parallelize sol liq

csadorf:
## Description

Parallelize the SolLiq order parameter calculations.

## Motivation and Context

The OP is currently not parallelized.

## How Has This Been Tested?

Existing unit tests and manual performance tests.

## Types of changes

## Checklist:


csadorf:
@vyasr The function performs fine in my manual tests, however the attribute unit tests fail. I could not determine what's wrong. Can you have a look at this?

csadorf:
Also, I have no idea why this branch is out-of-date with base. I just rebased on master and it says it's up-to-date.

bdice:
@csadorf Thanks for doing this! 🚀

jinsooihm:
@csadorf 
The issue is coming from fact that `computeClustersQdotNoNorm` did not call `resize()` on `m_qldot_ij` so we get a segfault.
I did few testing and the reason for failure at `test_attribute_access` but not at `test_identical_environments` is because the `SolLiq` object was getting initialized newly every time in the loop for `test_attribute_access`, but `test_identical_environments` calls `computeSolLiqNoNorm` after calling `compute` on the same `SolLiq` object.

csadorf:
> @csadorf
> The issue is coming from fact that `computeClustersQdotNoNorm` did not call `resize()` on `m_qldot_ij` so we get a segfault.
> I did few testing and the reason for failure at `test_attribute_access` but not at `test_identical_environments` is because the `SolLiq` object was getting initialized newly every time in the loop for `test_attribute_access`, but `test_identical_environments` calls `computeSolLiqNoNorm` after calling `compute` on the same `SolLiq` object.

@jinsoo960 Thank you very much for identifying the bug!!

csadorf:
The benchmark failures seem unrelated as far as I can tell.

cpp/order/SolLiq.cc
```diff
@@ -168,65 +173,69 @@ void SolLiq::computeClustersQdot(const locality::NeighborList* nlist, const vec3
...
+                          Qlmjnorm = sqrt(Qlmjnorm);
+                          Qdot = Qdot / real((Qlminorm * Qlmjnorm));
+                          m_qldot_ij[bond] = Qdot;
+                          // Check if we're bonded via the threshold criterion
+                          if (real(Qdot) > m_Qthreshold)
+                          {
+                              // Tick up counts of number of connections these particles
+                              // have
+                              m_number_of_connections.get()[i]++;
+                              m_number_of_connections.get()[j]++;
```
vyasr: I don't think this line is thread-safe. You could be adding to particle `j` on multiple different threads (since any `i` could have a given `j` as a neighbor). You would need to make an `enumerable_thread_specific<unsigned int *>` of size `Np`, then add to that on each thread, and then perform a reduction afterwards to get the right number of connections.

This task is much easier now than it used to be thanks to recent work by @jinsoo960. You can use the `util::ThreadStorage` class so that you don't need to know how to work with the relevant tbb class. For an example, see the RDF class. In particular, the following places are relevant: the declaration of [`m_local_bin_counts`](https://github.com/glotzerlab/freud/blob/master/cpp/density/RDF.h#L75); the way [it's initialized](https://github.com/glotzerlab/freud/blob/master/cpp/density/RDF.cc#L64); its usage [during accumulation](https://github.com/glotzerlab/freud/blob/master/cpp/density/RDF.cc#L251); [the reduction process](https://github.com/glotzerlab/freud/blob/master/cpp/density/RDF.cc#L110); and the [reset](https://github.com/glotzerlab/freud/blob/master/cpp/density/RDF.cc#L184). You should be able to copy that pattern almost exactly here to get thread-safe behavior here.

cpp/order/SolLiq.cc
```diff
@@ -238,47 +247,49 @@ void SolLiq::computeClustersQdotNoNorm(const locality::NeighborList* nlist, cons
...
+                          }
+                          m_qldot_ij[bond] = Qdot;
+                          //
+                          // Check if we're bonded via the threshold criterion
+                          if (real(Qdot) > m_Qthreshold)
+                          {
+                              // Tick up counts of number of connections these particles
+                              // have
+                              m_number_of_connections.get()[i]++;
+                              m_number_of_connections.get()[j]++;
```
vyasr: Same as above.

vyasr (changes requested):
Thanks for the contribution @csadorf! Most of the conversion looks right, but there is one issue with thread-safety that concerns me. Once we fix that, we can go ahead and merge this PR.

cpp/order/SolLiq.cc
csadorf: The reduction step for the RDF is also parallelized. That's not necessary in this case, right?

Edit: Thinking a bit harder about this, I could parallelize over the number of particles.

csadorf:
@vyasr Addressed the change requests! Would appreciate if you could review it once more, but should be ready to merge now.

bdice (approved):
This looks good to me.

PR #328, opened 2019-06-27
API changes mainly related to passing points=None

jinsooihm:
## Description

No more `exclude_ii=None` input for `make_default_nlist` and `make_default_nlist_nn`.
Also, some inconsistent names or orders for parameters were changed, for example, using `refPoints` instead of `ref_points`.
Tests have been changed.

MatchEnv
  - refPoints renamed to ref_points
  - mathcMotif
    - order of arguments changed from (points, refPoints, ...) to (ref_points, points, ...)
  - minRMSDMotif
    - order of arguments changed from (points, refPoints, ...) to (ref_points, points, ...)

LocalDescriptor
  - compute
    - renamed points_ref to ref_points
    - exclude_ii set to points is None now

## TODO Help Wanted
Many more tests have to be added so that the cases when `points` is different from `ref_points` are covered. Most of the current tests only use same set of points for both `points` and `ref_points`.

## Motivation and Context

Many `compute` methods allow the user to pass in `points=None` to use `ref_points` for points.
However, when computing the neighbor list using `make_default_nlist` or `make_default_nlist_nn`, the `exclude_ii` logic is not robust against malicious inputs, such as passing in `points=ref_points.copy()`, since when `exclude_ii=None`. This is because the neighbor list computing functions claim to use whether `ref_points` and `points` point to the same address (they were actually not even using them properly).
This problem can get worse if we start using `NeighborQuery` API since it is often the case that the address of the `ref_points` contained in `NeighborQuery` object is different from the input to the `ref_points`. 
Therefore, as a small step towards the API change, it should be now explicitly stated that passing in `points` as "same" set of points as `ref_points` may result in an undefined behavior, and that `compute` methods will assume that if `points` is not `None`, it will be considered as different points from `ref_points`

## How Has This Been Tested?

Passes all tests.

## Types of changes

## Checklist:


vyasr:
All of these changes look to be incorporated in #295, so we can close this PR, right @jinsoo960?

jinsooihm:
@vyasr Yeah, this can be closed.

vyasr:
This PR is subsumed by #295

PR #333, opened 2019-06-29
Fix cubatic replicates.

### LINKED ISSUE #325: CubaticOrderParameter replicates not behaving as expected.
**Describe the bug**
The number of replicates of CubaticOrderParameter gets set to 1 after the first `compute` call,
so that the result of first run and subsequent runs are different.

**To Reproduce**
```
t_initial = 5.0
t_final = 0.001
scale = 0.95
n_replicates = 10
seed = 0
cop = freud.order.CubaticOrderParameter(t_initial, t_final, scale, n_replicates, seed)
cop.compute(orientations)
print(cop.cubatic_order_parameter)
cop.compute(orientations)
print(cop.cubatic_order_parameter)
cop.compute(orientations)
print(cop.cubatic_order_parameter)
cop.compute(orientations)
print(cop.cubatic_order_parameter)
cop.compute(orientations)
print(cop.cubatic_order_parameter)
```

**Error output**
If applicable, add terminal output or screenshots to help explain your problem.
```
0.9993200898170471
0.9974551796913147
0.9974551796913147
0.9974551796913147
0.9974551796913147
```

 - Version of freud: 1.1.0

**Additional context**
It seems that the `CubaticOrderParameter.compute` is always passing in `1` for `replicates` into the C++ `CubaticOrderParameter::compute`.
In the first run, `m_replicates` is set by the constructor, so it may not be `1` and thus uses proper value. At the end of `compute`, `m_replicates` is assigned the given value for `replicates`, which is always `1` in the current implementation.
In the subsequent runs, `m_replicates` is fixed to be `1` forever.

It would make more sense to not have `m_replicates` initialized in the constructor, but to have it actually passed in from `compute`.

bdice:
## Description
The number of replicates used by the Cubatic Order Parameter was unintentionally reset after the first compute call.

## Motivation and Context
Resolves #325.

## How Has This Been Tested?
@jinsoo960 did you identify the original error by looking at the runtimes? I don't think there is a way we can make this appear in the freud tests, since the Python API behavior did not reflect this problem.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,6 +2,11 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## next
+
+### Fixed
+* Repeated Cubatic Order Parameter computations use the correct number of replicates.
+
 ## v1.2.0 - 2019-06-27
 
 ### Added
--- a/cpp/order/CubaticOrderParameter.cc
+++ b/cpp/order/CubaticOrderParameter.cc
@@ -157,8 +157,8 @@ quat<float> CubaticOrderParameter::calcRandomQuaternion(Saru& saru, float angle_
     return quat<float>::fromAxisAngle(axis, angle);
 }
 
-void CubaticOrderParameter::compute(quat<float> *orientations, unsigned int n)
-    {
+void CubaticOrderParameter::compute(quat<float>* orientations, unsigned int n)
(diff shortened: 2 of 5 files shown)


jinsooihm:
@bdice  It works now. I was trying to figure out why `CubaticOrderParameter` had fluctuating runtimes, so I think I was printing how many times some inner loops were being executed (I cannot remember exactly what I did). And I noticed that the first run is different from subsequent runs.

vyasr (approved):

PR #337, opened 2019-07-05
Adding more tests

jinsooihm:
## Description

Added more tests and removed some redundancy. 
Some tests were added for 
- `ComplexCF`
- `FloatCF`
- `RDF`
- `BondOrder`
- `PMFTR12`
- `PMFTXY2D`
- `PMFTXYT`

on some fabricated systems to test for when `ref_points` and `points` are different.

Help is needed from other people.

## Motivation and Context

Some modules do not have enough tests. For example, `LocalDescriptors` only test for the shapes of the output, not the actual value.
Also, most of the modules that accept `ref_point` and `point` do not test for when they are not same, so we cannot really test if we break something when we move to the new neighbor list API.

## How Has This Been Tested?

Added more tests.

## Types of changes

## Checklist:


tests/util.py
```diff
@@ -2,6 +2,85 @@
 import freud
 
 
+def make_box_and_random_points(box_size, num_points, is2D=False, seed=0):
+    R"""Helper function to make random points with a box.
```
vyasr: Can you just add a note that the box is always square/cube?

tests/util.py
```diff
@@ -2,6 +2,85 @@
...
+
+    # make alternating lattice of points_2 and points_1
+    for i in range(-extra_shell, lattice_size + extra_shell):
+        for j in range(-extra_shell, lattice_size + extra_shell):
+            p = np.array([i, j, 0])
+            p = rotation_matrix.dot(p)
+            if (i + j) % 2 == 0:
+                points_2.append(p)
+            else:
+                if 0 <= i < lattice_size and 0 <= j < lattice_size:
```
vyasr: Just make this `elif`.

tests/test_density_ComplexCF.py
```diff
@@ -188,10 +173,57 @@ def test_repr_png(self):
...
+        box = freud.box.Box.square(box_size)
+
+        ocf = freud.density.ComplexCF(rmax, dr)
+
+        points = []
+        values = []
+        supposed_RDF = []
+        N = 300
+
+        # We are essentially generating all n-th roots of unities
```
vyasr: `unity`

tests/test_density_ComplexCF.py
```diff
@@ -188,10 +173,57 @@ def test_repr_png(self):
...
+        # scalar multiplied by the each bin centers
+        # with the value of a point being its complex number representation.
+        # Therefore, the RDF should be uniformly zero
+        # since the roots of unities add up to zero, if we set our ref_point in
+        # the origin.
+        # Nice proof for this fact is that when the n-th roots of unities
```
vyasr: unity

tests/test_density_ComplexCF.py
```diff
@@ -188,10 +173,57 @@ def test_repr_png(self):
...
+        rmax = 10.0
+        dr = 0.1
+        box_size = rmax*5
+        box = freud.box.Box.square(box_size)
+
+        ocf = freud.density.ComplexCF(rmax, dr)
+
+        points = []
+        values = []
+        supposed_RDF = []
```
vyasr: Delete here, it's reassigned later

tests/test_density_LocalDensity.py
```diff
@@ -65,6 +75,24 @@ def test_refpoints(self):
...
+        diameter = 1
+        rcut = 2
+
+        v_around = 4/3 * (rcut**3) * np.pi
+
+        ld = freud.density.LocalDensity(rcut, volume, diameter)
+        ld.compute(box, ref_points, points)
+
+        cd0 = 2/v_around
+        cd1 = (1 + getFraction(np.sqrt(5), rcut, diameter))/v_around
```
vyasr: Can you replace `np.sqrt(5)` with `np.linalg.norm(ref_points[1] - points[1])`? Otherwise the 5 is a bit of a magic number.

tests/test_environment_BondOrder.py
```diff
@@ -96,6 +96,35 @@ def test_repr(self):
...
         self.assertEqual(str(bo), str(eval(repr(bo))))
 
+    def test_ref_points_ne_points(self):
+        lattice_size = 10
+        # big box to ignore periodicity
+        box = freud.box.Box.square(lattice_size*5)
+        angle = np.pi/30
+        ref_points, points = util.make_alternating_lattice(lattice_size, angle)
+
+        # actually not used
```
vyasr, line 106: Is this indicating that these values aren't actually used by BondOrder? Because you are still passing them in.

tests/test_environment_BondOrder.py
```diff
@@ -96,6 +96,35 @@ def test_repr(self):
...
+        rmax = 1.6
+        k = 0
+
+        # to make sure what we skip ref_points ref_points neighbors
```
vyasr: I think this comment is outdated, you're not doing anything to avoid ref_points that are neighbors of other ref_points, are you? In case the comment is still relevant, there's a typo, what->that.

tests/test_environment_BondOrder.py
```diff
@@ -96,6 +96,35 @@ def test_repr(self):
...
+        # orientations are not used in bod mode
+        ref_orientations = np.array([[1, 0, 0, 0]]*len(ref_points))
+        orientations = np.array([[1, 0, 0, 0]]*len(points))
+
+        bod.compute(box, ref_points, ref_orientations, points, orientations)
+
+        # we want to make sure that we get 12 nonzero places, so we can test
+        # whether we are not considering neighbors between ref_points
+        self.assertEqual(np.count_nonzero(bod.bond_order), 12)
+        self.assertEqual(len(np.unique(bod.bond_order)), 2)
```
vyasr, line 125: Let's add some extra checks for what happens if we call compute with swapped points and ref_points, none of these invariants should hold in that case.

tests/test_locality_NeighborQuery.py
```diff
@@ -205,10 +206,8 @@ def test_query_ball_to_nlist(self):
         L = 10  # Box Dimensions
         N = 400  # number of particles
 
-        box = freud.box.Box.cube(L)
-        np.random.seed(0)
-        ref_points = np.random.rand(N, 3)*L
-        points = np.random.rand(N, 3)*L
+        box, ref_points = make_box_and_random_points(L, N, seed=0)
+        _, points = make_box_and_random_points(L, N, seed=1)
```
vyasr: This is a little awkward, can you just call `np.random.rand(N, 3)*L` for the points arrays in this file instead?

vyasr (changes requested):
Some small changes to be made, then I think this is ready to merge.

tests/test_environment_BondOrder.py
jinsooihm: They are not used by `BondOrder` internally, whatever values are given for the constructor.

tests/test_environment_BondOrder.py
jinsooihm: Should I just do `assertNotEqual`? It seems that the actual values are not easy to compute by hand in those cases.

vyasr (approved):

PR #338, opened 2019-07-07
Fix LocalQlNear filter

### LINKED ISSUE #297: The LocalQlNear order parameter does not behave as expected
**Describe the bug**

The `LocalQlNear` order parameter according to the documentation promises:

> The number of included neighbors is determined by the kn parameter to the constructor.

That implies that the `rmax` value is a hint and should be expanded to find the actual number of required neighbors. The intended behavior has been confirmed by @xiyudu 
 in private communication. However, in reality the compute function on the C++ side uses a hard cut-off:

https://github.com/glotzerlab/freud/blob/master/cpp/order/LocalQl.cc#L89

**To Reproduce**

Compute the order parameter with a `rmax` value that is lower than the average bond length of the expected number of neighbors (`kn`). The value will substantially deviate in comparison to using a large `rmax` value, *e.g.*, half of the minimal box length.

 - OS: n/a
 - Version of Python: 3.7
 - Version of freud: 1.0

### LINKED ISSUE #344: Steinhardt averaged values are not calculated correctly
**Describe the bug**
In commit f51e3ac693847629e19444b5638dc1e94341d6c5, I introduced an error while refactoring the LocalQl code, by accidentally using the array `m_Qlmi` instead of `m_AveQlmi`. This change was [merged in this pull request on Bitbucket](https://bitbucket.org/glotzer/freud/pull-requests/249/clean-up-localql-and-localwl-preparing-to/diff) and released in v1.0.0. This affects the correctness of all results using the averaged Steinhardt parameters from v1.0.0 to present (v1.2.0)

See change: https://github.com/glotzerlab/freud/commit/f51e3ac693847629e19444b5638dc1e94341d6c5#diff-eaeab4497c6a5dc2521bc1c86d6160e4R197

This was discovered while testing #338 and will be resolved by #338.

**To Reproduce**
1. Start with an FCC system.
2. Perturb one particle position.
3. The averaged Ql parameters should be different for more than just the first neighbor shell of 12 neighbors.
4. The newly added tests in #338 catch this bug.

vyasr:
## Description
I removed the rmax/rmin filtering logic that's currently in the LocalQl.cc file.

## Motivation and Context
This resolves #297, resolves #344. The reason for the changes to the Python code as well as C++ is to ensure that the `rmin` argument to the `LocalQl` class is still respected. Since `LocalQlNear`always sets  `rmin=0`, I just filtered the `NeighborList` for all the `compute*` functions for `LocalQl`. While this isn't the most elegant implementation, since these classes are soon to be deprecated I think we should just get this fix in ASAP and worry about the newer code paths.

@b-butler can you check update the new `_Steinhardt` class to match this logic (in case it's not already updated)? Do that in a separate PR, and tag @jinsoo960 as a reviewer to make sure that the change will be compatible with the new `NeighborQuery` back-end when that finally gets merged into a release.

@csadorf can you check that this fixes your problem?

## How Has This Been Tested?
Existing tests pass. We will need to add new tests, but I'd prefer to do that on the new classes so that we can think about how these behaviors should work in the context of swapping between `NeighborList` and `NeighborQuery` objects.

## Types of changes

## Checklist:


csadorf:
>  Bug fix (non-breaking change which fixes an issue)

This is a breaking change, isn't it?

bdice:
> > Bug fix (non-breaking change which fixes an issue)
> 
> This is a breaking change, isn't it?

The current behavior is already “broken” from its original implementation. This fixes it. I checked the box for “breaking change” but we won’t wait to release it.

bdice (approved):
Looks good to me.

csadorf:
> > > Bug fix (non-breaking change which fixes an issue)
> > 
> > 
> > This is a breaking change, isn't it?
> 
> The current behavior is already “broken” from its original implementation. This fixes it. I checked the box for “breaking change” but we won’t wait to release it.

I'm not implying that you should wait, but you should transparently communicate to your users that previous results obtained with this class may be invalid.

bdice:
@csadorf I propose this changelog line, "Fixed filtering of neighbors in LocalQlNear and LocalWlNear to correct invalid results for some values of rmax." Is that sufficient communication?

vyasr:
I agree that we need to be clear in our release notes about the fact that this will change behavior, but also that this needs to get out ASAP. I assume that because this is a serious bug that is also of immediate use (at least to Simon), we will be making a bug fix release in which the issue will be highlighted.

csadorf:
> @csadorf I propose this changelog line, "Fixed filtering of neighbors in LocalQlNear and LocalWlNear to correct invalid results for some values of rmax." Is that sufficient communication?

I don't think that the wording is sufficiently transparent. I propose something along the lines of:

> Fixed critical bug resulting in the `LocalQlNear` and `LocalWlNear` class wrongly using a hard instead of a soft cut-off which may have led to calculations potentially being based on the incorrect number of neighbors. This would cause incorrect results especially for systems with an average *n*th nearest-neighbor distance smaller than `rmax`. This problem persisted from versions X.X.X to Y.Y.Y.

This kind of bug may have led to incorrect results that might even require an erratum depending on how critical data was for a given publication.

csadorf (changes requested):
I'd like to see the changelog message be a bit more descriptive.

bdice:
@vyasr @csadorf I have added tests and fixed everything I can think of. Note that I found another bug, #344, and resolved it. Let's see if we can get this reviewed/finalized and make a bugfix release in the next few days.

cpp/order/Steinhardt.cc
```diff
@@ -285,7 +285,7 @@ void Steinhardt::computeAve(const box::Box& box, const locality::NeighborList* n
                 m_QlmiAve.get()[index] /= neighborcount;
                 m_Qlm_local.local()[k] += m_QlmiAve.get()[index] / float(m_Np);
                 // Add the norm, which is the complex squared magnitude
-                m_QliAve.get()[i] += norm(m_Qlmi.get()[index]);
+                m_QliAve.get()[i] += norm(m_QlmiAve.get()[index]);
```
bdice, line 288: This is the line that resolves #344. See that issue for details.

tests/test_order_LocalWl.py
```diff
@@ -116,15 +130,16 @@ def test_shape(self):
...
+        comp = freud.order.LocalWlNear(box, 0.1, 6, 12)
         comp.compute(positions)
 
         npt.assert_equal(comp.Wl.shape[0], N)
 
     def test_identical_environments(self):
         (box, positions) = util.make_fcc(4, 4, 4)
 
-        comp = freud.order.LocalWlNear(box, 1.5, 6, 12)
+        # Use a really small rcut to ensure that it is used as a soft cutoff
```
csadorf: Good call! 👍 

Maybe add more of those comments to ensure that nobody changes that back at some point?

tests/test_order_LocalWl.py
```diff
@@ -4,6 +4,8 @@
 import unittest
 import util
 
+PERFECT_FCC_W6 = -0.00262604
```
csadorf: Where does this value come from?

tests/test_order_LocalQl.py
```diff
@@ -4,6 +4,8 @@
 import unittest
 import util
 
+PERFECT_FCC_Q6 = 0.57452416
```
csadorf: Where does this value come from?

cpp/order/Steinhardt.cc
csadorf: Are there potentially more areas where the same error might have been made?

csadorf (commented):
Not tested yet, but code looks good to me so far. I have a few questions.

tests/test_order_LocalQl.py
bdice: @csadorf This is a test against a known value for an FCC system. We have verified this value by manually calculating it and using other packages like `pyboo`:

```python
from util import make_fcc
import boo
import freud
import numpy as np


box, positions = make_fcc(8, 8, 8)
nlist = freud.locality.AABBQuery(box, positions).query(positions, k=12, exclude_ii=True).toNList()

q6m = boo.bonds2qlm(positions, nlist[:], l=6)
q6 = boo.ql(q6m)
w6 = boo.wl(q6m)

# Using the median will ignore the surface values,
# since pyboo can't natively handle periodic systems
print(np.median(q6))  # 0.5745242597140698
print(np.median(w6))  # -0.0026260383340077588
```

tests/test_order_LocalWl.py
bdice: See comment on `PERFECT_FCC_Q6`.

cpp/order/Steinhardt.cc
bdice: I looked back through the current code and the PR that introduced this error. I did not find any other copy-paste errors. The current code has been massively simplified, so there are fewer places such a mistake could hide.

csadorf (commented):
@vyasr Why did you remove the comment? I thought that was helpful.

vyasr:
@csadorf It was copied into multiple tests where it wasn't relevant. I left it in place in the only test where it actually mattered (the `test_identical_environments` tests). I removed it from the `repr` tests where it was copied.

csadorf:
There is currently **no** changelog entry for this bug, is there?

bdice:
> There is currently **no** changelog entry for this bug, is there?

You're right, I think that was an oversight. I'll fix this before releasing.

PR #339, opened 2019-07-07
Improve docstring for MSD

### LINKED ISSUE #303: MSD documentation is ambiguous on definition of N
Depending on the choice of mode, there may or may not be an averaging over frames in addition to an averaging over particles. The module documentation should be clearer about what the variable `N` is and which average is being performed.

vyasr:
Resolves #303.


bdice (approved):
Good!

PR #341, opened 2019-07-16
Refactor neighbor list order

jinsooihm:
## Description

Changed order of `ref_points` and `points` in Neighbor list and relevant modules.
It seems to work for now, but we would need more thorough testing to make sure.

## Motivation and Context

New API. Also required for #295.

## How Has This Been Tested?

Passes current tests. Some tests were modified. 

## Types of changes

## Checklist:


cpp/environment/LocalBondProjection.cc
```diff
@@ -90,21 +90,20 @@ void LocalBondProjection::compute(box::Box& box, const freud::locality::Neighbor
...
             {
                 const size_t j(neighbor_list[2 * bond + 1]);
+
                 // compute bond vector between the two particles
-                vec3<float> delta = box.wrap(pos[j] - p);
+                vec3<float> delta = box.wrap(pos[i] - ref_pos[j]);
                 vec3<float> local_bond(delta);
                 // rotate bond vector into the local frame of particle p
-                local_bond = rotate(conj(q), local_bond);
+                local_bond = rotate(conj(ref_ors[j]), local_bond);
```
vyasr, line 106: Let's leave this for now, but whether this should be `ref_ors[j]` or `ors[i]` seems relevant to #346.

cpp/environment/MatchEnv.cc
```diff
@@ -691,20 +691,23 @@ void MatchEnv::cluster(const freud::locality::NeighborList* env_nlist,
...
                 const size_t j(neighbor_list[2 * bond + 1]);
-                std::pair<rotmat3<float>, BiMap<unsigned int, unsigned int>> mapping
-                    = isSimilar(dj.s[i], dj.s[j], m_threshold_sq, registration);
-                rotmat3<float> rotation = mapping.first;
-                BiMap<unsigned int, unsigned int> vec_map = mapping.second;
-                // if the mapping between the vectors of the environments
-                // is NOT empty, then the environments are similar, so
-                // merge them.
-                if (!vec_map.empty())
+                if (i != j)
```
vyasr: Did you mean to add ii exclusion logic back in at this level?

cpp/environment/MatchEnv.h
```diff
@@ -104,6 +104,10 @@ class MatchEnv
     Environment buildEnv(const size_t* neighbor_list, size_t num_bonds, size_t& bond,
                          const vec3<float>* points, unsigned int i, unsigned int env_ind, bool hard_r);
 
+    //! Build EnvDisojintSet on given dj. 
+    //! This function is needed since new neighbor list is in points ref_points order.
+    EnvDisjointSet buildEnvDS(EnvDisjointSet& dj, const size_t* neighbor_list, size_t num_bonds, const vec3<float>* points);
```
vyasr: Where is this actually used? Is this a holdover, or is it used in a downstream PR?

cpp/locality/LinkCell.cc
```diff
@@ -170,27 +170,27 @@ void LinkCell::compute(const box::Box& box, const vec3<float>* ref_points, unsig
 {
```
vyasr, line 170: Once #349 is merged and we've recovered the performance, there's no need to keep the compute method around anymore, and it can be safely removed. For now, though, I'll go ahead and review these.

vyasr: As expected, the changes look fine, but the ordering may cause problems. Fortunately, on the next branch I think that's OK since we'll be getting rid of this soon enough.

cpp/locality/NearestNeighbors.cc
```diff
@@ -51,18 +51,18 @@ void NearestNeighbors::compute(const box::Box& box, const vec3<float>* ref_pos,
                                const vec3<float>* pos, unsigned int num_points, bool exclude_ii)
```
vyasr, line 51: This is another class that we plan to remove. We should probably do some benchmarking to compare this class against the query methods for AABBQuery and LinkCell first, but then I think we're in good shape to get rid of this as well.

cpp/locality/NeighborList.cc
```diff
@@ -2,6 +2,7 @@
 // This file is from the freud project, released under the BSD 3-Clause License.
 
 #include "NeighborList.h"
+#include <iostream>
```
vyasr: Needs to be removed.

vyasr (commented):
Most of the changes here look fine, but they're also clearly incomplete, and it makes it difficult to really review for correctness. I propose that once my change requests here are addressed, we merge these changes into feature/neighbor_iterator, change the target of that PR to next, and close this pull request in favor of that one.

vyasr:
Some requests that didn't fit into the review:
- The `loopOverNeighbors` function should validate the `NeighborList` before calling through to `loopOverNeighborList`.
- We should standardize the internal naming of the Np and Nref, but this is about the lowest priority item possible; I just wanted to note it, not requesting that it happen in this PR or any other PR in the foreseeable future.

cpp/environment/MatchEnv.cc
jinsooihm: I forgot to remove it.

cpp/environment/MatchEnv.h
jinsooihm: Some initialization used to get messed up when there was some error in the change of order `ref_points` and `points` and how the neighbor list was not sorted properly. It seems to work without it now.

vyasr:
This PR is superseded by #342

PR #342, opened 2019-07-17
More Neighbor Query iteration.

jinsooihm:
## Description

This branch assumes that #341 is correct.
NeighborQuery iterating logic is implemented for 
- `Steinhardt`
- `HexOrderParameter`
- `TransOrderParameter`
- `LocalDensity`

as mentioned in #295.
This is done with a new iterator interface.
Also, since `HexOrderParameter` and `TransOrderParameter` are so similar, they are now children of a common base class, and their own files were removed.

## Motivation and Context

More efficient when one does not want to build a whole neighbor list.

## How Has This Been Tested?

Passes current tests, but the actual correctness will depend on whether #341 is correct or not.

## Types of changes

## Checklist:


vyasr:
One note before I forget, but will be reviewing this thoroughly shortly: we should swap the internal order of the arguments passed to `loopOverNeighbors` to match what the API looks like, once we're happy with these other change. I'm fine with doing that in a separate PR, though.

cpp/density/LocalDensity.cc
```diff
@@ -3,6 +3,7 @@
 
 #include <complex>
 #include <stdexcept>
+#include <iostream>
```
vyasr: Remove.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -9,7 +9,180 @@
...
+        virtual bool end() const
+        {
+            return (m_returned_point_index != m_point_index) || at_end;
+        }
+
+    private:
+        const NeighborList* m_nlist;
+        size_t m_current_index;
+        size_t m_returned_point_index;
+        size_t m_point_index;
```
vyasr: Let's move this up to the parent, it's common and should be common since it's the only real item that's defined to be tracked by a NeighborIterator.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -9,7 +9,180 @@
...
+};
+
+//! NeighborIterator for NeighborQuery
+class NeighborQueryNeighborIterator : public NeighborIterator
+{
+public:
+    NeighborQueryNeighborIterator(const NeighborQuery* nq, const vec3<float> *points, unsigned int N, QueryArgs qargs)
+    {
+        m_qargs = qargs;
+        if(qargs.exclude_ii && (qargs.mode == QueryArgs::QueryType::nearest))
```
vyasr, line 117: Can you add a comment for why we do this? I can see this line be very confusing without context.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -94,6 +305,10 @@ template<typename ComputePairType>
 void loopOverNeighborQuery(const NeighborQuery* ref_points, const vec3<float>* points,
                            unsigned int Np, QueryArgs qargs, const ComputePairType& cf, bool parallel)
 {
+    if(qargs.exclude_ii && (qargs.mode == QueryArgs::QueryType::nearest))
```
vyasr, line 314: Just copy in whatever comment you write above into this.

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -9,7 +9,180 @@
 #include "NeighborList.h"
```
vyasr, line 9: The `loopOverNeighborList` function is currently wrong; since the order of the NeighborList has been reversed, you need to swap the arguments, otherwise you're currently calling `cf(point, ref_point...)` while the expected signature is `cf(ref_point, point...)`. As per my earlier comment on the PR (pre-review), I actually think the former signature makes more sense now that we've swapped the order, and I would like to change the `loopOverNeighbors` and `loopOverNeighborQuery` functions, but it might be easier for you to verify correctness for now by changing the `loopOverNeighborList` function and making sure everything works, then going back and changing everything in a follow-up. I'm OK with whichever version is less work for you @jinsoo960 .

cpp/locality/NeighborComputeFunctional.cc
```diff
@@ -0,0 +1,19 @@
+#include "NeighborComputeFunctional.h"
```
vyasr, line 1: Is there a reason that this needs to be in the cc file? Currently all the logic in the header, I'd prefer to keep it that way for consistency and split into two files later if needed. Is there something that isn't defined early enough in the header to put this in there?

cpp/locality/NeighborList.cc
```diff
@@ -2,6 +2,7 @@
 // This file is from the freud project, released under the BSD 3-Clause License.
 
 #include "NeighborList.h"
+#include <iostream>
```
vyasr: Remove.

cpp/environment/AngularSeparation.cc
```diff
@@ -77,7 +77,7 @@ void AngularSeparation::computeNeighbor(const freud::locality::NeighborList* nli
                                         const quat<float>* ref_equiv_ors, unsigned int Nref, unsigned int Np,
```
vyasr, line 77: Is there a reason we haven't update the `computeNeighbor` function to use the new `NeighborQuery` API? There are tests that at least check that behavior is correct in the queryBall case, and we can add new ones for the nearest neighbor-based querying.

cpp/environment/LocalDescriptors.cc
```diff
@@ -30,7 +30,7 @@ void LocalDescriptors::compute(const box::Box& box, const freud::locality::Neigh
                                const vec3<float>* r, unsigned int Np, const quat<float>* q_ref,
```
vyasr, line 30: I've written a test for this, let me push it and then we can adapt this code to the new method.

cpp/environment/MatchEnv.cc
```diff
@@ -691,20 +691,23 @@ void MatchEnv::cluster(const freud::locality::NeighborList* env_nlist,
...
                 const size_t j(neighbor_list[2 * bond + 1]);
-                std::pair<rotmat3<float>, BiMap<unsigned int, unsigned int>> mapping
-                    = isSimilar(dj.s[i], dj.s[j], m_threshold_sq, registration);
-                rotmat3<float> rotation = mapping.first;
-                BiMap<unsigned int, unsigned int> vec_map = mapping.second;
-                // if the mapping between the vectors of the environments
-                // is NOT empty, then the environments are similar, so
-                // merge them.
-                if (!vec_map.empty())
+                if (i != j)
```
vyasr: I also commented on this in the other PR: is this exclude_ii intentional?

cpp/environment/MatchEnv.h
```diff
@@ -104,6 +104,10 @@ class MatchEnv
     Environment buildEnv(const size_t* neighbor_list, size_t num_bonds, size_t& bond,
                          const vec3<float>* points, unsigned int i, unsigned int env_ind, bool hard_r);
 
+    //! Build EnvDisojintSet on given dj. 
+    //! This function is needed since new neighbor list is in points ref_points order.
+    EnvDisjointSet buildEnvDS(EnvDisjointSet& dj, const size_t* neighbor_list, size_t num_bonds, const vec3<float>* points);
```
vyasr: Also from the other PR: is this actually necessary?

cpp/order/OrderParameter.h
```diff
@@ -0,0 +1,167 @@
...
+    //         vec3<float> delta = m_box.wrap(points->getRefPoints()[j] - ref);
+
+    //         // Compute psi for neighboring particle
+    //         // (only constructed for 2d)
+    //         m_psi_array.get()[i] += func(delta);
+    //     },
+    //     [=](size_t i, int data)
+    //     {
+    //         m_psi_array.get()[i] /= std::complex<float>(m_k);
+    //     });
```
vyasr: Delete.

cpp/order/OrderParameter.h
```diff
@@ -0,0 +1,167 @@
...
+    //! Destructor
+    virtual ~OrderParameter() {}
+
+    //! Get the simulation box
+    const box::Box& getBox() const
+    {
+        return m_box;
+    }
+
+    //! Compute the hex order parameter
```
vyasr: Remove hex.

cpp/order/OrderParameter.h
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) 2010-2019 The Regents of the University of Michigan
+// This file is from the freud project, released under the BSD 3-Clause License.
+
+#ifndef ORDER_PARAMETER_H
+#define ORDER_PARAMETER_H
```
vyasr: We need to rename this to something less generic. As we've established, good naming is hard, and I don't want to waste your timing on coming up with good names, so for now I'm OK with literally naming the parent class HexTransOrderParameter or something equally banal and just getting things working.

cpp/order/OrderParameter.h
```diff
@@ -0,0 +1,167 @@
...
+    HexOrderParameter(unsigned int k = 6);
+
+    //! Destructor
+    ~HexOrderParameter();
+
+        //! Get a reference to the last computed psi
+    std::shared_ptr<std::complex<float>> getPsi()
+    {
+        return m_psi_array;
+    }
```
vyasr: Move this getter up to the parent class; since we're using a common array, I would remove the getDr function in TransOrderParameter and just use this function. Maybe rename it to `getOrder` or so. We definitely don't need to preserve API compatibility on the C++ side, so you can have the current Cython functions call the parent instead.

cpp/order/OrderParameter.h
```diff
@@ -0,0 +1,167 @@
...
+        unsigned int Np = points->getNRef();
+
+        // Reallocate the output array if it is not the right size
+        if (Np != m_Np)
+        {
+            m_psi_array = std::shared_ptr<std::complex<float>>(new std::complex<float>[Np],
+                                                          std::default_delete<std::complex<float>[]>());
+        }
+
+        freud::locality::loopOverNeighborsIterator(points, points->getRefPoints(), Np, qargs, nlist,
```
vyasr: For consistency, move this implementation into the cc file.

cpp/density/LocalDensity.cc
```diff
@@ -41,56 +41,38 @@ void LocalDensity::compute(const box::Box& box, const freud::locality::NeighborL
...
-                  // this is not particularly accurate for a single particle, but works well on average for
-                  // lots of them. It smooths out the neighbor count distributions and avoids noisy spikes
-                  // that obscure data
-                  num_neighbors += 1.0f + (m_rcut - (r + m_diameter/2.0f)) / m_diameter;
-              }
-          }
-
-          m_num_neighbors_array.get()[i] = num_neighbors;
-          if (m_box.is2D())
+    freud::locality::loopOverNeighborsIterator(ref_points, points, Np, qargs, nlist,
```
vyasr, line 43: While this looks right, I'm making a note here that we need to thoroughly review the ref_points/points ordering since it is not quite consistent everywhere yet.

cpp/order/Steinhardt.cc
```diff
@@ -2,6 +2,7 @@
 // This file is from the freud project, released under the BSD 3-Clause License.
```
vyasr, line 2: Since Steinhardt doesn't currently support two sets of points, these changes all seem fine; we'll just need to be extra careful with it when we do update it to enable that.

freud/_order.pxd
```diff
@@ -46,28 +46,26 @@ cdef extern from "NematicOrderParameter.h" namespace "freud::order":
         vec3[float] getNematicDirector()
 
 
-cdef extern from "HexOrderParameter.h" namespace "freud::order":
+cdef extern from "OrderParameter.h" namespace "freud::order":
```
vyasr: You'll need to change this to match whatever changes are made in C++.

freud/_order.pxd
```diff
@@ -46,28 +46,26 @@ cdef extern from "NematicOrderParameter.h" namespace "freud::order":
...
+                     const freud._locality.NeighborQuery*,
+                     freud._locality.QueryArgs) nogil except +
         # unsure how to pass back the std::complex,
         # but this seems to compile...
         shared_ptr[float complex] getPsi()
         unsigned int getNP()
         unsigned int getK()
 
-cdef extern from "TransOrderParameter.h" namespace "freud::order":
+cdef extern from "OrderParameter.h" namespace "freud::order":
```
vyasr: You don't need this line twice, as long as you keep the indent level the previous `cdef extern` just continues.

vyasr (changes requested):
This looks mostly right. In addition to my inline comments, I have three bigger picture things we need to address:
- We need to do a thorough survey of which classes implement the new neighbor API and which don't; I have not yet done that, but would like to look through it tomorrow.
- We need to develop a standard method for providing default query_args and define the appropriate resolution hierarchy.
- We need to identify exactly which methods still need tests so that we can write those. Between the two of us, I think we have substantial overlap but have also definitely added tests for many of the classes missing tests.

cpp/locality/NeighborComputeFunctional.cc
jinsooihm: There are some compilation errors for duplicate symbols if I have it in the header file.

cpp/environment/AngularSeparation.cc
jinsooihm: We need to know the neighbor list before hand. If we want to add nearest neighbor based querying, we can resolve that inside the Python side to find such neighbor list.

cpp/environment/MatchEnv.cc
jinsooihm: Nope, just forgot to remove.

cpp/locality/NeighborComputeFunctional.cc
vyasr: OK, yeah I figured that was the case. Never mind, then.

cpp/environment/AngularSeparation.cc
vyasr: Yes, that sounds fine to me.

cpp/environment/LocalDescriptors.cc
vyasr: The LocalDescriptors computation assumes knowledge of the number of bonds, so it will remain a computation on a NeighborList.

tests/test_environment_LocalBondProjection.py
```diff
@@ -197,18 +197,22 @@ def test_compute(self):
         # For the first particle, the bond is [1,0,0] and the orientation is
         # [1,0,0,0]. So the local bond is [1,0,0]. The projection onto
         # [0,0,1] is cos(pi/2)=0.
-        npt.assert_allclose(ang.projections[0], 0, atol=1e-6)
-        npt.assert_allclose(ang.normed_projections[0], 0, atol=1e-6)
+        # In the new neighbor list, the corresponding bond is at index 2
+        npt.assert_allclose(ang.projections[1], 0, atol=1e-6)
```
bdice: Comment doesn't seem to match index.

tests/test_environment_LocalBondProjection.py
```diff
@@ -197,18 +197,22 @@ def test_compute(self):
...
+        npt.assert_allclose(ang.projections[0], -1, atol=1e-6)
+        npt.assert_allclose(ang.normed_projections[0], -1, atol=1e-6)
         # For the third particle, the bond is [0,0,-1.5] and the orientation is
         # rotated about the z-axis by pi/2. So the local bond is [0,0,-1.5].
         # The projection onto [0,0,1] is 1.5*cos(pi)=-1.5
-        npt.assert_allclose(ang.projections[2], -1.5, atol=1e-6)
-        npt.assert_allclose(ang.normed_projections[2], -1, atol=1e-6)
+        # In the new neighbor list, the corresponding bond is at index 1
+        # and the bond becomes [0, 0, 1.5]
+        npt.assert_allclose(ang.projections[2], 1.5, atol=1e-6)
```
bdice: Comment doesn't seem to match index.

bdice (changes requested):
Mostly looks good. Please change back the util functions to `snake_case` instead of `camelCase`. Also check on the comments I left.

bdice:
Benchmarks of scipy vs. AABBQuery (NeighborQuery API) vs. LinkCell (old API) vs. LinkCell (NeighborQuery API):
image

Conclusion: We've recovered nearly all the performance hits by swapping the order of points. Performance losses are justifiable by doing actual work (e.g. storing distances) which saves time later in the pipeline of an analysis module. Overall speedups for common compute classes are awesome (~4x), per @jinsoo960's separate set of benchmarks.
Commit 13c99f5dfedb58002bac39ce5d1249f0b17957c1.

vyasr (approved):

PR #345, opened 2019-07-19
Enable query arguments for PMFTs. Test nearest neighbor behavior.

vyasr:
Adds tests for PMFTs and nearest neighbor-based queries.


jinsooihm (commented):
I think it looks good.

PR #349, opened 2019-07-22
Optimize link cell in case where cutoff is close to cell width.

vyasr:
This applies the optimization we discussed a while ago for LinkCell. @jinsoo960 can you quickly test whether this recovers the expected performance? If so, it should be fine to merge.

### DIFF
--- a/cpp/locality/LinkCell.h
+++ b/cpp/locality/LinkCell.h
@@ -542,7 +542,7 @@ public:
         // Upon querying, if the search radius is equal to the cell width, we
         // can guarantee that we don't need to search the cell shell past the
         // query radius. For simplicity, we store this value as an integer.
-        if (abs(m_r - neighbor_query->getCellWidth()) < 1e-6)
+        if (m_r == neighbor_query->getCellWidth())
         {
             m_extra_search_width = 0;
         }
--- a/cpp/locality/LinkCell.h
+++ b/cpp/locality/LinkCell.h
@@ -542,7 +542,7 @@ public:
         // Upon querying, if the search radius is equal to the cell width, we
         // can guarantee that we don't need to search the cell shell past the
         // query radius. For simplicity, we store this value as an integer.
-        if (abs(m_r - neighbor_query.getCellWidth()) < 1e-6)
+        if (abs(m_r - neighbor_query->getCellWidth()) < 1e-6)
         {
             m_extra_search_width = 0;
         }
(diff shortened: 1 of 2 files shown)


cpp/locality/LinkCell.h
```diff
@@ -538,7 +538,19 @@ class LinkCellQueryBallIterator : virtual public LinkCellIterator
                               float r, bool exclude_ii)
         : NeighborQueryIterator(neighbor_query, points, N, exclude_ii),
           LinkCellIterator(neighbor_query, points, N, exclude_ii), m_r(r)
-    {}
+    {
+        // Upon querying, if the search radius is equal to the cell width, we
+        // can guarantee that we don't need to search the cell shell past the
+        // query radius. For simplicity, we store this value as an integer.
+        if (abs(m_r - neighbor_query->getCellWidth()) < 1e-6)
```
bdice: @vyasr Why not `m_r == neighbor_query->getCellWidth()`? Or at least use a relative tolerance -- I'm hesitant to introduce a scale-dependent parameter like this.

vyasr: That's fair, checking for equality is probably fine. If for whatever reason the simple equality check fails due to some imprecision we're better off falling back to the slower but guaranteed correct version.

jinsooihm:
It is twice faster than calling `queryBall(...).toNList()` on the original implementation, but still twice slower compared to `compute(...)`. Hopefully, this difference should diminish once we change the order of `ref_points` and `points` as we just discussed @vyasr.

vyasr:
To provide a bit more information: when we originally benchmarked, there were two main slowdowns relative to the original `compute` method. The first was that many extra cells were being searched in the new method. This pull request addresses that problem, because in the specific use-case that the `compute` method supports, namely when the rcut is equal to the cell width, you can stop searching one cell earlier than you have to in the more general case. However, this pull request does not address the other reason for the slowdown, which is that the sorting to generate the NeighborList is slower in the `query...toNeighborList` approach  in the `compute` method. The reason for this is that the NeighborList is ordered by `ref_point, point`, and `compute` is designed around this order, whereas in freud 2.0 we are swapping that ordering and the `query*` methods are designed around that ordering. Once we perform the swap (in PR #342 into the `next` branch), these performance characteristics should be reversed and we should recover the original performance.

jinsooihm (approved):
Looks good.

bdice (approved):
👍

PR #350, opened 2019-07-22
Opt/rot acorr

vyasr:
Further optimizations for rotational autocorrelation. Gives another 5x speedup over the old method. Also gets rid of the usage of signed ints for quantities that should be unsigned.

### DIFF
--- a/cpp/order/RotationalAutocorrelation.h
+++ b/cpp/order/RotationalAutocorrelation.h
@@ -35,16 +35,6 @@ namespace freud { namespace order {
  */
 std::pair<std::complex<float>, std::complex<float>> quat_to_greek(const quat<float>& q);
 
-// This convenience function wraps std's gamma function for factorials, with
-// the appropriate shift by 1.
-inline unsigned int factorial(int n)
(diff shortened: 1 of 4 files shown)


cpp/order/RotationalAutocorrelation.h
```diff
@@ -7,7 +7,9 @@
 #include <cassert>
 #include <complex>
 #include <memory>
+#include <iostream>
```
bdice: Are `<iostream>` and `<map>` still needed? Remove if unused. Then this is good to go.

bdice (commented):
Looks good to me, one edit.

cpp/order/RotationalAutocorrelation.h
vyasr: Nope, I noticed those and then left them for you to find as a test :)

bdice: @vyasr I passed! 🥇

bdice (approved):
Looks good to me!

PR #351, opened 2019-07-22
Test/cf nn

vyasr:
Add tests for FloatCF and ComplexCF that test swapping points and ref_points, and using them in reverse order. Also enables query_args for these.


jinsooihm (approved):
Looks good.

PR #352, opened 2019-07-23
Test/localdescriptors

vyasr:
Adds tests of LocalDescriptors, including one that can be used to verify ref_points vs points behavior.

### DIFF
--- a/tests/test_environment_LocalDescriptors.py
+++ b/tests/test_environment_LocalDescriptors.py
@@ -116,7 +116,7 @@ class TestLocalDescriptors(unittest.TestCase):
         """Check if we can reproduce Steinhardt OPs."""
         def get_Ql(p, descriptors, nlist):
             """Given a set of points and a LocalDescriptors object (and the
-            underlying neighborlist, compute the per-particle Steinhardt order
+            underlying neighborlist), compute the per-particle Steinhardt order
             parameter for all :math:`l` values up to the maximum quantum number
             used in the computation of the descriptors."""
             Qbar_lm = np.zeros((p.shape[0], descriptors.sph.shape[1]),
@@ -163,8 +163,6 @@ class TestLocalDescriptors(unittest.TestCase):
                 steinhardt.compute(box, points, nlist=nl)
                 npt.assert_array_almost_equal(steinhardt.order, Ql[:, L])
 
-
-class TestLocalDescriptors2(unittest.TestCase):
     @skipIfMissing('scipy.special')
     def test_ld(self):
         """Verify the behavior of LocalDescriptors by explicitly calculating
--- a/tests/test_environment_LocalDescriptors.py
+++ b/tests/test_environment_LocalDescriptors.py
@@ -1,7 +1,8 @@
 import numpy as np
+import numpy.testing as npt
 import freud
 import unittest
-from util import make_box_and_random_points
+from util import make_box_and_random_points, make_sc, make_bcc, make_fcc
 
 
 class TestLocalDescriptors(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/util.py
```diff
@@ -118,6 +115,70 @@ def make_fcc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
...
+        ny: Number of repeats in the y direction, default is 1
+        nz: Number of repeats in the z direction, default is 1
+        scale: Amount to scale the unit cell by (in distance units)
+               (Default value = 1.0).
+        noise: Apply Gaussian noise with this width to particle positions
+               (Default value = 0.0).
+
+    Returns:
+        tuple (py:class:`freud.box.Box`, :class:`np.ndarray`): freud Box,
+            particle positions, shape=(nx*ny*nz, 3)
```
bdice: Probably should be `4*nx*ny*nz` but maybe there's a better way to write that across all of these functions.

tests/util.py
```diff
@@ -118,6 +115,70 @@ def make_fcc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
...
+    """
+    fractions = np.array([[0.5, .5, .5],
+                          [0, 0, 0]], dtype=np.float32)
+    return make_cubic(nx, ny, nz, fractions, scale, noise)
+
+
+def make_sc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
+    """Make an SC crystal for testing
```
bdice: SC ➡️ simple cubic; add period.

tests/util.py
```diff
@@ -118,6 +115,70 @@ def make_fcc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
...
+        scale: Amount to scale the unit cell by (in distance units)
+               (Default value = 1.0).
+        noise: Apply Gaussian noise with this width to particle positions
+               (Default value = 0.0).
+
+    Returns:
+        tuple (py:class:`freud.box.Box`, :class:`np.ndarray`): freud Box,
+            particle positions, shape=(nx*ny*nz, 3)
+    """
+    fractions = np.array([[0.5, .5, .5],
```
bdice: Be consistent with how the numbers are given. `0.5` vs `.5`

tests/util.py
```diff
@@ -118,6 +115,70 @@ def make_fcc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
...
+        ny: Number of repeats in the y direction, default is 1
+        nz: Number of repeats in the z direction, default is 1
+        scale: Amount to scale the unit cell by (in distance units)
+               (Default value = 1.0).
+        noise: Apply Gaussian noise with this width to particle positions
+               (Default value = 0.0).
+
+    Returns:
+        tuple (py:class:`freud.box.Box`, :class:`np.ndarray`): freud Box,
+            particle positions, shape=(nx*ny*nz, 3)
```
bdice: Shape as above.

tests/util.py
```diff
@@ -118,6 +115,70 @@ def make_fcc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
...
+    """
+    fractions = np.array([[.5, .5, 0],
+                          [.5, 0, .5],
+                          [0, .5, .5],
+                          [0, 0, 0]], dtype=np.float32)
+    return make_cubic(nx, ny, nz, fractions, scale, noise)
+
+
+def make_bcc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
+    """Make a BCC crystal for testing
```
bdice: body-centered cubic, add period.

tests/util.py
```diff
@@ -118,6 +115,70 @@ def make_fcc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
     return freud.box.Box(*box), positions
 
 
+def make_fcc(nx=1, ny=1, nz=1, scale=1.0, noise=0.0):
+    """Make a FCC crystal for testing
```
bdice: face-centered cubic; add period.

tests/util.py
```diff
@@ -80,27 +80,24 @@ def make_alternating_lattice(lattice_size, angle=0, extra_shell=2):
...
+        ny: Number of repeats in the y direction, default is 1.
+        nz: Number of repeats in the z direction, default is 1.
+        fractions: The basis to replicate using the lattice.
+        scale: Amount to scale the unit cell by (in distance units)
+               (Default value = 1.0).
+        noise: Apply Gaussian noise with this width to particle positions
+               (Default value = 0.0).
+
+    Returns:
+        tuple (py:class:`freud.box.Box`, :class:`np.ndarray`): freud Box,
```
bdice: Missing colon before `:py:class:`; inconsistent with use of `:py:class:` vs. `:class:` (assumed Python namespace).

tests/test_environment_LocalDescriptors.py
```diff
@@ -110,6 +112,181 @@ def test_repr(self):
         comp = freud.environment.LocalDescriptors(4, 8, 0.5, True)
         self.assertEqual(str(comp), str(eval(repr(comp))))
 
+    def test_ql(self):
+        """Check if we can reproduce Steinhardt OPs."""
+        def get_Ql(p, descriptors, nlist):
+            """Given a set of points and a LocalDescriptors object (and the
+            underlying neighborlist, compute the per-particle Steinhardt order
```
bdice: Missing a close parenthesis )

tests/test_environment_LocalDescriptors.py
```diff
@@ -110,6 +112,181 @@ def test_repr(self):
...
+        num_neighbors = 6
+        l_max = 12
+        r_max = 2
+
+        for struct_func in [make_sc, make_bcc, make_fcc]:
+            box, points = struct_func(5, 5, 5)
+
+            # In order to be able to access information on which particles are
+            # bonded to which ones, we precompute the neighborlist
+            nn = freud.locality.NearestNeighbors(r_max, num_neighbors)
```
bdice, line 152: Let's use a NeighborQuery API instead, so that we won't have to refactor this later.

tests/test_environment_LocalDescriptors.py
```diff
@@ -110,6 +112,181 @@ def test_repr(self):
...
+            Ql = get_Ql(points, ld, nl)
+
+            # Test all allowable values of l.
+            for L in range(2, l_max+1):
+                steinhardt = freud.order.Steinhardt(r_max*2, L)
+                steinhardt.compute(box, points, nlist=nl)
+                npt.assert_array_almost_equal(steinhardt.order, Ql[:, L])
+
+
+class TestLocalDescriptors2(unittest.TestCase):
```
bdice: Why make this a separate class?

tests/test_environment_LocalDescriptors.py
```diff
@@ -110,6 +112,181 @@ def test_repr(self):
...
+        N = 100
+        box, points = make_box_and_random_points(L, N)
+
+        num_neighbors = 1
+        r_max = 2
+        l_max = 2
+
+        # We want to provide the NeighborList ourselves since we need to use it
+        # again later anyway.
+        nn = freud.locality.NearestNeighbors(r_max, num_neighbors)
```
bdice, line 182: Use NeighborQuery API

bdice (changes requested):
Good to add these tests! I left some comments below.

tests/util.py
vyasr: It's fine, I'll just add it this way. It's a util helper, so more important to have the information than have it in a way that renders nicely.

tests/test_environment_LocalDescriptors.py
vyasr: Good call, I was thinking that we shouldn't use the new API on master but should be fine.

tests/test_environment_LocalDescriptors.py
vyasr: That was just for easily running just the tests I wanted. I'll remove.

vyasr:
I actually decided to hold off changing the API to use the new one for the NearestNeighbors. The reason is the (known) issue that the ordering of indices on master is still in the (ref_point, point) order, and for NearestNeighbors the results aren't symmetric. Once we merge #342, everything on the next branch will have a consistent ordering, and we'll get the expected results passing the data through, but at present, switching to the new API here would require also reversing the neighbor list indices, and we'd have to undo that later anyway. @bdice let me know if you're OK with that.

bdice (approved):
Looks good to me, after explaining the choice of older neighbor APIs.

PR #353, opened 2019-07-23
Force Cython rebuild on RTD.

bdice:
## Motivation and Context
Sometimes our builds on Read the Docs fail, if the Cythonized .cpp files in `freud/*.cpp` aren't up to date. This fixes that problem by forcing RTD to re-Cythonize on every build.

## How Has This Been Tested?
I ran RTD on a branch off of `next` which is currently failing from outdated cpp files, then I cherry-picked the commit back to `master` so it can be merged up to `next` later.

## Types of changes

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -163,7 +163,11 @@ sys.argv = ['setup.py'] + extras
 
 on_rtd = os.environ.get('READTHEDOCS') == 'True'
 if on_rtd:
+    logger.warning('Building on Read the Docs with Cython enabled.')
     args.use_cython = True
+    for cython_cpp_file in glob.glob('freud/*.cpp'):
+        logger.warning('Deleting {}'.format(cython_cpp_file))
+        os.remove(cython_cpp_file)
 
 
 ################################


setup.py
```diff
@@ -163,7 +163,11 @@ def _redirect_stderr(to_fd, original_stderr_fd):
 
 on_rtd = os.environ.get('READTHEDOCS') == 'True'
 if on_rtd:
+    logger.warning('Building on Read the Docs with Cython enabled.')
     args.use_cython = True
+    for cython_cpp_file in glob.glob('freud/*.cpp'):
+        logger.warning('Deleting {}'.format(cython_cpp_file))
```
bdice, line 169: We can make this silent if preferred, but I think having it loud is better because it will only ever show up on RTD and helps us track the code's behavior.

vyasr: I'm fine keeping it loud.

vyasr:
I thought I modified this a long time ago along with CircleCI, I just never checked back into it. Definitely OK with doing this, it adds minimal build time anyway compared to the actual compilation.

vyasr (approved):

PR #354, opened 2019-07-23
Fix/duplicate cell shells

### LINKED ISSUE #343: LinkCell query giving duplicate points
**Describe the bug**
`LinkCell.query` gives duplicate points. 

**To Reproduce**
```
import freud
points = [[1.5132198, 6.67087, 3.1856632], [1.3913784,  -2.3667011,   4.5227165], [-3.6133137, 9.043476, 0.8957424]]
box = freud.box.Box.cube(31)
rmax = 10
lc = freud.locality.LinkCell(box, rmax, points)
q = lc.queryBall(points[0], rmax)
list(q)
```

**Error output**
```
[(0, 0, 0.0),
 (2, 0, 6.095436096191406),
 (1, 0, 9.136752128601074),
 (1, 0, 9.136752128601074)]
```

 - OS: [macOS]
 - Version of Python [3.6.8]
 - Version of freud [1.2]

bdice:
## Description
This PR attempts to solve #343. I'm not yet fully convinced of this solution's correctness, but it passes tests and seems to work as I expect.

Questions for discussion:
1. @vyasr is there a reason why the QueryBall didn't have the same check to prevent the IteratorCellShell from duplicating a cell on the positive and negative sides? I had to add that from Query to QueryBall to make it work properly.
2. Should we refactor the `neighbor_cell` computation so that it's not duplicated several times? We would just need a single function to calculate the cell index from `vec3<unsigned int> point_cell` and the current state of `m_neigh_cell_iter`.
3. If the IteratorCellShell can prevent duplication in the positive and negative directions as mentioned above, do we even need to track the cells we've queried? It seems like we've already addressed this problem in a different way than tracking previously-queried cells, but it didn't seem to work (otherwise this bug wouldn't show up for the `query` methods too). Is there another case that I'm missing?

## Motivation and Context
Resolves #343.

## How Has This Been Tested?
The test case mentioned in #343 has been turned into an actual test.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/LinkCell.cc
+++ b/cpp/locality/LinkCell.cc
@@ -369,6 +369,7 @@ NeighborPoint LinkCellQueryBallIterator::next()
     while (cur_p < m_N)
     {
         vec3<unsigned int> point_cell(m_linkcell->getCellCoord(m_points[cur_p]));
+        m_searched_cells.clear();
 
         // Loop over cell list neighbor shells relative to this point's cell.
         while (true)
@@ -387,17 +388,21 @@ NeighborPoint LinkCellQueryBallIterator::next()
                 }
             }
 
-            // Determine the next neighbor cell to consider. We're done if we
-            // reach a new shell and the closest point of approach to the new
-            // shell is greater than our rcut.
-            ++m_neigh_cell_iter;
+            bool out_of_range = false;
 
-            if ((m_neigh_cell_iter.getRange() - m_extra_search_width) * m_linkcell->getCellWidth() > m_r)
+            while (true)
             {
-                break;
-            }
-            else
-            {
+                // Determine the next neighbor cell to consider. We're done if we
+                // reach a new shell and the closest point of approach to the new
+                // shell is greater than our rcut.
+                ++m_neigh_cell_iter;
+
+                if ((m_neigh_cell_iter.getRange() - m_extra_search_width) * m_linkcell->getCellWidth() > m_r)
+                {
+                    out_of_range = true;
+                    break;
+                }
+
                 const unsigned int neighbor_cell = m_linkcell->getCellIndexer()(
                     // Need to increment each dimension by the width to avoid taking the modulus
                     // of a negative number.
@@ -407,7 +412,21 @@ NeighborPoint LinkCellQueryBallIterator::next()
                         % m_linkcell->getCellIndexer().getH(),
                     (m_linkcell->getCellIndexer().getD() + point_cell.z + (*m_neigh_cell_iter).z)
                         % m_linkcell->getCellIndexer().getD());
-                m_cell_iter = m_linkcell->itercell(neighbor_cell);
+
+                auto searched_cell_iter = m_searched_cells.find(neighbor_cell);
+                if (searched_cell_iter == m_searched_cells.end())
+                {
+                    // This cell has not been searched yet, so we will iterate
+                    // over its contents. Otherwise, we loop back, increment
+                    // the cell shell iterator, and try the next one.
+                    m_searched_cells.insert(neighbor_cell);
+                    m_cell_iter = m_linkcell->itercell(neighbor_cell);
+                    break;
+                }
+            }
+            if (out_of_range)
+            {
+                break;
             }
         }
         cur_p++;
@@ -437,6 +456,7 @@ NeighborPoint LinkCellQueryIterator::next()
     while (cur_p < m_N)
     {
         vec3<unsigned int> point_cell(m_linkcell->getCellCoord(m_points[cur_p]));
+        m_searched_cells.clear();
 
         // Loop over cell list neighbor shells relative to this point's cell.
         if (!m_current_neighbors.size())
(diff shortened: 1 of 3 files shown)


vyasr:
I'm not sure I follow what the difference between (1) and (3) is. Are you referring to the max range check? If so, the bug that we've found is that the max range check is not sufficient if the max range is also exactly half the box width, because then you end up searching that cell twice. I had to address this in query because that can reach past half the box even without hitting any pathological cases.

Also yes, I'm open to refactoring that neighbor cell computation into a function.

vyasr:
Your test might be a bit too complicated, just to make sure there's no confusion, here's a simple one that illustrates the problem.

```
import freud
box = freud.box.Box.square(5)
points = [[-1.5, 0, 0]]
ref_points = [[0.9, 0, 0]]
rmax = 2.45
cell_width = 1

lc = freud.locality.LinkCell(box, 1, ref_points)
print(list(lc.queryBall(points, rmax)))
# [(0, 0, 2.4000000953674316), (0, 0, 2.4000000953674316)]
```

I would suggest using something like this, it makes the issue glaringly obvious.

bdice:
@vyasr I think the logic is fixed now -- both of the test cases now behave as I expect. I think it's ready for review.

cpp/locality/LinkCell.cc
```diff
@@ -387,32 +418,49 @@ NeighborPoint LinkCellQueryBallIterator::next()
...
+                {
+                    out_of_range = true;
+                    break;
+                }
+
+                // In cases where we need to check the entire box, make sure
+                // that we don't check the same cell on the positive and
+                // negative sides of the IteratorCellShell cube.
+                if (check_double_range_shell(m_linkcell, *m_neigh_cell_iter))
+                    continue;
```
bdice: I think this check is actually not necessary, since we're tracking (and skipping) the previously-searched cells in `m_searched_cells`. If you agree, @vyasr, I'd consider removing it. (The tests pass if it's removed.)

cpp/locality/LinkCell.cc
```diff
@@ -465,29 +515,35 @@ NeighborPoint LinkCellQueryIterator::next()
...
+                    if (m_neigh_cell_iter == IteratorCellShell(max_range, m_neighbor_query->getBox().is2D()))
+                    {
+                        break;
+                    }
+
+                    // In cases where we need to check the entire box, make sure
+                    // that we don't check the same cell on the positive and
+                    // negative sides of the IteratorCellShell cube.
+                    if (check_double_range_shell(m_linkcell, *m_neigh_cell_iter))
+                        continue;
```
bdice: See comment above. Could be removed.

cpp/locality/LinkCell.cc
```diff
@@ -137,6 +137,35 @@ bool compareFirstNeighborPairs(const std::vector<std::tuple<size_t, size_t, floa
...
+    unsigned int d = linkcell->getCellIndexer().getD();
+    if (2 * cell_delta.x + 1 > (int) w)
+        range_check = true;
+    else if (2 * cell_delta.y + 1 > (int) h)
+        range_check = true;
+    else if (2 * cell_delta.z + 1 > (int) d)
+        range_check = true;
+    return range_check;
+}
+
```
vyasr: I would make these two functions into LinkCell class methods.

cpp/locality/LinkCell.cc
vyasr: I agree, you can remove this. I think I understand what your point (1) was referring to. I probably just overlooked this possibility for ball queries; it would have come up naturally in a sparse system for k-nearest neighbor queries since cutoffs don't matter there, but I didn't test large enough cutoffs to see that scenario with ball queries. However, it's also clearly not strict enough to catch e.g. the test case I posted in this PR's thread because you have to search that cell, you just need to avoid searching it twice.

cpp/locality/LinkCell.cc
vyasr: Same, agree.

cpp/locality/LinkCell.cc
vyasr: Actually, just the first one since you're removing the second function.

cpp/locality/LinkCell.cc
```diff
@@ -465,29 +515,35 @@ NeighborPoint LinkCellQueryIterator::next()
...
+                    if (searched_cell_iter == m_searched_cells.end())
+                    {
+                        // This cell has not been searched yet, so we will iterate
+                        // over its contents. Otherwise, we loop back, increment
+                        // the cell shell iterator, and try the next one.
+                        m_searched_cells.insert(neighbor_cell);
+                        m_cell_iter = m_linkcell->itercell(neighbor_cell);
+                        break;
+                    }
+                }
```
vyasr, line 510: I almost want to try and move this into a function of the LinkCellIterator, but not worth trying in this PR. Just noting it so that we remember the possibility.

vyasr: There would need to be some virtual function dispatch to iterator specific termination routines, but most of the code is common.

vyasr (changes requested):
Looks good to me. I requested one change, plus the ones that you suggested, then we're good to go.

vyasr:
I now understand what you meant with points (1) and (3). The IteratorCellShell can track how far you've gone in each direction, but it's not going to catch the specific edge case that we're testing for, which is when we're at the edge of the valid range but we check the same cell twice, once in each direction. 

I do have some ideas for how we can do a better job with that check than what we're doing now, but I'd prefer to just get this working fix into freud and optimize later. There are a couple of obvious optimizations that can be made to the iterator code now that I'm reasonably confident in its stability, so those can come in a future PR. Once you address my comments, let's merge this PR and be ready for 1.2.1

cpp/locality/LinkCell.cc
bdice: Removed.

cpp/locality/LinkCell.cc
bdice: Removed.

vyasr (approved):
Looks fine to me.

PR #359, opened 2019-07-24
Update PR template checklist

bdice:
The wording was confusing on some checklist items, which were dependent on each other. This PR fixes that.


vyasr (approved):
As discussed offline, we should also start being strict about updating the changelog with PRs.

PR #361, opened 2019-07-27
Next update v1.2.1

bdice:
This PR includes the changes released in v1.2.1 into `next`. I think we can just merge this. I also reformatted C++ since we've made some huge changes there.


bdice:
@vyasr If you're cool with this, we can merge it. It was a fairly significant update so I wanted your review.

vyasr:
I started reviewing then went back a few commits to try and avoid looking at clang formatting changes, all the substance looks fine to me.

vyasr:
Many if not most of the clang formatting changes will get wiped out when I merge in the variable renaming branch, but since at least some of the clang formatting changes were far enough upstream in this branch (i.e. pre-version bump) I figure it's easier to just merge this then undo as needed on the relevant downstream merge.

Issue #305, opened 2019-06-10
Improve Voronoi Computation


yzjin:
**Is your feature request related to a problem? Please describe.**
Previously, we have refactored `locality.Voronoi` in freud by merging functions with similar functionality and moving most of computation to c++, including `compute()` and `computeNeighbors()`. 
However, `qhull()` is still implemented in cython using `Scipy.spatial.Voronoi` and slows down the `computeVolume()`. 

**Describe the solution you'd like**
We could implement the volume computation in C++ via `qhull()` from `Scipy.spatial.Voronoi` or this quickhull implementation: https://github.com/akuukka/quickhull to have the entire computation in c++.

bdice:
We're going to skip this in favor of a new voro++ implementation.

PR #362, opened 2019-07-30
Use voro++ for Voronoi computations.

bdice:
## Description
This PR adds the [voro++](http://math.lbl.gov/voro++/) library written by Chris Rycroft (@chr1shr) to freud. This will replace the existing Voronoi methods, which used less sophisticated algorithms and were quite slow.

## Motivation and Context
Issue #305 describes challenges with Voronoi performance -- this continues to be a bottleneck for my current analyses, so I found it worthwhile to rewrite this entire feature using voro++.

## How Has This Been Tested?
I have adapted existing Voronoi tests, which work as expected in both 2D and 3D systems. Some changes (like reordering of vertices in a Voronoi cell) are expected, so I added functions to sort the vertices in a deterministic way for the purpose of testing. Currently, this new approach is about 32x faster for a system of 5000 random points in a cubic box of side length 10.

## Types of changes

## Checklist:


bdice:
To-Do:
- [x] Expand test suite
- [x] Replace previous Voronoi code (rename `freud.locality._VoroPlusPlus` to `freud.locality.Voronoi`)
- [x] Decide if other cell properties should be added (since they're almost free)
  - Not for now.
- [x] Revise API
- [x] Remove unneeded C++ code (debugging, extra computed properties, etc.)
- [x] Update documentation

bdice:
@vyasr Currently this is using a direct copy-paste of the source tarball. I think this is the definitive repo source: https://repo.or.cz/voro++.git We should probably change this into a git submodule -- should we add it directly, or as a mirror via the glotzerlab organization? We have Eigen mirrored, and we control all other submodules.

vyasr:
Even that git repo looks like a mirror. The "active" repo (last commit was 2017-09-29, so almost 2 years ago) is the [Subversion repository](https://code.lbl.gov/scm/?group_id=108). Of course, maintaining an svn repository inside a git repository would be a huge pain, and I wouldn't advocate that. I think the relevant question is whether you think any further updates to the code are forthcoming, and if so, whether they will get into the git repository.

I don't think that we should maintain our own mirror. It introduces unnecessary maintenance overhead that I don't think is terribly helpful, and I believe that we're moving away from that with Eigen too (IIRC Josh plans to remove that mirror when HOOMD gets the 3.0 version bump and stops using submodules). So I'd say that copying in the code is actually not the worst option, unless you think that future updates will make it into the git repository, in which case a submodule would work fine.

chr1shr:
I am actually planning to convert the old Subversion repository into a Git repo in the coming month. I recently got a grant to further Voro++ development and this is the first step to doing that. Hence, if you want to incorporate the library as a submodule, that could work.

There are a number of developments that I've already done, or plan to incorporate as part of the grant: (A) 2D computations, (B) more reliable 3D computations, (C) multi-threading, (D) parallelization via domain decomposition. A & B are already mainly done, C is ongoing, and D is future work.

bdice:
@chr1shr Wonderful to hear! Glad I tagged you. :) I implemented fake 2D computations by just putting the points in the z=0 plane and using the top surface of a box with height 1. Please do let us know when you make a git repo!

In particular with multithreading, we use Intel Threading Building Blocks in our code. I think it would work immediately for us if I could get the loop object over all particles in the container to be easier to index, so that blocks of cells could be assigned to each thread.

bdice:
I'm closing this PR so that I can use its final set of C++/Cython files but re-do the branch with a git submodule. I was not successful in rebasing the original "source copy" commit by replacing it with a git submodule addition.

PR #363, opened 2019-07-31
Fix compiler warnings

vyasr:
Removes some of the warnings that have appeared over the course of our changes. It intentionally does not fix all of them. In general, there are two classes of warnings that I didn't fix. One is cases that will be fixed by adding proper support for double precision, for example because there are floats multiplying the outputs of dot products. The other is implicit size_t <--> unsigned int conversions; I don't think there's anything we should do about this, it's simply the way that tbb exposes the iterator variable in parallel for loops.

### DIFF
--- a/cpp/environment/AngularSeparation.cc
+++ b/cpp/environment/AngularSeparation.cc
@@ -35,7 +35,7 @@ float computeSeparationAngle(const quat<float> ref_q, const quat<float> q)
         R.s = -1.0;
     }
 
-    float theta = 2.0 * acos(R.s);
+    float theta = float(2.0 * acos(R.s));
 
     return theta;
 }
--- a/cpp/environment/LocalDescriptors.cc
+++ b/cpp/environment/LocalDescriptors.cc
@@ -130,7 +130,7 @@ void LocalDescriptors::compute(const box::Box& box,
                 const float magR(sqrt(rsq));
                 float theta(atan2(bond_ij.y, bond_ij.x)); // theta in [-pi..pi] initially
                 if (theta < 0)
-                    theta += 2 * M_PI;             // move theta into [0..2*pi]
+                    theta += float(2 * M_PI);             // move theta into [0..2*pi]
                 float phi(acos(bond_ij.z / magR)); // phi in [0..pi]
 
                 // catch cases where bond_ij.z/magR falls outside [-1, 1]
--- a/cpp/locality/AABBQuery.h
+++ b/cpp/locality/AABBQuery.h
@@ -114,13 +114,13 @@ protected:
                 throw std::runtime_error("You must set nn in the query arguments.");
             if (args.scale == -1)
             {
-                args.scale = 1.1;
+                args.scale = float(1.1);
             }
             if (args.rmax == -1)
             {
                 vec3<float> L = this->getBox().getL();
                 float rmax = std::min(L.x, L.y);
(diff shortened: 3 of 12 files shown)


cpp/environment/AngularSeparation.cc
```diff
@@ -35,7 +35,7 @@ float computeSeparationAngle(const quat<float> ref_q, const quat<float> q)
         R.s = -1.0;
     }
 
-    float theta = 2.0 * acos(R.s);
+    float theta = float(2.0 * acos(R.s));
```
bdice, line 38: Does `float theta(2.0f * acos(R.s));` trigger a warning?

cpp/environment/LocalDescriptors.cc
```diff
@@ -130,7 +130,7 @@ void LocalDescriptors::compute(const box::Box& box,
                 const float magR(sqrt(rsq));
                 float theta(atan2(bond_ij.y, bond_ij.x)); // theta in [-pi..pi] initially
                 if (theta < 0)
-                    theta += 2 * M_PI;             // move theta into [0..2*pi]
+                    theta += float(2 * M_PI);             // move theta into [0..2*pi]
```
bdice, line 133: Would prefer `2.0f * M_PI`

cpp/locality/AABBQuery.h
```diff
@@ -114,13 +114,13 @@ class AABBQuery : public NeighborQuery
                 throw std::runtime_error("You must set nn in the query arguments.");
             if (args.scale == -1)
             {
-                args.scale = 1.1;
+                args.scale = float(1.1);
```
bdice, line 117: `1.1f`

cpp/locality/AABBQuery.h
```diff
@@ -114,13 +114,13 @@ class AABBQuery : public NeighborQuery
...
             }
             if (args.rmax == -1)
             {
                 vec3<float> L = this->getBox().getL();
                 float rmax = std::min(L.x, L.y);
-                args.rmax = this->getBox().is2D() ? 0.1 * rmax : 0.1 * std::min(rmax, L.z);
+                args.rmax = this->getBox().is2D() ? float(0.1) * rmax : float(0.1) * std::min(rmax, L.z);
```
bdice, line 123: `0.1f`

cpp/order/Steinhardt.cc
```diff
@@ -130,7 +130,7 @@ void Steinhardt::baseCompute(const freud::locality::NeighborList* nlist,
                              const freud::locality::NeighborQuery* points,
                              freud::locality::QueryArgs qargs)
 {
-    const float normalizationfactor = 4 * M_PI / (2 * m_l + 1);
+    const float normalizationfactor = float(4 * M_PI / (2 * m_l + 1));
```
bdice, line 133: Try `4.0f` and see if that fixes it

vyasr:
Regarding your comments, my thought process is to make changes that will be easily compatible with allowing freud to offer single or double precision options. Hardcoding something like `2.0f` doesn't work for that, whereas `float(2)` can easily be replaced with a typedef (_a la_ `Scalar` in HOOMD).

vyasr:
Also I can't trigger most of these warnings on my compiler. You'll have to try it on your Windows machine to let me know :D

bdice:
@vyasr Good call on the hardcoding. We have a lot of `0.5f`s _floating_ around in our codebase already so we'll need to watch out for that if we change the precision.

bdice (approved):

Issue #266, opened 2019-05-07
Standardize class attribute access


vyasr:
Currently, class attributes (essentially the constructor arguments) are sometimes stored in Python, and sometimes not. In the latter case, the getter functions (properties) typically access the data directly from the underlying C++ class. We should reduce this redundancy and establish a standard for where data should be stored.

vyasr:
This will be done as part of our API rewrite, see #179.

vyasr:
Resolved by #467 and #473.

PR #367, opened 2019-08-06
Add optional bond weights to Steinhardt order parameters.

### LINKED ISSUE #45: Implement Minkowski Structure Metric
**[Original report](https://bitbucket.org/glotzer/freud/issue/45) by Eric Harper (Bitbucket: [harperic](https://bitbucket.org/harperic), GitHub: [harperic](https://github.com/harperic)).**

bdice:
## Description
Weighted bonds enable the use of _Minkowski structure metrics_ by defining Steinhardt order parameters over a Voronoi neighborhood weighted by Voronoi polytope facet areas, discussed in ref. 1 below.

## Motivation and Context
This PR resolves #45.

This PR changes the normalization done by the Steinhardt class to use bond weights if the `weighted` flag is set to True. Alternatively, we could _always_ enable weights without expecting a significant change in behavior, since the NeighborQuery and NeighborList default behavior is to assign a weight of 1 to each bond.

## How Has This Been Tested?
New tests have been added (todo: not yet finalized/pushed).

## References
[1] Shortcomings of the bond orientational order parameters for the analysis of disordered particulate matter
J. Chem. Phys. **138**, 044501 (2013); https://doi.org/10.1063/1.4774084
Walter Mickel, Sebastian C. Kapfer, Gerd E. Schröder-Turk, and Klaus Mecke

## Types of changes

## Checklist:

### DIFF
--- a/cpp/order/Steinhardt.h
+++ b/cpp/order/Steinhardt.h
@@ -122,6 +122,24 @@ public:
         return m_norm;
     }
 
+    //!< Whether to take a second shell average
+    bool isAverage()
+    {
+        return m_average;
+    }
+
+    //!< Whether to use the third-order invariant Wl
+    bool isWl()
+    {
+        return m_Wl;
+    }
+
+    //!< Whether to use neighbor weights in computing Qlmi
+    bool isWeighted()
+    {
+        return m_weighted;
+    }
+
     //! Compute the order parameter
     virtual void compute(const freud::locality::NeighborList* nlist,
                                   const freud::locality::NeighborQuery* points, freud::locality::QueryArgs qargs);
--- a/cpp/order/Steinhardt.h
+++ b/cpp/order/Steinhardt.h
@@ -65,8 +65,8 @@ public:
      *                         Allows looking at, for instance, only the second shell,
      *                         or some other arbitrary rdf region.
      */
-    Steinhardt(float rmax, unsigned int l, float rmin = 0, bool average = false, bool Wl = false)
-        : m_Np(0), m_rmax(rmax), m_l(l), m_rmin(rmin), m_average(average), m_Wl(Wl), m_Qlm_local(2 * l + 1)
+    Steinhardt(float rmax, unsigned int l, float rmin = 0, bool average = false, bool Wl = false, bool weighted = false)
+        : m_Np(0), m_rmax(rmax), m_l(l), m_rmin(rmin), m_average(average), m_Wl(Wl), m_weighted(weighted), m_Qlm_local(2 * l + 1)
     {
         // Error Checking
         if (m_rmax < 0.0f || m_rmin < 0.0f)
@@ -167,6 +167,7 @@ private:
     // Flags
     bool m_average; //!< Whether to take a second shell average (default false)
     bool m_Wl;      //!< Whether to use the third-order invariant Wl (default false)
+    bool m_weighted;      //!< Whether to use neighbor weights in computing Qlmi (default false)
 
     std::shared_ptr<std::complex<float>> m_Qlmi; //!< Qlm for each particle i
     std::shared_ptr<std::complex<float>> m_Qlm;  //!< Normalized Qlm(Ave) for the whole system
--- a/freud/order.pyx
+++ b/freud/order.pyx
@@ -573,7 +573,7 @@ cdef class Steinhardt(Compute):
     cdef num_neighbors
 
     def __cinit__(self, r_max, l, r_min=0, average=False, Wl=False,
-                  weighted=False, num_neighbors=0, *args, **kwargs):
+                  weighted=False, num_neighbors=0):
         if type(self) is Steinhardt:
             self.r_max = r_max
             self.sph_l = l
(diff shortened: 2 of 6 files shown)


tests/test_order_Steinhardt.py
```diff
@@ -135,6 +135,43 @@ def test_identical_environments_Wl_near(self):
...
+
+            for wt in [0, 0.1, 0.9, 1.1, 10, 1e6]:
+                # Change the weight of the first bond for each particle
+                nlist.weights[nlist.segments] = wt
+
+                comp = freud.order.Steinhardt(rmax, 6, weighted=True)
+                comp.compute(box, ts[0], nlist=nlist)
+
+                # Unequal neighbor weighting in a perfect FCC structure
+                # appears to increase the Q6 order parameter
```
vyasr, line 158: Can we add a precise test? You can copy the Steinhardt tests that I added for LocalDescriptors (I think?) and just add in weights. It would be helpful to verify the behavior against another implementation rather than just check a single observed behavior.

freud/order.pyx
```diff
@@ -671,6 +677,34 @@ cdef class Steinhardt(Compute):
...
+        return plot.histogram_plot(self.order,
+                                   title="Steinhardt Order Parameter",
+                                   xlabel=xlabel,
+                                   ylabel=r"Number of particles",
+                                   ax=ax)
+
+    def _repr_png_(self):
+        import plot
+        try:
+            return plot.ax_to_bytes(self.plot(mode=self.plot_mode))
```
vyasr: I think this is copied from other OPs, you'll want to get rid of the plot mode.

cpp/order/Steinhardt.h
```diff
@@ -65,8 +65,8 @@ class Steinhardt
      *                         Allows looking at, for instance, only the second shell,
      *                         or some other arbitrary rdf region.
      */
-    Steinhardt(float rmax, unsigned int l, float rmin = 0, bool average = false, bool Wl = false)
-        : m_Np(0), m_rmax(rmax), m_l(l), m_rmin(rmin), m_average(average), m_Wl(Wl), m_Qlm_local(2 * l + 1)
+    Steinhardt(float rmax, unsigned int l, float rmin = 0, bool average = false, bool Wl = false, bool weighted = false)
```
vyasr, line 68: For 2.0 we should probably do a review to make sure that default parameter values are properly mirrored everywhere in Cython/C++. It's annoying, but probably the best approach if we want to move towards allowing C++ usage of freud.

Just a note, not a to-do.

vyasr (changes requested):
I think leaving weights as an option rather than always using them is fine for now. We should probably think about how bond weights should be used in general, and whether there are other cases we should enable them. That might help us clarify what the API should look like long term

tests/test_order_Steinhardt.py
bdice: Good idea, I added a test for this.

cpp/order/Steinhardt.h
```diff
@@ -122,6 +122,24 @@ class Steinhardt
...
+    {
+        return m_Wl;
+    }
+
+    //!< Whether to use neighbor weights in computing Qlmi
+    bool isWeighted()
+    {
+        return m_weighted;
+    }
+
```
vyasr, line 142: Do we actually need these? They're not even exposed in the Cython API. This is related to #266, I think we need further discussion there on whether or not we want to consistently expose all constructor arguments via properties. I think it's not a bad thing, but we should be more consistent about it. In this case, please either remove these methods or add properties to the class that use them.

vyasr (approved):
Approved pending addressing of my one request. Once you've addressed in either of the ways I suggested, feel free to merge once tests pass.

ArvDee:
Hi Bradly & the freud team,

a while ago I wanted to calculate these exactly these Minkowski structure metrics for some output data from hoomd, and ended up writing my own code for it, after not finding mention of weighting in the freud docs. Only after the code was pretty much done did I find this pull request stating you were working on it. D'oh. If you'll forgive a bit of shameless self-promotion, you might be interested in seeing [my implementation](https://github.com/ArvDee/Minkowski-structure-metrics-calculator). Perhaps it can be of use if you're looking to test the output to another independent implementation. 

Additionally, while testing this code I looked around in literature for some reference values to check if my output was correct. Given that you're working on this, I assume you've found these same papers already, but in case you haven't:

[Rein ten Wolde et al., J. Chem. Phys (1996)](http://aip.scitation.org/doi/10.1063/1.471721) reports values for the Steinhardt w4 and w6 in Table 1.
[Dietz et al., Phys. Rev. E (2016)](https://link.aps.org/doi/10.1103/PhysRevE.94.033207) report the Minkowski structure metric values q4' and q6' for a BCC lattice in section IIc.

I hope this can be of use to you.

Kind regards,
R

vyasr:
Fantastic, thanks for the information! Glad to see that you're finding freud useful. I definitely think validating against your implementation would be great for us, we've been working to beef up our testing throughout freud so that will be very helpful. We'll be in touch if we have any questions!

bdice:
@ArvDee, thanks for your comment! I really appreciate having another code for comparison. There's also a draft PR #362 that replaces our existing Voronoi implementation with Voro++. We're currently adding these features for v2.0, which comes with a variety of API changes / simplifications as well.

Would you consider licensing your code using a permissive license like BSD-3 or MIT? We can't incorporate GPL code into freud without also making freud GPL. There might be some pieces of your library that we would like to use, if you're open to that (with credit, of course). It looks like all your dependencies (Eigen, Voro++) are permissively licensed and you're the only contributor to the repository, so this should be straightforward for you to change if you're open to it.

cpp/order/Steinhardt.h
bdice: I exposed the properties and added them to the repr and plot labels.

ArvDee:
@bdice Sure thing, I've changed the license to MIT. Feel free to use it as you see fit.

PR #370, opened 2019-08-07
Feature/common array

### LINKED ISSUE #151: Fix some instances of reference rot by setting base attribute of arrays
**[Original report](https://bitbucket.org/glotzer/freud/issue/151) by Matthew Spellings (Bitbucket: [mspells](https://bitbucket.org/mspells), GitHub: [klarh](https://github.com/klarh)).**

Currently, the arrays returned by freud (for example `RDF.getRDF()`) are invalidated in two cases:

1. When the parent object (the RDF object in the example above) gets garbage collected
2. When the arrays are reallocated (if you would be able to change rmax in the RDF, for example)

I think we should be able to prevent (1), at least, by setting the `numpy.ndarray.base` flag on the generated arrays to `self` (whichever compute object is generating them). Solving (2) is more complicated, but also not totally impossible I think.

An example of this sort of usage is at https://gist.github.com/GaelVaroquaux/1249305 .

vyasr:
## Description
This pull request adds a standard array object to be used throughout freud. The array is intended for both passing in arguments and as a class member for array data that will be available through the Python API. The array object contains extra logic to minimize data copying while also providing support for passing around ownership of data. The Python wrapper class is responsible for owning the data, relinquishing control to the C++ object on calls to compute, and creating new arrays if the existing array is being used in Python.

Developers using this will need to be aware of a few important conventions, for instance the fact that arrays will now need to be returned by reference for this method to work. I plan to add some additional assertions in the `ManagedArrayContainer` Python class to help developers debug when they run into problems with this, but I think that level of complexity is to be expected if we want to address our reference counting issues. Once this is implemented, copying the pattern shouldn't be hard for new freud developers.

The C++ `ManagedArray` class could still use a little cleaning up, and then it needs to be implemented everywhere else in freud. For now, only the `freud.cluster.Cluster` class has been updated to use it, so we can work out the kinks there. @joaander @klarh I'm interested to hear your thoughts on possible issues with this approach, and if anything can be done to mitigate the complexity.

## Motivation and Context
This pull request resolves #151. It also simplifies our C++ APIs, as we no longer need to pass around array lengths along with pointers.

## How Has This Been Tested?
I've added a simple test to check that copying is no longer necessary for saving attributes. I'm also happy to add additional checks anyone has to suggest.

## Types of changes

## Checklist:


vyasr:
@thomasjpfan in case you're interested, this PR is attempting to resolve the array ownership issues we discussed.

vyasr:
Ignore the 3.6 failure, Codecov is currently down.

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,305 @@
...
+     */
+    ManagedArray(const ManagedArray &other) : m_data(other.m_data), m_managed(false), m_external_ptr(other.m_external_ptr)
+    {
+        if (m_external_ptr)
+        {
+            m_size = new unsigned int(*other.m_size);
+        }
+        else
+        {
+            m_size = other.m_size;
```
klarh: Without adding any reference counting, this leaves the implementation open to a segfault if you make a copy of a parent object that later goes out of scope (when it deallocates its `m_size`), right?

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,305 @@
...
+    /*! Similar to the copy constructor, this operator must be defined to
+     *  ensure that the original object always owns its own memory.
+     *
+     *  \param first ManagedArray instance to copy.
+     */
+    ManagedArray& operator= (const ManagedArray &other)
+    {
+        if (m_managed)
+        {
+            throw std::runtime_error("You cannot assign to a ManagedArray that is currently managing its own memory.");
```
klarh: Maybe you had a different use case in mind, but why not? Can't you just deallocate the memory this object owns and adjust the flags appropriately?

klarh:
Having two separate flags seems like a lot of configuration space to debug. Do you have a use case in mind for when we wouldn't know whether an array input will be coming from freud or numpy? I'm wondering if it would make sense to put the `managed=False` cases into a much simpler `SizedArrayWrapper` struct or something that just stores a pointer and a size, and move the other use cases into a class with reference counting capabilities (like the SharedArray at https://github.com/glotzerlab/libgetar/blob/master/src/SharedArray.hpp#L60 and https://github.com/glotzerlab/libgetar/blob/master/gtar/_gtar.pyx#L98 ).

klarh:
I would also recommend adding tests for memory usage patterns with valgrind, but without C++-level tests it may be hard to filter out the python cruft.

vyasr:
@bdice raised some very valid concerns offline about this perhaps being too much complexity and introducing a heavy maintenance burden on the (very likely less experienced) future maintainers of freud. That's something we should definitely consider when deciding whether to move forward with this PR. If he's already having trouble understanding it, I expect that the problem will only be worse for future maintainers unless we get developers with strong C/C++ backgrounds.

vyasr:
Fundamentally, this PR is actually resolving three separate problems:
1. Providing an abstraction layer around array data structure for member variables in C++. We can always change the implementation for this on the back-end, but defining a class like this one helps standardize how developers do things like resize and clear memory when compute is called; right now developers have to actually call memset on the right array size.
2. Fixing the reference rot issues identified in #151.
3. Simplifying the Python/C++ bindings by encapsulating array sizes as part of the class.

I think that items 1 and 2 must be resolved in concert, because they're intimately tied to the underlying data representations. However, in working through this PR I think that item 3 is maybe something we shouldn't do anything about. Trying to resolve item 3 as part of this pull request introduces additional complexity for basically no gain, since you end up having to construct the ManagedArray object in the Cython compute wrapper anyway. The underlying problem is that we're interfacing with NumPy, which is basically a C library, so there's no way around using raw pointers for that interaction for inputs. For outputs and internal storage, we have more flexibility since we only need to deal with numpy storage layouts at the final step of writing the array.

I think that we should either resolve item 3 separately (by introducing a much simpler class that's just a convenience wrapper around a pointer and a size) or leave it as is. Trying to change the input array interfacing seems like adding complexity for very little gain. That change would considerably simplify some of the internals of this class. Additionally, we could start making use of smart pointers, which would substantially reduce the chances of improper memory management. This is a MAJOR plus because I think it is a major simplification of all of the hoops I'm jumping through to avoid creating memory leaks. Whether or not it makes it simple enough to be worth including is still up for debate.

cpp/util/ManagedArray.h
vyasr: Correct. As noted in the C++ class documentation, there are currently two cases where a ManagedArray can be invalidated: 
```
There are two specific scenarios when ManagedArrays are invalidated.
    1. ManagedArrays in state 2 will be invalidated if the data source
       array is deleted.
    2. ManagedArrays in state 3 will be invalidated if the data source is
        deleted or changes in size.
```

That's true not just for deallocating `m_size`, but also the underlying `m_data` pointer to the array itself. If we do what I suggested in my longer comment and stop using this object as a pointer to NumPy data, then we could write this class as pure C++ and use smart pointers to alleviate this issue. Currently the problem is that trying to point to a numpy.ndarray means that we can't use smart pointers without adding other complexity because they'll delete the data even if the numpy array should be owning it. If we stop pointing to numpy arrays some of these problems go away.

cpp/util/ManagedArray.h
vyasr: The reason I forbade this is to minimize the possibility of ManagedArrays lying around that pointed to invalidated memory (because the ManagedArray owning the memory was deallocated in this fashion). But I agree that it's perhaps a confusing restriction. If we switch to using smart pointers as discussed in the other comments, this issue goes away.

vyasr:
> Having two separate flags seems like a lot of configuration space to debug. Do you have a use case in mind for when we wouldn't know whether an array input will be coming from freud or numpy? I'm wondering if it would make sense to put the `managed=False` cases into a much simpler `SizedArrayWrapper` struct or something that just stores a pointer and a size, and move the other use cases into a class with reference counting capabilities (like the SharedArray at https://github.com/glotzerlab/libgetar/blob/master/src/SharedArray.hpp#L60 and https://github.com/glotzerlab/libgetar/blob/master/gtar/_gtar.pyx#L98 ).

I'm just reading this after all of my other comments. It seems slightly different, but largely in line with my other suggestion; stop using ManagedArray to point to numpy inputs so that we can make use of smart pointers. That would significantly reduce the memory management burden, simplifying the copy constructor and the assignment operator logic, it would fix the cases where pointers become invalidated, and it would shorten a lot of the code that's currently in place to delet arrays. So maybe that's the way to go.

vyasr:
@bdice @klarh I've simplified things quite a bit by removing usage of this class for inputs. It's now only intended as a compute class member that may need to be preserved on reallocation/resize. I no longer find the code complexity particularly concerning, and there aren't really any major gotchas anymore except for the usual Cython problems (type resolution for templates, etc). Let me know your thoughts.

cpp/cluster/Cluster.cc
```diff
@@ -25,26 +25,27 @@ Cluster::Cluster(float rcut) : m_rcut(rcut), m_num_particles(0), m_num_clusters(
...
+                      const freud::locality::NeighborList* nlist,
+                      const vec3<float> *points,
+                      unsigned int Np,
+                      freud::locality::QueryArgs qargs)
 {
     assert(points);
     assert(Np > 0);
 
     // reallocate the cluster_idx array if the size doesn't match the last one
     if (Np != m_num_particles)
```
klarh: Does it make sense to remove these guards, since resize is making the same check?

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,146 @@
...
+    unsigned int size() const
+    {
+        return *m_size;
+    }
+
+    //! Make a deep copy of this array (i.e. copy the underlying data).
+    /*! This function returns by value since all memory ownership information
+     *  is transmitted via shared pointers. 
+     */
+    void *deepCopy()
```
klarh: Is there any particular reason this is a void pointer instead of `ManagedArray<T>*`?

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,146 @@
...
+    {
+        ManagedArray<T> *deep_copy = new ManagedArray<T>(size());
+        // This *m_data will be a different shared_ptr than the current
+        // instance's, so assigning to it directly shouldn't affect any other
+        // arrays pointing to the memory of the current one and reallocating
+        // the memory of others won't affect this one.
+        *((*deep_copy).m_data) = *m_data;
```
klarh: This is doing `shared_ptr = shared_ptr`, which makes a shallow copy of the data after all, right? Am I missing something here?

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,146 @@
...
+        return deep_copy;
+    }
+        
+private:
+    //! Reallocate the data array into a new chunk of memory.
+    void create_new_array(unsigned int size)
+    {
+        // Always allocate at least size 1
+        unsigned int new_size = size > 1 ? size : 1;
```
klarh: Can you help me understand why we potentially change the size from 0?

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,146 @@
...
+{
+public:
+    //! Default constructor with optional size for thread local arrays
+    /*! \param size Size of the array to allocate.
+     */
+    ManagedArray(unsigned int size=0)
+    {
+        m_size = std::make_shared<unsigned int>(size);
+        // Hard code making an array of at least size 1.
+        m_data = std::shared_ptr<std::shared_ptr<T> >(
```
klarh, line 53: This seems very odd, can you help me understand why another allocation is necessary?

cpp/cluster/Cluster.cc
vyasr: Yes, probably. I was actually thinking that I could avoid exposing reallocation to the Cython side at all by just adding a flag that also indicates whether or not reallocation needs to happen, and then every compute class would always just call some function (called `prepArray` or something... names are hard) that would resize if needed, otherwise reallocate if needed. Does that seem reasonable? I don't know if there's any way to automate that process, but at least simplifying the logic to a single function call per compute class would be nice.

cpp/util/ManagedArray.h
vyasr: No. At one point the output of `deepCopy` was getting passed through additional function calls on the Cython side and I was just being lazy and saving myself a cast from `ManagedArray<T> *` to `void *` in Cython. Definitely no reason to leave this long term.

cpp/util/ManagedArray.h
vyasr: Yeah, the function is named poorly, it's a shallow copy.

Just in case the purpose isn't obvious to everyone, let me explain so you all can provide some guidance on how I could improve the comments/docstrings or if we should maybe rework the current structure to be a bit more transparent as to why the code is designed the way it is (apologies in advance for the verbosity):

Normally a `ManagedArray`'s data is structured as `outer_ptr1->inner_ptr1->data1`. If you pass around `ManagedArray`s by value, you create other instances, where each instance has the structure `outer_ptrX->inner_ptr1->data`, which means that any of them can resize/reallocate/whatever the data and all the others will still be pointing to the right data (and same goes for updating the size).

Now, what we want is that if compute is called and there are numpy arrays referencing the current data, we need to keep that data exactly where it is because the numpy arrays are only passed the raw pointer to that data. The first line of `deepCopy` creates a new `ManagedArray` which now contains `outer_ptr2->inner_ptr2->data2`. The assignment above then sets `inner_ptr2 = inner_ptr1`, so the pointer sequence is now `outer_ptr2->inner_ptr2->data1` (and `data2` goes out of scope and is freed). We can now safely call `reallocate` on the original array, at which point we now have `outer_ptr[1|X]->inner_ptr1->data3` while the original `ManagedArrayManager` (which had its reference count incremented for every numpy array created) now contains the `ManagedArray` with the pointer sequence `outer_ptr2->inner_ptr2->data1`, so it's keeping the original data alive and is the only pointer to it.

Hopefully you already got that from the code and were just clarifying the poor naming, but if not there's my super long description.

cpp/util/ManagedArray.h
vyasr: Ah this is a holdover from when I couldn't remember whether the C++ standard permitted 0 length arrays. I didn't want to write non-portable code and this was my kludgy solution. But I believe I looked at the standard afterward and found that 0-length _heap_ arrays are allowed post C++0x, but the only valid (and required) operation on it is deletion (dereferencing is UB). Correct me if I'm wrong.

cpp/util/ManagedArray.h
vyasr: This is the same kludge as above.

cpp/util/ManagedArray.h
klarh: Yeah, I think you can create them, but you must also delete them.

cpp/cluster/ClusterProperties.cc
```diff
@@ -104,38 +97,36 @@ void ClusterProperties::computeProperties(const box::Box& box, const vec3<float>
...
-        G[0 * 3 + 0] += delta.x * delta.x;
-        G[0 * 3 + 1] += delta.x * delta.y;
-        G[0 * 3 + 2] += delta.x * delta.z;
-        G[1 * 3 + 0] += delta.y * delta.x;
-        G[1 * 3 + 1] += delta.y * delta.y;
-        G[1 * 3 + 2] += delta.y * delta.z;
-        G[2 * 3 + 0] += delta.z * delta.x;
-        G[2 * 3 + 1] += delta.z * delta.y;
-        G[2 * 3 + 2] += delta.z * delta.z;
+        unsigned int start_idx = c*9;
```
bdice: This block and the one below seems like it should be replaced by an `Index3D` and/or a `for` loop.

cpp/cluster/ClusterProperties.h
```diff
@@ -49,30 +50,28 @@ class ClusterProperties
     }
 
     //! Get a reference to the last computed cluster_com
-    std::shared_ptr<vec3<float>> getClusterCOM()
+    util::ManagedArray<vec3<float>> getClusterCOM()
```
bdice: Why is `getClusterIdx()` returning a reference but this one (and others below) don't return a reference? It's not clear to me which signature is preferable.

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,136 @@
+#ifndef NUMERICAL_ARRAY_H
```
bdice: Replace with `MANAGED_ARRAY_H` here and at the bottom.

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,136 @@
...
+
+    //! Update size of the array.
+    /*! \param size New size of the array.
+     */
+    void resize(unsigned int size)
+    {
+        if (size != *m_size)
+        {
+            *m_size = size;
+            reallocate();
```
bdice: `resize(N)` calls `reallocate()` which calls `reset()`, which I think is an unexpected effect if comparing to std-like containers. I would prefer to make resets explicit.

In my understanding, std::vector behaves like this: `[1, 2, 3] -> resize(5) -> [1, 2, 3, 0, 0]` while the current implementation of `ManagedArray` behaves like this: `[1, 2, 3] -> resize(5) -> [0, 0, 0, 0, 0]`. That is, resizing retains or copies the elements up to the length of the container.

https://stackoverflow.com/questions/9591840/does-resizing-an-stl-vector-erase-invalidate-its-previous-contents

freud/cluster.pyx
```diff
@@ -70,15 +73,20 @@ cdef class Cluster(Compute):
...
     """
     cdef freud._cluster.Cluster * thisptr
-    cdef r_max
+    cdef float r_max
+    cdef public freud.util.ManagedArrayManager __cluster_idx
 
     def __cinit__(self, float r_max):
         self.thisptr = new freud._cluster.Cluster(r_max)
         self.r_max = r_max
+        self.__cluster_idx = freud.util.ManagedArrayManager.init(
```
bdice: This property `__cluster_idx` and its management feels super hacky. I'm not sure if the Cython side is as clean as it could be, yet. I'd like to think about this some more before I can make a recommendation for improvement. A few possible strategies:
1. Simplify more of the logic by leveraging the base `Compute` class?
2. Reduce the number of times we have to specify array dtypes?
3. Wrap more logic into `util`?

freud/util.pyx
```diff
@@ -0,0 +1,158 @@
...
+
+    def __array__(self):
+        """Convert the underlying data array into a read-only numpy array."""
+        if self.shape == tuple():
+            raise ValueError("You must specify the shape of the numpy array "
+                             "to be created by calling set_shape.")
+        cdef unsigned int ndim = len(self.shape)
+
+        # These arrays must be allocated at compile time, so we make separate
+        # branches for different dimensionalities.
```
bdice: Is there a better way? This is gross. 😕

cpp/cluster/ClusterProperties.cc
vyasr: While I agree, I think that should be saved for a separate PR.

cpp/cluster/ClusterProperties.h
vyasr: That reference return is a holdover from an earlier version where it mattered. It's no longer necessary for correctness, so I'd probably just change it back to the original signature (return by value).

cpp/util/ManagedArray.h
vyasr: You are correct that the behavior is different. However, when we resize we always wipe the memory because it's fresh computation, that's the standard use-case in freud. We could use a different function name if you prefer, but copying by default on resize is an unnecessary expense.

freud/cluster.pyx
vyasr: I'm open to any ideas on how to simplify this part of the code, I agree that I"m not a huge fan of how the Cython side of things is currently structured. Suggestions 1 and 3 are kind of vague though, I'm not really sure what you're proposing. Regarding 2, I don't think dtypes are specified in very many places... unless you're referring to template types? There's not much we can do about needing to specify the template parameters for ManagedArray, but I've tried to encapsulate all of that into util.pxd.

freud/util.pyx
vyasr: Lol not that I'm aware of. What would you suggest?

freud/util.pyx
```diff
@@ -0,0 +1,158 @@
...
+
+    def __dealloc__(self):
+        if self.var_typenum == np.NPY_UINT32:
+            del self.thisptr.uint_ptr
+
+    def __array__(self):
+        """Convert the underlying data array into a read-only numpy array."""
+        if self.shape == tuple():
+            raise ValueError("You must specify the shape of the numpy array "
+                             "to be created by calling set_shape.")
```
bdice: "setting the shape property" -- right? There isn't a set_shape method.

freud/util.pyx
bdice: @vyasr I thought some more about this -- try creating it as a one-dimensional array with size `np.prod(self.shape)` and calling `arr.reshape(self.shape)` after you've created the array object? This returns a view, so it should avoid copies.

freud/cluster.pyx
bdice: Suggestions for improvement:

1. Use single-underscores instead of double-underscores for the private variable.
2. Create an array "registration" method, inherited from `Compute`, and use it in the `__cinit__`.
```python
self.register_array(self._cluster_idx, self.thisptr.getClusterIdx())
```
3. Make the array know its intended shape on the C++ side, perhaps in a member variable like `vector<size_t> m_shape`, and use this in conjunction with the `reshape` method I proposed below. This could be integrated with Index2D and Index3D so they can return a vector, like `m_array.set_shape(index.get_shape())`.

bdice (commented):
I have some ideas for moving this forward. @vyasr If you want to discuss in person or have other thoughts, let me know.

vyasr:
> I have some ideas for moving this forward. @vyasr If you want to discuss in person or have other thoughts, let me know.

Sure let's chat tomorrow.

bdice:
@vyasr I think it would make more sense to move this new code into `freud.common` instead of `freud.util`, since they are more like the parent `Compute` class and the associated conversion logic currently in that module.

freud/util.pyx
vyasr: Good call, this definitely cleans up the code in `__array__`.

vyasr:
I actually think `freud.common` and `freud.util` should be merged (under whichever name). Now that both are Cython modules instead of Python modules, there's not really a clear separation between them IMO. I was thinking about this while making this PR; I created a `freud._util` module which contains things that used to be in freud.util but really belong in an "underscore module" based on our typical conventions.

Regarding the `Compute` class, we should revisit that as part of this PR once we get closer to a final version. I think the logic in there on whether or not something has been computed could and should be logged as part of the `ManagedArray` class. This method would also allow determination of such things at the CPP level, at the expense of a bit more code duplication since I don't think there's as nice of a way at the C++ level to globally set up these kinds of compute flags. We can discuss the possibility at a later point in this PR.

cpp/cluster/ClusterProperties.cc
```diff
@@ -42,16 +42,9 @@ void ClusterProperties::computeProperties(const box::Box& box, const vec3<float>
...
-    memset((void*) m_cluster_com.get(), 0, sizeof(vec3<float>) * m_num_clusters);
-
-    m_cluster_G = std::shared_ptr<float>(new float[m_num_clusters * 3 * 3], std::default_delete<float[]>());
-    memset((void*) m_cluster_G.get(), 0, sizeof(float) * m_num_clusters * 3 * 3);
-
-    m_cluster_size = std::shared_ptr<unsigned int>(new unsigned int[m_num_clusters],
-                                                   std::default_delete<unsigned int[]>());
-    memset((void*) m_cluster_size.get(), 0, sizeof(unsigned int) * m_num_clusters);
+    m_cluster_com.prepare(m_num_clusters);
+    m_cluster_G.prepare(m_num_clusters*3*3);
```
bdice:
```suggestion
    m_cluster_G.prepare({m_num_clusters, 3, 3});
```

bdice: Let's move towards indexing the ManagedArray instead of carrying around Index2D/3D objects. (@vyasr has logic for this already.)

freud/_cluster.pxd
```diff
@@ -2,27 +2,26 @@
...
+        void compute(const freud._locality.NeighborQuery*,
+                     const freud._locality.NeighborList*,
+                     const vec3[float] *,
+                     unsigned int,
+                     freud._locality.QueryArgs) nogil except +
         void computeClusterMembership(const unsigned int*) nogil except +
         unsigned int getNumClusters()
         unsigned int getNumParticles()
-        shared_ptr[unsigned int] getClusterIdx()
+        const freud.util.ManagedArray[unsigned int] &getClusterIdx()
```
bdice, line 24:
```suggestion
        const freud.util.ManagedArray[unsigned int] getClusterIdx()
```

cpp/cluster/ClusterProperties.h
bdice: Resolving this -- see suggested edit on the other line.

freud/cluster.pyx
```diff
@@ -138,10 +143,11 @@ cdef class Cluster(Compute):
...
     def cluster_idx(self):
-        cdef unsigned int n_particles = self.thisptr.getNumParticles()
-        cdef const unsigned int[::1] cluster_idx = \
-            <unsigned int[:n_particles]> self.thisptr.getClusterIdx().get()
-        return np.asarray(cluster_idx)
+        cdef freud.util.ManagedArray[uint] ma = self.thisptr.getClusterIdx()
+        cdef freud.util.ManagedArrayManager mam = \
+            freud.util.ManagedArrayManager.init(
+                &ma, freud.util.arr_type_t.UNSIGNED_INT)
+        return np.asarray(mam)
```
bdice: Is there a way to rewrite this logic into one line? Perhaps using templates? Something like
```python
@Compute._computed_property("compute")
def cluster_idx(self):
    return make_array[uint](self.thisptr.getClusterIdx())
```

bdice (commented):
@vyasr Please remove the changes to Cython-generated files. A few comments below.

freud/cluster.pyx
vyasr: Cython doesn't support templates, which is really what makes the Cython code for this PR as ugly as it is. I've tried to avoid it as much as possible for that reason, and I'll try to streamline it further over time as I incorporate ManagedArrays into more classes.

cpp/util/ManagedArray.h
```diff
@@ -0,0 +1,316 @@
...
+
+namespace freud { namespace util {
+
+//! Class to handle the storage of all arrays of numerical data used in freud.
+/*! The purpose of this class is to handle standard memory management, and to
+ *  provide an abstraction around the implementation-specific choice of
+ *  underlying data structure for arrays of data in freud. These arrays are
+ *  specifically designed for numerical data calculated by a compute class,
+ *  particularly for arrays that must be made accessible through the Python API
+ *  of freud. To support multidimensional arrays, the underlying data is stored
```
bdice: Change this last sentence:
The array shape is stored and used to enable multidimensional indexing.

freud/cluster.pyx
```diff
@@ -143,10 +145,9 @@ cdef class Cluster(PairCompute):
 
     @Compute._computed_property("compute")
     def cluster_idx(self):
-        cdef unsigned int n_particles = self.thisptr.getNumParticles()
-        cdef const unsigned int[::1] cluster_idx = \
-            <unsigned int[:n_particles]> self.thisptr.getClusterIdx().get()
-        return np.asarray(cluster_idx)
+        return freud.util.make_managed_numpy_array(
+            <const void *> &self.thisptr.getClusterIdx(),
```
bdice:
```suggestion
            &self.thisptr.getClusterIdx(),
```

PR #371, opened 2019-08-08
Fix Condon-Shortley Phase resulting in incorrect Wl order parameters.

### LINKED ISSUE #369: Steinhardt Wl parameters are computed incorrectly
**Describe the bug**
I expected that the results from the Wl order parameter would be rotationally invariant. It doesn't seem like the result I get are what I expect.

**To Reproduce**
1. Start with a single particle at the origin surrounded by 12 neighbors in an FCC arrangement.
2. Create a `Steinhardt` or `LocalWl` object with a manually-constructed neighborlist.
3. Compute the order parameters. The real part of W6 for the particle at the origin is -0.00262604 for Steinhardt, LocalWl, and pyboo.
4. Rotate the points around the origin and compute the order parameters again.
5. Results disagree with those of `pyboo`. The results from freud appear to be different after the rotation.

```python
import rowan
import freud
import numpy as np
import boo

box = freud.box.Box.cube(10)
positions = np.array([[0, 0, 0],
                      [-1, -1, 0],
                      [-1, 1, 0],
                      [1, -1, 0],
                      [1, 1, 0],
                      [-1, 0, -1],
                      [-1, 0, 1],
                      [1, 0, -1],
                      [1, 0, 1],
                      [0, -1, -1],
                      [0, -1, 1],
                      [0, 1, -1],
                      [0, 1, 1]])
index_i = np.zeros(12)
index_j = np.arange(1, 13)
nlist = freud.locality.NeighborList.from_arrays(13, 13, index_i, index_j)
nlist.distances[:] = np.sqrt(2)
ql = freud.order.Steinhardt(r_max=1.5, l=6, Wl=True)
lwl = freud.order.LocalWl(box, 1.5, 6)

print('FCC:')
ql.compute(box, positions, nlist=nlist)
print('Steinhardt:', ql.order)
lwl.compute(positions, nlist=nlist)
print('LocalWl:', lwl.Wl)
q6m = boo.bonds2qlm(positions, nlist[:], l=6)
w6 = boo.wl(q6m)
print('pyboo:', w6)

# Again but rotated
quat = rowan.random.rand()
positions_rot = rowan.rotate(quat, positions)

print('\nROTATED FCC:')
ql.compute(box, positions_rot, nlist=nlist)
print('Steinhardt:', ql.order)
lwl.compute(positions_rot, nlist=nlist)
print('LocalWl:', lwl.Wl)
q6m = boo.bonds2qlm(positions_rot, nlist[:], l=6)
w6 = boo.wl(q6m)
print('pyboo:', w6)
```

**Error output**
```python
FCC:
Steinhardt: [-0.00262604-4.1720927e-10j         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj]
LocalWl: [-0.00262604-4.1720927e-10j         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj         nan          +nanj
         nan          +nanj]
pyboo: [-0.00262604 -0.09791764 -0.09791764 -0.09791764 -0.09791764 -0.09791764
 -0.09791764 -0.09791764 -0.09791764 -0.09791764 -0.09791764 -0.09791764
 -0.09791764]

ROTATED FCC:
Steinhardt: [-0.00854041-0.00187109j         nan       +nanj         nan       +nanj
         nan       +nanj         nan       +nanj         nan       +nanj
         nan       +nanj         nan       +nanj         nan       +nanj
         nan       +nanj         nan       +nanj         nan       +nanj
         nan       +nanj]
LocalWl: [-0.00854041-0.00187108j         nan       +nanj         nan       +nanj
         nan       +nanj         nan       +nanj         nan       +nanj
         nan       +nanj         nan       +nanj         nan       +nanj
         nan       +nanj         nan       +nanj         nan       +nanj
         nan       +nanj]
pyboo: [-0.00262604 -0.09791764 -0.09791764 -0.09791764 -0.09791764 -0.09791764
 -0.09791764 -0.09791764 -0.09791764 -0.09791764 -0.09791764 -0.09791764
 -0.09791764]
```

 - OS: Ubuntu
 - Version of Python: 3.7
 - Version of freud: `next`, commit 554290969eb998ecc21563ebac4bcefeaa0ddf2b

bdice:
## Description
As described in #369, the Steinhardt Wl parameters computed by the `Steinhardt` and `LocalWl` classes are rotationally _variant_ -- when they should be rotationally _invariant_. This fixes the incorrect computations by adding a factor for the Condon-Shortley phase to the spherical harmonics _Ylm_ for all positive, odd values of _m_. Thanks to @klarh for the recommendation and @csadorf for catching a similar issue in https://github.com/glotzerlab/pythia/issues/9!

## Motivation and Context
Resolves #369. I manually tested v0.8.2 and v0.6.1 of freud (the earliest versions I could get via conda-forge and via the glotzer conda channel) and this problem with rotational invariance appeared in both versions. I can't go back any further in time than this without extensive effort, but I will document this problem clearly in the Changelog entry.

Before the introduction of the `fsph` module to compute spherical harmonics, the code used `boost::math::spherical_harmonic`. This function is [documented here, as of boost 1.53 which was used in freud v0.4.0](https://www.boost.org/doc/libs/1_53_0/libs/math/doc/sf_and_dist/html/math_toolkit/special/sf_poly/sph_harm.html). This function uses Condon-Shortley phases implicitly in its definition of the associated Legendre polynomials, and it says it adheres to the conventions used in Physics (which differ from the "math" conventions used in fsph). I would guess that a translation error was made when moving to fsph, between v0.4.1 and v0.5.0.

## How Has This Been Tested?
I added tests for rotational invariance, and applied them to all Steinhardt classes (LocalQl, LocalWl, and the new Steinhardt class).

## Types of changes

## Checklist:


cpp/order/Steinhardt.cc
```diff
@@ -303,17 +292,23 @@ float Steinhardt::normalize()
 std::complex<float> Steinhardt::normalizeWl()
 {
     // Get Wigner 3j coefficients:
-    // j1 from -l to l
-    // j2 from max(-l-j1, -l) to min(l-j1, l)
+    // j1 = j2 = j3 = l
+    // m1 from -l to l
+    // m2 from max(-l-m1, -l) to min(l-m1, l)
+    // m3 = -m1 - m2
```
bdice: I rewrote this indexing math so it was clearer to anyone reading about Wigner 3j symbols and follows the conventions of these two pages much more clearly:
http://mathworld.wolfram.com/Wigner3j-Symbol.html
https://en.wikipedia.org/wiki/3-j_symbol

cpp/order/wigner3j.h
```diff
@@ -16,32 +16,40 @@ vector<float> getWigner3j(unsigned int l);
...
-    W_l = np.zeros(counter, dtype=np.float64)
-
-    j = 0
-    for u1 in range(2*l+1):
-        for u2 in range(max(l-u1,0),min(2*l+1,3*l-u1+1)):
-            u3 = 3*l-u2-u1
-            W_l[j] = float(wigner_3j(l,l,l,u1-l,u2-l,u3-l))
-            j += 1
-    wigner['l'+str(l)] = W_l.tolist()
+for l in range(21):
```
bdice: This new (commented) Python script can be used to regenerate the Wigner 3j coefficients with greater automation than the previous version. Almost all the code in `wigner3j.cc` is generated by this script. I regenerated that file because I didn't trust its results at first and also wanted to add the ability to compute Wl for small (0, 1) and odd (3, 5, 7, ..., 19) values of l.

klarh (approved):
Looks good!

vyasr:
Unfortunately, the use of the Condon-Shortley phase is common in many tools, including both sympy and SciPy IIRC. Makes sense that the quantity should be real, though; I always thought it was strange that there was an imaginary component but never followed up. Thanks for tracking this down!

I'm OK pushing through this PR quickly since it's a bug fix, but can you prioritize writing a more thorough validation test by adapting the Ql calculation I wrote in #352, similar to my request for your other open PR #367? I know it might seem like overkill to reimplement the calculation in Python, but I'm very uncomfortable with how our lack of testing allows these kinds of bugs to slip through the cracks unnoticed for long periods of time on relatively high-leverage components of the code base. I would prefer to take these bug fix PRs as opportunities to properly buff up our validation. The calculations are also relatively straightforward to implement in pure Python if you make use of LocalDescriptors to compute the spherical harmonics. For now, I'm OK having cross-dependencies in the testing suite. At worst it means that test failures might initially be slightly misleading, but it also effectively means higher coverage.

bdice:
@vyasr I added the explicit test with a Python computation of the Wls. I had to add sympy as a requirement because it's the simplest way to get the Wigner 3j symbols.

I think this is ready to merge after we address one question: should we change the return type of the Wl arrays to be `np.float32` instead of `np.complex64` and drop the imaginary components internally, since we should have real-valued results?

ChangeLog.md
```diff
@@ -5,10 +5,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Next
 
 ### Fixed
+* Fixed missing Condon-Shortley phase affecting LocalWl and Steinhardt Wl
+  computations. This missing factor of -1 caused results for third-order (Wl)
+  Steinhardt order parameters to be incorrect, shown by their lack of
+  rotational invariance. This problem was introduced in v0.5.0.
```
vyasr, line 11: Minor nitpick: we should be consistent in formatting new lines in the Changelog. We haven't typically included any, so for purely historical reasons I would prefer not to change that rather than reformat the document, but if we want to start doing it we should reformat thoroughly.

appveyor.yml
```diff
@@ -13,7 +13,7 @@ build_script:
   - conda config --set always_yes yes --set changeps1 no
   - conda config --add channels conda-forge
   - conda update -q conda
-  - conda install pip numpy scipy cython tbb tbb-devel
+  - conda install pip cython matplotlib numpy rowan scipy sympy tbb tbb-devel
```
vyasr, line 16: If we're going to introduce rowan as a testing dependency, we should also update other tests (e.g. Cubatic) that implement their own quaternion generation logic, etc.

vyasr: Can you make a separate issue for that? No need to address in this PR.

cpp/order/LocalWl.cc
```diff
@@ -34,13 +34,13 @@ void LocalWl::computeYlm(const float theta, const float phi, std::vector<std::co
     for (typename fsph::PointSPHEvaluator<float>::iterator iter(sph_eval.begin_l(m_l, 0, false));
          iter != sph_eval.end(); ++iter)
     {
-        Ylm[(j + m_l) % (2 * m_l + 1)] = *iter;
+        // Manually add the Condon-Shortley phase, (-1)^m, to positive odd m
+        float phase = (j % 2 == 1) ? -1 : 1;
```
vyasr, line 38: Given our discussions a few weeks ago, it's probably worth adding a note that `(j % 2) == -1` if j is a _negative_ odd number since that's not the behavior of Python, only C(++).

cpp/order/LocalWl.cc
```diff
@@ -34,13 +34,13 @@ void LocalWl::computeYlm(const float theta, const float phi, std::vector<std::co
...
+        Ylm[m_l + j] = phase * (*iter);
```
vyasr, line 39: Do we not need the old modulo math? It would make sense that we just get the positive values from the iterator, but I'm confused about why the modulus used to be there at all.

cpp/order/LocalWl.cc
```diff
@@ -34,13 +34,13 @@ void LocalWl::computeYlm(const float theta, const float phi, std::vector<std::co
...
+        if (j > 0)
+            Ylm[m_l - j] = std::conj(*iter);
```
vyasr, line 41: Is the lack of conjugation from before also an error?

cpp/order/LocalWl.cc
```diff
@@ -79,7 +79,7 @@ void LocalWl::compute(const locality::NeighborList* nlist, const vec3<float>* po
             {
                 const unsigned int index = (2 * m_l + 1) * i;
                 const unsigned int u3 = 3 * m_l - u1 - u2;
-                m_Wli.get()[i] += m_wigner3jvalues[counter] * m_Qlmi.get()[index + u1]
+                m_Wli.get()[i] += float(m_wigner3jvalues[counter]) * m_Qlmi.get()[index + u1]
```
vyasr: Why are all these float conversions necessary? The object is a std::vector, it should already return floats.

cpp/order/wigner3j.cc
```diff
@@ -9,1306 +9,2466 @@
...
+        };
     }
+
     case 12:
     {
-        float wigner3j12[469]
-            = {0.00764132640913,  -0.0194816362349,  0.035645677572,    -0.0537378808656,  0.0703594019295,
-               -0.0820524576002,  0.0862629766784,   -0.0820524576002,  0.0703594019295,   -0.0537378808656,
-               0.035645677572,    -0.0194816362349,  0.00764132640913,  -0.0194816362349,  0.0420272952502,
-               -0.0633234617366,  0.0752330332119,   -0.0718102638895,  0.0517033900004,   -0.018824124
```
vyasr: "Are not"

cpp/order/LocalWl.cc
vyasr: OK, I see that you changed that too. Why? Were you getting compiler warnings from hardcoding numbers in the wigner3j.cc file that overflow float precision?

cpp/order/LocalWl.cc
vyasr: Wow OK yes it is. That's pretty egregious.

cpp/order/Steinhardt.h
```diff
@@ -157,6 +157,13 @@ class Steinhardt
     void aggregateWl(std::shared_ptr<std::complex<float>> target,
                      std::shared_ptr<std::complex<float>> source);
 
+    //! Compute array index for m values, indexed like:
+    //  [0, 1, ..., l, -1, -2, ..., -l]
+    int mIndex(int m) const
+    {
+        return m < 0 ? m_l - m : m;
+    }
+
```
vyasr: @bdice @klarh Can we make a PR to fsph that creates a C++ function equivalent to the Python function `get_LMs` that it exposes? It would make a freud's indexing w.r.t. spherical harmonics much more transparent if fsph actually told you the order in which it returned spherical harmonics, ideally via a simple iterator.

vyasr (changes requested):
Some questions and some change requests.

ChangeLog.md
bdice: I am planning to wrap all long Changelog lines in the next release branch.

cpp/order/LocalWl.cc
bdice: I know... but it's fixed now.

cpp/order/LocalWl.cc
bdice: No, that modulus was unneeded. I tried testing with and without it, and it made no difference.

cpp/order/LocalWl.cc
bdice: `j` is nonnegative -- it's an unsigned int. We use j to index over m=0, 1, 2, 3, and we compute the negative m values by conjugating the positive m values below. (We only apply the Condon-Shortley phase to positive odd m.)

cpp/order/LocalWl.cc
bdice: Yeah, I was getting errors (not just warnings, if I recall correctly). I changed the return type to `vector<float>` and it was interpreting the vector literals `{0.0, 0.1, ...}` as doubles. This was the easier thing to change.

cpp/order/Steinhardt.h
bdice: That's a great idea -- I considered doing that but I felt it was out of scope for this PR. I created https://github.com/glotzerlab/fsph/issues/1.

cpp/order/wigner3j.cc
bdice: @vyasr I'm confused, is this comment meant to apply to this part of the file?

bdice:
@vyasr Please re-review when you get a chance. I think I have responded to all of your comments. Please mark them as resolved when you've seen my replies, or reply if you think additional actions are needed.

cpp/order/LocalWl.cc
vyasr: My mistake, I think I got confused between this and Steinhardt with respect to whether the iterator was including the negative `m` values. I think the comment is a bit confusing since you add the Condon-Shortley phase to all positive `m`, it's just that the phase is positive for even `m`, but that's a minor nitpick. Update if you feel like it.

cpp/order/LocalWl.cc
vyasr: Yeah that sounds right. Another random old oversight I suppose.

cpp/order/LocalWl.cc
vyasr: What caused errors that didn't cause errors before? Was it because you increased the precision of the numbers hard-coded in the file?

cpp/order/Steinhardt.h
vyasr: Agreed, that can be a separate PR.

vyasr:
Just one question left regarding the double <--> float interconversion, I think it's a little ugly and am confused as to why it's necessary (but am OK leaving it as is if it is needed).

PR #373, opened 2019-08-09
Make Wl real.

bdice:
## Description
Incorrect computations led us to treat Steinhardt Wl order parameters as complex-valued. This PR corrects that by using real-valued arrays and return types for Wl order parameters.

I refactored the Wigner 3j reduction operations into the `wigner3j.[cc,h]` files, so that this is implemented exactly once and is re-used within the Steinhardt class for computing the order parameter and the norm over all particles.

The refactoring of the LocalWl class was less extensive -- I just changed the types of the arrays and dropped the imaginary parts, since we will remove this class in the 2.0 release.

## Motivation and Context
Related to #371, #369.

## How Has This Been Tested?
Previous tests only addressed the real part of the computation -- the existing tests are sufficient.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/order/LocalWl.h
+++ b/cpp/order/LocalWl.h
@@ -5,7 +5,7 @@
 #define LOCAL_WL_H
 
 #include "LocalQl.h"
-#include "wigner3j.h"
+#include "Wigner3j.h"
 
 /*! \file LocalWl.h
     \brief Compute a Wl per particle
@@ -19,7 +19,7 @@ namespace freud { namespace order {
  * Steinhardt that can aid in distinguishing between FCC, HCP, BCC.
  *
  * For more details see PJ Steinhardt (1983) (DOI: 10.1103/PhysRevB.28.784)
- * Uses a Python wrapper to pass the wigner3j coefficients to C++
+ * Uses a Python wrapper to pass the Wigner 3j coefficients to C++
  */
 //! Added first/second shell combined average Wl order parameter for a set of points
 /*!
--- a/cpp/order/LocalWl.h
+++ b/cpp/order/LocalWl.h
@@ -19,7 +19,6 @@ namespace freud { namespace order {
  * Steinhardt that can aid in distinguishing between FCC, HCP, BCC.
  *
  * For more details see PJ Steinhardt (1983) (DOI: 10.1103/PhysRevB.28.784)
- * Uses a Python wrapper to pass the Wigner 3j coefficients to C++
  */
 //! Added first/second shell combined average Wl order parameter for a set of points
 /*!
--- a/cpp/order/LocalWl.h
+++ b/cpp/order/LocalWl.h
@@ -62,22 +62,22 @@ public:
 
     //! Get a reference to the last computed Wl/WlNorm for each particle.
     //  Returns NaN for particles with no neighbors.
-    std::shared_ptr<std::complex<float>> getWl()
+    std::shared_ptr<float> getWl()
     {
         return m_Wli;
     }
-    std::shared_ptr<std::complex<float>> getWlNorm()
+    std::shared_ptr<float> getWlNorm()
     {
         return m_WliNorm;
     }
 
     //! Get a reference to the last computed AveWl/AveWlNorm for each particle.
     //  Returns NaN for particles with no neighbors.
-    std::shared_ptr<std::complex<float>> getAveWl()
+    std::shared_ptr<float> getAveWl()
     {
         return m_AveWli;
     }
-    std::shared_ptr<std::complex<float>> getAveNormWl()
+    std::shared_ptr<float> getAveNormWl()
     {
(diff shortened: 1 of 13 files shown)


cpp/order/LocalWl.h
```diff
@@ -19,7 +19,7 @@ namespace freud { namespace order {
  * Steinhardt that can aid in distinguishing between FCC, HCP, BCC.
  *
  * For more details see PJ Steinhardt (1983) (DOI: 10.1103/PhysRevB.28.784)
- * Uses a Python wrapper to pass the wigner3j coefficients to C++
+ * Uses a Python wrapper to pass the Wigner 3j coefficients to C++
```
vyasr: This line isn't actually true. I assume that at some point we actually passed Wigner coefficients in, but now they're just hardcoded so I would remove this comment altogether.

cpp/order/Wigner3j.cc
```diff
@@ -1,17 +1,67 @@
 // Copyright (c) 2010-2019 The Regents of the University of Michigan
 // This file is from the freud project, released under the BSD 3-Clause License.
 
+#include <algorithm>
```
vyasr, line 4: I don't think you're actually using this anywhere are you?

cpp/order/Steinhardt.cc
```diff
@@ -278,83 +274,35 @@ void Steinhardt::computeAve(const box::Box& box, const locality::NeighborList* n
...
-                target.get()[i] += float(wigner3jvalues[counter])
-                    * source.get()[particle_index + m1_index]
-                    * source.get()[particle_index + m2_index]
-                    * source.get()[particle_index + m3_index];
-                counter++;
-            }
-        } // Ends loop over Wigner 3j coefficients
-    } // Ends loop over particles
+            const unsigned int particle_index = (2 * m_l + 1) * i;
+            target.get()[i] = reduceWigner3j(&source.get()[particle_index], m_l, wigner3jvalues);
```
vyasr: Can you add some parenthesis around `source.get()[particle_index]` just to avoid any future reading confusion about operator precedence?

cpp/order/Steinhardt.h
```diff
@@ -96,28 +100,27 @@ class Steinhardt
...
-    //! Get whether the Wl flag was set
-    bool getUseWl()
-    {
-        return m_Wl;
-    }
-
-    //! Get Wl norm
-    std::complex<float> getNormWl()
+    //! Get the last calculated Ql
+    std::shared_ptr<float> getQl()
```
vyasr, line 104: This function name is misleading if it can return a Wl value. Is that actually what it should do? I would expect it to return either `m_Qli` or `m_Qlmi`.

cpp/order/Steinhardt.h
```diff
@@ -182,9 +174,8 @@ class Steinhardt
     std::shared_ptr<float> m_QliAve;           //!< Averaged Ql with 2nd neighbor shell for each particle i
     std::shared_ptr<complex<float>> m_QlmiAve; //!< Averaged Qlm with 2nd neighbor shell for each particle i
     std::shared_ptr<std::complex<float>> m_QlmAve;   //!< Normalized QlmiAve for the whole system
-    float m_Norm;                                    //!< System normalized norm over all Qlm(Ave)
-    std::complex<float> m_NormWl;                    //!< System normalized norm over all Wlm(Ave)
-    std::shared_ptr<std::complex<float>> m_WliOrder; //!< Wl order parameter for each particle i
+    float m_norm;                                    //!< System normalized order parameter
+    std::shared_ptr<float> m_Wli; //!< Wl order parameter for each particle i
```
vyasr: Can you add a comment thatWli is reused for average as well?

vyasr (changes requested):

cpp/order/LocalWl.h
bdice: I removed it.

cpp/order/Wigner3j.cc
bdice: It is a required `#include` for `min` and `max` on Windows.

cpp/order/Steinhardt.cc
bdice: Done.

cpp/order/Steinhardt.h
bdice: This was a mistake on my part. Fixed.

cpp/order/Steinhardt.h
bdice: Done.

bdice:
@vyasr I think this PR is fixed up, please re-review when convenient.

vyasr (approved):

vyasr:
This looks good. We can merge into the other branch whenever you think it'll be easiest w.r.t. finalizing that branch and merging into master.

PR #374, opened 2019-08-12
Fix/gaussian args

vyasr:
## Description
I changed the constructor signature for GaussianDensity to accept various types of width arguments rather than having two distinct signatures.

## Motivation and Context
This change simplifies the API into a more Pythonic form. Users would expect widths to accept a list if we want different widths in different directions rather than have different arguments as in the current signature.

## How Has This Been Tested?
Updating existing tests to use the new API, and they pass.

## Types of changes

## Checklist:


cpp/density/GaussianDensity.h
```diff
@@ -39,28 +37,30 @@ class GaussianDensity
         return m_box;
     }
 
+    //! Get the simulation box
```
bdice: Fix comment

freud/_density.pxd
```diff
@@ -6,6 +6,8 @@ from libcpp.memory cimport shared_ptr
 cimport freud._box
 cimport freud._locality
 
+ctypedef unsigned int uint
```
bdice: Is this necessary? If not, I would remove it. I have had luck with just using `unsigned int` recently, it may be a newer fix in Cython.

freud/density.pyx
```diff
@@ -24,6 +24,8 @@ cimport numpy as np
 # _always_ do that, or you will have segfaults
 np.import_array()
 
+ctypedef unsigned int uint
```
bdice, line 29: Similarly, I would remove this typedef if it isn't needed to make the code work.

freud/density.pyx
```diff
@@ -528,19 +515,29 @@ cdef class GaussianDensity(Compute):
...
-            self.thisptr = new freud._density.GaussianDensity(
-                args[0], args[1], args[2], args[3], args[4])
-            self.arglist = [args[0], args[1], args[2], args[3], args[4]]
+    cdef float r_max
+
+    def __cinit__(self, width, r_max, sigma):
+        cdef vec3[uint] width_vector
+        if isinstance(width, int):
+            width_vector = vec3[uint](width, width, width)
+        elif isinstance(width, (list, tuple)):
```
bdice: Can we use `Sequence` from `collections.abc`? https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

freud/density.pyx
```diff
@@ -528,19 +515,29 @@ cdef class GaussianDensity(Compute):
...
+            if len(width) == 2:
+                width_vector = vec3[uint](width[0], width[1], 1)
+            elif len(width) == 3:
+                width_vector = vec3[uint](width[0], width[1], width[2])
+            else:
+                raise ValueError("The width must be either a number of pixels "
```
bdice: pixels vs. bins? I tend to prefer "bins" since we aren't really dealing with displays.

freud/density.pyx
```diff
@@ -528,19 +515,29 @@ cdef class GaussianDensity(Compute):
...
+                                 "or a list of length 3 (2) indicating the "
+                                 "widths in the 3 (2) spatial dimensions.")
```
bdice: in each spatial dimension (don't repeat the 3/2 thing)

freud/density.pyx
```diff
@@ -528,19 +515,29 @@ cdef class GaussianDensity(Compute):
...
+        if isinstance(width, int):
+            width_vector = vec3[uint](width, width, width)
+        elif isinstance(width, (list, tuple)):
+            if len(width) == 2:
+                width_vector = vec3[uint](width[0], width[1], 1)
+            elif len(width) == 3:
+                width_vector = vec3[uint](width[0], width[1], width[2])
+            else:
+                raise ValueError("The width must be either a number of pixels "
+                                 "or a list of length 3 (2) indicating the "
```
bdice: I would make it more obvious that this number matches the system dimensionality.

freud/density.pyx
```diff
@@ -571,38 +568,34 @@ cdef class GaussianDensity(Compute):
 
     @Compute._computed_property()
     def gaussian_density(self):
-        cdef unsigned int width_x = self.thisptr.getWidthX()
-        cdef unsigned int width_y = self.thisptr.getWidthY()
-        cdef unsigned int width_z = self.thisptr.getWidthZ()
-        cdef unsigned int array_size = width_x * width_y
+        cdef vec3[uint] width = self.thisptr.getWidth()
+        cdef unsigned int array_size = width.x * width.y
```
bdice: Can you flip the box declaration with the `array_size` declaration so that the width math stays together? Or:

`cdef unsigned int array_size = width.x * width.y * (1 if box.is2D() else width.z)`

ChangeLog.md
```diff
@@ -4,9 +4,24 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 ## Upcoming: Changes for freud 2.0
 
+### Added
+* Ability to specify NeighborQuery objects as points for neighbor-based pair computes.
+* Various validation tests.
+
+### Changed
+* All compute objects that perform neighbor computations now use NeighborQuery internally.
+* All compute objects that perform neighbor computations now loop over NeighborBond objects.
+* Renamed ref\_points, points to points, query\_points to clarify their usage.
```
bdice: Add parens around these pairs. `(ref_points, points) to (points, query_points)`

ChangeLog.md
```diff
@@ -4,9 +4,24 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+* Standardized naming of various standard parameters across freud.
```
bdice: image

bdice (changes requested):
Great PR - some minor suggestions below.

freud/_density.pxd
vyasr: Interesting, looks like it is no longer necessary. I wonder when they implemented that, it used to cause a Cython parsing error to use `unsigned int` as a template parameter. Good catch.

freud/density.pyx
vyasr: Actually, it still fails in this place. It looks like they've got it working in some contexts, but not others (I would guess lvalues but not rvalues?). I'll remove it from the pxd but need to keep it here.

freud/density.pyx
vyasr: I actually considered this, but I decided I wanted to be a bit more proscriptive with what we permit. Supporting arbitrary sequences here provides no real benefit IMO, so I would prefer not to be super ducktyping friendly in this context. We could always add that support later if we found a compelling reason to do so.

ChangeLog.md
vyasr: What's redundant? The second line is referring to things like rcut, num_neighbors, etc. Could make that explicit if you like.

freud/density.pyx
bdice: @vyasr If we don't, this will fail with NumPy arrays of input, which I think is a problem. We've made strides to use / allow NumPy arrays explicitly for all list-like numeric inputs and outputs elsewhere in the package.

ChangeLog.md
bdice: Sorry, I didn't make my comment clear; "Standardized ... standard parameters" sounded redundant.

freud/density.pyx
vyasr: That's fair. Sequence still seems a bit too permissive to me, but not a huge deal.

ChangeLog.md
vyasr: Got it.

freud/density.pyx
```diff
@@ -521,19 +523,20 @@ cdef class GaussianDensity(Compute):
...
+                                 "or a list indicating the widths in each "
+                                 "spatial dimension (length 2 in 2D, length 3 "
+                                 "in 3D).")
         else:
-            raise ValueError("The width must be either a number of pixels "
-                             "or a list of length 3 (2) indicating the "
-                             "widths in the 3 (2) spatial dimensions.")
+            raise ValueError("The width must be either a number of bins or a "
+                             "list indicating the widths in each " "spatial "
+                             "dimension (length 2 in 2D, length 3 " "in 3D).")
```
bdice: Fix the weird spacing here between consecutive strings, they're broken up by double quotes for no reason.

bdice (approved):
Please fix the one thing I marked and then this can be merged.

_edit:_ I fixed it.

bdice:
@vyasr A style error caused early failure with previous commits, but now with the `collections.abc` module, we're Python 2 incompliant. Can we drop Python 2 tests for `next`? See #380.

vyasr:
This should be merged after #380

PR #383, opened 2019-08-15
Add citations

### LINKED ISSUE #214: Export citations
Some of freud's methods are quite novel and are developed in other papers. We should make it easy to export citations for the relevant papers when a given method is being used. Most of the relevant citations should already exist in the freud documentation.

bdice:
## Description
This PR adds instructions for citing freud, with a page in the docs modeled after that of signac.

## Motivation and Context
Resolves #214. When the paper is published, we will update the citation info.

## How Has This Been Tested?
No tests needed.

## Types of changes
Docs only. 📜 

## Checklist:


doc/source/index.rst
```diff
@@ -34,13 +34,24 @@ The freud library uses single-precision `NumPy arrays <https://www.numpy.org/>`_
...
    modules
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Reference
+
    development
-   citations
+   citing
+   references
```
vyasr, line 54: I think our references section is a little strange to be honest. I think we should put all citations inline (i.e. in the context of whatever method they're there for) and get rid of this section. We should also get more consistent with having citations, but that's a longer term things (maybe good for a hackathon to ask someone to identify features that need a citation).

vyasr (changes requested):
One request relating to references, I'd like to move them inline. I don't think having a references section really makes sense for this.

doc/source/index.rst
vyasr: Discussed offline, we will do this in a separate PR.

Issue #179, opened 2018-06-21
Standardize the freud API


bdice:
From roadmap planning session with @vyasr and @bdice. To be assigned after #176 is closed.

# Standardize the freud API

Once #176 (doc cleaning) is done, we should identify what we want APIs to look like for everything in freud.

1. Review existing APIs
2. Determine a standard for all modules to follow, with minimal exceptions
3. Create a list of cases where the standard is not currently followed

The behavior of compute/accumulate/property getters is one notable case that should be standardized.

For methods that we want to remove (e.g. getRDF, which should be replaced by a property), we will remove that method from the documentation and add deprecation warnings for version 2.0.

For any class/function where the signature has to change, we will make use of `*args, **kwargs` to take variable APIs and then dynamically resolve them. Deprecation warnings will be issued wherever appropriate.

csadorf:
I would expect that similar to the different classes in `sklearn`, for instance all the different classes in the `order` module expose the exact same API (as much as possible) so to make it easier to substitute those in one workflow.

bdice:
`LinkCell.getCell` should probably be `LinkCell.getCellIndex`

bdice:
The correlation functions expose their data in the property `RDF` but that is confusing to many users because it's not actually the radial distribution function (it's the correlation function).

bdice:
- [x] Computing cluster membership should just be part of the main cluster compute call.

jinsooihm:
`AngularSeparation.computeNeighbor` does not allow `points=None` default argument, while most of the other methods do. 
Also, some of the documentation is wrong. The arguments of `computeGlobal` are ordered in  `global_ors`, `ors`, `equiv_quats`, but the documentation does not agree with this.

jinsooihm:
`BondOrder` claims to use cutoff distance, which is passed in as an argument for the constructor as `rmax`.
However, it is never really used anywhere other than for the `rmax_guess` for `make_default_nlist_nn`.
This might lead to unexpected behavior if the user wanted to use a strict cutoff.

jinsooihm:
`LocalDescriptors` documentation seem to be misleading.
1. `num_neighbors` and `rmax` does not seem to be necessary for the constructor, as they are never really used.
[...]

PR #348, opened 2019-07-22
Docs/add tutorial

vyasr:
## Description
Adds more documentation to the package, starting with an introduction to what freud is, a quick start guide, and a more detailed tutorial. This PR should not be merged until all of the neighbor finding refactoring is complete, since we need to verify that all the APIs, etc are correct.

## Motivation and Context
Our current documentation only includes API information, but has no guidance for users on how to start out using freud. This PR addresses that need.

Also resolves #274 as much as we can; I suggest fleshing that section out as we receive questions on specific other data formats people want to read into freud.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

PR #395, opened 2019-08-21
Feature/standardize query args

vyasr:
## Description
This PR cleans up some of the C++ code for the `QueryArgs` class, introducing more constants into the namespace and improving code consistency. Query argument names have been standardized further, and updated throughout freud. In particular, most references to `rmax`, `nn`, `num_neigh`, etc have been removed in favor of the standard names `r_max` and `num_neighbors`. On the Python side, more of the `QueryArgs` logic has been standardized by introducing a `PairCompute` subclass of `Compute` that encapsulates the standard code that goes at the beginning of every `compute` or `accumulate` call. In the process, I've also defined a standard method for classes to expose a set of default query arguments.

## Motivation and Context
This PR is necessary to standardize the usage of query arguments (especially default query arguments) throughout freud. It's also consistent with #179 and related changes for improving the API.

## How Has This Been Tested?
No new tests have been added yet, they will need to be added once the code is closer to a final state. The main tests we will need to add are ensuring that default query arguments work as expected, and that invalid `QueryArgs` objects are properly rejected.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/AABBQuery.h
+++ b/cpp/locality/AABBQuery.h
@@ -110,18 +110,18 @@ protected:
         inferMode(args);
         if (args.mode == QueryArgs::ball)
         {
-            if (args.r_max == -1)
+            if (args.r_max == QueryArgs::DEFAULT_R_MAX)
                 throw std::runtime_error("You must set r_max in the query arguments when performing ball queries.");
         }
         else if (args.mode == QueryArgs::nearest)
         {
-            if (args.num_neigh == -1)
+            if (args.num_neigh == QueryArgs::DEFAULT_NUM_NEIGH)
                 throw std::runtime_error("You must set num_neigh in the query arguments when performing number of neighbor queries.");
-            if (args.scale == -1)
+            if (args.scale == QueryArgs::DEFAULT_SCALE)
             {
                 args.scale = float(1.1);
             }
-            if (args.r_max == -1)
+            if (args.r_max == QueryArgs::DEFAULT_R_MAX)
             {
                 vec3<float> L = this->getBox().getL();
                 float r_max = std::min(L.x, L.y);
--- a/cpp/locality/AABBQuery.h
+++ b/cpp/locality/AABBQuery.h
@@ -123,9 +123,11 @@ protected:
             }
             if (args.r_max == QueryArgs::DEFAULT_R_MAX)
             {
+                // By default, we use 1/10 the smallest box dimension as the guessed query distance.
                 vec3<float> L = this->getBox().getL();
                 float r_max = std::min(L.x, L.y);
-                args.r_max = this->getBox().is2D() ? float(0.1) * r_max : float(0.1) * std::min(r_max, L.z);
+                r_max = this->getBox().is2D() ? r_max : std::min(r_max, L.z);
+                args.r_max = float(0.1) * r_max;
             }
         }
     }
--- a/cpp/locality/AABBQuery.h
+++ b/cpp/locality/AABBQuery.h
@@ -59,7 +59,7 @@ public:
         }
         else if (args.mode == QueryArgs::nearest)
         {
-            return query(query_points, n_query_points, args.num_neigh, args.r_max, args.scale, args.exclude_ii);
+            return query(query_points, n_query_points, args.num_neighbors, args.r_max, args.scale, args.exclude_ii);
         }
         else
         {
@@ -115,8 +115,8 @@ protected:
         }
         else if (args.mode == QueryArgs::nearest)
         {
-            if (args.num_neigh == QueryArgs::DEFAULT_NUM_NEIGH)
-                throw std::runtime_error("You must set num_neigh in the query arguments when performing number of neighbor queries.");
+            if (args.num_neighbors == QueryArgs::DEFAULT_NUM_NEIGHBORS)
+                throw std::runtime_error("You must set num_neighbors in the query arguments when performing number of neighbor queries.");
             if (args.scale == QueryArgs::DEFAULT_SCALE)
             {
                 args.scale = float(1.1);
--- a/cpp/locality/AABBQuery.h
+++ b/cpp/locality/AABBQuery.h
@@ -55,11 +55,11 @@ public:
         this->validateQueryArgs(args);
         if (args.mode == QueryArgs::ball)
         {
-            return queryBall(query_points, n_query_points, args.rmax, args.exclude_ii);
+            return queryBall(query_points, n_query_points, args.r_max, args.exclude_ii);
         }
         else if (args.mode == QueryArgs::nearest)
         {
-            return query(query_points, n_query_points, args.nn, args.rmax, args.scale, args.exclude_ii);
+            return query(query_points, n_query_points, args.num_neigh, args.r_max, args.scale, args.exclude_ii);
         }
         else
         {
@@ -76,7 +76,7 @@ public:
                                                          unsigned int num_neighbors, bool exclude_ii = false) const
     {
         throw std::runtime_error("AABBQuery k-nearest-neighbor queries must use the function signature that "
-                                 "provides rmax and scale guesses.");
+                                 "provides r_max and scale guesses.");
     }
 
     std::shared_ptr<NeighborQueryIterator> query(const vec3<float>* query_points, unsigned int n_query_points, unsigned int num_neighbors,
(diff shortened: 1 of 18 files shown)


cpp/locality/AABBQuery.h
```diff
@@ -101,26 +101,31 @@ class AABBQuery : public NeighborQuery
...
                 args.scale = float(1.1);
             }
-            if (args.rmax == -1)
+            if (args.r_max == QueryArgs::DEFAULT_R_MAX)
             {
                 vec3<float> L = this->getBox().getL();
-                float rmax = std::min(L.x, L.y);
-                args.rmax = this->getBox().is2D() ? float(0.1) * rmax : float(0.1) * std::min(rmax, L.z);
+                float r_max = std::min(L.x, L.y);
+                args.r_max = this->getBox().is2D() ? float(0.1) * r_max : float(0.1) * std::min(r_max, L.z);
```
bdice: Make the magic constant `float(0.1)` only appear one time.
```suggestion
                if (this->getBox().is2D())
                    r_max = std::min(r_max, L.z);
                args.r_max = float(0.1) * r_max;
```

freud/common.pyx
```diff
@@ -125,6 +133,70 @@ cdef class Compute:
         return wrapper
 
 
+cdef class PairCompute(Compute):
+    R"""Parent class for all compute classes in freud that depend on finding
+    nearest neighbors.
```
bdice, line 138:
```suggestion
    neighbor pairs.
```

freud/common.pyx
```diff
@@ -125,6 +133,70 @@ cdef class Compute:
...
+    well as dealing with boxes and query arguments.
+
+    .. moduleauthor:: Vyas Ramasubramani <vramasub@umich.edu>
+    """
+    def preprocess_arguments(self, box, points, query_points=None, nlist=None,
+                              query_args=None):
+        """Process standard compute arguments into freud's internal types by
+        calling all the required internal functions.
+
+        The preprocessing of boxes and points into NeighborQuery objects, the
```
bdice: Reword: "This function handles preprocessing..." instead of "The preprocessing... are handled by this function."

bdice: Please add `:py:class:` designations for classes and write argument names like `query_points` in code font.

freud/common.pyx
```diff
@@ -125,6 +133,70 @@ cdef class Compute:
...
+        cdef freud.locality._QueryArgs qargs = \
+            freud.locality._QueryArgs.from_dict(
+                query_args if query_args else
+                self.get_default_query_args(query_points is None))
+        if query_points is None:
+            query_points = nq.points
+        query_points = freud.common.convert_array(
+            query_points, shape=(None, 3))
+        cdef const float[:, ::1] l_query_points = query_points
+        cdef unsigned int n_p = l_query_points.shape[0]
```
bdice: It's confusing to use the name `n_p` for the number of query points.

freud/density.pyx
```diff
@@ -858,24 +858,15 @@ cdef class RDF(Compute):
...
-        qargs.update(query_args)
-        points = nq.points
-
-        if query_points is None:
-            query_points = points
-        query_points = freud.common.convert_array(
-            query_points, shape=(None, 3))
-        cdef const float[:, ::1] l_query_points = query_points
-        cdef unsigned int n_p = l_query_points.shape[0]
+        cdef:
```
bdice, line 861: I'm not familiar with this syntax. Does this `cdef` a list of following (indented) lines? If so, why is `cdef` repeated in the other lines?

freud/common.pyx
```diff
@@ -125,6 +133,70 @@ cdef class Compute:
...
+            nlist (:class:`freud.locality.NeighborList`, optional):
+                NeighborList to use to find bonds (Default value =
+                :code:`None`).
+            query_args (dict): A dictionary of query arguments (Default value =
+                :code:`None`).
+        """
+        cdef freud.box.Box b = freud.common.convert_box(box)
+
+        cdef freud.locality.NeighborQuery nq = freud.locality.make_default_nq(
+            box, points)
```
bdice, line 177: This doesn't implement the choice of `NeighborQuery` or tuple syntax `(box, points)` yet, but we'll do that in this function when we get there, right?

freud/common.pyx
```diff
@@ -125,6 +133,70 @@ cdef class Compute:
...
+                self.get_default_query_args(query_points is None))
+        if query_points is None:
+            query_points = nq.points
+        query_points = freud.common.convert_array(
+            query_points, shape=(None, 3))
+        cdef const float[:, ::1] l_query_points = query_points
+        cdef unsigned int n_p = l_query_points.shape[0]
+        return (b, nq, nlistptr, qargs, l_query_points, n_p)
+
+    def get_default_query_args(self, exclude_ii):
```
bdice: Does it make sense to have this `exclude_ii` parameter, or should it be left undefined in `get_default_query_args` and then just updated by the `preprocess_arguments` method? As far as I can remember, we always choose `exclude_ii` based on whether there is a second set of points provided, so I don't think this option is needed.

freud/environment.pyx
```diff
@@ -340,18 +339,20 @@ cdef class LocalDescriptors(Compute):
...
             The number of points passed to the last call to :meth:`~.compute`.
-        num_neighbors (unsigned int):
-            The number of neighbors used by the last call to compute. Bounded
-            from above by the number of reference points multiplied by the
-            lower of the num_neighbors arguments passed to the last compute
-            call or the constructor.
+        num_sphs (unsigned int):
+            The last number of spherical harmonics computed. This is equal to
+            the number of bonds in the last computation, which is bounded from
+            above by the number of reference points multiplied by the lower of
```
bdice: Fix "reference points" to new nomenclature. Also I'd prefer to use less mathematical language here:
"which is at most the number of points times ..."
 ...
"(it may be less if there are not enough neighbors for every particle)."

freud/environment.pyx
```diff
@@ -470,21 +471,20 @@ cdef class LocalDescriptors(Compute):
...
+        return ("freud.environment.{cls}(num_neighbors={num_neighbors}, "
                 "lmax={lmax}, r_max={r_max}, "
-                "negative_m={negative_m})").format(cls=type(self).__name__,
-                                                   num_neigh=self.num_neigh,
-                                                   lmax=self.lmax,
-                                                   r_max=self.r_max,
-                                                   negative_m=self.negative_m)
+                "negative_m={negative_m})").format(
+                    cls=type(self).__name__, num_neighbors=self.num_neighbors,
+                    lmax=self.lmax, r_max=self.r_max,
```
bdice: Does this even work? (Make sure we have a `repr` test for this.) The property above is `l_max`. We need to be consistent with `r_max` name conventions.

freud/locality.pxd
```diff
@@ -30,9 +30,9 @@ cdef class NeighborQueryResult:
     @staticmethod
     cdef inline NeighborQueryResult init(
             NeighborQuery nq, const float[:, ::1] points,
-            cbool exclude_ii, float r=0, unsigned int k=0):
+            cbool exclude_ii, float r_max=0, unsigned int num_neighbors=0):
         # Internal API only
-        assert r != 0 or k != 0
+        assert r_max != 0 or num_neighbors != 0
```
bdice:
```suggestion
        assert r_max > 0 or num_neighbors > 0
```

freud/locality.pxd
```diff
@@ -59,9 +59,11 @@ cdef class AABBQueryResult(NeighborQueryResult):
     @staticmethod
     cdef inline AABBQueryResult init_aabb_nn(
             AABBQuery aabbq, const float[:, ::1] points,
-            cbool exclude_ii, unsigned int k, float r, float scale):
+            cbool exclude_ii,
+            unsigned int num_neighbors, float r_max,
+            float scale):
         # Internal API only
-        assert k != 0
+        assert num_neighbors != 0
```
bdice:
```suggestion
        assert num_neighbors > 0
```

freud/order.pyx
```diff
@@ -428,13 +428,13 @@ cdef class TransOrderParameter(Compute):
...
     cdef freud._order.TransOrderParameter * thisptr
-    cdef num_neigh
+    cdef num_neighbors
     cdef r_max
 
     def __cinit__(self, r_max, k=6.0, num_neighbors=0):
         self.thisptr = new freud._order.TransOrderParameter(k)
         self.r_max = r_max
-        self.num_neigh = (num_neighbors if num_neighbors else int(k))
+        self.num_neighbors = (num_neighbors if num_neighbors else int(k))
```
bdice: Be explicit:
```suggestion
        self.num_neighbors = (num_neighbors if num_neighbors > 0 else int(k))
```

tests/test_density_ComplexCF.py
```diff
@@ -212,34 +212,34 @@ def test_query_nn(self):
 
         cf = freud.density.ComplexCF(rmax, dr)
         cf.compute(box, ref_points, ref_values, points, values,
-                   query_args={'mode': 'nearest', 'nn': 1})
+                   query_args={'mode': 'nearest', 'num_neighbors': 1})
```
bdice, line 215: Some tests should use mode inference, by removing `'mode': 'nearest'`.

bdice (changes requested):
Very helpful! @vyasr you said this PR would be hard to review, but I think it went smoothly. 😄
Some suggestions for edits below:

tests/test_density_ComplexCF.py
bdice: Maybe that should be tested in the locality module tests instead of here.

freud/common.pyx
vyasr: I don't know if neighbor pairs is quite accurate, it's really about finding all the nearest neighbors (by whatever specification) of each particle. It's not just a 1-1 mapping.

freud/density.pyx
vyasr: Yes it does. And because I was testing you. Obviously.

freud/common.pyx
vyasr: Correct, that will come in a future PR.

freud/common.pyx
vyasr: I was debating about this. I originally had this as a property, then made it a function that depended on `exclude_ii` because I didn't want developers to have to update it everywhere in the compute method. Now that I've moved this logic into the common module, I can change it back to a property and just update the `exclude_ii` logic afterwards.

tests/test_density_ComplexCF.py
vyasr: Yes, definitely in locality and not here.

vyasr:
I will draft relevant documentation and add to #348 soon.

bdice (approved):
Great work! I approve this.

PR #403, opened 2019-08-23
Add plot to __init__.

bdice:
Added plot module to the package import.


vyasr:
How do you anticipate this being used outside the context of the class plotting methods?

PR #408, opened 2019-08-25
Change clustering algorithm to use connected components

bdice:
## Description
In previous versions of freud, the cluster algorithm has looped over bonds in the NeighborList/NeighborQuery and used a cutoff radius to filter those bonds. That is, two particles are in the same cluster if and only if there is a path between them in the network of bonds, where every bond distance along that walk is less than `r_cut`. This is annoying for two reasons:

1. If this behavior is desired, it's easy to just filter the network of bonds using a cutoff distance before passing it into `Cluster.compute`.
2. It is not straightforward to use a network of bonds formed by some criterion other than distance (unless you fake it by producing a NeighborList and using an arbitrarily high value of `r_cut`).

## Motivation and Context
By rewriting this, it becomes much easier to use this module for the clustering performed by the Solid-Liquid order parameter.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/cluster/Cluster.cc
+++ b/cpp/cluster/Cluster.cc
@@ -44,7 +44,7 @@ void Cluster::compute(const freud::locality::NeighborQuery* nq,
     // particle index.
     vector<size_t> cluster_label(m_num_particles, -1);
     vector<size_t> cluster_label_count(m_num_particles);
-    vector<size_t> cluster_min_id(m_num_particles);
+    vector<size_t> cluster_min_id(m_num_particles, -1);
 
     // Loop over every particle.
     m_num_clusters = 0;
@@ -69,7 +69,9 @@ void Cluster::compute(const freud::locality::NeighborQuery* nq,
 
     // Resize label counts and min ids to the number of unique clusters found.
     cluster_label_count.resize(m_num_clusters);
+    cluster_label_count.shrink_to_fit();
     cluster_min_id.resize(m_num_clusters);
+    cluster_min_id.shrink_to_fit();
 
     // Get a permutation that reorders clusters, largest to smallest.
     vector<size_t> cluster_reindex = sort_indexes_inverse(cluster_label_count, cluster_min_id);
--- a/cpp/cluster/Cluster.cc
+++ b/cpp/cluster/Cluster.cc
@@ -3,13 +3,12 @@
 
 #include <algorithm>
 #include <cassert>
-#include <map>
 #include <numeric>
 #include <stdexcept>
 
 #include "Cluster.h"
-#include "NeighborComputeFunctional.h"
 #include "NeighborBond.h"
+#include "NeighborComputeFunctional.h"
 #include "dset/dset.h"
 
 using namespace std;
@@ -17,7 +16,7 @@ using namespace std;
 //! Finds clusters using a network of neighbors.
 namespace freud { namespace cluster {
 
-Cluster::Cluster() : m_num_particles(0), m_num_clusters(0) {}
+Cluster::Cluster() {}
 
 void Cluster::compute(const freud::locality::NeighborQuery* nq,
                       const freud::locality::NeighborList* nlist,
@@ -36,7 +35,7 @@ void Cluster::compute(const freud::locality::NeighborQuery* nq,
     freud::locality::loopOverNeighbors(
         nq, points, Np, qargs, nlist,
         [&dj](const freud::locality::NeighborBond& neighbor_bond) {
-            // merge the two sets using the disjoint set
+            // Merge the two sets using the disjoint set
             if (!dj.same(neighbor_bond.ref_id, neighbor_bond.id))
             {
                 dj.unite(neighbor_bond.ref_id, neighbor_bond.id);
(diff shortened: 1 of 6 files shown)


bdice:
## To-do:
- [x] Update documentation
- [x] Improve description of behavior with `keys`
- ...

cpp/cluster/Cluster.cc
```diff
@@ -45,65 +45,108 @@ void Cluster::computeClusters(const freud::locality::NeighborQuery* nq,
...
             cur_set++;
         }
 
-        // label this point in cluster_idx
-        m_cluster_idx.get()[i] = label_map[s];
+        // Increment the counter for this cluster label
+        label_counts[label_map[s]]++;
+
+        // Track the smallest particle index in this cluster
+        label_min_id[label_map[s]] = std::min(label_min_id[label_map[s]], i);
```
vyasr: Since you're iterating over the particles in order, won't the smallest index always be the first one you find? So in the conditional you can just set `label_min_id[cur_set] = i` and be done?

cpp/cluster/Cluster.cc
```diff
@@ -45,65 +45,108 @@ void Cluster::computeClusters(const freud::locality::NeighborQuery* nq,
...
     }
 
-    // cur_set is now the number of clusters
+    // cur_set is now the total number of clusters
     m_num_clusters = cur_set;
```
vyasr: Get rid of `cur_set` and just use `m_num_clusters`.

cpp/cluster/Cluster.cc
```diff
@@ -45,65 +45,108 @@ void Cluster::computeClusters(const freud::locality::NeighborQuery* nq,
...
-    uint32_t cur_set = 0;
-    for (uint32_t i = 0; i < m_num_particles; i++)
+    // Done looping over points. All clusters are now determined.
+    // Next, we renumber clusters from zero to num_clusters-1.
+    // These new cluster indices are then sorted by cluster size from largest
+    // to smallest, with equally-sized clusters sorted based on their minimum
+    // particle index.
+    map<size_t, size_t> label_map;
+    map<size_t, size_t> label_counts;
+    map<size_t, size_t> label_min_id;
```
vyasr: Make `label_min_id` a vector to start with, then you don't copy it into a vector again later. Requires that you allocate it to be the size of the number of particles to start with (and probably resize to m_num_clusters later), but in general freud assumes we're compute and not memory bound anyway. You can call `shrink_to_fit` afterwards if you're concerned about memory usage.

vyasr: I think you can do the same thing with `label_counts` as well.

cpp/cluster/Cluster.cc
```diff
@@ -45,65 +45,108 @@ void Cluster::computeClusters(const freud::locality::NeighborQuery* nq,
...
+    // Clear the cluster keys
     m_cluster_keys.resize(m_num_clusters);
-
-    for (unsigned int i = 0; i < m_num_clusters; i++)
-        m_cluster_keys[i].clear();
-
-    // add members to the sets
-    for (unsigned int i = 0; i < m_num_particles; i++)
+    for (auto v : m_cluster_keys)
+        v.clear();
```
vyasr: Honestly this might be simpler if you just recreate the cluster keys every time (or at least if the number of clusters changes).

freud/cluster.pyx
```diff
@@ -103,52 +105,43 @@ cdef class Cluster(PairCompute):
             self.preprocess_arguments(box, points, nlist=nlist,
                                       query_args=query_args)
 
+        cdef unsigned int* l_keys_ptr = NULL
+        cdef unsigned int[::1] l_keys
+        if keys is not None:
+            l_keys = freud.common.convert_array(
+                keys, shape=(num_query_points, ), dtype=np.uint32)
+            l_keys_ptr = &l_keys[0]
```
vyasr, line 112: I propose treating `keys=None` as "don't bother to compute keys". We could add some special key like " 'clusters' " or "{auto}" that indicates the behavior you've implemented. That would allow us to avoid doing the extra work for building the keys if the user won't ever use them.

vyasr: What's currently being done in C++ in the default case could be accomplished by just creating an array in Python with `np.arange(points.shape[0])` and passing that through.

vyasr (changes requested):

freud/cluster.pyx
bdice: Discussed further in person - assuming keys are the particle IDs seems like a useful default behavior, based on what current users get from this class.

cpp/cluster/Cluster.cc
```diff
@@ -1,34 +1,28 @@
...
-        throw invalid_argument("Cluster requires that r_max must be non-negative.");
-}
+Cluster::Cluster() {}
 
 void Cluster::compute(const freud::locality::NeighborQuery* nq,
                       const freud::locality::NeighborList* nlist,
-                      const vec3<float> *points,
-                      unsigned int Np,
-                      freud::locality::QueryArgs qargs)
+                      const vec3<float>* points, unsigned int Np,
```
vyasr: No need to pass in these extra points arguments, just pull out of the `nq` object.

cpp/cluster/Cluster.cc
```diff
@@ -40,67 +34,98 @@ void Cluster::compute(const freud::locality::NeighborQuery* nq,
...
-    uint32_t cur_set = 0;
-    for (uint32_t i = 0; i < m_num_particles; i++)
+    // Done looping over points. All clusters are now determined.
+    // Next, we renumber clusters from zero to num_clusters-1.
+    // These new cluster indexes are then sorted by cluster size from largest
+    // to smallest, with equally-sized clusters sorted based on their minimum
+    // particle index.
+    vector<size_t> cluster_label(m_num_particles, -1);
+    vector<size_t> cluster_label_count(m_num_particles);
+    vector<size_t> cluster_min_id(m_num_particles);
```
vyasr: For consistency with `cluster_label`, do you also want to initialize this to `-1`? It should never be necessary, but would let you sanity check things if needed (since 0 is a valid id).

cpp/cluster/Cluster.cc
```diff
@@ -40,67 +34,98 @@ void Cluster::compute(const freud::locality::NeighborQuery* nq,
...
-        m_cluster_idx[i] = label_map[s];
+        // Increment the counter for this cluster label.
+        cluster_label_count[cluster_label[s]]++;
     }
 
-    // cur_set is now the number of clusters
-    m_num_clusters = cur_set;
+    // Resize label counts and min ids to the number of unique clusters found.
+    cluster_label_count.resize(m_num_clusters);
+    cluster_min_id.resize(m_num_clusters);
```
vyasr, line 73: Doesn't really matter, but as I mentioned when I first requested this rewrite you could call [shrink_to_fit](https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit) if you're still concerned about memory overhead. Up to you.

cpp/cluster/Cluster.h
```diff
@@ -20,43 +20,19 @@
...
-    on top of the \c cluster_idx array. Given a key value per particle (i.e. the
-    polymer id), the computeClusterMembership function will process cluster_idx
-    with the key values in mind and provide a list of keys that are present in
-    each cluster.
-
-    <b>2D:</b><br>
-    Cluster properly handles 2D boxes. As with everything else in freud, 2D
-    points must be passed in as 3 component vectors x,y,0. Failing to set 0 in
-    the third component will lead to undefined behavior.
-*/
```
vyasr, line 44: Please put the docstring back, updated as needed.

cpp/cluster/Cluster.h
```diff
@@ -20,43 +20,19 @@
...
     //! Compute the point clusters
     void compute(const freud::locality::NeighborQuery* nq,
                  const freud::locality::NeighborList* nlist,
-                 const vec3<float> *points,
-                 unsigned int Np,
-                 freud::locality::QueryArgs qargs);
-
-    //! Compute clusters with key membership
-    void computeClusterMembership(const unsigned int* keys);
+                 const vec3<float>* points, unsigned int Np,
```
vyasr: Update signature, same as in cc file.

cpp/cluster/Cluster.h
```diff
@@ -70,24 +46,29 @@ class Cluster
         return m_num_particles;
     }
 
-    //! Get a reference to the last computed cluster_idx
+    //! Get a reference to the last computed cluster ids
```
vyasr: Add a period.

cpp/cluster/Cluster.h
```diff
@@ -70,24 +46,29 @@ class Cluster
...
     const util::ManagedArray<unsigned int> &getClusterIdx()
     {
         return m_cluster_idx;
     }
 
-    //! Returns the cluster keys last determined by computeClusterKeys
-    const std::vector<std::vector<unsigned int>>& getClusterKeys()
+    //! Returns the last computed cluster keys
```
vyasr: Add a period.

freud/_cluster.pxd
```diff
@@ -12,13 +12,13 @@ cimport freud.util
 
 cdef extern from "Cluster.h" namespace "freud::cluster":
     cdef cppclass Cluster:
-        Cluster(float) except +
+        Cluster() except +
         void compute(const freud._locality.NeighborQuery*,
                      const freud._locality.NeighborList*,
-                     const vec3[float] *,
+                     const vec3[float]*,
                      unsigned int,
```
vyasr: Update signature.

vyasr (changes requested):

vyasr (approved):

PR #413, opened 2019-08-27
Histogram Class

### LINKED ISSUE #385: RDF should use "bins" argument
The RDF and CorrelationFunction codes should allow a bin definition similar to `np.histogram`:

`bins : int or sequence of scalars or str, optional`

An argument name `bins` could be defined as "integer number of bins" for now and generalized to "or an array of bin edges" later.

_Originally suggested by @csadorf in https://github.com/glotzerlab/freud/issues/179#issuecomment-514261487_

bdice:
## Description
I implemented a Histogram class that supports arbitrary dimensionality and easy binning of values into the histogram. As a test-case, I've implemented its usage in the RDF class. I've also made other simplifications to the RDF class to further simplify it and improve performance.

One breaking change that was made along with this PR is that `freud.density.RDF` now accepts a number of bins as an argument rather than a bin width. One major fix that results from this is that we always completely capture the range from `r_min` to `r_max`.

## Motivation and Context
Many methods in freud amount to computing a histogram, typically of bond distances. A preliminary attempt was made to refactor the common elements of these classes using the NdHistogram, but that class is still quite specialized to the RDF/PMFT use-cases and does not abstract away the core logic we want to remove, which is the binning of the data into a histogram. The new histogram class is a simple data structure that supports this goal.

Resolves #385.

## How Has This Been Tested?
Existing tests of the RDF pass. I had to adapt them because the argument order changed and the bin spacing is now also subtly different because we no longer neglect a final partial bin.

## Types of changes

## Checklist:


vyasr:
Closing because this PR cannot be reviewed by its creator

PR #417, opened 2019-08-28
Use ManagedArray for NeighborList

bdice:
## Description
This PR replaces the NeighborList memory management with ManagedArray. I also changed all of its properties to match the new nomenclature in v2.0, which makes `index_i` and `index_j` unambiguous.

## Motivation and Context
To update the compute classes that compute and utilize per-bond quantities, it is helpful to first refactor the NeighborList class.

## How Has This Been Tested?
Existing tests pass after slight alterations. For now, weights can't be altered.

## Types of changes

## Checklist:


cpp/locality/NeighborList.cc
```diff
@@ -6,185 +6,201 @@
...
 }
 
-size_t NeighborList::getNumJ() const
+NeighborList::NeighborList(unsigned int num_bonds)
+    : m_num_query_points(0), m_num_points(0)
 {
-    return m_num_j;
+    m_neighbors.prepare({num_bonds, 2});
+    m_weights.prepare(num_bonds);
+    m_distances.prepare(num_bonds);
```
vyasr: Just call the constructor as part of the initializer list.

cpp/locality/NeighborList.cc
```diff
@@ -6,185 +6,201 @@
...
-size_t* NeighborList::getNeighbors()
+NeighborList::NeighborList(unsigned int num_bonds, const unsigned int* query_point_index,
+                 unsigned int num_query_points, const unsigned int* point_index,
+                 unsigned int num_points, const float* distances, const float* weights)
+    : m_num_query_points(num_query_points), m_num_points(num_points)
 {
-    return m_neighbors.get();
+    m_neighbors.prepare({num_bonds, 2});
+    m_weights.prepare(num_bonds);
+    m_distances.prepare(num_bonds);
```
vyasr: Same as above, use initializer list.

cpp/locality/NeighborList.h
```diff
@@ -9,94 +9,147 @@
...
-    void resize(size_t max_bonds, bool force = false);
+    const util::ManagedArray<float> &getWeights() const
+    {
+        return m_weights;
+    }
+    //! Access the counts array for reading
+    const util::ManagedArray<unsigned int> &getCounts() const
+    {
+        updateSegmentCounts();
+        return m_segments;
```
vyasr: Should be m_counts.

cpp/locality/NeighborList.cc
```diff
@@ -6,185 +6,201 @@
...
-void NeighborList::setNumBonds(size_t num_bonds, size_t num_i, size_t num_j)
+NeighborList::NeighborList(const NeighborList& other)
+    : m_num_query_points(other.getNumQueryPoints()), m_num_points(other.getNumPoints())
 {
-    m_num_bonds = num_bonds;
-    m_num_i = num_i;
-    m_num_j = num_j;
+    m_neighbors = other.getNeighbors();
+    m_weights = other.getWeights();
+    m_distances = other.getDistances();
```
vyasr: You don't want to do this here, access the members of `other` directly or you'll have problems if `other` gets deleted.

cpp/locality/NeighborList.cc
```diff
@@ -6,185 +6,201 @@
...
+    unsigned int index(0);
+    for (unsigned int i = 0; i < num_bonds; i++)
+    {
+        index = query_point_index[i];
+        if (index < last_index)
+            throw std::runtime_error("NeighborList query_point_index must be sorted.");
+        if (index >= m_num_query_points)
+            throw std::runtime_error("NeighborList query_point_index values must be less than num_query_points.");
+        if (point_index[i] >= m_num_points)
+            throw std::runtime_error("NeighborList point_index values must be less than num_points.");
```
vyasr, line 41: Probably should also track the `last_query_point_index` to check sort order of `query_points` for a given `point`, if we're error checking we should be thorough.

tests/test_locality_NeighborList.py
```diff
@@ -26,52 +27,108 @@ def test_writable(self):
...
-            self.cl.nlist.index_j[:] = 0
+            self.cl.nlist.point_index[:] = 0
 
         # the weights array may be useful to write to, though
-        self.cl.nlist.weights[18] = 3
-        self.assertEqual(self.cl.nlist.weights[18], 3)
+        # TODO: weights aren't writable since changing to ManagedArray
+        with self.assertRaises(ValueError):
+            self.cl.nlist.weights[18] = 3
+        # self.assertEqual(self.cl.nlist.weights[18], 3)
```
vyasr: I assume this should be deleted.

cpp/locality/NeighborList.cc
```diff
@@ -6,185 +6,201 @@
...
 {
     // number of good (unfiltered-out) elements so far
-    size_t num_good(0);
-    size_t* neighbors(m_neighbors.get());
-    float* weights(m_weights.get());
+    unsigned int num_good(0);
+    const unsigned int old_size(getNumBonds());
 
-    for (size_t i(0); i < m_num_bonds; ++i)
+    for (unsigned int i(0); i < getNumBonds(); ++i)
```
vyasr: Use `old_size` here.

cpp/locality/NeighborList.cc
```diff
@@ -6,185 +6,201 @@
...
+unsigned int NeighborList::filter_r(float r_max, float r_min)
 {
-    // number of good (unfiltered-out) elements so far
-    size_t num_good(0);
-    size_t* neighbors(m_neighbors.get());
-    float* weights(m_weights.get());
-    float* distances(m_distances.get());
-
-    for (size_t bond(0); bond < m_num_bonds; ++bond)
+    // Can't use vector<bool> because that is specialized for compact storage
```
vyasr, line 132: The problem is really that we need to pass a contiguous raw array to `filter` here, right? If `filter` accepted an `std::vector` then it would be fine, but there's no meaningful way to access `front` or `data` in this case?

cpp/locality/NeighborList.cc
```diff
@@ -6,185 +6,201 @@
...
-    std::copy(src_neigh, src_neigh + 2 * other.m_num_bonds, dst_neigh);
-    std::copy(src_weights, src_weights + other.m_num_bonds, dst_weights);
-    std::copy(src_distances, src_distances + other.m_num_bonds, dst_distances);
-    m_num_bonds = other.m_num_bonds;
-    m_num_i = other.m_num_i;
-    m_num_j = other.m_num_j;
+    setNumBonds(other.getNumBonds(), other.getNumQueryPoints(), other.getNumPoints());
+    m_neighbors = other.getNeighbors();
+    m_weights = other.getWeights();
+    m_distances = other.getDistances();
```
vyasr: Same change as for the constructor from other, access the members directly.

freud/_cluster.pxd
```diff
@@ -1,11 +1,8 @@
 # Copyright (c) 2010-2019 The Regents of the University of Michigan
 # This file is from the freud project, released under the BSD 3-Clause License.
 
-from libcpp cimport bool
 from freud.util cimport vec3, uint
-from libcpp.memory cimport shared_ptr
 from libcpp.vector cimport vector
```
vyasr, line 5: Remove the extra line here.

freud/_locality.pxd
```diff
@@ -60,30 +61,29 @@ cdef extern from "RawPoints.h" namespace "freud::locality":
...
+        freud.util.ManagedArray[float] &getWeights()
+        freud.util.ManagedArray[float] &getSegments()
+        freud.util.ManagedArray[float] &getCounts()
+
+        unsigned int getNumBonds() const
+        unsigned int getNumPoints() const
+        unsigned int getNumQueryPoints() const
+        void setNumBonds(unsigned int, unsigned int, unsigned int)
+        unsigned int filter(const bool*) except +
+        unsigned int filter_r(float, float) except +
```
vyasr, line 81: Can either of these actually throw exceptions.

freud/environment.pyx
```diff
@@ -888,9 +888,6 @@ cdef class AngularSeparation(Compute):
             The global angles in radians. **This field is only populated
             after** :meth:`~.computeGlobal` **is called.** The angles
             are stored in the order of the neighborlist object.
-
-    .. todo Need to figure out what happens if you use a neighborlist with
-            strict_cut=True
```
vyasr, line 889: What happens?

freud/locality.pyx
```diff
@@ -303,137 +303,118 @@ cdef class NeighborQuery:
...
-
-    For efficiency, all bonds for a particular reference particle :math:`i`
-    are contiguous and bonds are stored in order based on reference
-    particle index :math:`i`. The first bond index corresponding to a given
-    particle can be found in :math:`\log(n_{bonds})` time using
+    R"""Class representing bonds between two sets of points.
+
+    Compute classes contain a set of bonds between two sets of position
+    arrays ("query points" and "points") and hold a set of index pairs
+    :math:`\left(i, j\right): i < N_{querypoints}, j < N_{points}`
```
vyasr: Add curly braces for proper set notation.

freud/locality.pyx
```diff
@@ -303,137 +303,118 @@ cdef class NeighborQuery:
...
-            A segment array, which is an array of length :math:`N_{ref}`
-            indicating the first bond index for each reference particle from
-            the last set of points this object was evaluated with.
-        neighbor_counts ((:math:`N_{points}`) :class:`np.ndarray`):
+            The weights for each bond. By default, bonds have a weight of 1.
+        distances ((:math:`N_{bonds}`) :class:`np.ndarray`):
+            The distances for each bond.
+        segments ((:math:`N_{querypoints}`) :class:`np.ndarray`):
+            A segment array, which is an array of length
+            :math:`N_{querypoints}` indicating the first bond index for each
```
vyasr: No need to put length in the text if it's also in the type.

freud/locality.pyx
```diff
@@ -303,137 +303,118 @@ cdef class NeighborQuery:
...
+            query point.
+        neighbor_counts ((:math:`N_{querypoints}`) :class:`np.ndarray`):
             A neighbor count array, which is an array of length
-            :math:`N_{ref}` indicating the number of neighbors for each
-            reference particle from the last set of points this object was
-            evaluated with.
+            :math:`N_{querypoints}` indicating the number of neighbors for each
```
vyasr: Same, remove this.

freud/locality.pyx
```diff
@@ -303,137 +303,118 @@ cdef class NeighborQuery:
...
+            query point.
 
     Example::
 
        # Assume we have position as Nx3 array
-       lc = LinkCell(box, 1.5).compute(box, positions)
-       nlist = lc.nlist
+       aq = freud.locality.AABBQuery(box, positions)
+       nlist = aq.query(r_max=3).toNList()
```
vyasr: Need to query with positions.

vyasr: Also this is now `toNeighborList`.

freud/locality.pyx
```diff
@@ -303,137 +303,118 @@ cdef class NeighborQuery:
...
                 value of 1 for each weight) (Default value = :code:`None`).
         """
-        index_i = freud.common.convert_array(
-            index_i, shape=(None,), dtype=np.uint64)
-        index_j = freud.common.convert_array(
-            index_j, shape=index_i.shape, dtype=np.uint64)
+        query_point_index = freud.common.convert_array(
+            query_point_index, shape=(None,), dtype=np.uint32)
+        point_index = freud.common.convert_array(
+            point_index, shape=query_point_index.shape, dtype=np.uint32)
```
vyasr: Is there a reason to change these types?

freud/util.pxd
```diff
@@ -19,19 +19,19 @@ ctypedef float complex fcomplex
 ctypedef double complex dcomplex
 
 ctypedef enum arr_type_t:
-    UNSIGNED_INT
     FLOAT
     DOUBLE
     COMPLEX_FLOAT
     COMPLEX_DOUBLE
+    UNSIGNED_INT
```
vyasr, line 26: Why did you reorder this? If you wanted to be alphabetical you need to change all the others too, it was virtually in reverse alphabetical order before.

freud/locality.pyx
```diff
@@ -1104,7 +1016,8 @@ cdef class NearestNeighbors:
         result[:] = self.UINTMAX
         cdef unsigned int start_idx = self.nlist.find_first_index(i)
         cdef unsigned int end_idx = self.nlist.find_first_index(i + 1)
-        result[:end_idx - start_idx] = self.nlist.index_j[start_idx:end_idx]
+        result[:end_idx - start_idx] = \
```
vyasr: Can you add parenthesis around the index here?

cpp/locality/NeighborList.cc
```diff
@@ -6,185 +6,201 @@
...
-        const bool good(distances[bond] > r_min && distances[bond] < r_max);
-        if (good)
-        {
-            neighbors[2 * num_good] = neighbors[2 * bond];
-            neighbors[2 * num_good + 1] = neighbors[2 * bond + 1];
-            weights[num_good] = weights[bond];
-            distances[num_good] = distances[bond];
-            ++num_good;
-        }
+        dist_filter[i] = (m_distances[i] >= r_min && m_distances[i] < r_max);
```
vyasr, line 136: This is a change from existing behavior. It's very minor, but while we're looking at this is there any reason to prefer one convention over another?

tests/test_locality_NeighborList.py
```diff
@@ -26,52 +27,108 @@ def test_writable(self):
...
-        self.cl.nlist.filter_r(self.fbox, self.points, points2, 2.5)
+        old_size = len(self.cl.nlist)
+
+        # Make sure weights are 1
+        npt.assert_equal(self.cl.nlist.weights, 1)
+
+        # Compute neighbor pairs that we expect to be kept after the filter
+        kept_neighbors = self.cl.nlist[np.logical_and(
+            self.cl.nlist.distances < filter_max_distance,
+            self.cl.nlist.distances >= filter_min_distance)]
```
vyasr: This may need to change based on whether [this](https://github.com/glotzerlab/freud/pull/417/files#diff-24ba3759a6350d78b88b62042f6ed536R143) changes.

vyasr (changes requested):

cpp/locality/NeighborList.cc
bdice: We don't enforce ordering along the second index. I always thought we did, but that isn't true.

freud/_locality.pxd
bdice: I think so -- I added this for a reason, I think it was to catch an error thrown by ManagedArray.

freud/environment.pyx
bdice: We're not using `strict_cut` anymore (this is implemented by the `r_max` cutoff) so this comment is just out of date.

freud/util.pxd
bdice: This is just the result of a conflict resolution -- I didn't do this on purpose. I don't think the order matters as long as we're consistent, which is currently true.

cpp/locality/NeighborList.cc
bdice: The default `r_min` is zero, and if the user doesn't provide a minimum, then I expect to retain bonds with distance 0.

freud/locality.pyx
bdice: Everything uses `unsigned int` instead of `size_t` internally now.

vyasr: I meant from uint64 to uint32, or just uint.

cpp/locality/NeighborList.cc
vyasr: Makes sense.

cpp/locality/NeighborList.cc
vyasr: You didn't actually do this :)

freud/locality.pyx
```diff
@@ -303,137 +303,117 @@ cdef class NeighborQuery:
...
-            the last set of points this object was evaluated with.
-        neighbor_counts ((:math:`N_{points}`) :class:`np.ndarray`):
-            A neighbor count array, which is an array of length
-            :math:`N_{ref}` indicating the number of neighbors for each
-            reference particle from the last set of points this object was
-            evaluated with.
+            The weights for each bond. By default, bonds have a weight of 1.
+        distances ((:math:`N_{bonds}`) :class:`np.ndarray`):
+            The distances for each bond.
+        segments ((:math:`N_{querypoints}`) :class:`np.ndarray`):
```
vyasr: `query_points`

freud/locality.pyx
```diff
@@ -303,137 +303,117 @@ cdef class NeighborQuery:
...
+            A segment array indicating the first bond index for each query
+            point.
+        neighbor_counts ((:math:`N_{querypoints}`) :class:`np.ndarray`):
```
vyasr: `query_points`

freud/locality.pyx
```diff
@@ -482,94 +462,48 @@ cdef class NeighborList:
...
-            result = np.asarray(neighbors[:, :], dtype=np.uint64)
-        result.flags.writeable = False
-        return result[key]
+        return freud.util.make_managed_numpy_array(
+            &self.thisptr.getNeighbors(),
+            freud.util.arr_type_t.UNSIGNED_INT)[key]
 
     @property
-    def index_i(self):
+    def query_point_index(self):
```
vyasr: Should we also make this plural for correctness? `query_point_indices` seems better (and `point_indices` below).

vyasr: Answering my own question: yes, we should. The other attributes are all plural too.

vyasr (changes requested):
Few more small changes, then should be good to go.

freud/locality.pyx
bdice: `uint32` is the appropriate type for `unsigned int`.

vyasr (approved):

Issue #410, opened 2019-08-26
Unify neighbor querying methods for v2.0


bdice:
To make the user experience more consistent between compute classes and NeighborQuery objects, I propose to alter the `query` method behavior (which current performs nearest-neighbor queries) to accept query arguments, e.g. `{'num_neighbors': 12}` and remove the `queryBall` method.

This makes the user experience clearer, since users would use query arguments in locality classes as well as all compute classes.

bdice:
Move towards: `nq.query(**kwargs)` where `**kwargs` is a dictionary of query arguments like `r_max = 3.0`.

Also we should make `nlist` and `qargs` into one argument called `neighbors`, to be resolved at the Cython level.

vyasr:
Resolved in #419.

PR #419, opened 2019-08-30
Refactor/single query iterator

### LINKED ISSUE #412: Simplify neighbor finding logic in C++
@vyasr and @bdice need to review the C++ API to determine if parts can be simplified or made private, to prevent accidentally using the wrong methods. Also names should be clarified.

vyasr:
## Description
I created new iterators for neighbors that operate on a per-point basis. I moved all of the logic into these iterators, and just left one at the top level (NeighborQuery) that still does a full iteration. I removed the now redundant iterators in the NeighborComputeFunctional. I removed the query and queryBall functions, and created a single-function API based on querying with QueryArgs.

## Motivation and Context
The per-point iterators allowed me to move all of the logic for exclude_ii and num_neighbors into the iterator in all cases, rather than having add hacks in various other iterators because of index mismatches. Wrapping these per-point iterators in the existing NeighborQueryIterator allowed me to maintain most of the existing API with respect to querying for multiple points. The code duplication and convoluted code paths in the NeighborComputeFunctional have also been removed, and everything is more straightforward now.

Resolves #412 and #410.

## Types of changes

## Checklist:


cpp/locality/NeighborComputeFunctional.h
```diff
@@ -8,192 +8,70 @@
...
-        }
-        else
-        {
-            m_nqiter = nq->queryWithArgs(points, N, m_qargs);
-        }
+        NeighborBond nb = NeighborBond(m_nlist->getNeighbors()[2 * m_current_index],
+                                        m_nlist->getNeighbors()[2 * m_current_index + 1], 
+                                        m_nlist->getDistances()[m_current_index],
+                                        m_nlist->getWeights()[m_current_index]);
+        ++m_current_index;
```
bdice, line 51: Check if next bond is valid and set `m_finished` instead of returning an empty bond on the next call to `next()`?

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -207,53 +85,104 @@ template<typename Body> void forLoopWrapper(size_t begin, size_t end, const Body
...
+                cf(i, niter);
+            }
+        }, parallel);
+    }
+    else
+    {
+        // if nlist does not exist, check if neighbor_query is an actual NeighborQuery
+        std::shared_ptr<NeighborQueryIterator> iter;
+        std::shared_ptr<AABBQuery> abq;
+        // check if neighbor_query is a pointer to a RawPoints object
```
bdice: Too verbose -- just indicate that dynamic_cast returns a null pointer.

freud/locality.pyx
```diff
@@ -42,14 +42,22 @@ except ImportError:
 # _always_ do that, or you will have segfaults
 np.import_array()
 
-cdef class _QueryArgs:
-    # This class is temporarily included for testing and may be
-    # removed in future releases.
+cdef class QueryArgs:
+    R"""POD class to contain query arguments.
```
bdice: Take out POD. Container for query arguments.

freud/locality.pyx
```diff
@@ -289,89 +260,32 @@ cdef class NeighborQuery:
...
             query_points ((:math:`N`, 3) :class:`numpy.ndarray`):
                 Points to query for.
-            num_neighbors (int, optional):
-                The number of nearest neighbors to find.
-                (Default value = :code:`1`)
-            exclude_ii (bool, optional):
-                Set this to :code:`True` if pairs of points with identical
-                indices to those in :code:`self.points` should be excluded.
-                (Default value = :code:`False`).
+            **query_args:
```
bdice: make a dict

cpp/locality/AABBQuery.h
```diff
@@ -160,19 +114,17 @@ class AABBIterator : virtual public NeighborQueryIterator
...
 
-//! Iterator that gets nearest neighbors from AABB tree structures
-class AABBQueryIterator : virtual public NeighborQueryQueryIterator, virtual public AABBIterator
+//! Iterator that gets a specified number of nearest neighbors from AABB tree structures.
+class AABBQueryIterator : public AABBIterator
 {
 public:
-    // Explicitly indicate which toNeighborList function is used.
-    using NeighborQueryQueryIterator::toNeighborList;
+    //// Explicitly indicate which toNeighborList function is used.
```
bdice: // // //

bdice: delete

cpp/locality/AABBQuery.h
```diff
@@ -160,19 +114,17 @@ class AABBIterator : virtual public NeighborQueryIterator
...
 
     //! Constructor
-    AABBQueryIterator(const AABBQuery* neighbor_query, const vec3<float>* points, unsigned int N,
-                      unsigned int k, float r, float scale, bool exclude_ii)
-        : NeighborQueryIterator(neighbor_query, points, N, exclude_ii),
-          NeighborQueryQueryIterator(neighbor_query, points, N, exclude_ii, k),
-          AABBIterator(neighbor_query, points, N, exclude_ii), m_search_extended(false), m_r(r), m_r_cur(r),
+    AABBQueryIterator(const AABBQuery* neighbor_query, const vec3<float> query_point, unsigned int query_point_idx,
+                      unsigned int num_neighbors, float r, float scale, bool exclude_ii)
+        : AABBIterator(neighbor_query, query_point, query_point_idx, exclude_ii), m_count(0), m_num_neighbors(num_neighbors), m_search_extended(false), m_r(r), m_r_cur(r),
```
bdice: maybe replace `m_r` and `m_r_cur` with one variable

cpp/locality/AABBQuery.h
```diff
@@ -198,18 +150,17 @@ class AABBQueryIterator : virtual public NeighborQueryQueryIterator, virtual pub
                                                    //!< used when searching beyond maximum safe AABB distance.
 };
 
-//! Iterator that gets neighbors in a ball of size r using AABB tree structures
-class AABBQueryBallIterator : virtual public AABBIterator
+//! Iterator that gets neighbors in a ball of size r using AABB tree structures.
```
bdice: of size r_max

cpp/locality/LinkCell.h
```diff
@@ -454,24 +454,23 @@ class LinkCell : public NeighborQuery
...
-    //  that are the nearest neighbors for each point.
-    virtual std::shared_ptr<NeighborQueryIterator> query(const vec3<float>* query_points, unsigned int n_query_points,
-                                                         unsigned int num_neighbors, bool exclude_ii = false) const;
-
-    //! Given a set of points, find all elements of this data structure
-    //  that are within a certain distance r.
-    virtual std::shared_ptr<NeighborQueryIterator> queryBall(const vec3<float>* query_points, unsigned int n_query_points,
-                                                             float r_max, bool exclude_ii = false) const;
+    //! Perform a query based on a set of query parameters.
+    /*! Given a QueryArgs object and a set of points to perform a query
```
bdice: fix docstring

cpp/locality/LinkCell.h
```diff
@@ -524,19 +523,17 @@ class LinkCellIterator : virtual public NeighborQueryIterator
...
 
-//! Iterator that gets nearest neighbors from LinkCell tree structures
-class LinkCellQueryIterator : virtual public NeighborQueryQueryIterator, virtual public LinkCellIterator
+//! Iterator that gets specified numbers of nearest neighbors from LinkCell tree structures.
+class LinkCellQueryIterator : public LinkCellIterator
 {
 public:
-    // Explicitly indicate which toNeighborList function is used.
-    using NeighborQueryQueryIterator::toNeighborList;
+    //// Explicitly indicate which toNeighborList function is used.
```
bdice: Fix comment, delete

cpp/order/Steinhardt.cc
```diff
@@ -173,17 +173,42 @@ void Steinhardt::baseCompute(const freud::locality::NeighborList* nlist,
 void Steinhardt::computeAve(const freud::locality::NeighborList* nlist,
                                   const freud::locality::NeighborQuery* points, freud::locality::QueryArgs qargs)
 {
-    std::shared_ptr<freud::locality::NeighborIterator> niter =
-        freud::locality::getNeighborIterator(points, points->getPoints(), m_Np, qargs, nlist);
+    std::shared_ptr<locality::NeighborQuery> nq;
+    std::shared_ptr<locality::NeighborQueryIterator> iter;
+    if (nlist == NULL)
+    {
+        if (dynamic_cast<const locality::RawPoints*>(points))
```
bdice: Should just be in RawPoints -> query

bdice: bool made_nlist = false;
nlist = points->query(qargs).toNeighborList()
something like that
Later:
if (made_nlist)
   delete stuff

bdice: Or just resolve to NeighborList on the Cython side.

vyasr: As I mentioned earlier, we can clean this up further by just implementing an iterator for NeighborLists. Not a per-point iterator, but one that goes over the entire NeighborList and implements the `query` function to return `PerPointIterator`s. Since the NeighborListPerPointIterator has already been implemented, it should be pretty easy to do that. I don't think that's super necessary, so I'm not going to do this in this PR; if we find this use-case popping up in many places, though, we can implement that later and it should be pretty easy.

cpp/locality/RawPoints.h
```diff
@@ -0,0 +1,48 @@
...
+ *  logic, which is primary encapsulated in the NeighborComputeFunctional.h
+ *  file, helps provide a nice Python API as well.
+ */
+class RawPoints : public NeighborQuery
+{
+public:
+    RawPoints() {}
+
+    RawPoints(const box::Box& box, const vec3<float>* points, unsigned int n_points)
+        : NeighborQuery(box, points, n_points), aq(box, points, n_points) {}
```
bdice, line 31: Should the AABBQuery be made on construction, or only if `querySingle` is called?

Maybe initialize `aq` as a null pointer. When `querySingle` is called, then construct the AABBQuery before querying if needed. Unless this needs to be parallel-safe... in which case just leave it as-is unless you have a better idea.

bdice (approved):

PR #422, opened 2019-09-04
Use ManagedArray for ThreadStorage

vyasr:
## Description
Replace raw array with ManagedArray in ThreadStorage.

## Motivation and Context
While this isn't as important as it is for output arrays (which was what ManagedArray was create for re: #151), there are dual benefits to this change. Moving to a standard array representation _everywhere_ in freud helps us encapsulate that logic (especially to isolate any memory management to a single class), and making this change immediately allows the usage of the convenient indexing functions in ManagedArray when accessing ThreadStorage objects.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,7 @@ Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to
 * Ability to specify NeighborQuery objects as points for neighbor-based pair computes.
 * Various validation tests.
 * Added standard method for preprocessing arguments of pair computations.
+* New internal array object that allows data persistence and improves indexing in C++.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


bdice (commented):
@vyasr I approve generally, but is there a reason why we should track the shape and size of the `ThreadStorage` when that information is readily available from the `ManagedArray` itself? I would remove those since there aren't even getters for (or internal uses of) those properties at the `ThreadStorage` level.

vyasr:
I agree, those are holdovers from when they were used in the raw array size allocation and deletion, but I think they are no longer needed and can be safely removed.

bdice (approved):
🧵 Looks good to me!

PR #427, opened 2019-09-07
Refactor/managedarray order

vyasr:
## Description
I changed the order classes to use ManagedArray, except for SolLiq and Steinhardt which are both currently undergoing significant refactoring elsewhere.

This PR also serves as a _significant_ refactoring of the `CubaticOrderParameter` code. In order to make it work with `ManagedArray` I already needed to make some changes to data types deeper in the code, so I took the opportunity to take the code apart and understand how the calculation actually worked. I added extra documentation directly referencing the paper, and created functions that directly map to such equations to make the process much clearer. Some of the changes are just moving code around, including moving the tensor4 class into the CubaticOrderParameter header file because that's the only place it gets used and removed the TensorMath header. I also removed a lot of unused operators from the tensor4 class, made the existing ones class members and defined a couple new ones for convenience. In the CubaticOrderParameter class, I was able to remove a number of class members that were redundant, and I reduced the number of accessible attributes, many of which just introduce confusion and are not actually useful metrics that need to be exposed to the user.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:


cpp/order/CubaticOrderParameter.cc
```diff
@@ -6,148 +6,175 @@
...
-    return m_sp_global_tensor;
+    memcpy(ma.get(), (void*) &data, sizeof(float) * 81);
 }
 
-std::shared_ptr<float> CubaticOrderParameter::getCubaticTensor()
+//! Complete tensor contraction.
+/*! This function is simply a sum-product over two tensors. For reference, see eq. 4.
+ *
+ *  \param a The first tensor.
+ *  \param a The second tensor.
```
bdice:
```suggestion
 *  \param b The second tensor.
```

cpp/order/CubaticOrderParameter.cc
```diff
@@ -6,148 +6,175 @@
...
+ *  a scaled sum of some delta function products. For convenience, its
+ *  calculation is performed in a single function.
+ */ 
+tensor4 genR4Tensor()
 {
-    memcpy(m_sp_gen_r4_tensor.get(), (void*) &m_gen_r4_tensor.data, sizeof(float) * 81);
-    return m_sp_gen_r4_tensor;
-}
+    // Construct the identity matrix to build the delta functions.
+    util::ManagedArray<float> identity({3, 3});
```
bdice, line 113: This probably doesn't demand a change, but `identity` is a function in the standard library for C++20: https://en.cppreference.com/w/cpp/utility/functional/identity

(I only noticed because the syntax highlighting was different for that word.)

cpp/order/CubaticOrderParameter.cc
```diff
@@ -6,148 +6,175 @@
...
-}
 
-float CubaticOrderParameter::getTFinal()
+CubaticOrderParameter::CubaticOrderParameter(float t_initial, float t_final, float scale,
+                                             unsigned int replicates, unsigned int seed)
+    : m_t_initial(t_initial), m_t_final(t_final), m_scale(scale), m_n(0), m_replicates(replicates),
+      m_seed(seed)
 {
-    return m_t_final;
+    // sanity checks, should be caught in python
```
bdice: Is this still true? If not, remove the comment. I think we should defer our Python checks and just throw C++ exceptions if possible.

cpp/order/CubaticOrderParameter.cc
```diff
@@ -6,148 +6,175 @@
...
+    if (m_t_initial < m_t_final)
+        throw std::invalid_argument("CubaticOrderParameter requires that t_initial must be greater than t_final.");
+    if (t_final < 1e-6)
+        throw std::invalid_argument("CubaticOrderParameter requires that t_final must be >= 1e-6.");
+    if ((scale > 1) || (scale < 0))
+        throw std::invalid_argument("CubaticOrderParameter requires that scale must be between 0 and 1.");
+
+    // required to not have memory overwritten
```
bdice: Can you clarify this? I'm not sure what this means.

cpp/order/CubaticOrderParameter.cc
```diff
@@ -284,50 +306,40 @@ void CubaticOrderParameter::compute(quat<float>* orientations, unsigned int n)
...
-            tensor4<float> l_particle_tensor;
-            float l_particle_op;
-            quat<float> l_orientation = orientations[i];
-            calcCubaticTensor((float*) &l_particle_tensor.data, l_orientation);
-            calcCubaticOrderParameter(l_particle_op, (float*) &l_particle_tensor.data);
-            m_particle_order_parameter.get()[i] = l_particle_op;
+            // The per-particle order parameter is defined as the value of the
+            // cubatic order parameter if the global oreintation was the
+            // particle orientation, so we can reuse the same machinery.
+            m_particle_order_parameter[i] = calcCubaticOrderParameter(calcCubaticTensor(orientations[i]), global_tensor);;
```
bdice:
```suggestion
            m_particle_order_parameter[i] = calcCubaticOrderParameter(calcCubaticTensor(orientations[i]), global_tensor);
```

cpp/order/CubaticOrderParameter.h
```diff
@@ -8,23 +8,59 @@
...
+
+    float data[81];
+};
+
+
 //! Compute the cubatic order parameter for a set of points
-/*!
+/*! The cubatic order parameter is defined according to the paper "Strong
+ * orientational coordinates and orientational order parameters for symmetric
+ * objects" by Amir Haji-Akbar
```
bdice:
```suggestion
 * objects" by Amir Haji-Akbari
```

cpp/order/CubaticOrderParameter.h
```diff
@@ -34,37 +70,95 @@ class CubaticOrderParameter
...
+     *
+     *  \param cubatic_tensor The cubatic tensor (denoted M_{\omega} in the paper), overwritten by reference.
+     *  \param orientation The orientation that will be used to determine the vectors used in the calculation.
+     */
+    tensor4 calcCubaticTensor(quat<float> &orientation);
+
+    //! Calculate the scalar cubatic order parameter.
+    /*! Implements eq. 22
+     *
+     *  \param cubatic_order_parameter The output value (updated as a reference)
```
bdice: This comment is inaccurate. I'm not sure what to suggest.

bdice: I think it's supposed to return `cubatic_order_parameter` and there's an argument `global_tensor` that's missing.

cpp/order/CubaticOrderParameter.h
```diff
@@ -34,37 +70,95 @@ class CubaticOrderParameter
...
+     *  corresponding to each basis vector rotated by the provided orientation
+     *  and then summing all these resulting tensors.
+     *
+     *  \param cubatic_tensor The cubatic tensor (denoted M_{\omega} in the paper), overwritten by reference.
+     *  \param orientation The orientation that will be used to determine the vectors used in the calculation.
+     */
+    tensor4 calcCubaticTensor(quat<float> &orientation);
+
+    //! Calculate the scalar cubatic order parameter.
+    /*! Implements eq. 22
```
bdice:
```suggestion
    /*! Implements eq. 22.
```

cpp/order/CubaticOrderParameter.h
```diff
@@ -76,20 +170,16 @@ class CubaticOrderParameter
...
-    std::shared_ptr<float> m_particle_tensor;
-    std::shared_ptr<float>
-        m_sp_gen_r4_tensor; //!< Shared pointer for r4 tensor, only used to return values to Python.
+    tensor4 m_gen_r4_tensor;  //!< The sum of various products of Kronecker deltas that is stored as a member for convenient reuse.
 
+    util::ManagedArray<float> m_particle_order_parameter; //!< The per-particle value of the order parameter.
+    util::ManagedArray<float>
+        m_global_tensor; //!< Copy of global tensor used to return persistent data.
+    util::ManagedArray<float>
+        m_cubatic_tensor; //!< Shared pointer for cubatic tensor, only used to return values to Python.
```
bdice:
```suggestion
        m_cubatic_tensor; //!< Cubatic tensor
```
... or something more informative.

cpp/order/NematicOrderParameter.h
```diff
@@ -28,33 +29,31 @@ class NematicOrderParameter
...
+
 private:
     unsigned int m_n;                //!< Last number of points computed
     vec3<float> m_u;                 //!< The molecular axis
     float m_nematic_order_parameter; //!< Current value of the order parameter
     vec3<float> m_nematic_director;  //!< The director (eigenvector corresponding to the OP)
-    float m_nematic_tensor[9];       //!< The Q tensor
 
-    std::shared_ptr<float> m_sp_nematic_tensor; //!< Pointer to nematic tensor that is passed back
+    util::ManagedArray<float> m_nematic_tensor; //!< Pointer to nematic tensor that is passed back
```
bdice: Not a pointer. Update comment.

cpp/util/diagonalize.cc
```diff
@@ -2,6 +2,14 @@
 
 namespace freud { namespace util {
 
+void diagonalize33SymmetricMatrix(const util::ManagedArray<float> &mat, float eigen_vals[3], float eigen_vecs[9])
```
bdice, line 5: Does this act the same as if you passed `&mat[0]` (or some raw pointer to the first element) to the other constructor?

freud/_order.pxd
```diff
@@ -55,7 +55,7 @@ cdef extern from "HexTransOrderParameter.h" namespace "freud::order":
                      freud._locality.QueryArgs) except +
         # unsure how to pass back the std::complex,
```
bdice: Remove comment.

freud/order.pyx
```diff
@@ -1142,10 +1114,9 @@ cdef class RotationalAutocorrelation(Compute):
 
     @Compute._computed_property()
     def ra_array(self):
```
bdice, line 1116: Surely there's a better name for this. `order`? `ra`? `rotational_autocorrelation`? `autocorrelation`?

bdice (changes requested):
Good rewrite of Cubatic! I would recommend including the actual content of the most relevant equations in the documentation and/or C++ source, in LaTeX form.

cpp/order/CubaticOrderParameter.cc
vyasr: Oh interesting. That's cool. But yeah not worth changing, even C++17 support isn't commonly or completely out of beta yet depending on which compiler you use.

cpp/order/CubaticOrderParameter.cc
vyasr: It's wrong, it was there for old lines of code (for which that was also wrong), but it's no longer even relevant to the line below.

cpp/util/diagonalize.cc
vyasr: Yes, it would, but currently the only reason I didn't change the function altogether is because it's also used in LocalDescriptors. Once I update that as well, I will remove the old function signature here. I'm trying to remove references to raw arrays as much as possible so that everything can use `ManagedArray`. We may at some point want to update `ManagedArray` to support stack allocated arrays for certain specific fixed sizes for convenience, though, but I don't think that's urgent.

freud/order.pyx
vyasr: Almost certainly, but I figured we'd save renaming for later when we decide how to be more consistent with API across the module.

bdice (approved):

PR #429, opened 2019-09-09
Refactor/rename r guess

### LINKED ISSUE #393: Add strict_cut filter to query arguments.
For nearest neighbor queries, it should be possible to set a flag similar to the current NearestNeighbors class that will not find any neighbors beyond a certain distance.

vyasr:
## Description
Rename `r_max` to `r_guess` for AABBQuery, use `r_max` as hard cutoff when provided.

This PR also removes the NearestNeighbors class. The new method for estimating `r_guess` requires the box, and implementing it revealed a subtle issue where both `LinkCell` and `AABBQuery` contained their own member boxes that shadowed the `NeighborQuery` box. Removing it was trivial in `AABBQuery`, but because `NearestNeighbors` had a dependence on modifying `LinkCell` in ways that don't make sense for `NeighborQuery` objects (changing the box and the cell width) there were various conflicts to be resolved when that box was removed. The easiest solution was to just remove `NearestNeighbors` entirely, which also simplified the `LinkCell` code.

## Motivation and Context
Resolves #393. 

## How Has This Been Tested?
Added new test verifying behavior of `r_max` for nearest neighbor query. Also added tests verifying that `scale` and `r_guess` don't affect the correctness of results.

## Types of changes

## Checklist:


bdice (changes requested):
A few comments:

1. The behavior of the `r_max`, `r_guess`, `scale` arguments needs to be documented. If you want to push that to a later PR, that's fine, but we need to make sure we have a corresponding issue on the 2.0 milestone.

2. We need to validate that `scale > 1` or throw an exception. I think otherwise we can have infinite loops...

3. If `r_guess` > `r_max`, we should set `r_guess = r_max`. This helps in the case of large systems where 0.1*L is not appropriate.

4. I would also like to improve that default choice of `r_guess` based on assuming a homogeneous system density and choosing a ball size that is likely to contain `num_neighbors` particles -- I can work out the math for that and do that as a subsequent PR. I'll file an issue for now.

vyasr:
Responding:

1. Yes, agreed. I made #433 to track this.
1. Yes, I'll make that change in this PR.
1. Sure that seems like a reasonable optimization.
1. Sounds good, that is straightforward enough that I'm happy to include that estimate in this code.

freud/locality.pyx
```diff
@@ -749,23 +759,12 @@ def make_default_nlist_nn(box, points, query_points, num_neighbors, nlist=None,
...
-    if nlist is None:
-        cnlist.base = None
+    cdef AABBQuery aq = AABBQuery(box, points)
+    cdef NeighborList aq_nlist = aq.query(
+        query_points, dict(num_neighbors=num_neighbors, r_guess=r_max_guess,
+                           exclude_ii=exclude_ii)).toNeighborList()
 
-    # Return the owner of the neighbor list as well to prevent gc problems
-    return nn.nlist, nn
+    return aq_nlist, aq
```
bdice, line 767: We need to follow these changes with a PR that just returns the nlist -- no need for the owning object anymore.

tests/test_environment_LocalDescriptors.py
```diff
@@ -156,8 +156,10 @@ def get_Ql(p, descriptors, nlist):
 
             # In order to be able to access information on which particles are
             # bonded to which ones, we precompute the neighborlist
-            nn = freud.locality.NearestNeighbors(r_max, num_neighbors)
-            nl = nn.compute(box, points).nlist
+            lc = freud.locality.LinkCell(box, r_max/10, points)
```
bdice: Our default choice should always be AABBQuery -- fewer parameters and faster performance.

tests/test_locality_NeighborList.py
```diff
@@ -1,174 +1,148 @@
...
+        self.N = 40
+
+        self.query_args = dict(mode="nearest", num_neighbors=6, r_guess=3,
+                               exclude_ii=True)
 
         # Initialize Box and cell list
-        self.cl = locality.NearestNeighbors(self.r_max, self.num_neighbors)
-        self.fbox, self.points = make_box_and_random_points(L, N)
+        box, points = make_box_and_random_points(self.L, self.N)
+        self.nq = freud.locality.LinkCell(box, self.query_args['r_guess']/10,
```
bdice, line 19: Use AABBQuery.

cpp/locality/NeighborQuery.h
```diff
@@ -167,6 +169,10 @@ class NeighborQuery
         {
             if (args.num_neighbors == QueryArgs::DEFAULT_NUM_NEIGHBORS)
                 throw std::runtime_error("You must set num_neighbors in the query arguments when performing number of neighbor queries.");
+            if (args.r_max == QueryArgs::DEFAULT_R_MAX)
```
bdice, line 172: Seems like we're missing a check that `scale > 1` in the `args.mode == QueryArgs::nearest` section here.

bdice (changes requested):

freud/locality.pyx
vyasr: See #434.

cpp/locality/NeighborQuery.h
vyasr: Discussed offline, we want to keep this check in `AABBQuery`. However, the most consistent thing to do here would be to also add explicit checks that users aren't passing in arguments that are unused and therefore implicitly ignored (e.g. passing `r_guess` or `scale` to `LinkCell.query`).

vyasr:
Attempting to change the uses of `LinkCell` to `AABBQuery` revealed what appears to be a bug in `Steinhardt`, perhaps due to tolerances. The `NeighborList` objects generated through both methods are identical, except that the distances are slightly off, so I suspect that the problem is caused by this, but can't be sure. However, the test I added (`test_wl_new` in `test_environment_LocalDescriptors.py`) does check that `LocalDescriptors` produces essentially identical results for both methods. @bdice can you have a look?

vyasr:
I fixed one problem, which was that in certain edge cases where particles are aligned on the z-axis we were calling `acos` with a number that was outside the acceptable bounds due to problems with machine precision, e.g. `-2/2 = -1 - epsilon` for some nonzero positive epsilon. This fixes the specific test that I introduced, but the `test_ql` test still fails without increasing tolerances. This is because the structures are too perfect, so the numerical imprecisions dominate the results. I just increased the tolerances to fix this.

bdice (approved):

PR #428, opened 2019-09-08
Fix/linkcell issue pr409

vyasr:
## Description
Make sure LinkCell::query sorts the neighbors before deciding how many sufficiently close points have been found.

## Motivation and Context
Fixes the bug found in #409.

## How Has This Been Tested?
Added the test from #409 and it now passes.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] I have updated the documentation (if relevant).
- [x] I have added tests that cover my changes (if relevant).
- [x] All new and existing tests passed.
- [x] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

PR #430, opened 2019-09-09
Modernize BondOrder

bdice:
## Description
This PR updates the API of the BondOrder class to match more of the rest of freud v2.

## Motivation and Context
Related to #406.

## How Has This Been Tested?
Tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -13,7 +13,7 @@ import numpy as np
 import warnings
 import freud.locality
 
-from freud.common cimport Compute
+from freud.common cimport Compute, PairCompute
 from freud.util cimport vec3, quat
 from libcpp.vector cimport vector
 from libcpp.map cimport map
@@ -29,7 +29,7 @@ cimport numpy as np
 np.import_array()
 
 
-cdef class BondOrder(Compute):
+cdef class BondOrder(PairCompute):
     R"""Compute the bond orientational order diagram for the system of
     particles.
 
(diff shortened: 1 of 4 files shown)


freud/environment.pyx
```diff
@@ -137,6 +137,11 @@ cdef class BondOrder(Compute):
     def __dealloc__(self):
         del self.thisptr
 
+    @property
+    def default_query_args(self):
+        return dict(mode="nearest", r_max=self.r_max,
+                    num_neighbors=self.num_neighbors)
```
vyasr, line 143: Does it make sense for `self.r_max` to exist anymore? I think that once we're closer to finalizing the API it may make sense to make it possible for users to set the default query arguments (probably will require renaming to just `self.query_args` or so), but in general I don't think query argument-like things should be class members anymore unless they're strictly bounds on a histogram or something similar (like `r_max` for RDFs).

vyasr (commented):
One question, depending on what you think there may be more changes to make here.

freud/environment.pyx
```diff
@@ -137,6 +137,11 @@ cdef class BondOrder(Compute):
...
+
     @Compute._compute()
     def accumulate(self, box, points, orientations, query_points=None,
                    query_orientations=None, str mode="bod", nlist=None):
```
vyasr, line 147: Actually one more comment, `query_args={}` should be added as an optional argument here.

vyasr: Also why is mode typed?

vyasr (commented):
Couple comments I missed.

vyasr:
One of the tests is failing, any idea why?

bdice:
> One of the tests is failing, any idea why?

The LinkCell fix in #428 resolved the issue, once I rebased.

freud/environment.pyx
bdice: Punting this for a later rewrite to support QueryArgs.

freud/environment.pyx
bdice: Punting this for a later rewrite to support QueryArgs.

vyasr (approved):

PR #437, opened 2019-09-13
Feature/r min filter

vyasr:
## Description
I added the ability to filter neighbors based on a minimum distance.

I also made a few other minor changes to the code to clean things up, such as renaming outstanding instances of NeighborBond objects named `np`.

## Motivation and Context
Resolves #356. 

## How Has This Been Tested?
I added a new test that adapts existing tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] I have updated the documentation (if relevant).
- [x] I have added tests that cover my changes (if relevant).
- [x] All new and existing tests passed.
- [x] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

PR #438, opened 2019-09-14
Feature/histogram rewrite

### LINKED ISSUE #385: RDF should use "bins" argument

vyasr:
## Description
I implemented a Histogram class that supports arbitrary dimensionality and easy binning of values into the histogram. As a test-case, I've implemented its usage in the RDF class. I've also made other simplifications to the RDF class to further simplify it and improve performance.

One breaking change that was made along with this PR is that `freud.density.RDF` now accepts a number of bins as an argument rather than a bin width. One major fix that results from this is that we always completely capture the range from `r_min` to `r_max`.

## Motivation and Context
Many methods in freud amount to computing a histogram, typically of bond distances. A preliminary attempt was made to refactor the common elements of these classes using the NdHistogram, but that class is still quite specialized to the RDF/PMFT use-cases and does not abstract away the core logic we want to remove, which is the binning of the data into a histogram. The new histogram class is a simple data structure that supports this goal.

Resolves #385.

## How Has This Been Tested?
Existing tests of the RDF pass. I had to adapt them because the argument order changed and the bin spacing is now also subtly different because we no longer neglect a final partial bin.

## Types of changes

## Checklist:


cpp/density/RDF.cc
```diff
@@ -1,157 +1,100 @@
...
 
 namespace freud { namespace density {
 
-RDF::RDF(float r_max, float dr, float r_min) : util::NdHistogram(), m_r_max(r_max), m_r_min(r_min), m_dr(dr)
+RDF::RDF(unsigned int bins, float r_max, float r_min) : m_box(box::Box()), m_frame_counter(0), m_n_points(0),
+    m_n_query_points(0), m_reduce(true), m_r_max(r_max), m_r_min(r_min), m_bins(bins)
 {
-    if (dr <= 0.0f)
-        throw invalid_argument("RDF requires dr to be positive.");
+    if (bins <= 0)
```
bdice:
```suggestion
    if (bins == 0)
```

cpp/density/RDF.cc
```diff
@@ -1,157 +1,100 @@
...
-    assert(query_points);
-    assert(m_n_points > 0);
-    assert(n_query_points > 0);
-
-    float dr_inv = 1.0f / m_dr;
-    accumulateGeneral(neighbor_query, query_points, n_query_points, nlist, qargs,
+    m_box = neighbor_query->getBox();
+    locality::loopOverNeighbors(neighbor_query, query_points, n_query_points, qargs, nlist,
         [=](const freud::locality::NeighborBond& neighbor_bond) {
         if (neighbor_bond.distance < m_r_max && neighbor_bond.distance > m_r_min)
```
bdice: Should this be `>= r_min`? This effectively drops all bonds with distance zero. Not sure what the appropriate physical limit is, but typically I think we're using `min <= dist < max` for criteria in other modules.

cpp/density/RDF.cc
```diff
@@ -1,157 +1,100 @@
...
 
-    m_N_r_array.get()[0] = m_avg_counts.get()[0];
-    for (unsigned int i = 1; i < m_nbins; i++)
-    {
-        m_N_r_array.get()[i] = m_N_r_array.get()[i-1] + m_avg_counts.get()[i];
-    }
-
-    for (unsigned int i = 0; i < m_nbins; i++)
+    // Define prefactors with appropriate types to simplify and speed later code.
+    float ndens = float(m_n_query_points) / m_box.getVolume();
```
bdice, line 56: When using NPT with accumulation, do we reduce on every frame or just at the end? Do we need to adjust how we handle this reduction step for correctness with variable box sizes?

cpp/density/RDF.h
```diff
@@ -32,53 +29,99 @@ class RDF : public util::NdHistogram
...
     }
 
-    unsigned int getNBins();
+    float getRMin() const
+    {
+        return m_r_min;
+    }
+
+    //! Get bin centers.
+    std::vector<float> getR()
```
bdice: Any reason not to return a `ManagedArray`?

cpp/density/RDF.h
```diff
@@ -32,53 +29,99 @@ class RDF : public util::NdHistogram
...
+    {
+        // RDFs are always 1D histograms, so we just return the first element.
+        return m_histogram.getBinCenters()[0];
+    }
+
+    //! Return the bin boundaries.
+    std::vector<float> getBins() const
```
bdice, line 101: Any reason not to return a `ManagedArray`?

bdice (commented):
A few comments below. Please review them and decide if further action is needed. I think some of those tasks can be deferred as issues if this PR is blocking any other progress.

cpp/density/RDF.cc
vyasr: I thought about changing this, but once #437 is merged all of this logic will be subsumed into that anyway so I didn't bother. Histograms should of course still internally validate that provided values are within bounds.

cpp/density/RDF.cc
vyasr: We reduce at the very end. Right now, I don't think the RDF handles changing boxes. You'd have to keep track of the box passed to every call to `accumulate`. That seems like a reasonable feature to add, though, if there's any interest.

cpp/density/RDF.h
vyasr: We could change the behavior specifically for RDF, but in general multi-dimensional histograms could have different numbers of bins in different dimensions, so it makes more sense to return a vector of vectors (which becomes a list of lists in Python). If we ever actually implement the feature to specify bin locations for RDFs, then I'd also expect us to enable that feature for e.g. PMFTs, and those would definitely be lists of lists. I'm not even sure if the property on the Python side should be a NumPy array, but I left it for now as a later API question when the API rewrite is happening. I'm open to ideas on how else we could do it, though.

vyasr:
It's OK if we make edits on this PR, it's not urgent to get it merged. There's still plenty of orthogonal progress to be made. However, to make further progress generalizing this to more classes, I'll need to make changes to the corresponding classes on the Python side, and I know that could conflict with #362 since you moved and maybe changed `SpatialHistogram`. How do you want to proceed on that front? Should I wait until that PR is done before continuing here?

vyasr:
Actually it might be beneficial to just use this histogram in a few other places without trying to generalize the compute logic for histogram methods just yet, so I'd say we can move forward on this PR once you think it's good without immediately needing to resolve the Cython-side changes involved in #362.

cpp/density/RDF.cc
bdice: #437 has been merged, and I think this should be changed to `>= m_r_min`.

cpp/density/RDF.cc
bdice: Please document that we don't give correct answers for accumulations over varying box sizes.

cpp/density/RDF.cc
bdice: @vyasr Apply the change I suggested to `==`?

cpp/density/RDF.h
bdice: I want a NumPy array as the user-facing API, but it's fine to return a vector here to Cython and then convert it with `np.asarray` so long as it returns a copy and/or cannot be invalidated.

cpp/density/RDF.h
bdice: Same as above: it's fine to return a vector here to Cython and then convert it with `np.asarray` so long as it returns a copy and/or cannot be invalidated.

cpp/pmft/PMFT.h
```diff
@@ -74,48 +108,64 @@ class PMFT : public util::NdHistogram
...
     }
 
-    //! Helper function to reduce two dimensionally with appropriate Jaocobian.
-    template<typename JacobFactor>
-    void reduce2D(unsigned int first_dim, unsigned int second_dim, JacobFactor jf)
-    {
-        reduce3D(1, first_dim, second_dim, jf);
-    }
-
     //! Helper function to reduce three dimensionally with appropriate Jaocobian.
```
bdice:
```suggestion
    //! Helper function to reduce three dimensionally with appropriate Jacobian.
```

bdice (changes requested):
@vyasr More requests related to previous comments.

cpp/density/RDF.h
vyasr: How would you do this for PMFTs? I think there are some API questions to sort out regarding how we specify bins for multi-dimensional arrays, which should correspond naturally to how we return the objects.

cpp/density/RDF.cc
vyasr: I don't know if I made my point clear; this check should _removed entirely_ here. The check is done by the histogram, it does not need to happen here any more at all.

cpp/density/RDF.h
vyasr: As discussed offline, we're mostly going with my proposal of returning tuples of bins. Main modification is that each set of bins will be a numpy array rather than a list, so the cast makes sense. This is already how RDF is behaving, so I'll update PMFTs and call it good.

cpp/pmft/PMFTR12.cc
```diff
@@ -33,70 +27,51 @@ PMFTR12::PMFTR12(float r_max, unsigned int n_r, unsigned int n_t1, unsigned int
...
-    if (m_dt2 > m_t2_max)
-        throw invalid_argument("PMFTR12 requires that dt2 is less than or equal to t2_max.");
-
-    // precompute the bin center positions for r
-    m_r_array = precomputeArrayGeneral(m_n_r, m_dr, [](float r, float nextr) {
-        return 2.0f / 3.0f * (nextr * nextr * nextr - r * r * r) / (nextr * nextr - r * r);
-    });
+    // Construct the Histogram object that will be used to keep track of counts of bond distances found.
+    util::Histogram::Axes axes;
+    float angle_max = 2.0 * M_PI;
```
bdice:
```suggestion
    const float TWO_PI = 2.0 * M_PI;
```

cpp/pmft/PMFTR12.cc
```diff
@@ -33,70 +27,51 @@ PMFTR12::PMFTR12(float r_max, unsigned int n_r, unsigned int n_t1, unsigned int
...
+    axes.push_back(std::make_shared<util::RegularAxis>(n_r, 0, r_max));
+    axes.push_back(std::make_shared<util::RegularAxis>(n_t1, 0, angle_max));
```
bdice:
```suggestion
    axes.push_back(std::make_shared<util::RegularAxis>(n_t1, 0, TWO_PI));
```

cpp/pmft/PMFTR12.cc
```diff
@@ -33,70 +27,51 @@ PMFTR12::PMFTR12(float r_max, unsigned int n_r, unsigned int n_t1, unsigned int
...
+    axes.push_back(std::make_shared<util::RegularAxis>(n_t2, 0, angle_max));
```
bdice:
```suggestion
    axes.push_back(std::make_shared<util::RegularAxis>(n_t2, 0, TWO_PI));
```

cpp/pmft/PMFTR12.cc
```diff
@@ -33,70 +27,51 @@ PMFTR12::PMFTR12(float r_max, unsigned int n_r, unsigned int n_t1, unsigned int
...
+    m_histogram = util::Histogram(axes);
+    m_local_histograms = util::Histogram::ThreadLocalHistogram(m_histogram);
 
     // calculate the jacobian array; computed as the inverse for faster use later
-    m_inv_jacobian_array.prepare({m_n_r, m_n_t1, m_n_t2});
-    for (unsigned int i = 0; i < m_n_r; i++)
+    m_inv_jacobian_array.prepare({n_r, n_t1, n_t2});
+    std::vector<float> bins_r = m_histogram.getBinCenters()[0];
+    float dr = r_max / float(n_r);
+    float dt1 = angle_max / float(n_t1);
```
bdice:
```suggestion
    float dt1 = TWO_PI / float(n_t1);
```

cpp/pmft/PMFTR12.cc
```diff
@@ -33,70 +27,51 @@ PMFTR12::PMFTR12(float r_max, unsigned int n_r, unsigned int n_t1, unsigned int
...
+    float dt2 = angle_max / float(n_t2);
```
bdice:
```suggestion
    float dt2 = TWO_PI / float(n_t2);
```

cpp/pmft/PMFTR12.cc
```diff
@@ -119,26 +94,7 @@ void PMFTR12::accumulate(const locality::NeighborQuery* neighbor_query,
             {
                 t2 += 2 * M_PI;
```
bdice: Make `TWO_PI` a constant that can be used here too?

cpp/pmft/PMFTXYT.cc
```diff
@@ -32,60 +26,42 @@ PMFTXYT::PMFTXYT(float x_max, float y_max, unsigned int n_x, unsigned int n_y, u
         throw invalid_argument("PMFTXYT requires that x_max must be positive.");
     if (y_max < 0.0f)
         throw invalid_argument("PMFTXYT requires that y_max must be positive.");
+    float angle_max = 2.0 * M_PI;
```
bdice: Maybe do the same thing with pi here?

freud/pmft.pyx
```diff
@@ -258,49 +283,20 @@ cdef class PMFTR12(_PMFT):
...
     def __repr__(self):
+        bounds = self.bounds
         return ("freud.pmft.{cls}(r_max={r_max}, n_r={n_r}, n_t1={n_t1}, "
                 "n_t2={n_t2})").format(cls=type(self).__name__,
                                        r_max=self.r_max,
-                                       n_r=self.n_bins_R,
-                                       n_t1=self.n_bins_T1,
-                                       n_t2=self.n_bins_T2)
+                                       n_r=bounds[0][1],
+                                       n_t1=bounds[2][1],
```
bdice:
```suggestion
                                       n_t1=bounds[1][1],
```

freud/pmft.pyx
```diff
@@ -465,48 +457,20 @@ cdef class PMFTXYT(_PMFT):
...
         return ("freud.pmft.{cls}(x_max={x_max}, y_max={y_max}, n_x={n_x}, "
                 "n_y={n_y}, n_t={n_t})").format(cls=type(self).__name__,
-                                                x_max=self.xmax,
-                                                y_max=self.ymax,
-                                                n_x=self.n_bins_X,
-                                                n_y=self.n_bins_Y,
-                                                n_t=self.n_bins_T)
+                                                x_max=bounds[0][1],
+                                                y_max=bounds[1][1],
+                                                n_x=self.nbins[0],
```
bdice: ...and similarly below.
```suggestion
                                                n_x=nbins[0],
```

bdice (changes requested):

cpp/density/RDF.h
```diff
@@ -32,53 +29,99 @@ class RDF : public util::NdHistogram
...
+    {
+        // RDFs are always 1D histograms, so we just return the first element.
+        return m_histogram.getBinBoundaries()[0];
```
bdice: Tests failed, I think this will fix it.
```suggestion
        return m_histogram.getBinEdges()[0];
```

bdice (approved):
@vyasr I resolved some of my own comments for obvious fixes. The only possible change worth considering would be to clarify how we handle the constant of 2 pi in PMFTs with angles. I think `TWO_PI` is more descriptive than `angle_max`. This isn't a blocking issue, so I am approving this PR and you can choose whether you want to do something or nothing about that constant of 2 pi.

vyasr:
I took your suggestion and made file-level constants for 2*pi in both those places.

PR #439, opened 2019-09-15
Refactor/managedarray steinhardt

vyasr:
## Description
Steinhardt now uses ManagedArrays, which have been optimized.

## Motivation and Context
More ManagedArray usage. Also, Steinhardt revealed some cases where ManagedArray indexers were performance bottlenecks, so the optimization was necessary.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:


cpp/order/Steinhardt.cc
```diff
@@ -152,31 +130,32 @@ void Steinhardt::baseCompute(const freud::locality::NeighborList* nlist,
                     theta = 0;
                 }
 
-                std::vector<std::complex<float>> Ylm(2 * m_l + 1);
+                std::vector<std::complex<float>> Ylm(m_num_ms);
                 this->computeYlm(theta, phi, Ylm); // Fill up Ylm
 
                 for (unsigned int k = 0; k < Ylm.size(); ++k)
```
bdice: For consistency:
```suggestion
                for (unsigned int k = 0; k < m_num_ms; ++k)
```

cpp/order/Steinhardt.cc
```diff
@@ -240,43 +222,44 @@ float Steinhardt::normalize()
     if (m_Wl)
     {
         auto wigner3jvalues = getWigner3j(m_l);
+        //TODO Change this from using a pointer
```
bdice: I'd prefer to resolve that TODO in this PR. Let me know if you'd like me to do it.

cpp/order/Steinhardt.cc
```diff
@@ -240,43 +222,44 @@ float Steinhardt::normalize()
...
     });
 }
 
 void Steinhardt::reduce()
 {
-    parallel_for(tbb::blocked_range<size_t>(0, 2 * m_l + 1), [=](const blocked_range<size_t>& r) {
+    parallel_for(tbb::blocked_range<size_t>(0, m_num_ms), [=](const tbb::blocked_range<size_t>& r) {
         for (size_t i = r.begin(); i != r.end(); i++)
         {
             for (util::ThreadStorage<complex<float>>::const_iterator Ql_local = m_Qlm_local.begin();
```
bdice: Try `auto` here?
```suggestion
            for (auto Ql_local = m_Qlm_local.begin();
```

cpp/util/ManagedArray.h
```diff
@@ -154,67 +152,80 @@ template<typename T> class ManagedArray
...
+    // variadic template arguments.
+    //
+    // Performance note: the variadic indexers can be expensive
+    // since they have to build up a std::vector through a sequence of
+    // (hopefully compiler optimized) function calls. In general, this is
+    // unimportant since performance critical code paths in freud do not
+    // involve large amounts of indexing into ManagedArrays. However, in cases
+    // where critical code paths do index into the array, it may be
+    // advantageous to directly call the function with a std::vector of indices
+    // and bypass the variadic indexers.
```
bdice, line 167: Given the performance constraints we're seeing on variadic indexing, I am somewhat inclined to remove variadic indexing and just supply `operator()` overloads for 1, 2, 3, and 4-dimensional ManagedArrays (similar to the `IndexND` approach of the previous codebase) and require vector notation for anything higher. The operator overload would become considerably simpler and more maintainable (though more repetitive), and it wouldn't cause these performance issues (as far as I can tell):
```c++
// 4D case
inline T &operator()(int i, int j, int k, int l)
{
    return (*this)({i, j, k, l});
}
```

I would include up to 4D so that we can handle cubatic tensors, and that's the highest dimensionality I'm aware of in freud.

bdice: Just a suggestion -- if the problem is adequately solved and you don't see the variadic slowdown being a problem elsewhere, then we can leave this as-is. But the comments you wrote seem to indicate this is a significant slowdown, so I thought I'd pose an alternative.

bdice (changes requested):
Comments below.

cpp/order/Steinhardt.cc
vyasr: The reason I haven't done it yet is that I'm not quite sure how yet. In particular, I'm not sure what the best way is to adapt it for the use-case in `aggregateWl`. Let me know if you have ideas.

cpp/util/ManagedArray.h
vyasr: I don't think there's a strong reason overall to prefer either approach (there are pros and cons to both with respect to complexity vs duplication/generality), but I do think the problem is adequately solved. In general, Steinhardt is the first time that I'm finding this problem, and it's specifically because in `computeAve` we use the indexer inside like 4 nested for loops _and_ the indexing is the only operation in there. I haven't noticed the problem in any other changes. We can definitely revisit this if we start seeing the slowdown more often, but for now I don't think it's worth overreacting to this one use-case.

I would consider this difference to be similar to the difference between directly indexing into vs use `at` for `std::vector`, there are tradeoffs to both approaches and I think it's OK to just have the options available.

bdice (approved):
lgtm!

PR #441, opened 2019-09-18
Use ManagedArray in ParticleBuffer

bdice:
## Description
Switched ParticleBuffer to use ManagedArrays.

## Motivation and Context
More ManagedArray conversions.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/box/ParticleBuffer.cc
+++ b/cpp/box/ParticleBuffer.cc
@@ -1,12 +1,8 @@
 // Copyright (c) 2010-2019 The Regents of the University of Michigan
 // This file is from the freud project, released under the BSD 3-Clause License.
 
-#include <cassert>
-#include <memory>
 #include <stdexcept>
-#include <vector>
 
-#include "ManagedArray.h"
 #include "ParticleBuffer.h"
 
 using namespace std;
@@ -28,9 +24,6 @@ void ParticleBuffer::compute(const vec3<float>* points, const unsigned int Np, c
         throw invalid_argument("Buffer y distance must be non-negative.");
     if (buff.z < 0)
         throw invalid_argument("Buffer z distance must be non-negative.");
-
-    vector<vec3<float>> buffer_parts;
-    vector<unsigned int> buffer_ids;
 
     // Get the box dimensions
     vec3<float> L(m_box.getL());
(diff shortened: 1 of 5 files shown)


bdice:
@vyasr Following up on our offline discussion about whether this should be a ManagedArray or not, I prefer to keep it as-is in this PR and use a ManagedArray. If I recall correctly, it's not straightforward to convert a `vector<vec3<float>>` to an `(N, 3)` NumPy array without some memory view magic (potentially involving some raw pointer casting), since it involves `vec3`s. That would be easy if it were a vector of primitive types, but I would prefer to avoid retaining memory view logic in freud's Cython layer if we can avoid it.

vyasr:
Maybe it wasn't clear in our offline discussion, what I was proposing was just returning a copy. There's no reason not to if we're already forced to copy internally to generate the ManagedArray. So there wouldn't be any memoryviews, we'd just return a copy of the vector and use:

```
        particles = self.thisptr.getBufferParticles()
        return np.asarray([[p.x, p.y, p.z] for p in particles])
```

vyasr:
I actually already implemented the change and just saw this comment when I came back here. I'm going to push, but we can revert my commit if we decide to stick with ManagedArray here.

bdice:
@vyasr Tests are failing for Voronoi because of a lack of memory contiguity in the input array of buffer particles. Shall we treat this as an error for ParticleBuffer (changing from contiguous to non-contiguous arrays) or the Voronoi class (not ensuring its data is contiguous) or both? Should we call `freud.common.convert_array` on the output?

vyasr:
> @vyasr Tests are failing for Voronoi because of a lack of memory contiguity in the input array of buffer particles. Shall we treat this as an error for ParticleBuffer (changing from contiguous to non-contiguous arrays) or the Voronoi class (not ensuring its data is contiguous) or both? Should we call `freud.common.convert_array` on the output?

From a quick glance, it seems like the most natural place to fix this is in `_qhull_compute`. Users of `ParticleBuffer` may not care about contiguity, but `Voronoi` does, so `Voronoi` should fix it. Does that seem reasonable?

Also, feel free to resolve the other discussion as you see fit.

vyasr (approved):

Issue #435, opened 2019-09-11
Inline references


vyasr:
I think our references section is a little strange to be honest. I think we should put all citations inline (i.e. in the context of whatever method they're there for) and get rid of this section. We should also get more consistent with having citations, but that's a longer term things (maybe good for a hackathon to ask someone to identify features that need a citation).

_Originally posted by @vyasr in https://github.com/glotzerlab/freud/pull/383/files_

bdice:
We should use this package: https://sphinxcontrib-bibtex.readthedocs.io/en/latest/quickstart.html#minimal-example

vyasr:
After attempting this with #520 (where we did switch to bibtex), we decided that inlining citations was not worth the extra complications associated with avoiding duplicate references in bibtex.

PR #445, opened 2019-09-18
SolidLiquid rewrite

bdice:
## Description
This PR is a complete rewrite of the SolidLiquid order parameter code. It uses other modules (Steinhardt, Cluster) internally, instead of re-implementing those functionalities.

## Motivation and Context
Bringing SolidLiquid code up to date for freud v2.0.

## How Has This Been Tested?
Existing tests will be adapted.

## Types of changes

## Checklist:


bdice:
The tests will probably pass on this, but I don't think the tests are performing adequate validation for this class to be considered "ready." I should also verify the results against the old code.

bdice:
@vyasr I have finished with this and it's ready for review.

bdice:
@vyasr I added an inline citation (#435) which looks like this. I'm not sure if I like the formatting very well. Suggestions welcome.
image

cpp/order/Steinhardt.h
```diff
@@ -124,29 +130,29 @@ class Steinhardt
     }
 
     //! Compute the order parameter
-    virtual void compute(const freud::locality::NeighborList* nlist,
+    void compute(const freud::locality::NeighborList* nlist,
+                                  const freud::locality::NeighborQuery* points, freud::locality::QueryArgs qargs);
+
+    //! Calculates Qlms and the Ql order parameter before any further modifications
+    void baseCompute(const freud::locality::NeighborList* nlist,
```
vyasr: I think this method should still be private.

vyasr: (or at least protected)

freud/order.pyx
```diff
@@ -607,351 +606,172 @@ cdef class Steinhardt(PairCompute):
...
+            (:class:`matplotlib.axes.Axes`): Axis with the plot.
         """
-        cdef freud.locality.NeighborList nlist_
-        nlist_ = freud.locality.make_default_nlist(
-            self.m_box, points, None, dict(num_neighbors=self.num_neighbors,
-                                           r_guess=self.r_max), nlist)
-        return SolLiq.computeSolLiqNoNorm(self, points, nlist_)
+        import freud.plot
+        try:
+            count = np.unique(self.cluster_idx, return_counts=True)
```
vyasr: Unpack the tuple here and pass it into the plot function.

cpp/order/SolidLiquid.h
```diff
@@ -0,0 +1,129 @@
...
+
+#ifndef SOLID_LIQUID_H
+#define SOLID_LIQUID_H
+
+#include <algorithm>
+#include <complex>
+#include <memory>
+#include <stdexcept>
+#include <stdint.h>
+#include <vector>
```
vyasr, line 8: You should be able to remove memory and stdint. Not sure about algorithm (I don't see sort the usual suspects like `sort` or `max`). Move `stdexcept` to the cc file.

cpp/order/SolidLiquid.h
```diff
@@ -0,0 +1,129 @@
...
+#include "ManagedArray.h"
+#include "NeighborList.h"
+#include "Steinhardt.h"
+#include "ThreadStorage.h"
+#include "VectorMath.h"
+#include "utils.h"
+
+namespace freud { namespace order {
+
+//! Computes dot products of Qlm between particles and uses these for clustering
```
vyasr: Missing period.

cpp/order/SolidLiquid.h
```diff
@@ -0,0 +1,129 @@
...
+     *  \param l Spherical harmonic number l.
+     *  \param Q_threshold Value of dot product threshold when evaluating
+     *     \f$Q_{lm}(i) Q_{lm}^*(j)\f$ to determine if a neighbor pair is
+     *     a solid-like bond. (For l=6, 0.7 is generally good for FCC or BCC
+     *     structures)
+     *  \param S_threshold Minimum required number of adjacent solid-like bonds
+     *     for a particle to be considered solid-like for clustering. (For
+     *     l=6, 6-8 is generally good for FCC or BCC structures)
+     *  \param normalize_Q Whether to normalize the per-bond dot products of Qlm.
+     */
```
vyasr, line 63: In our C++ docstrings we usually put general descriptions of the class above the class rather than in the constructor documentation. I'd remove the "After creation..." stuff and move that to a separate comment outside the class (i.e. add to the comment where I asked you to add a period above). It would also be helpful at that point to add some detail about how this interacts with Cluster and Steinhardt.

cpp/order/SolidLiquid.cc
```diff
@@ -0,0 +1,101 @@
...
+
+void SolidLiquid::compute(const freud::locality::NeighborList* nlist,
+        const freud::locality::NeighborQuery* points, freud::locality::QueryArgs qargs)
+{
+    // Make NeighborList from NeighborQuery if not provided
+    if (nlist == NULL)
+    {
+        auto nqiter(points->query(points->getPoints(), points->getNPoints(), qargs));
+        nlist = nqiter->toNeighborList();
+    }
```
vyasr, line 29: We should standardize the way we deal with computes that require NeighborList objects somewhere. It would be convenient to do in Python, but as we move towards exposing our C++ API it probably makes sense to add the logic in C++ instead. Either way, needs to be possible.

cpp/order/SolidLiquid.cc
```diff
@@ -0,0 +1,101 @@
...
+                // Optionally normalize dot products by points' Ql values,
+                // accounting for the normalization of Ql values
+                if (m_normalize_Q)
+                {
+                    bond_Ql_ij *= normalizationfactor / (Ql[i] * Ql[j]);
+                }
+                m_Ql_ij[bond] = bond_Ql_ij.real();
+            }
+        }
+    }, true);
```
vyasr, line 67: I'm wondering if we should add a variation on the existing `loopOver*` functions that's specifically for `NeighborLists` and in addition to providing the `NeighborBond` instance, also provides the index in the `NeighborList` of the current bond. The logic would be very similar to what's already in the `NeighborListPerPointIterator`, and it would make these kinds of loops simpler since they're in multiple places (`LocalDescriptors`, `LocalBondProjection`, `AngularSeparation`, etc). Thoughts?

cpp/order/SolidLiquid.cc
```diff
@@ -0,0 +1,101 @@
...
+
+    // Filter neighbors to contain only solid-like bonds
+    unique_ptr<bool[]> solid_filter(new bool[num_bonds]);
```
vyasr: Is there a `using namespace std` hiding somewhere to make this work?

cpp/order/SolidLiquid.cc
```diff
@@ -0,0 +1,101 @@
...
+            unsigned int bond(nlist->find_first_index(i));
+            for (; bond < num_bonds && nlist->getNeighbors()(bond, 0) == i; ++bond)
+            {
+                const unsigned int j(nlist->getNeighbors()(bond, 1));
+
+                // Accumulate the dot product over m of Qlmi and Qlmj vectors
+                std::complex<float> bond_Ql_ij = 0;
+                for (unsigned int k = 0; k < m_num_ms; k++)
+                {
+                    bond_Ql_ij += Qlm(i, k) * conj(Qlm(j, k));
```
vyasr: There must be a `using namespace` hiding somewhere for this `conj` to work.

vyasr: I think I found it in wigner3j.h. Please remove it and fix the errors that crop up with it.

freud/order.pyx
```diff
@@ -607,351 +606,172 @@ cdef class Steinhardt(PairCompute):
...
+    If a particle has more than :math:`S_{threshold}` solid-like bonds, then
+    the particle is considered solid-like. Finally, solid-like particles are
+    clustered.
+
+    .. moduleauthor:: Bradley Dice <bdice@bradleydice.com>
     .. moduleauthor:: Richmond Newman <newmanrs@umich.edu>
 
+    .. [Frenkel1995] ten Wolde, P. R., Ruiz-Montero, M. J., & Frenkel, D
+       (1995).  Numerical Evidence for bcc Ordering at the Surface of a
+       Critical fcc Nucleus. Phys. Rev. Lett., 75 (2714).
```
vyasr, line 633: Can you add a citation for the other paper that you showed me? The one that had a more extensive mathematical description (and was properly normalized for our purposes of testing) and had reference plots for FCC, BCC, and SC?

cpp/order/SolidLiquid.cc
```diff
@@ -0,0 +1,101 @@
...
+                }
+
+                // Optionally normalize dot products by points' Ql values,
+                // accounting for the normalization of Ql values
+                if (m_normalize_Q)
+                {
+                    bond_Ql_ij *= normalizationfactor / (Ql[i] * Ql[j]);
+                }
+                m_Ql_ij[bond] = bond_Ql_ij.real();
```
vyasr, line 64: Does the complex component not mean anything? I don't think it will be zero since you are taking products over the various Qlms, but I don't remember the math in the paper well enough to be sure offhand.

tests/test_order_SolidLiquid.py
```diff
@@ -0,0 +1,79 @@
+import numpy.testing as npt
+import freud
+import unittest
+import util
+
+# Validated against manual calculation and pyboo
+PERFECT_FCC_Q6 = 0.57452416
```
vyasr: Remove this.

tests/test_order_SolidLiquid.py
```diff
@@ -0,0 +1,79 @@
...
+        comp.cluster_sizes
+        comp.cluster_idx
+        comp.num_connections
+        comp._repr_png_()
+
+    def test_repr(self):
+        comp = freud.order.SolidLiquid(6, Q_threshold=.7, S_threshold=6)
+        self.assertEqual(str(comp), str(eval(repr(comp))))
+
+
```
vyasr, line 74: While I know this isn't really the point of this code (since it's supposed to be insensitive to what type of structure you have), there are known characteristic values for FCC that were in the paper, right? Could we test against that? Also this would be a good system to add a test against a known simulation output by just storing a final frame in a numpy array or so as a testing dataset.

vyasr (changes requested):
Thanks for the hard work! This is a huge improvement over the old code. Just a little more to be done.

cpp/order/SolidLiquid.cc
bdice: I'm just creating a NeighborList here if one wasn't provided. I think this is a reasonable choice for standard behavior.

cpp/order/SolidLiquid.cc
bdice: Good idea, could be helpful, but not something I want in this PR and not a blocker for 2.0.

freud/order.pyx
bdice: The paper with reference plots isn't normalized properly, but I added a second citation with a robust description of the method and recommendations on how to use its parameters effectively.

cpp/order/SolidLiquid.cc
bdice: The imaginary part is generally not zero, but all the equations I found specify to take the real part so I did the same here.

tests/test_order_SolidLiquid.py
bdice: I am not exposing the Ql(i, j) to Python so I can't test this. At one point I was, but I don't think it helps the user API because it's a per-bond quantity. Then we would have to expose a NeighborList to make that quantity usable, etc.

cpp/order/SolidLiquid.cc
vyasr: Right, there are some methods where we always want to use NeighborLists. That's another general feature we should add but probably isn't critical to 2.0. I'll make an issue for it.

cpp/order/SolidLiquid.h
vyasr: @bdice is this resolved?

tests/test_order_SolidLiquid.py
vyasr: That's fair. This raises a different point though; is there no concept of per-particle value? The determination of whether a bond is solid checks whether each bond is solid like using the S threshold. Is the per-particle determination a useful quantity? It seems like it would be, although it's effectively given by belonging to a cluster unless there is a singleton cluster (which wouldn't get caught).

freud/order.pyx
```diff
@@ -607,351 +605,180 @@ cdef class Steinhardt(PairCompute):
...
+
+    Next, the bonds are filtered to keep only "solid-like" bonds with
+    :math:`Q_l(i, j)` above a cutoff value :math:`Q_{threshold}`.
+
+    If a particle has more than :math:`S_{threshold}` solid-like bonds, then
+    the particle is considered solid-like. Finally, solid-like particles are
+    clustered.
+
+    .. moduleauthor:: Bradley Dice <bdice@bradleydice.com>
     .. moduleauthor:: Richmond Newman <newmanrs@umich.edu>
```
vyasr: I don't know if you want to remove these here or in a future PR (in case it conflicts with the changes you made in #461), but it should be moved in one or the other.

cpp/order/SolidLiquid.cc
vyasr: So I guess `conj` is being imported into this namespace? I wonder what's doing it, maybe `std::complex`, but I guess nothing else to be done about it.

vyasr (approved):
There are still a couple minor open questions, but I'm approving. Feel free to merge when you feel you've addressed them suitably.

freud/order.pyx
bdice: I'll fix it in whichever PR is merged second.

bdice: Fixing here.

tests/test_order_SolidLiquid.py
bdice: The per-particle values are Qlmi (which is just from Steinhardt) and the number of solid-like bonds (which is used to filter the bonds to only include (i, j) iff both i and j have S or more solid-like bonds).

bdice: I test the number of solid-like bonds in the `num_connections` property, and the `Qlmi` values aren't exposed. We're covered well enough here.

PR #458, opened 2019-09-21
Fix/iterator terminator

vyasr:
## Description
Explicitly specify the const terminator in NeighborQuery.

## Motivation and Context
For some reason having one constant reference the other leads to inconsistent failures. I can't be 100% certain this was the problem, but prior to this fix if I recompiled the code multiple times and reran the tests I would experience failures roughly every other time. With this change I've been able to recompile and run the code tens of times with no failures.

To be more specific, the C++ code seemed to always work fine, the iterators were correctly identifying neighbors and returning them. However, the corresponding Python iterator seemed to have problems comparing to the sentinel properly, even though printing out the terminator typically showed the expected sentinel values (see [here](https://github.com/glotzerlab/freud/blob/master/freud/locality.pyx#L173)). @joaander any thoughts on what could cause this problem? The fact that I cannot reproduce this on my laptop, my desktop, or the linux workstations suggests that this is a very environment-specific compiler bug, but just want to get a second opinion before calling it done

## How Has This Been Tested?
Existing tests now pass consistently.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/NeighborQuery.cc
+++ b/cpp/locality/NeighborQuery.cc
@@ -5,7 +5,7 @@
 
 namespace freud { namespace locality {
 
-const NeighborBond NeighborQueryIterator::ITERATOR_TERMINATOR(NeighborPerPointIterator::ITERATOR_TERMINATOR);
+const NeighborBond NeighborQueryIterator::ITERATOR_TERMINATOR(-1, -1, 0);
 
 const QueryArgs::QueryType QueryArgs::DEFAULT_MODE(QueryArgs::none);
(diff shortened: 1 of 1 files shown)


bdice:
@vyasr FYI, I fixed the error that caused Appveyor to fail on `next`.

joaander (approved):
This is not a compiler bug. The fix is required. See:
* https://stackoverflow.com/questions/40107792/c-static-on-static-member-variable-dependent-initialization-with-int-vs-struc
* https://isocpp.org/wiki/faq/ctors#static-init-order

bdice:
@joaander Great links, thanks for the explanation!

PR #459, opened 2019-09-24
Feature/histogram compute

vyasr:
## Description
I removed the FloatCF class and converted ComplexCF into a generic CorrelationFunction class that tracks whether any complex inputs were provided, and if so, converts the output to real.

I created a parent compute class for all histogram calculations and applied it to RDF, CorrelationFunction, all PMFTs, and BondOrder.

I also removed various properties on the Python side that are now redundant.

## Motivation and Context
Substantially reduces code duplication and makes it much easier to create new histogram-based computations. Also makes API standardization much easier on the Python side now that there's a single way of providing histogram quantities.

This PR is still in draft until we can decide what the APIs for all of these should look like.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:


cpp/density/CorrelationFunction.cc
```diff
@@ -29,77 +29,51 @@ CorrelationFunction<T>::CorrelationFunction(float r_max, float dr)
...
 {
-    // zero the bin counts for totaling
-    m_local_rdf_array.reset();
-    m_local_bin_counts.reset();
-    // reset the frame counter
-    m_frame_counter = 0;
-    m_reduce = true;
+    BondHistogramCompute::reset();
+
+    // zero the rdf as well
```
bdice: Do you mean Correlation Function instead of RDF?

cpp/environment/BondOrder.cc
```diff
@@ -67,57 +68,34 @@ BondOrder::BondOrder(unsigned int n_bins_theta, unsigned int n_bins_phi)
...
-        }
-    }
+    m_histogram.reduceOverThreadsPerBin(m_local_histograms, [&] (size_t i) {
+            m_bo_array[i] = m_histogram[i] / m_sa_array[i] / static_cast<float>(m_frame_counter);
+        });
 }
 
 const util::ManagedArray<float> &BondOrder::getBondOrder()
 {
     if (m_reduce == true)
```
bdice: Is there a "reduce and return" to be used here?

cpp/environment/BondOrder.cc
```diff
@@ -182,29 +154,8 @@ void BondOrder::accumulate(
         // NOTE that the below has replaced the commented out expression for phi.
         float phi = acos(v.z / sqrt(v.x * v.x + v.y * v.y + v.z * v.z)); // 0..Pi
```
bdice: Please replace with `sqrt(dot(v, v))`

cpp/environment/BondOrder.h
```diff
@@ -82,20 +75,15 @@ class BondOrder
...
     float m_dp;
     unsigned int m_n_bins_theta;  //!< number of bins for theta
     unsigned int m_n_bins_phi;    //!< number of bins for phi
-    unsigned int m_frame_counter; //!< number of frames calculated
-    bool m_reduce;                //!< Whether arrays need to be reduced across threads
 
-    util::ManagedArray<unsigned int> m_bin_counts; //!< bin counts computed
     util::ManagedArray<float> m_bo_array;          //!< bond order array computed
     util::ManagedArray<float> m_sa_array;          //!< surface area array computed
     util::ManagedArray<float> m_theta_array;       //!< theta array computed
```
bdice: Can these be removed and replaced with results from the axes?

cpp/locality/BondHistogramCompute.h
```diff
@@ -0,0 +1,133 @@
...
+    //! Reduce thread-local arrays onto the primary data arrays.
+    virtual void reduce() = 0;
+
+    //! Get the simulation box
+    const box::Box& getBox() const
+    {
+        return m_box;
+    }
+
+    //! Return :code:`thing_to_return` after reducing.
```
bdice: Remove Sphinx style from this comment.

cpp/locality/BondHistogramCompute.h
```diff
@@ -0,0 +1,133 @@
...
+        \param nlist Neighbor List. If not NULL, loop over it. Otherwise, use neighbor_query
+           appropriately with given qargs.
+        \param qargs Query arguments
+        \param cf An object with operator(NeighborBond) as input.
+    */
+    template<typename Func>
+    void accumulateGeneral(const locality::NeighborQuery* neighbor_query, 
+                           const vec3<float>* query_points, unsigned int n_query_points,
+                           const locality::NeighborList* nlist,
+                           freud::locality::QueryArgs qargs,
```
bdice: We're very inconsistent about namespaces throughout freud -- lots of places specify `freud::locality::something`. Above, it just uses `locality::NeighborQuery`, but that's very uncommon in the source. This is just a reminder in case we decide that standardization is important.

cpp/pmft/PMFTXY2D.cc
```diff
@@ -32,11 +32,11 @@ PMFTXY2D::PMFTXY2D(float x_max, float y_max, unsigned int n_x, unsigned int n_y)
     m_pcf_array.prepare({n_x, n_y});
 
     // Construct the Histogram object that will be used to keep track of counts of bond distances found.
-    util::Histogram::Axes axes;
+    BHAxes axes;
```
bdice, line 35: Does some of this belong in the parent PMFT class? I see it's repeated.

cpp/util/Histogram.h
```diff
@@ -15,6 +15,30 @@
...
+template <typename T>
+struct Weight
+{
+    Weight() : value(1), is_default(true) {}
+    Weight(T value) : value(value), is_default(false) {}
+
+    Weight &operator=(Weight other)
+    {
+        if (!is_default)
+            throw std::runtime_error("Weight can only be assigned to once.");
```
bdice:
```suggestion
            throw std::runtime_error("Weight can only be assigned once.");
```

cpp/util/Histogram.h
```diff
@@ -133,11 +157,7 @@ class RegularAxis : public Axis
 #else
         unsigned int bin = (unsigned int)(val);
 #endif
-        // Avoid rounding leading to overflow.
```
bdice, line 136: So is this rounding issue no longer a problem?

vyasr:
This PR also resolves #385.

freud/density.pyx
```diff
@@ -358,19 +157,15 @@ cdef class ComplexCF(SpatialHistogram):
...
-    def box(self):
-        return freud.box.BoxFromCPP(self.thisptr.getBox())
+        return output if self.is_complex else np.real(output)
 
     @Compute._reset
     def reset(self):
-        R"""Resets the values of the correlation function histogram in
-        memory.
-        """
+        """Overrides parent since resetting here requires additional logic."""
```
bdice: This is going to show as a docstring but should just be a comment.

bdice (approved):
Overall this looks great. Some comments below.

cpp/locality/BondHistogramCompute.h
vyasr: Namespaces are one area where I'm not concerned with standardization tbh. I don't think there's necessarily a clear best practice or need to refer to the least qualified namespace. In general I try to use only the required amount of qualification to reduce verbosity, but I don't think it's something that's worth putting effort to standardize.

cpp/pmft/PMFTXY2D.cc
vyasr: I'm not sure exactly which lines you're referring to, but the construction of the histogram has to be specific to a given PMFT based on the provided coordinates.

cpp/util/Histogram.h
vyasr: I've never actually hit it, yeah. The existing check has actually been checking slightly incorrectly anyway, so I would imagine we would have hit this bug in practice if it was an issue. With ManagedArray it should be easy to diagnose if it actually ever comes up, though.

Issue #456, opened 2019-09-20
Normalization of Steinhardt Order Parameter Wl


yzjin:
**Is your feature request related to a problem? Please describe.**
I was trying to verify the computation of wl computed by freud with the values given from [this link]. (https://github.com/ArvDee/Minkowski-structure-metrics-calculator/tree/master/testLattices)
The current computation of Steinhart Wl Order parameter in ```freud.local.Steinhardt``` doesn't enable it to be normalized by the value over all particles.

**Describe the solution you'd like**
Add an option for Wl normalization.

bdice:
This is the formula to implement:
![image](https://user-images.githubusercontent.com/3943761/65439666-8cc5a580-ddf5-11e9-8af1-0e1daea30e13.png)
Citation: https://aip.scitation.org/doi/abs/10.1063/1.1638740

vyasr:
Resolved by #460.

PR #460, opened 2019-09-24
Feature/wl normalization

yzjin:
## Description
This PR allows Steinhardt order parameter Wl to be normalized. The formula used to implement, and the issue it resolves can be found here: #456.

## How Has This Been Tested?
it passes all existing tests. 
Wl computed by freud differs with values given from [this link](https://github.com/ArvDee/Minkowski-structure-metrics-calculator/tree/master/testLattices) in average Ql for hcp structure, normalized Wl for l =2 , and average Wl for bcc, fcc, hcp, sc structures. 

## Types of changes

## Checklist:


tests/validation/test_Minkowski_Structure_Metrics.py
```diff
@@ -0,0 +1,29 @@
...
+import garnett
+import freud
+
+
+structures = ['fcc', 'bcc', 'hcp', 'sc']
+
+
+@pytest.mark.parametrize("filename", structures)
+def test_minkowski_structure_metric_ql(filename):
+    expected_ql = np.genfromtxt('files/{}_q.txt'.format(filename))
```
vyasr: Use `os.path.join(__file__, 'files/{}_q.txt'...)` so that the test works no matter where it's run from.

Also do this below in the `garnett.read` line.

tests/validation/test_Minkowski_Structure_Metrics.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import pytest
+from pytest import approx
+import garnett
```
vyasr, line 3: I prefer to just use `gsd` directly. This is partly just personal preference, but also because frankly I trust its longevity as a package more (it's also nice but almost certainly irrelevant for these small files that using the package would let you use the C reader rather than the python reader). I'm sure @bdice differs on this, so I'll leave you two to decide what to do, but in either case the package should be added to the `tests_require` in setup.py.

tests/validation/test_Minkowski_Structure_Metrics.py
```diff
@@ -0,0 +1,29 @@
...
+        voro = freud.locality.Voronoi()
+        voro.compute(box, positions)
+        for sph_l in range(expected_ql.shape[1]):
+            comp = freud.order.Steinhardt(sph_l, weighted=True)
+            comp.compute(box, positions, nlist=voro.nlist)
+            assert comp.order == approx(expected_ql[:, sph_l],
+                                        rel=5e-5, abs=1e-3)
+
+
+if __name__ == '__main__':
```
vyasr, line 73: I assume we still need to add corresponding tests for Wl, the averaged OPs, and the normalized?

tests/validation/test_Minkowski_Structure_Metrics.py
```diff
@@ -0,0 +1,29 @@
...
+        frame = traj[0]
+        box = frame.box
+        positions = frame.positions
+        voro = freud.locality.Voronoi()
+        voro.compute(box, positions)
+        for sph_l in range(expected_ql.shape[1]):
+            comp = freud.order.Steinhardt(sph_l, weighted=True)
+            comp.compute(box, positions, nlist=voro.nlist)
+            assert comp.order == approx(expected_ql[:, sph_l],
+                                        rel=5e-5, abs=1e-3)
```
vyasr: This test should still use either `unittest` or `np.testing` functions (in this case probably `np.testing.assertAllclose` or so). `pytest` is nice, but probably not the appropriate solution here.

freud/order.pyx
```diff
@@ -485,6 +485,9 @@ cdef class Steinhardt(PairCompute):
             spherical harmonics over neighbors. If enabled and used with a
             Voronoi neighbor list, this results in the Minkowski Structure
             Metrics :math:`Q'_l`. (Default value = :code:`False`)
+        Wl_normalize (bool, optional):
+            Determines whether to normalize the :math:`W_l` version
+            of the Steinhardt order parameter. (Default value = :code:`False`)
```
vyasr, line 454: We still need to finalize the API for this. I suggested converting Ql, Wl, and Wl_normalize into mode arguments, but no final decisions were made. Thoughts? The orthogonality of the arguments should be preserved if at all possible.

cpp/order/Steinhardt.h
```diff
@@ -152,12 +160,13 @@ class Steinhardt
                                   const freud::locality::NeighborQuery* points, freud::locality::QueryArgs qargs);
 
     //! Normalize the order parameter
-    float normalize();
+    float normalizeSystem();
```
vyasr, line 177: Can you add to this docstring? Ideally we'd want to see a reference to a paper here discuss what the normalization does. It would be hard to evaluate the quality of the name without knowing exactly what the normalization is doing, for someone unfamiliar with the method or reading this code for the first time.

vyasr (changes requested):
Looks good! In addition to the inline comments, I have a couple of requests:
1. All of the text output files should be condensed into a npz archive. Unless we plan to add comments to the text files with more information (which seems very unnecessary since it's pretty obvious what they're doing) I don't want to bloat the repo with lots of validation files.
2. The validation folder should have an extra `minkowski` subdirectory so that we can separate these files from anything that gets added in the future as validation for other methods.
3. We need to add a reference for this normalization method. I wasn't able to verify correctness because I don't know exactly what the formula is that's being implemented. I can tell roughly what's going on, but I don't know if it's the expected behavior or not.

freud/order.pyx
yzjin: The implementation of this is much closer to a boolean flag, in terms of how "normalization" affects the math.

cpp/order/Steinhardt.h
yzjin: This is a separate issue, unrelated to Wl normalization. @bdice will make an issue describing this.

freud/order.pyx
bdice: @vyasr Do you still think mode arguments are the best approach here, or should we just fail if `Wl is False and Wl_normalize is True`?

cpp/order/Steinhardt.h
bdice: I added a more descriptive comment instead of filing an issue. It's somewhat better. I'm not sure what references I would cite here but I'll keep it in mind if I see a paper defining that quantity.

freud/order.pyx
vyasr: To be honest I'm not sure. I find both versions a little bit awkward, but I don't see a good way around it. If this is the only holdup , then I say we can stick with what you have.

requirements-testing.txt
```diff
@@ -0,0 +1,10 @@
+coverage==4.5.4
+cython==0.29.13
+garnett==0.5.0
+GitPython==3.0.2
```
vyasr, line 4: We don't actually need GitPython for anything, do we?

vyasr:
I'd still prefer to have all the validation data stored in a NumPy archive or something similar, unless there's a strong benefit to being able to actually see the values (which I kind of doubt since the absolute values of these quantities are only known based on actually calculating for a particular reference). Other than that, this PR looks mostly fine now aside from one comment I made regarding the requirements-testing file.

requirements-testing.txt
vyasr: @bdice indicated that he intends to use this file for CI purposes. In anticipation of performing benchmarks on CI, we will keep GitPython in this file for now.

vyasr:
In the interest of not letting things get stale, I'm going to merge this PR. Since there is future work anticipated on testing parts of this method, we can finalize cleanup at a later date.

vyasr (approved):

Issue #463, opened 2019-09-25
Simpler looping over NeighborLists


vyasr:
**Is your feature request related to a problem? Please describe.**
Certain internal methods are not reductions over bonds, but actual per-bond calculations that also store outputs on a per-bond basis. Currently there's some friction with writing this code because compute classes have to keep track of bond ids and loop indices.

**Describe the solution you'd like**
It should be simple to generalize the existing NeighborList iterator in NeighborComputeFunctional to also provide a bond id, so the lambda function would have signature `(size_t bond_idx, NeighborBond nb)`.

vyasr:
As part of this, we would also want a standard way of making the NeighborList from NeighborQuery objects if no NeighborList is provided.

vyasr:
The standard function was implemented in #472 as `makeDefaultNlist` in `NeighborComputeFunctional.cc`.

vyasr:
As of now there aren't enough use-cases for the generic looping over a `NeighborList` to be necessary. The three classes using it are `LocalDescriptors`, `LocalBondProjection`, and `AngularSeparationNeighbor`, but `LocalDescriptors` works a bit differently because it needs to loop per point and then for each bond per point, whereas the other two are truly per-bond calculations. If future uses come up for this, we can reopen, but for now it's not a common enough calculation to be worth generalizing.

PR #472, opened 2019-09-27
Refactor/neighbors arg

### LINKED ISSUE #466: Standardize query argument usage in compute methods
All neighbor-based compute methods in freud need to accept query arguments in addition to NeighborList objects.

vyasr:
## Description
Standardize the usage of a `neighbors` argument everywhere. This argument can be either a dictionary or a `NeighborList` (or `None` if a class has default query arguments), and this object is resolved by standard preprocessing in the `PairCompute` class.

Note that I have intentionally chosen _not_ to update the docstrings. #348 will help in providing a standard place to provide documentation, which all these docstrings need to point to. I prefer to do this all at once so that we can put some thought into the appropriate docstring.

## Motivation and Context
This resolves #466 and the second part of #463. Also contributes to #266 and #179. 

## How Has This Been Tested?
Existing tests pass after being adapted to use the new API.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/environment/AngularSeparation.cc
+++ b/cpp/environment/AngularSeparation.cc
@@ -3,6 +3,7 @@
 
 #include "AngularSeparation.h"
 #include "utils.h"
+#include "NeighborComputeFunctional.h"
 
 #if defined _WIN32
 #undef min // std::min clashes with a Windows header
@@ -55,25 +56,28 @@ float computeMinSeparationAngle(const quat<float> ref_q, const quat<float> q, co
     return min_angle;
 }
 
-void AngularSeparationNeighbor::compute(const quat<float>* orientations, unsigned int n_points,
-                         const quat<float>* query_orientations, unsigned int n_query_points,
-                         const quat<float>* equiv_orientations, unsigned int n_equiv_orientations,
-                         const freud::locality::NeighborList* nlist)
+void AngularSeparationNeighbor::compute(const locality::NeighborQuery *nq, const quat<float>* orientations,
+                 const vec3<float>* query_points,
+                 const quat<float>* query_orientations, unsigned int n_query_points,
+                 const quat<float>* equiv_orientations, unsigned int n_equiv_orientations,
+                 const freud::locality::NeighborList* nlist, locality::QueryArgs qargs)
 {
-    nlist->validate(n_query_points, n_points);
+    // This function requires a NeighborList object, so we always make one and store it locally.
+    locality::NeighborList m_nlist = locality::makeDefaultNlist(nq, nlist, query_points, n_query_points, qargs);
+    m_nlist.validate(n_query_points, nq->getNPoints());
 
-    const size_t tot_num_neigh = nlist->getNumBonds();
+    const size_t tot_num_neigh = m_nlist.getNumBonds();
     m_angles.prepare(tot_num_neigh);
 
-    util::forLoopWrapper(0, n_points, [=](size_t begin, size_t end) {
-        size_t bond(nlist->find_first_index(begin));
+    util::forLoopWrapper(0, nq->getNPoints(), [=](size_t begin, size_t end) {
+        size_t bond(m_nlist.find_first_index(begin));
         for (size_t i = begin; i < end; ++i)
         {
             quat<float> q = orientations[i];
 
-            for (; bond < tot_num_neigh && nlist->getNeighbors()(bond, 0) == i; ++bond)
+            for (; bond < tot_num_neigh && m_nlist.getNeighbors()(bond, 0) == i; ++bond)
             {
-                const size_t j(nlist->getNeighbors()(bond, 1));
+                const size_t j(m_nlist.getNeighbors()(bond, 1));
                 quat<float> query_q = query_orientations[j];
 
                 float theta = computeMinSeparationAngle(q, query_q, equiv_orientations, n_equiv_orientations);
--- a/cpp/environment/AngularSeparation.cc
+++ b/cpp/environment/AngularSeparation.cc
@@ -63,7 +63,7 @@ void AngularSeparationNeighbor::compute(const locality::NeighborQuery *nq, const
                  const freud::locality::NeighborList* nlist, locality::QueryArgs qargs)
 {
     // This function requires a NeighborList object, so we always make one and store it locally.
-    locality::NeighborList m_nlist = locality::makeDefaultNlist(nq, nlist, query_points, n_query_points, qargs);
+    m_nlist = locality::makeDefaultNlist(nq, nlist, query_points, n_query_points, qargs);
     m_nlist.validate(n_query_points, nq->getNPoints());
 
     const size_t tot_num_neigh = m_nlist.getNumBonds();
--- a/cpp/environment/AngularSeparation.cc
+++ b/cpp/environment/AngularSeparation.cc
@@ -15,10 +15,6 @@
 
 namespace freud { namespace environment {
 
-AngularSeparation::AngularSeparation() {}
-
-AngularSeparation::~AngularSeparation() {}
-
 float computeSeparationAngle(const quat<float> ref_q, const quat<float> q)
 {
     quat<float> R = q * conj(ref_q);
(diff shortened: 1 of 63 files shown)


cpp/environment/AngularSeparation.cc
```diff
@@ -42,7 +39,7 @@ float computeMinSeparationAngle(const quat<float> ref_q, const quat<float> q, co
     float min_angle = computeSeparationAngle(ref_q, q);
 
     // loop through all equivalent rotations and see if they have smaller angles with ref_q
-    for (unsigned int i = 0; i < n_equiv_quats; i++)
+    for (unsigned int i = 1; i < n_equiv_quats; ++i)
```
bdice: Is this change from 0 to 1 fixing a bug? I'm not 100% sure how this is initialized.

bdice: Also is there a reason to prefer `++i` over `i++`?

cpp/environment/AngularSeparation.h
```diff
@@ -5,57 +5,83 @@
...
+                 const quat<float>* orientations, unsigned int n_points,
+                 const quat<float>* equiv_orientations, unsigned int n_equiv_orientations);
+
+    //! Returns the last computed global angle array
+    const util::ManagedArray<float> &getAngles() const
+    {
+        return m_angles;
+    }
+
+protected:
```
bdice: I think this should be private for consistency with other classes, unless there's a benefit to `protected` that I'm not aware of.
```suggestion
private:
```

cpp/environment/LocalBondProjection.cc
```diff
@@ -52,32 +52,34 @@ float computeMaxProjection(const vec3<float> proj_vec, const vec3<float> local_b
...
     const vec3<float>* query_points, unsigned int n_query_points,
+    const vec3<float>* proj_vecs,  unsigned int n_proj,
     const quat<float>* equiv_orientations, unsigned int n_equiv_orientations,
-    const freud::locality::NeighborList* nlist)
+    const freud::locality::NeighborList* nlist, locality::QueryArgs qargs)
 {
-    nlist->validate(n_query_points, n_points);
+    // This function requires a NeighborList object, so we always make one and store it locally.
+    m_nlist = locality::makeDefaultNlist(nq, nlist, query_points, n_query_points, qargs);
+    m_nlist.validate(n_query_points, nq->getNPoints());
```
bdice: Might as well stick this validation call into `makeDefaultNlist`?

freud/_box.pxd
```diff
@@ -35,13 +35,17 @@ cdef extern from "Box.h" namespace "freud::box":
...
+        void makeCoordinates(vec3[float]*, unsigned int) const
+        void makeFraction(vec3[float]*, unsigned int) const
+        void getImage(vec3[float]*, unsigned int, vec3[int]*) const
+        # Note that getLatticeVector is a const function, but due to Cython
+        # parsing limitations we cannot have it both be const and pass the
+        # exception back to Cython so we choose to capture the exception since
+        # constness is less important on the Cython side.
         vec3[float] getLatticeVector(unsigned int i) except +
-        void wrap(vec3[float]* vs, unsigned int Nv) except +
+        void wrap(vec3[float]* vs, unsigned int Nv) const
```
bdice: We don't include variable names in these declarations.
```suggestion
        void wrap(vec3[float]*, unsigned int) const
```

freud/_box.pxd
```diff
@@ -35,13 +35,17 @@ cdef extern from "Box.h" namespace "freud::box":
...
-        void makeFraction(vec3[float]*, unsigned int) except +
-        void getImage(vec3[float]*, unsigned int, vec3[int]*) except +
+        void makeCoordinates(vec3[float]*, unsigned int) const
+        void makeFraction(vec3[float]*, unsigned int) const
+        void getImage(vec3[float]*, unsigned int, vec3[int]*) const
+        # Note that getLatticeVector is a const function, but due to Cython
+        # parsing limitations we cannot have it both be const and pass the
+        # exception back to Cython so we choose to capture the exception since
+        # constness is less important on the Cython side.
         vec3[float] getLatticeVector(unsigned int i) except +
```
bdice: We don't include variable names in these declarations.
```suggestion
        vec3[float] getLatticeVector(unsigned int) except +
```

freud/environment.pyx
```diff
@@ -133,8 +131,7 @@ cdef class BondOrder(SpatialHistogram):
 
     @Compute._compute()
     def accumulate(self, box, points, orientations, query_points=None,
-                   query_orientations=None, str mode="bod", nlist=None,
-                   query_args=None):
+                   query_orientations=None, str mode="bod", neighbors=None):
```
bdice: Should `mode` be a constructor argument? Probably. It wouldn't make sense to accumulate multiple frames with different modes.

freud/environment.pyx
```diff
@@ -300,30 +288,29 @@ cdef class LocalDescriptors(Compute):
...
 
     def __dealloc__(self):
         del self.thisptr
 
     @Compute._compute()
-    def compute(self, box, unsigned int num_neighbors, points,
-                query_points=None,
-                orientations=None, mode='neighborhood', nlist=None):
+    def compute(self, box, points,
+                query_points=None, orientations=None, mode='neighborhood',
```
bdice: Same here -- should `mode` be a constructor argument?

freud/environment.pyx
```diff
@@ -385,32 +362,29 @@ cdef class LocalDescriptors(Compute):
...
+            nq.get_ptr(),
+            <vec3[float]*> &l_query_points[0, 0], num_query_points,
             l_orientations_ptr, l_mode,
-            nlist_.get_ptr())
+            nlistptr.get_ptr(), dereference(qargs.thisptr))
         return self
 
+    @Compute._computed_property()
+    def nlist(self):
+        return freud.locality.nlist_from_cnlist(self.thisptr.getNList())
```
bdice, line 382: This method is very similar to the `BoxFromCPP` method. Would be nice to standardize the naming of these somehow. Or just see if they can be class methods of the `Box` and `NeighborList` classes?

freud/locality.pyx
```diff
@@ -560,6 +550,25 @@ cdef class NeighborList:
         return self
 
 
+cdef NeighborList nlist_from_cnlist(freud._locality.NeighborList *c_nlist):
+    """Create a Python NeighborList object that points to an existing C++
+    NeighborList object.
+
+    This functions generally serves two purposes. Any special locality
+    NeighborList generators, like :class:`~.Voronoi`, should use this as a way
+    to point to the C++ Neighborlist they generate internally. Additionally,
```
```suggestion
    to point to the C++ NeighborList they generate internally. Additionally,
```

tests/test_environment_AngularSeparation.py
```diff
@@ -1,61 +1,74 @@
...
 import numpy as np
 import freud
-import rowan
 import unittest
 from util import make_box_and_random_points
 
 
 class TestAngularSeparation(unittest.TestCase):
-    def test_getN(self):
+    def test_getN_global(self):
```
bdice: This should probably be `TestAngularSeparationGlobal.test_getN` instead of `TestAngularSeparation.test_getN_global`.

bdice (changes requested):

cpp/environment/AngularSeparation.cc
vyasr: No it isn't fixing a bug. Thanks for the check. I originally thought it was unnecessary to check 0, but then I realized that it actually does need to get checked and I just forgot to change it back.

cpp/environment/AngularSeparation.h
vyasr: There's not really a reason to prefer one over the other for classes with no children. I changed it because I was originally going to have a common parent then realized there was no logic in common. I can change it back to be consistent.

cpp/environment/LocalBondProjection.cc
vyasr: Tbh I'm not sure we even need to validate, it's really designed for `NeighborLists` that are constructed elsewhere and need to be verified. But I'll put it in there for now.

freud/environment.pyx
vyasr: I'm fine with something like that. I'd prefer doing that in a separate PR since we need to do a lot of renaming.

bdice (approved):

PR #486, opened 2019-09-30
Use voro++ to compute Voronoi diagrams.

bdice:
**This PR replaces #362, which I rewrote to use a git submodule.** PR #362 was too far out of sync with `next` to salvage.

## Description
This PR adds the voro++ library ([homepage](http://math.lbl.gov/voro++/), [GitHub repo](https://github.com/chr1shr/voro)) written by Chris Rycroft (@chr1shr) to freud. This will replace the existing Voronoi methods, which used less sophisticated algorithms and were quite slow.

## Motivation and Context
Issue #305 describes challenges with Voronoi performance -- this continues to be a bottleneck for my current analyses, so I found it worthwhile to rewrite this entire feature using voro++.

## How Has This Been Tested?
I have adapted existing Voronoi tests, which work as expected in both 2D and 3D systems. Some changes (like reordering of vertices in a Voronoi cell) are expected, so I added functions to sort the vertices in a deterministic way for the purpose of testing. Currently, this new approach is about 32x faster for a system of 5000 random points in a cubic box of side length 10.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/Voronoi.cc
+++ b/cpp/locality/Voronoi.cc
@@ -1,195 +1,170 @@
 // Copyright (c) 2010-2019 The Regents of the University of Michigan
 // This file is from the freud project, released under the BSD 3-Clause License.
 
-#include <stdexcept>
 #include <tbb/tbb.h>
 #include <cmath>
 #include <vector>
 
 #include "Voronoi.h"
-
-#if defined _WIN32
-#undef min // std::min clashes with a Windows header
-#undef max // std::max clashes with a Windows header
-#endif
+#include "NeighborBond.h"
 
 /*! \file Voronoi.cc
-    \brief Build a cell list from a set of points.
+    \brief Computes Voronoi neighbors for a set of points.
 */
 
 namespace freud { namespace locality {
 
-// Default constructor
-Voronoi::Voronoi()
-{}
-
-// A compare function used to sort NeighborBonds
-bool compareNeighborPairs(const NeighborBond &n1, const NeighborBond &n2)
-{
-    return n1.less_id_ref_weight(n2);
-}
-
-typedef tbb::enumerable_thread_specific< std::vector<NeighborBond> > BondVector;
-
-void add_valid_bonds(BondVector::reference local_bonds,
-    unsigned int i, unsigned int expanded_i,
-    unsigned int j, unsigned int expanded_j,
-    unsigned int N, float weight, float distance)
-{
-    // Make sure we only add bonds with real particles as the reference
-    if (i < N && distance != 0)
+// Voronoi calculations should be kept in double precision.
+void Voronoi::compute(const box::Box &box, const vec3<double>* points, unsigned int n_points)
     {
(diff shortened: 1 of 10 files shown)


vyasr:
Good call on making a new PR. This approach also has the ancillary approach of avoiding having our history cluttered with all the Voro++ files. I'll aim to review later today.

cpp/locality/Voronoi.cc
```diff
@@ -1,195 +1,170 @@
...
+            boxLatticeVectors[2] = box.getLatticeVector(2);
+        }
+        voro::container_periodic container(
+            boxLatticeVectors[0].x,
+            boxLatticeVectors[1].x,
+            boxLatticeVectors[1].y,
+            boxLatticeVectors[2].x,
+            boxLatticeVectors[2].y,
+            boxLatticeVectors[2].z,
+            3, 3, 3, 3
```
vyasr, line 42: Why do we have 3 blocks in each direction? And what is the logic for the choice of the initial memory allocation?

cpp/locality/Voronoi.cc
```diff
@@ -1,195 +1,170 @@
...
+        );
+
+        for (size_t query_point_id = 0; query_point_id < n_points; query_point_id++) {
+            container.put(query_point_id, points[query_point_id].x, points[query_point_id].y, points[query_point_id].z);
+        }
```
vyasr, line 47: I'm personally inclined to avoid using the `query_point` terminology at all in this code, since Voronoi constructions don't really make sense for two distinct sets of points. I could be convinced otherwise, though, if you think it's a helpful mental model. Thoughts?

vyasr: I also don't think it's a good enough reason to hold up this PR. So if nothing else major crops up I'm not going to bother asking for a change no matter where we come down.

tests/test_locality_Voronoi.py
```diff
@@ -5,46 +5,81 @@
 import util
 
 
-@util.skipIfMissing('scipy.spatial')
+def sort_rounded_xyz_array(arr, decimals=4):
```
vyasr, line 8: Add a comment on the reason for the sort order.

vyasr (approved):
This looks great! I have a couple of questions I'd like answered before you merge, and one comment I'd like added to the tests, but I'm approving contingent on that happening.

cpp/locality/Voronoi.cc
bdice: I'm just trying to keep it consistent. I don't know what else I would call it without being _more_ strange, since we've standardized terms elsewhere in the codebase.

cpp/locality/Voronoi.cc
bdice: I added a comment about this marked as `TODO`.

tests/test_locality_Voronoi.py
bdice: Done.

Issue #420, opened 2019-09-02
RawPoints shouldn't make an AABBQuery unless needed by querySingle


bdice:
An internally-held AABBQuery should not be made on construction of a RawPoints object, only if `querySingle` is called.

Maybe initialize `aq` as a null pointer. When `querySingle` is called, then construct the AABBQuery before querying if needed. Being parallel-safe may be tricky.

_Originally posted by @bdice in https://github.com/glotzerlab/freud/pull/419_

vyasr:
Resolved by #489.

PR #489, opened 2019-09-30
Refactor/update locality matchenv apis

### LINKED ISSUE #278: Environment Matching helper function
As suggested by @bvansade, it could be useful to create a helper function that assists users in selecting parameters for `MatchEnv`. In my experience, it's rather difficult to find settings that work well without a deep understanding of the module and its algorithms, so any sort of helpful heuristic would be nice. Perhaps this would work by scanning a range of settings and doing some kind of optimization.

@erteich @jglaser @xiyudu I'd be happy to hear your thoughts on this!

vyasr:
## Description
This PR has a few unrelated changes in it that unfortunately sort of cascaded.
- The LinkCell constructor argument order has been swapped, and `cell_width` is now optional and computed for the user if none is provided. I'm open to ideas on improving the calculation, but I would prefer to push that to a separate PR (presumably after the 2.0 beta release) since it's not API-related.
- Changes `RawPoints` to create its `AABBQuery` only when needed. The key is to do it in `query` rather than `querySingle` so that there are no parallelism issues. It will error out if a user manually calls `querySingle` rather than having it called through the `query->NeighborQueryIterator::query->querySingle` chain, but I think that's OK since that's not currently part of the Python API anyway.
- Simplifies the `Compute` class by removing many of the decorators and inferring that information based on function names. This change is made possible by further standardization of the API.
- That standardization requires that `MatchEnv` be brought into conformance with the rest of freud. The largest part of this PR is a significant rewrite of `MatchEnv` that splits it into 3 separate classes that are each responsible for a single calculation. I also did a lot of cleanup to make the classes a bit easier to understand.

Remaining to-dos:
- The `getEnvironment` function should be removed in favor of just having the cluster environments all just exposed through a `ManagedArray` or similar interface.
- I recommend making everything other than `EnvironmentCluster` private on the Python side for now, until we hear from more users about whether or not they actually make use of `minRMSDMotif` 
 and `matchMotif`.
- These classes should be switched to using `NeighborQuery`. It won't be too large a change if we still just use `makeDefaultNList` on the C++ side.
- I need to add tests for the two new compute classes. I have some basic ones written, but not in the form of `unittest` class methods yet.

## Motivation and Context

Resolves: #420. I think this also resolves #278. Splitting into three classes clarifies the use of each calculation, there's no longer any confusion of which attributes are populated or what they mean (and how to use them), and the only free parameter left is the threshold, for which I added documentation along the lines of what @erteich suggested in #278. The threshold will always be a free variable, so I think some guidance regarding how to pick it is the most we can do.

It's also a significant step towards making it easier to actually understand the code in MatchEnv.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/environment/MatchEnv.cc
+++ b/cpp/environment/MatchEnv.cc
@@ -603,7 +603,7 @@ void MatchEnv::cluster(const freud::locality::NeighborList* env_nlist,
 
     // done looping over points. All clusters are now determined. Renumber
     // them from zero to num_clusters-1.
-    populateEnv(dj, true);
+    populateEnv(dj, m_env_index, m_env, m_tot_env, true);
 }
 
 void MatchEnv::matchMotif(const freud::locality::NeighborList* nlist, const vec3<float>* points,
@@ -675,7 +675,7 @@ void MatchEnv::matchMotif(const freud::locality::NeighborList* nlist, const vec3
     // DON'T renumber the clusters in the disjoint set from zero to
     // num_clusters-1. The way I have set it up here, the "0th" cluster
     // is the one that matches the motif.
-    populateEnv(dj, false);
+    populateEnv(dj, m_env_index, m_env, m_tot_env, false);
 }
 
 std::vector<float> MatchEnv::minRMSDMotif(const freud::locality::NeighborList* nlist,
@@ -754,12 +754,12 @@ std::vector<float> MatchEnv::minRMSDMotif(const freud::locality::NeighborList* n
     // DON'T renumber the clusters in the disjoint set from zero to
     // num_clusters-1. The way I have set it up here, the "0th" cluster
     // is the one that matches the motif.
-    populateEnv(dj, false);
+    populateEnv(dj, m_env_index, m_env, m_tot_env, false);
 
     return min_rmsd_vec;
 }
 
-void MatchEnv::populateEnv(EnvDisjointSet dj, bool reLabel)
+void MatchEnv::populateEnv(EnvDisjointSet dj, util::ManagedArray<unsigned int> &env_index, std::map<unsigned int, std::vector<vec3<float> > > &env, util::ManagedArray<vec3<float> > &tot_env, bool reLabel)
 {
     std::map<unsigned int, unsigned int> label_map;
 
@@ -783,7 +783,7 @@ void MatchEnv::populateEnv(EnvDisjointSet dj, bool reLabel)
                 label_map[c] = cur_set;
                 std::vector<vec3<float>> vecs = dj.getAvgEnv(c);
                 label_ind = reLabel ? label_map[c] : c;
-                m_env[label_ind] = vecs;
+                env[label_ind] = vecs;
                 cur_set++;
             }
             else
@@ -792,10 +792,10 @@ void MatchEnv::populateEnv(EnvDisjointSet dj, bool reLabel)
             }
 
             // label this particle in m_env_index
-            m_env_index[particle_ind] = label_ind;
+            env_index[particle_ind] = label_ind;
             for (unsigned int m = 0; m < part_vecs.size(); m++)
             {
-                m_tot_env(particle_ind, m) = part_vecs[m];
+                tot_env(particle_ind, m) = part_vecs[m];
             }
             particle_ind++;
         }
--- a/cpp/environment/MatchEnv.cc
+++ b/cpp/environment/MatchEnv.cc
@@ -485,7 +485,6 @@ std::map<unsigned int, unsigned int> minimizeRMSD(const box::Box &box, const vec
 /************
  * MatchEnv *
  ************/
-
 MatchEnv::MatchEnv(const box::Box& box, float r_max, unsigned int num_neighbors) : m_box(box), m_r_max(r_max), m_num_neighbors(num_neighbors)
 {
     m_max_num_neighbors = 0;
@@ -667,9 +666,9 @@ unsigned int EnvironmentCluster::populateEnv(EnvDisjointSet dj)
     return cur_set;
 }
 
-/**********************
- * EnvironmentCluster *
- **********************/
+/*************************
+ * EnvironmentMotifMatch *
+ *************************/
 void EnvironmentMotifMatch::compute(const freud::locality::NeighborList* nlist, const vec3<float>* points,
                           unsigned int Np, const vec3<float>* motif, unsigned int motif_size, float threshold,
                           bool registration)
--- a/cpp/environment/MatchEnv.cc
+++ b/cpp/environment/MatchEnv.cc
@@ -269,11 +269,10 @@ std::vector<vec3<float>> EnvDisjointSet::getIndividualEnv(const unsigned int m)
  * Convenience functions *
  *************************/
 std::pair<rotmat3<float>, BiMap<unsigned int, unsigned int>>
-isSimilar(Environment& e1, Environment& e2, float r_max, float threshold_sq, bool registration)
+isSimilar(Environment& e1, Environment& e2, float threshold_sq, bool registration)
 {
     BiMap<unsigned int, unsigned int> vec_map;
     rotmat3<float> rotation = rotmat3<float>(); // this initializes to the identity matrix
-    float r_max_sq = r_max*r_max;
 
     // If the vector sets do not have equal numbers of vectors, just return
     // an empty map since the 1-1 bimapping will be too weird in this case.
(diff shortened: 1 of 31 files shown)


cpp/environment/MatchEnv.cc
```diff
@@ -614,45 +610,88 @@ void MatchEnv::cluster(const freud::locality::NeighborList* env_nlist,
...
+            particle_ind++;
+        }
+    }
+
+    // specify the number of cluster environments
+    return cur_set;
+}
+
+/**********************
+ * EnvironmentCluster *
```
```suggestion
 * EnvironmentMotifMatch *
```

cpp/locality/LinkCell.cc
```diff
@@ -213,9 +213,23 @@ LinkCell::LinkCell()
...
+    // Want n_points/num_cells = 10
+    if (cell_width == 0)
+    {
+        // This number is arbitrary because there is no way to determine an
+        // appropriate cell density for an arbitrary triclinic box.
+        const unsigned int num_particle_per_cell = 10;
+        vec3<float> L = box.getNearestPlaneDistance();
+        unsigned int desired_num_cells = std::max(n_points/num_particle_per_cell,
+                                                  static_cast<unsigned int>(1));
+        m_cell_width = std::cbrtf(float(L.x*L.y*L.z)/desired_num_cells);
```
bdice: Don't you just want the volume? Also this probably fails for 2D where `L.z == 0`?
```suggestion
        m_cell_width = std::cbrtf(box.getVolume()/desired_num_cells);
```

cpp/locality/LinkCell.h
```diff
@@ -184,11 +184,8 @@ class LinkCell : public NeighborQuery
     //! Null Constructor
     LinkCell();
 
-    //! Old constructor
-    LinkCell(const box::Box& box, float cell_width);
-
     //! New constructor
```
bdice:
```suggestion
    //! Constructor
```

cpp/locality/RawPoints.h
```diff
@@ -28,19 +30,45 @@ class RawPoints : public NeighborQuery
...
+    }
+
     // dummy implementation for pure virtual function in the parent class
     virtual std::shared_ptr<NeighborQueryPerPointIterator> querySingle(const vec3<float> query_point, unsigned int query_point_idx,
                                                          QueryArgs qargs) const
     {
-        return aq.querySingle(query_point, query_point_idx, qargs);
+        if (!aq)
+        {
+            throw std::runtime_error("The underlying AAABBQuery object has not yet been initialized. This likely developer failure.");
```
bdice:
```suggestion
            throw std::runtime_error("The underlying AABBQuery object has not yet been initialized. This is likely developer failure.");
```

freud/common.pyx
```diff
@@ -42,91 +39,41 @@ cdef class Compute:
     """
 
     def __cinit__(self):
-        self._called_compute = {"compute": False}
-
-    def _set_compute_flag(self, flag_name):
-        self._called_compute[flag_name] = False
+        self._called_compute = False
+
+    def __getattribute__(self, attr):
```
bdice, line 45: I'm not sure if I like removing the decorators and relying on "magic names," since it's less transparent for new developers. But I don't have strong feelings on it.

freud/common.pyx
```diff
@@ -24,15 +24,12 @@ cdef class Compute:
     .. code-block:: python
         class Cluster(Compute):
 
-            @Compute._compute("computeClusters")
-            def computeClusters(...)
+            def compute(...)
                 ...
 
-            @Compute._computed_property("computeClusters")
```
bdice, line 31: I'm not sure if you meant to remove this line. I think it should just have no argument?

freud/environment.pyx
```diff
@@ -373,17 +369,17 @@ cdef class LocalDescriptors(PairCompute):
             nlist.get_ptr(), dereference(qargs.thisptr))
         return self
 
-    @Compute._computed_property()
+    @Compute._computed_property
     def nlist(self):
-        return freud.locality.nlist_from_cnlist(self.thisptr.getNList())
+        return freud.locality._nlist_from_cnlist(self.thisptr.getNList())
```
bdice, line 374: Should this be a static method / class method of `NeighborList`? Similarly for `BoxFromCPP` and `freud.box.Box`.

freud/environment.pyx
```diff
@@ -409,9 +405,104 @@ cdef class LocalDescriptors(PairCompute):
                     negative_m=self.negative_m, mode=self.mode)
 
 
-cdef class MatchEnv(Compute):
-    R"""Clusters particles according to whether their local environments match
-    or not, according to various shape matching metrics.
+def minimizeRMSD(box, ref_points, points, registration=False):
+    R"""Get the somewhat-optimal RMSD between the set of vectors ref_points
```
bdice, line 409: Don't say "somewhat-optimal" -- it's at least a local minimum, right?

cpp/locality/LinkCell.cc
vyasr: Good call.

freud/common.pyx
vyasr: I had the same thought for a bit. But I do like that it provides a tiny bit of extra enforcement for our API standards. It also significantly simplifies things given that aside from properties all the other methods are very standard (`compute`, `reset`, and `plot` are well-defined standards in the context of the new freud API IMO).

freud/environment.pyx
vyasr: Yes, I think that makes sense. However, since neither of these methods is part of the public API I just made an issue, we can wait on this for now.

freud/environment.pyx
vyasr: Yes, it finds a local minimum in some region of the landscape, but there's some random number generation in there, so running it multiple times can give you different results. So it seems pretty inaccurate to remove the extra qualifications.

freud/common.pyx
bdice: Good enough justification -- resolving this.

freud/environment.pyx
```diff
@@ -409,353 +405,404 @@ cdef class LocalDescriptors(PairCompute):
                     negative_m=self.negative_m, mode=self.mode)
 
 
-cdef class MatchEnv(Compute):
-    R"""Clusters particles according to whether their local environments match
-    or not, according to various shape matching metrics.
+def _minimizeRMSD(box, ref_points, points, registration=False):
```
bdice, line 408:
```suggestion
def _minimize_RMSD(box, ref_points, points, registration=False):
```

freud/environment.pyx
```diff
@@ -409,353 +405,404 @@ cdef class LocalDescriptors(PairCompute):
...
+    cdef map[unsigned int, unsigned int] results_map = \
+        freud._environment.minimizeRMSD(
+            dereference(b.thisptr),
+            <vec3[float]*> &l_ref_points[0, 0],
+            <vec3[float]*> &l_points[0, 0],
+            nRef1, min_rmsd, registration)
+    return [min_rmsd, np.asarray(l_points), results_map]
+
+
+def _isSimilar(box, ref_points, points, threshold, registration=False):
```
bdice, line 454:
```suggestion
def _is_similar_motif(box, ref_points, points, threshold, registration=False):
```

freud/environment.pyx
```diff
@@ -409,353 +405,404 @@ cdef class LocalDescriptors(PairCompute):
...
+
+    cdef const float[:, ::1] l_ref_points = ref_points
+    cdef const float[:, ::1] l_points = points
+    cdef unsigned int nRef1 = l_ref_points.shape[0]
+    cdef unsigned int nRef2 = l_points.shape[0]
+    cdef float threshold_sq = threshold*threshold
+
+    if nRef1 != nRef2:
+        raise ValueError(
+            ("The number of vectors in ref_points must MATCH"
```
bdice, line 495: Easy there, cowboy. 🤠 
```suggestion
            ("The number of vectors in ref_points must match"
```

freud/environment.pyx
```diff
@@ -409,353 +405,404 @@ cdef class LocalDescriptors(PairCompute):
...
             registration, global_search)
         return self
 
-    @Compute._compute()
-    def matchMotif(self, points, ref_points, threshold, registration=False,
-                   nlist=None):
-        R"""Determine clusters of particles that match the motif provided by
-        ref_points.
+    @Compute._computed_property
+    def clusters(self):
```
bdice, line 611: For consistency with other clustering methods?
```suggestion
    def cluster_idx(self):
```

bdice (approved):
Approving. Let's merge and put change requests into a new PR.

freud/environment.pyx
vyasr: :)

PR #494, opened 2019-10-01
Use NeighborQuery in more classes

bdice:
## Description
I updated:
- ClusterProperties
- GaussianDensity
- Voronoi

I could not update:
- PeriodicBuffer (Takes box in constructor. Not sure if we still desire that behavior.)
- Interface (Needs two disjoint sets of points, and has no clear use for QueryArgs)
- MSD (Takes box in constructor. Also uses "positions" instead of "points.")

## Motivation and Context
We're moving to using `neighbor_query` or `(box, points)` instead of `box, points` everywhere.

## How Has This Been Tested?
Updated tests to use new syntax.

## Types of changes

## Checklist:


vyasr:
As discussed offline, changes to `PeriodicBuffer` and `InterfaceMeasure` will occur in a separate PR. Since MSD is a fundamentally different calculation than others in freud (it's a time-series rather than a spatial analysis), we will leave its API as-is.

cpp/cluster/ClusterProperties.cc
```diff
@@ -62,7 +61,7 @@ void ClusterProperties::compute(const box::Box& box, const vec3<float>* points,
         // are done, we can add the computed center to ref_pos to get the
         // center in the space frame.
         vec3<float> delta = pos - ref_pos[c];
```
vyasr: This seems like a place to use `utils::bondVector`.

cpp/cluster/ClusterProperties.cc
```diff
@@ -76,16 +75,16 @@ void ClusterProperties::compute(const box::Box& box, const vec3<float>* points,
...
     // Now that we have determined the centers of mass for each cluster, tally
     // up the gyration tensor. This has to be done in a loop over the points.
-    for (unsigned int i = 0; i < Np; i++)
+    for (unsigned int i = 0; i < nq->getNPoints(); i++)
     {
         unsigned int c = cluster_idx[i];
-        vec3<float> pos = points[i];
-        vec3<float> delta = box.wrap(pos - m_cluster_centers[c]);
+        vec3<float> pos = (*nq)[i];
+        vec3<float> delta = nq->getBox().wrap(pos - m_cluster_centers[c]);
```
vyasr, line 86: This is the bondVector calculation, but may be a little odd since you'll have to pass in the `cluster_centers` where you'd normally put query_points. May also want to think about proper directionality here; shouldn't it point from the center out to the point? That would affect the bondVector call

vyasr (approved):
Couple small requests regarding vector directionality in ClusterProperties. Otherwise all set, can merge whenever those changes are done.

PR #518, opened 2019-10-11
Fix/issue500

### LINKED ISSUE #500: Histograms don't preserve ManagedArray data
**Describe the bug**
All histogram classes will not have the expected behavior if users access the property in Python and do not copy it.

**To Reproduce**
1. Make an RDF and compute it
2. Save the data: `saved_RDF = rdf.RDF`
3. Recompute the RDF for a different set of points
4. See that `saved_RDF` has changed.

**Additional context**
This problem is closely related to what was originally reported in #151 and solved in #370. The problem specifically arises because of the features enabled in #438 and #459, which created standard histogram classes and ways to perform computations on histograms. The underlying problem is that while the thread-local `m_local_histograms` variable is reset when `reset` is called and otherwise accumulates over time, what the user sees is a reflection of the `m_histogram` object. This object needs a way to copy it's underlying data into a new `ManagedArray` when accumulate is called in case the old version is still being referred to. The solution would be to add a `copy` flag to `ManagedArray::prepare` that would copy the data to the new array if a new array needed to be allocated.

vyasr:
## Description
I made the histograms use `prepare` for the underlying array so that it would be preserved between calls.

## Motivation and Context

Resolves: #500 

## How Has This Been Tested?

I extracted the test in Cluster that was a ManagedArray test and then added that test to all the histograms.

## Types of changes

## Checklist:


bdice (approved):

bdice:
@vyasr Just a couple minor fixes for style/tests, otherwise this is fine.

PR #523, opened 2019-10-12
Fix/issue198

### LINKED ISSUE #198: PMFT inconsistency
The PMFTXY2D class and the PMFTXYT classes should produce identical results when the number of theta bins is set to 1 for the latter class, but this does not currently happen. Similar cross-checking with e.g. PMFTXYZ may also be in order.

vyasr:
## Description
I added tests to check that the different coordinate systems don't affect the result of the PMFT. I also changed the Jacobians of PMFTXY and PMFTXYZ to add in multiplicative prefactors corresponding to the Jacobian of the orientational coordinates that are implicitly integrated out by these calculations.

## Motivation and Context

Resolves: #198. 

After some discussion with @joaander, we decided that adding the prefactors of 2pi to PMFTXY and 8pi^2 to PMFTXYZ was the right approach to ensure consistency across all the classes.

## How Has This Been Tested?

Added new tests, which pass but are based on the current behavior.

## Types of changes

## Checklist:


tests/test_pmft.py
```diff
@@ -1001,5 +1001,66 @@ def compute(self):
...
+    def test_XY_XYZ(self):
+        """Check that 2D and 3D PMFTs give the same results."""
+        x_max = 2.5
+        y_max = 2.5
+        z_max = 1
+        nbins = 4
+        num_points = 100
+        L = 10
+
+        box2 = freud.box.Box.square(L)
```
bdice:
```suggestion
        box2d = freud.box.Box.square(L)
```

tests/test_pmft.py
```diff
@@ -1001,5 +1001,66 @@ def compute(self):
...
+        box3 = freud.box.Box.cube(L)
```
bdice:
```suggestion
        box3d = freud.box.Box.cube(L)
```

bdice: ...And similar below.

bdice (approved):
Tests look fine - just rename `2` to `2d` and `3` to `3d`. I'd also like to hear from @jamesaan or @harperic about the normalization we should expect from XY vs. XY with theta.

Issue #446, opened 2019-09-18
hard to tell where import util is supposed to come from


jennyfothergill:
**Describe the bug**
Following along with this demo
https://freud.readthedocs.io/en/stable/examples/module_intros/Density-LocalDensity.html
It has `import util`, and it's easy to think this means import freud.util. Anyway, it took me a bit to figure out that util is from https://github.com/glotzerlab/freud-examples

**To Reproduce**
n/a

**Error output**
n/a
**System configuration (please complete the following information):**
n/a
**Additional context**
Explaining where util is from or providing setup for the examples would help.

jennyfothergill:
OK, the main examples page https://freud.readthedocs.io/en/stable/examples.html gives this information, but it might be easier for the user if all the information was on the page they're on.

bdice:
@jennyfothergill Thanks for filing an issue! It's really helpful to know what is confusing for new users, so please keep us in the loop as you explore the library and its features. We've heard this feedback from others as well, so we're planning to remedy this in the near future. We're preparing for a 2.0 release, and will be updating our examples.

bdice:
Related issue: #390.

bdice:
Most functions have been added to `freud.data` or `freud.plot` as of PR #532. The corresponding PR to freud-examples includes other changes that should help address this. https://github.com/glotzerlab/freud-examples/pull/9

Issue #390, opened 2019-08-19
Add common utility functions into freud.util module


bdice:
**Is your feature request related to a problem? Please describe.**
Many new users are confused by the use of a "util" module in the freud examples.

**Describe the solution you'd like**
There are a number of utility methods used for testing and examples that should be incorporated into the main package, in `freud.util`. This includes generation of common systems, e.g. fcc, bcc, and random points. Some other pieces of plotting-related code currently found in the examples would fit into `freud.plot`.

bdice:
Most functions have been added to `freud.data` or `freud.plot` as of PR #532. The corresponding PR to freud-examples includes other changes that should help address this. https://github.com/glotzerlab/freud-examples/pull/9

PR #524, opened 2019-10-13
Feature/add data

vyasr:
## Description
Added a new data module that provides convenience functions for generating data (specifically various common cubic lattices). These are currently present and used in both tests and examples.

## Motivation and Context

This partially addresses #390 and #446. I'm open to moving more of the sample system generation functions in `tests/util.py` into this as well, specifically the `make_alternating_lattice` and `make_box_and_random_points` functions, but I also don't see any urgency to add those unless and until they're used in examples. I'm less certain about what to do with the other functions in `examples/util.py`. I definitely to not want to move any of those plotting functions into freud proper, except maybe `box_2d_to_points`, which would be a reasonable fit for `plot.py` for the purposes of allowing boxes to have a natural representation.

We could either switch from using `cubeellipse` to one of the cyclic colormaps in `matplotlib`, or we can start using `plato` in our tests. I'm more inclined to the former even if it doesn't look quite as good since maintaining `plato` as a dependency on the examples could be painful. One option for the other functions would either be to simplify our other plotting codes across the examples. It might also be sufficient to just rename `util` to something like `examples_util` to disambiguate from anything in `freud` (to address the confusion in #446).

## How Has This Been Tested?

Tests pass.

## Types of changes

## Checklist:


freud/data.py
```diff
@@ -0,0 +1,138 @@
...
+            1.0
+        noise:
+            Apply Gaussian noise with this width to particle positions (Default
+            value = 0.0)
+
+    Returns:
+        tuple (py:class:`freud.box.Box`, :class:`np.ndarray`):
+            freud Box, particle positions, shape=(nx*ny*nz, 3)
+    """
+    box, positions = make_cubic(nx, ny, 1, fractions, scale, noise)
```
bdice: Subtle thing: please verify that this is drawing from the appropriate distribution when considering noise. If it's drawing from a 3D gaussian and then setting z=0, I'm not sure if that's the same as drawing from a 2D gaussian to begin with.

tests/test_environment_LocalDescriptors.py
```diff
@@ -281,13 +284,13 @@ def test_ql_weighted(self):
                 npt.assert_allclose(
                     steinhardt.particle_order,
                     ql[:, L],
-                    atol=1e-3 if struct_func == make_sc else 1e-6,
+                    atol=1e-3 if struct_func == freud.data.make_sc else 1e-6,
                     err_msg="Failed for {}, L = {}".format(
                         struct_func.__name__, L))
 
     @unittest.skipIf(sys.version_info < (3, 2),
```
bdice: Remove this "skip" -- we dropped support for everything below this (2.7).

bdice (changes requested):

freud/data.py
bdice: Overall we may need to verify if we're drawing x, y, z independently or not. We probably need to use the `multivariate_normal` function in both places: https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.random.multivariate_normal.html

vyasr: Assuming that the different dimensions are independent (i.e. assuming that the covariance matrix is diagonal), it is equivalent to sample them like this versus using the multivariate_normal function. The marginalization over each dimension trivially reduces in this case. However, there's no harm in using the multivariate_normal method.

vyasr:
@bdice thoughts on my new solution? After our discussion I thought about it a bit more, and I think defining this new UnitCell class is a cleaner approach than either our old solution or the `make_structure` function that you have. We can add other structures as classmethods as needed.

I'll update tests, etc, if you like this version.

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+    unit cell as well as enabling the generation of systems of points
+    (optionally with some noise) from the unit cell.
+
+    Args:
+        box:
+            A box-like object (see :meth:`~freud.box.Box.from_box`) containing
+            the lattice vectors of the unit cell.
+        basis_positions ((:math:`N_{points}`) :class:`numpy.ndarray`):
+            The basis of the unit cell in fractional coordinates
+            (Default value = ``[[0, 0, 0]]``.
```
bdice:
```suggestion
            (Default value = ``[[0, 0, 0]]``).
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+    This class represents the unit cell of a crystal structure, which is
+    defined by a lattice and a basis. It provides the basic attributes of the
+    unit cell as well as enabling the generation of systems of points
+    (optionally with some noise) from the unit cell.
+
+    Args:
+        box:
+            A box-like object (see :meth:`~freud.box.Box.from_box`) containing
+            the lattice vectors of the unit cell.
+        basis_positions ((:math:`N_{points}`) :class:`numpy.ndarray`):
```
bdice:
```suggestion
        basis_positions ((:math:`N_{points}`, 3) :class:`numpy.ndarray`):
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+            The basis of the unit cell in fractional coordinates
+            (Default value = ``[[0, 0, 0]]``.
+    """
+    def __init__(self, box, basis_positions=[[0, 0, 0]]):
+        self._box = freud.box.Box.from_box(box)
+        self._basis_positions = np.dot(basis_positions, self.box.to_matrix().T)
```
bdice: This is misleading to accept an argument in fractional coordinates and return a property in real coordinates. Also, this should definitely use the box's `make_fractional` and `make_absolute` methods.

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def to_system(self, num_replicas=1, scale=1, sigma_noise=0):
+        """Generate a system from the unit cell.
+
+        Args:
+            num_replicas (:class:`tuple` or `int`):
```
bdice:
```suggestion
            num_replicas (:class:`tuple` or int):
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+            The basis of the unit cell in fractional coordinates
+            (Default value = ``[[0, 0, 0]]``.
+    """
+    def __init__(self, box, basis_positions=[[0, 0, 0]]):
+        self._box = freud.box.Box.from_box(box)
+        self._basis_positions = np.dot(basis_positions, self.box.to_matrix().T)
+
+    def to_system(self, num_replicas=1, scale=1, sigma_noise=0):
+        """Generate a system from the unit cell.
+
```
bdice, line 42: Needs a bit of a description of how the scaling / replication / noise addition works.

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+                             "2D unit cell.")
+
+        basis = np.tile(
+            self.basis_positions[np.newaxis, np.newaxis, np.newaxis, ...],
+            (nx, ny, nz, 1, 1))
+        basis[..., 0] += np.arange(nx)[:, np.newaxis, np.newaxis, np.newaxis]
+        basis[..., 1] += np.arange(ny)[np.newaxis, :, np.newaxis, np.newaxis]
+        basis[..., 2] += np.arange(nz)[np.newaxis, np.newaxis, :, np.newaxis]
+
+        positions = (basis - 0.5).reshape(-1, 3).dot(
```
bdice: Use `make_absolute`!

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+    def to_system(self, num_replicas=1, scale=1, sigma_noise=0):
+        """Generate a system from the unit cell.
+
+        Args:
+            num_replicas (:class:`tuple` or `int`):
+                If provided as a single number, the number of replicas in all
+                dimensions. If a tuple, the number of times replicated in each
+                dimension. Must be of the form (nx, ny, 1) for 2D boxes
+                (Default value = 1).
+            scale (float): Amount to scale the unit cell by (in distance units)
```
bdice: The multiplicative factor is unitless, not in distance units...
```suggestion
            scale (float): Factor by which to scale the unit cell
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+        box = freud.box.Box.from_matrix(
+            self.box.to_matrix() * scale * [[nx, ny, nz]])
+
+        if sigma_noise != 0:
+            mean = [0]*3
+            var = sigma_noise*sigma_noise
+            cov = np.diag([var, var, var if self.dimensions == 3 else 0])
+            positions += np.random.multivariate_normal(
+                mean, cov, size=positions.shape[:-1])
+
```
bdice, line 103: This needs a final step: `positions = box.wrap(positions)` to account for noise pushing particles over the boundaries.

freud/data.py
bdice: e.g.
```
The basis positions are transformed into absolute coordinates for every replica requested.
The box and positions are expanded by a factor of ``scale`` about the origin.
Finally, noise with standard deviation ``sigma_noise`` is added to the positions and the
positions are wrapped back into the box.
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+    @property
+    def lattice_vectors(self):
+        """:math:`(3, 3)` :class:`np.ndarray`: The matrix of lattice
+        vectors."""
+        return self._box.to_matrix()
+
+    @property
+    def basis_positions(self):
+        """:math:`(N_{points}, 3)` :class:`np.ndarray`: The basis positions."""
```
bdice, line 121:
```suggestion
        """:math:`(N_{points}, 3)` :class:`np.ndarray`: The basis positions in fractional coordinates."""
```
(Or absolute coordinates?)

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+        return self._basis_positions
+
+    @property
+    def a1(self):
```
bdice, line 125: These should be properties of `freud.box.Box`, I think. Perhaps we replace `get_lattice_vector`?

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @property
+    def a3(self):
+        """:math:`(3, )` :class:`np.ndarray`: The third lattice vector."""
+        return self.box.to_matrix()[:, 2]
+
+    @property
+    def dimensions(self):
+        """int: The dimensionality of the unit cell."""
+        return self.box.dimensions
```
bdice, line 142: Is it better to use `self.box` or `self._box`? In `lattice_vectors`, it uses `self._box`. Be consistent.
```suggestion
        return self.box.dimensions
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+    def fcc(cls):
+        """Create an FCC crystal.
+
+        Returns:
+            :class:`~.UnitCell`: An FCC unit cell.
+        """
+        fractions = np.array([[.5, .5, 0],
+                              [.5, 0, .5],
+                              [0, .5, .5],
+                              [0, 0, 0]], dtype=np.float32)
```
bdice: Shouldn't add the dtypes.
```suggestion
                              [0, 0, 0]])
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @classmethod
+    def bcc(cls):
+        """Create an BCC crystal.
+
+        Returns:
+            :class:`~.UnitCell`: An BCC unit cell.
+        """
+        fractions = np.array([[.5, .5, .5],
+                              [0, 0, 0]], dtype=np.float32)
```
bdice:
```suggestion
                              [0, 0, 0]])
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+        return cls([1, 1, 1], fractions)
+
+    @classmethod
+    def sc(cls):
+        """Create an SC crystal.
```
bdice:
```suggestion
        """Create a simple cubic (sc) crystal.
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+        return self.box.to_matrix()[:, 2]
+
+    @property
+    def dimensions(self):
+        """int: The dimensionality of the unit cell."""
+        return self.box.dimensions
+
+    @classmethod
+    def fcc(cls):
+        """Create an FCC crystal.
```
bdice: Note that `fcc` should not be capitalized, according to Wikipedia, some literature, and our favorite crystallographer.
```suggestion
        """Create a face-centered cubic (fcc) crystal.
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+
+        Returns:
+            :class:`~.UnitCell`: An FCC unit cell.
```
bdice:
```suggestion
            :class:`~.UnitCell`: An fcc unit cell.
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        fractions = np.array([[.5, .5, 0],
+                              [.5, 0, .5],
+                              [0, .5, .5],
+                              [0, 0, 0]], dtype=np.float32)
+        return cls([1, 1, 1], fractions)
+
+    @classmethod
+    def bcc(cls):
+        """Create an BCC crystal.
```
bdice:
```suggestion
        """Create a body-centered cubic (bcc) crystal.
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+
+        Returns:
+            :class:`~.UnitCell`: An BCC unit cell.
```
bdice:
```suggestion
            :class:`~.UnitCell`: A bcc unit cell.
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+        return cls([1, 1, 1], fractions)
+
+    @classmethod
+    def sc(cls):
+        """Create an SC crystal.
+
+        Returns:
+            :class:`~.UnitCell`: An SC unit cell.
+        """
+        fractions = np.array([[0, 0, 0]], dtype=np.float32)
```
bdice:
```suggestion
        fractions = np.array([[0, 0, 0]], dtype=np.float32)
```
```suggestion
        fractions = np.array([[0., 0., 0.]])
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+        fractions = np.array([[.5, .5, .5],
+                              [0, 0, 0]], dtype=np.float32)
+        return cls([1, 1, 1], fractions)
+
+    @classmethod
+    def sc(cls):
+        """Create an SC crystal.
+
+        Returns:
+            :class:`~.UnitCell`: An SC unit cell.
```
bdice:
```suggestion
            :class:`~.UnitCell`: An sc unit cell.
```

freud/data.py
```diff
@@ -0,0 +1,163 @@
...
+        return cls([1, 1, 1], fractions)
+
+    @classmethod
+    def square(cls):
+        """Create a square crystal.
+
+        Returns:
+            :class:`~.UnitCell`: A square unit cell.
+        """
+        fractions = np.array([[0, 0, 0]], dtype=np.float32)
```
bdice:
```suggestion
        fractions = np.array([[0., 0., 0.]])
```

bdice (commented):
@vyasr I left some comments on the new class. I think this is useful.

Be aware that `freud.locality.PeriodicBuffer` replicates a lot of the functionality developed in this PR (pun intended). I would prefer to use that class for a lot of the math developed here, with the arguments `buffer=[nx, ny, nz], images=True`. That would simplify a lot of the code in `to_system`:

```python
pbuff = freud.locality.PeriodicBuffer()
pbuff.compute((self.box, self.basis_positions), buffer=(nx, ny, nz), images=True)
system_box = pbuff.buffer_box
system_points = np.concatenate((self.basis_positions, pbuff.buffer_points))  # Check the syntax of this line
system_box *= scale
system_points *= scale
# Add noise
system_points = system_box.wrap(system_points)
return system_box, system_points

**Comment by bdice** [comment]
@vyasr Also be aware that this `UnitCell` class is similar to the internal `crystaldb` code developed by @yannahm. This would be helpful for anyone seeking a similar repository of crystal structures in the future: https://bitbucket.org/glotzer/crystaldb/

File: freud/data.py (hunk 5 above)
Comment (vyasr) [reply freud/data.py]: Agreed on all counts.

File: freud/data.py (hunk 12 above)
Comment (vyasr) [reply freud/data.py:125]: Actually now that I'm thinking about it, I think we should get rid of `box.get_lattice_vector`. If people want it, it's easy enough in Python, just `box.to_matrix()[:, i]`. We could even implement `__getitem__` to index into the box matrix directly if we wanted. At the very least, though, we should rename it to `get_box_vector` so that it's not misleading.

**Review by vyasr — COMMENTED** [review]
I agree, definitely should be using `PeriodicBuffer`. Now that we've agreed that this is worthwhile I can refactor.

Regarding `crystaldb`, it's a good note, but not a problem for this, right? The duplication definitely makes sense given the relative maturity of the packages and the fact that one is not open source. We're also duplicating the logic in `hoomd.lattice` to some extent, but again I think that's reasonable.

**Comment by bdice** [comment]
1. Rename `getLatticeVector` to `getBoxVector` in C++
2. Remove `get_lattice_vector` from Cython and replace with properties `v1`, `v2`, `v3`
3. Fix `a1` to `self.box.v1` for `UnitCell`
4. Document that `v1` is `[Lx, 0, 0]`, etc.

File: freud/data.py (hunk 14 above)
Comment (vyasr) [reply freud/data.py]: Why not add them? Otherwise freud is going to convert them internally anyway, right?

Comment (bdice) [reply freud/data.py]: I don't want to introduce lower precision into the Cython code unless required, in anticipation of moving to double precision and minimizing the total amount of code. The specifier isn't necessary for correctness and I would prefer to allow the internal conversion to do its thing.

File: freud/box.pyx [hunk 24]
```diff
@@ -247,22 +247,37 @@ cdef class Box:
...
-        cdef vec3[float] result = self.thisptr.getLatticeVector(i)
-        if self.is2D:
-            result.z = 0.0
-        return np.asarray([result.x, result.y, result.z])
+        return self.to_matrix()[:, i]
+
+    @property
+    def v1(self):
+        """:math:`(3, )` :class:`np.ndarray`: The first box vector
+        :math:`[Lx, 0, 0]`."""
```
bdice: Please match the notation formats used here: https://hoomd-blue.readthedocs.io/en/stable/box.html
Also render the docs and make sure the right symbols are escaped.
```suggestion
        :math:`(L_x, 0, 0)`."""
```

freud/data.py
```diff
@@ -0,0 +1,178 @@
...
+        The box and the positions are expanded by ``scale``, and then Gaussian
+        noise with standard deviation ``sigma_noise`` is added to the
+        positions.  All points are wrapped back into the box before being
+        returned.
+
+        Args:
+            num_replicas (:class:`tuple` or int):
+                If provided as a single number, the number of replicas in all
+                dimensions. If a tuple, the number of times replicated in each
+                dimension. Must be of the form (nx, ny, 1) for 2D boxes
```
bdice:
```suggestion
                dimension. Must be of the form :code:`(nx, ny, 1)` for 2D boxes
```

freud/data.py
```diff
@@ -0,0 +1,178 @@
...
+                (Default value = 1).
+            scale (float):
+                Factor by which scale the unit cell (Default value = 1).
```
bdice:
```suggestion
                Factor by which to scale the unit cell (Default value = 1).
```

freud/data.py
```diff
@@ -0,0 +1,178 @@
...
+            sigma_noise (float):
+                The standard deviation of the normal distribution used to add
+                noise to the positions in the system (Default value = 0).
+            seed (int):
+                If provided, used to seed the random noise generation. Not used
+                unless ``sigma_noise`` > 0 (Default value = None).
```
bdice: I think we use ``:code:`sigma_noise` `` instead of double backticks everywhere else in freud. Also check how we format `None` elsewhere -- I think we also put that in code font.
```suggestion
                unless :code:`sigma_noise` > 0 (Default value = None).
```

freud/data.py
```diff
@@ -0,0 +1,178 @@
...
+            pbuff.compute((self.box, self.basis_positions),
+                          buffer=(nx-1, ny-1, nz-1),
+                          images=True)
+            box = pbuff.buffer_box*scale
+            positions = np.concatenate((self.basis_positions,
+                                        pbuff.buffer_points))
+        else:
+            box = self.box*scale
+            positions = self.basis_positions.copy()
+        positions *= scale
```
bdice, line 90: Am I missing the part where you convert from fractional to absolute coordinates? I can't find it anywhere.

freud/data.py
```diff
@@ -0,0 +1,178 @@
...
+                The standard deviation of the normal distribution used to add
+                noise to the positions in the system (Default value = 0).
+            seed (int):
+                If provided, used to seed the random noise generation. Not used
+                unless ``sigma_noise`` > 0 (Default value = None).
+
+        Returns:
+            tuple (:class:`freud.box.Box`, :class:`np.ndarray`):
+                A system-like object (see
+                :class:`~freud.locality.NeighborQuery.from_system`.).
```
bdice:
```suggestion
                :class:`~freud.locality.NeighborQuery.from_system`).
```

freud/data.py
```diff
@@ -0,0 +1,178 @@
...
+            mean = [0]*3
+            var = sigma_noise*sigma_noise
+            cov = np.diag([var, var, var if self.dimensions == 3 else 0])
+            positions += np.random.multivariate_normal(
+                mean, cov, size=positions.shape[:-1])
+            if seed is not None:
+                np.random.set_state(random_state)
+
+        positions = box.wrap(positions)
+        return box, positions
```
bdice, line 105: I think the return type should be `RawPoints`. We should accept `(box, points)` as inputs to things but I think it's more appropriate to return a `RawPoints` object than a tuple when we intend to give the user a system-like quantity.

bdice (changes requested):

freud/data.py
vyasr: After discussing, we decided against making this change now. We can revisit at a later point if we'd like to, particularly if we actually create a parent `System` class to encapsulate these objects.

PR #529, opened 2019-10-17
Feature/rdf normalization

### LINKED ISSUE #396: Incorrect RDF normalization
**Describe the bug**
The radial distribution function g(r) that I accumulated for a trajectory does not have the correct limiting value. The system is dilute and so g -> 1 for large r, but freud's result is consistently smaller. I recomputed g using VMD and obtained the behavior I physically expected. I can rescale the freud result onto VMD's by multiplying by a factor of ~1.05, suggesting a possible normalization issue. (This factor could be specific to my system though: 20 particles, cubic box with L = 50, 1000 frames.)

**To Reproduce**
Minimum reproducer script (`gr.py`) and a sample trajectory (`cm.gsd`) are attached below. The script reads the trajectory and accumulates g(r) with `rmax = 20` and `dr = 0.2`. The freud output (`freud_gr.dat`) should match the VMD output (`vmd_gr.dat`), and both should approach 1.

[gr_repro.tar.gz](https://github.com/glotzerlab/freud/files/3530602/gr_repro.tar.gz)

**Error output**
Comparison of freud and VMD output
compare

**System configuration**
 - OS: Ubuntu 16
 - python 3.7.4
 - freud 1.2.2
(I confirmed this on other system configurations too.)

vyasr:
## Description
Enable normalization of RDF to 1 in the large R limit. Also fix a bug in Jupyter plotting code.

## Motivation and Context

Resolves: #396. Also fixes an unrelated bug.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -22,6 +22,7 @@ and this project adheres to
 * Extensive new documentation including tutorial for new users and reference sections on crucial topics.
 * ClusterProperties computes radius of gyration from the gyration tensor for each cluster.
 * `freud.data` module.
+* Optional normalization for RDF, useful for small systems.
 
 ### Changed
 * All compute objects that perform neighbor computations now use NeighborQuery internally.
--- a/cpp/density/RDF.cc
+++ b/cpp/density/RDF.cc
@@ -48,7 +48,11 @@ void RDF::reduce()
     m_N_r.prepare(getAxisSizes()[0]);
 
     // Define prefactors with appropriate types to simplify and speed later code.
-    float number_density = (m_normalize ? float(m_n_points) : float(m_n_query_points)) / m_box.getVolume();
+    float number_density = float(m_n_query_points) / m_box.getVolume();
+    if (m_normalize)
+    {
+        number_density *= static_cast<float>(m_n_query_points-1)/(m_n_query_points);
+    }
     float np = static_cast<float>(m_n_points);
     float prefactor = float(1.0)/(np*number_density*m_frame_counter);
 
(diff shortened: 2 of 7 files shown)


vyasr:
@joaander see #396 for the relevant discussion, I wanted to get your input before finalizing this.

mphoward (approved):
Thanks for fixing! This looks like it should work to me.

joaander (approved):
Looks good!

[Levine, Stone, and Kohlmeyer](http://www-s.ks.uiuc.edu/Research/gpu/files/rdfgpu/rdfgpu.pdf) have a clean way of expressing g(r) normalization with arbitrary particle selections if you would like to generalize this or use their notation in the documentation.

vyasr:
Thanks for the paper! I think we could implement that generally here, it would just require maintaining a counter inside the loop in compute to determine the total number of pairs that we computed for and then normalizing for that case. However, it seems like a pretty esoteric use-case, so I don't think it's a priority and I'm happy to punt on it for now. @bdice thoughts? If you agree with me, then go ahead and merge this PR when you're happy and we can reconsider that at a future date.

mphoward:
Josh makes a good point. It actually probably isn't too hard to treat this normalization exactly here because freud does not have such a strong sense of "selections" / "groups" or "tags" as other programs. The number of pairs in the normalization is `N*(N-1)` iff `query_points is neighbors` AND `exclude_ii is True`. Otherwise, the normalization is for `N*N` pairs, some of which may be excluded. The `normalize` flag would activate this pair-counting logic (and could include a link to that paper); otherwise, it would fall back on the old logic and always use the `N*N` pairs.

bdice (approved):
Looks good!

PR #532, opened 2019-10-20
Feature/system plotting

bdice:
## Description
Added plotting functions for `Box` and `NeighborQuery` objects.

## Motivation and Context
This feature is super helpful for freud examples. Also, by using the `c` or `s` arguments, the points can be colored or sized according to the results of some analysis.

## How Has This Been Tested?
I tested locally.

## Screenshots

image

image

(data credit: [drawdata.xyz](https://drawdata.xyz/))

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/Voronoi.cc
+++ b/cpp/locality/Voronoi.cc
@@ -113,9 +113,10 @@ void Voronoi::compute(const freud::locality::NeighborQuery* nq)
 
                 // Save polytope vertices in system coordinates
                 std::vector<vec3<double> > system_vertices;
+                vec3<double> query_point_system_coords((*nq)[query_point_id]);
                 for (auto vertex_iter = relative_vertices.begin(); vertex_iter != relative_vertices.end(); vertex_iter++)
                 {
-                    system_vertices.push_back((*vertex_iter) + query_point);
+                    system_vertices.push_back((*vertex_iter) + query_point_system_coords);
                 }
                 m_polytopes[query_point_id] = system_vertices;
 
--- a/freud/box.pyx
+++ b/freud/box.pyx
@@ -189,7 +189,7 @@ cdef class Box:
             :math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`:
                 Absolute coordinate vector(s).
         """  # noqa: E501
-        fractions = np.asarray(fractional_coordinates)
+        fractions = np.asarray(fractional_coordinates).copy()
         flatten = fractions.ndim == 1
         fractions = np.atleast_2d(fractions)
         fractions = freud.util._convert_array(fractions, shape=(None, 3))
@@ -211,7 +211,7 @@ cdef class Box:
             :math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`:
                 Fractional coordinate vector(s).
         """  # noqa: E501
-        vecs = np.asarray(absolute_coordinates)
+        vecs = np.asarray(absolute_coordinates).copy()
         flatten = vecs.ndim == 1
         vecs = np.atleast_2d(vecs)
         vecs = freud.util._convert_array(vecs, shape=(None, 3))
--- a/freud/box.pyx
+++ b/freud/box.pyx
@@ -323,8 +323,11 @@ cdef class Box:
         vecs = np.asarray(vecs)
         flatten = vecs.ndim == 1
         vecs = np.atleast_2d(vecs)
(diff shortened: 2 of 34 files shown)


vyasr:
@bdice let me know when you think you've made all the changes necessary to fix the bugs that you've found and I'll review.

cpp/locality/Voronoi.cc
```diff
@@ -113,9 +113,10 @@ void Voronoi::compute(const freud::locality::NeighborQuery* nq)
 
                 // Save polytope vertices in system coordinates
                 std::vector<vec3<double> > system_vertices;
+                vec3<double> query_point_system_coords((*nq)[query_point_id]);
                 for (auto vertex_iter = relative_vertices.begin(); vertex_iter != relative_vertices.end(); vertex_iter++)
                 {
-                    system_vertices.push_back((*vertex_iter) + query_point);
+                    system_vertices.push_back((*vertex_iter) + query_point_system_coords);
```
vyasr, line 119: Is the issue here that `cell.vertices` returns the coordinates in a different coordinate system?

freud/cluster.pyx
```diff
@@ -140,15 +140,14 @@ cdef class Cluster(_PairCompute):
...
+            values, counts, num_clusters_to_plot=10, ax=ax)
 
     def _repr_png_(self):
-        import freud.plot
         try:
-            return freud.plot.ax_to_bytes(self.plot())
-        except AttributeError:
+            import freud.plot
+            return freud.plot._ax_to_bytes(self.plot())
+        except (AttributeError, ImportError):
```
vyasr, line 150: Why do we need this? I don't think it's necessary to check that freud.plot imports successfully, but if it is for some reason necessary, we should be doing this everywhere.

freud/box.pyx
```diff
@@ -462,6 +465,28 @@ cdef class Box:
...
+                (Default value = :code:`None`).
+            image (list):
+                The periodic image location at which to draw the box (Default
+                value = :code:`[0, 0, 0]`).
+            ``*args``, ``**kwargs``:
+                All other arguments are passed on to
+                :meth:`mpl_toolkits.mplot3d.Axes3D.plot` or
+                :meth:`matplotlib.axes.Axes.plot`.
+        """
+        import freud.plot
```
vyasr, line 486: Here's one example where we don't check for an `ImportError`.

tests/test_box_Box.py
```diff
@@ -142,6 +142,22 @@ def test_unwrap(self):
...
+
+        testpoints = [[10, 0, 0], [11, 0, 0]]
+        imgs = [10, 1, 2]
+        npt.assert_allclose(
+            box.unwrap(testpoints, imgs), [[20, 1, 2], [21, 1, 2]], rtol=1e-6)
+
+        # Test broadcasting one vector with multiple images
+        box = freud.box.Box.cube(1)
+
+        testpoints = [[10, 0, 0]]
```
vyasr: For completeness, make this a 1D array so that the `atleast_2D` conversion also gets tested.

freud/plot.py
```diff
@@ -31,6 +30,148 @@ def ax_to_bytes(ax):
...
+            :code:`None`).
+    """
+    # Adapted from https://stackoverflow.com/a/50664367
+
+    if limits is None:
+        limits = np.array([ax.get_xlim3d(), ax.get_ylim3d(), ax.get_zlim3d()])
+    else:
+        limits = np.asarray(limits)
+    origin = np.mean(limits, axis=1)
+    radius = 0.5 * np.max(np.abs(limits[:, 1] - limits[:, 0]))
```
vyasr: Why do you need `np.abs` here? Aren't the limits guaranteed to be ordered?

vyasr (changes requested):

cpp/locality/Voronoi.cc
vyasr: As discussed offline with @bdice, the reason for this change is that voro++ shifts the box by L/2 (i.e. different origin placement), and this change is the easiest way to undo that shift.

freud/box.pyx
vyasr: As discussed offline with @bdice, this is an API choice to have the `obj.plot()` methods fail if users request them and `matplotlib` is not available, but the `_repr_png` methods should not fail since a user is not necessarily actively requesting a plot, so the fallback to `__repr__` is reasonable.

vyasr (approved):
There are a couple of very minor changes I requested, after that feel free to merge.

PR #549, opened 2019-11-08
Convenience method to center particles

lyrivera:
## Description

Adds convenience method to the `Box` class to center the particles using periodic boundaries.

## Motivation and Context

Resolves: None

## How Has This Been Tested?

Tested by visual inspection, no tests has been added yet.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] I have updated the documentation (if relevant).
- [ ] I have added tests that cover my changes (if relevant).
- [x] All new and existing tests passed.
- [ ] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/credits.rst).
- [ ] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

PR #553, opened 2019-11-10
Add periodic center-of-mass computations to Box class.

### LINKED ISSUE #552: ClusterProperties centers of mass are wrong in specific circumstances
**Describe the bug**
I constructed a system where `ClusterProperties` gives the wrong center of mass. The current implementation (v2.0.1) uses a "reference point" chosen as the first point seen in the cluster, and then finds the center of mass using the vectors to other particles via the minimum image convention from that reference frame. The example constructed below gives an incorrect result because the cluster spans the box and the reference point is chosen unwisely (symmetrically placed and opposite to the true center of mass -- though I'm not immediately sure which of those are necessary/sufficient conditions).

**To Reproduce**
```
import freud
import numpy as np

box = freud.Box.cube(3)

# Center of mass is actually on the periodic boundary in x, not in the center
points = [[0, 0, 0],
          [-1, 0, 0], [-1, 0.5, 0], [-1, -0.5, 0],
          [1, 0, 0], [1, 0.5, 0], [1, -0.5, 0]]

clp = freud.cluster.ClusterProperties()
clp.compute((box, points), np.zeros(len(points)))

# Should be [[-1.5, 0, 0]] but gives [[0, 0, 0]] in freud v2.0.1
print(clp.centers)
```

bdice:
## Description
While reviewing #549, I realized that there were some instances in the C++ side of freud's code that also need to do center-of-mass computations. I recognized a bug in the implementation of this for `ClusterProperties` and rewrote #549 in C++ to solve the bug and enable this new feature. This replaces the `ClusterProperties` class's computations of center-of-mass, which is wrong in specific, unusual cases (see #552).

## Motivation and Context
Replaces #549. Resolves #552.

## How Has This Been Tested?
Added a test for the case in #552.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,14 @@ The format is based on
 and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## next
+
+### Added
+* The Box class has methods `center_of_mass` and `center` for periodic-aware center of mass and shifting points to center on the origin.
+
+### Fixed
+* Corrected `ClusterProperties` calculation of centers of mass in specific systems (issue #552).
+
 ## v2.0.1 - 2019-11-08
 
 ### Added
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -12,7 +12,7 @@ and this project adheres to
 ### Fixed
 * The from\_box method correctly passes user provided dimensions to from\_matrix it if is called.
 * Correctly recognize Ovito DataCollection objects in from\_system.
-* Corrected `ClusterProperties` calculation of centers of mass in specific systems (issue #552).
+* Corrected `ClusterProperties` calculation of centers of mass in specific systems.
 
 ## v2.0.1 - 2019-11-08
 
--- a/cpp/box/Box.h
+++ b/cpp/box/Box.h
@@ -353,12 +353,12 @@ public:
      *  \param masses Optional array of masses, of length Nvecs
      *  \return Center of mass as a vec3<float>
      */
-    vec3<float> centerOfMass(vec3<float>* vecs, unsigned int Nvecs, float* masses = NULL) const
+    vec3<float> centerOfMass(vec3<float>* vecs, size_t Nvecs, float* masses = NULL) const
     {
         // This roughly follows the implementation in
         // https://en.wikipedia.org/wiki/Center_of_mass#Systems_with_periodic_boundary_conditions
         float total_mass(0);
-        vec3<std::complex<float>> xi_mean(vec3<std::complex<float>>(0.0, 0.0, 0.0));
+        vec3<std::complex<float>> xi_mean;
 
         for (size_t i = 0; i < Nvecs; ++i) {
             vec3<float> phase(TWO_PI * makeFractional(vecs[i]));
--- a/cpp/box/Box.h
+++ b/cpp/box/Box.h
@@ -5,6 +5,7 @@
 #define BOX_H
 
 #include "utils.h"
+#include <complex>
 #include <sstream>
 #include <stdexcept>
 
@@ -15,6 +16,9 @@
 */
 
 namespace freud { namespace box {
+
+// namespace-level constant 2*pi for convenient use everywhere.
+constexpr float TWO_PI = 2.0 * M_PI;
 
 //! Stores box dimensions and provides common routines for wrapping vectors back into the box
 /*! Box stores a standard HOOMD simulation box that goes from -L/2 to L/2 in each dimension, allowing Lx, Ly,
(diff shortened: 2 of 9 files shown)


tests/test_box_Box.py
```diff
@@ -171,6 +171,55 @@ def test_images(self):
...
+        box = freud.box.Box.cube(5)
+
+        points = [[0, 0, 0], -box.L/4]
+        masses = [2, 1]
+        phases = np.exp(2*np.pi*1j*box.make_fractional(points))
+        com_angle = np.angle(phases.T @ masses / np.sum(masses))
+        com = box.make_absolute(com_angle / (2*np.pi))
+        npt.assert_allclose(box.center(points, masses),
+                            box.wrap(points - com), atol=1e-6)
+
```
lyrivera, line 222: After centering, can probably compute the center of mass without considering the masses and check that it no longer lies in the origin.

freud/box.pyx
```diff
@@ -341,6 +341,59 @@ cdef class Box:
...
+                provided or :code:`None` (Default value = :code:`None`).
+
+        Returns:
+            :math:`\left(3\right)` :class:`numpy.ndarray`:
+                Center of mass.
+        """  # noqa: E501
+        vecs = freud.util._convert_array(vecs, shape=(None, 3))
+        if masses is None:
+            masses = np.ones(len(vecs), dtype=np.float32)
+        masses = freud.util._convert_array(masses, shape=(len(vecs), ))
```
lyrivera: Why explicitly handle `masses==None` here instead of relaying to the C++ function?

freud/box.pyx
```diff
@@ -341,6 +341,59 @@ cdef class Box:
...
+
+        cdef const float[:, ::1] l_points = vecs
+        cdef const float[::1] l_masses = masses
+        cdef unsigned int Np = l_points.shape[0]
+        cdef vec3[float] result = self.thisptr.centerOfMass(
+            <vec3[float]*> &l_points[0, 0], Np, <float*> &l_masses[0])
+        return np.asarray([result.x, result.y, result.z])
+
+    def center(self, vecs, masses=None):
```
lyrivera, line 372: Same as above

lyrivera (commented):
Left a few comments/questions but other than that looks good to me. I'm deferring the approval until a more experienced freud developer reviews and approves this PR.

freud/box.pyx
bdice: Good idea. It's a little awkward to use `NULL` in Cython, but I agreed with you and changed the behavior to use `NULL` because I've used the same behavior in other places (e.g. cluster keys).

tests/test_box_Box.py
bdice: Test added.

bdice:
@lyrivera Thanks for the review. We'll merge after @vyasr takes a look at it.

lyrivera:
> @lyrivera Thanks for the review. We'll merge after @vyasr takes a look at it.

NP, thanks for enabling this.

lyrivera (approved):
LGTM!

cpp/box/Box.h
```diff
@@ -343,6 +347,53 @@ class Box
         });
     }
 
+    //! Compute center of mass for vectors
+    /*! \param vecs Vectors to compute center of mass
+     *  \param Nvecs Number of vectors
+     *  \param masses Optional array of masses, of length Nvecs
+     *  \return Center of mass as a vec3<float>
+     */
+    vec3<float> centerOfMass(vec3<float>* vecs, unsigned int Nvecs, float* masses = NULL) const
```
vyasr: Should `Nvecs` be a `size_t` as well if that's what you're using in the loop? If not, is there a reason to use `size_t` in the loop counter?

cpp/box/Box.h
```diff
@@ -343,6 +347,53 @@ class Box
...
+    {
+        // This roughly follows the implementation in
+        // https://en.wikipedia.org/wiki/Center_of_mass#Systems_with_periodic_boundary_conditions
+        float total_mass(0);
+        vec3<std::complex<float>> xi_mean(vec3<std::complex<float>>(0.0, 0.0, 0.0));
```
vyasr: More a stylistic point than anything else, but just `0, 0, 0` should be enough since it's typed. Also you could just use the default constructor here.

cpp/box/Box.h
```diff
@@ -16,6 +17,9 @@
 
 namespace freud { namespace box {
 
+// namespace-level constant 2*pi for convenient use everywhere.
+constexpr float TWO_PI = 2.0 * M_PI;
```
vyasr, line 21: Should we just define this in one place (e.g. VectorMath) for use everywhere? This is now duplicated in 3 or 4 places.

cpp/box/Box.h
```diff
@@ -343,6 +347,53 @@ class Box
...
+            total_mass += mass;
+            xi_mean += std::complex<float>(mass, 0) * xi;
+        }
+        xi_mean /= std::complex<float>(total_mass, 0);
+
+        return wrap(makeAbsolute(vec3<float>(
+                std::arg(xi_mean.x),
+                std::arg(xi_mean.y),
+                std::arg(xi_mean.z)) / TWO_PI));
+    }
```
vyasr, line 379: According to [the documentation](https://en.cppreference.com/w/cpp/numeric/math/atan2) for `atan2` (which governs the error cases for `std::arg`), whether or not `std::arg(std::complex(0))` throws a domain error is implementation-specific. I tested on my Mac and the vislab and it seems fine, but it might be worth either checking if any of the components of `xi_mean` are 0 to avoid NaNs here (or use the actual error-checking functionality as documented [here](https://en.cppreference.com/w/cpp/numeric/math/math_errhandling).

ChangeLog.md
```diff
@@ -6,9 +6,13 @@ and this project adheres to
 
 ## next
 
+### Added
+* The Box class has methods `center_of_mass` and `center` for periodic-aware center of mass and shifting points to center on the origin.
+
 ### Fixed
 * The from\_box method correctly passes user provided dimensions to from\_matrix it if is called.
 * Correctly recognize Ovito DataCollection objects in from\_system.
+* Corrected `ClusterProperties` calculation of centers of mass in specific systems (issue #552).
```
vyasr: I'd prefer to just stick with not documenting issues/PRs in the changelog.

vyasr (changes requested):

cpp/box/Box.h
bdice: It's not vector math, though. If we centralize it, I would want a sufficiently generic header name. Maybe `Constants.h`. Or just declare it in a `FreudMath.h` file that includes `VectorMath.h`?

cpp/box/Box.h
bdice: @vyasr I don't understand what you want me to do here. I read this [page about atan2](https://en.cppreference.com/w/cpp/numeric/math/atan2) but I'm not sure how to proceed. I would appreciate help.

bdice: The page you linked seems to indicate that `std::atan2(0, 0) = 0` and is not implementation-dependent...?

cpp/box/Box.h
bdice: Changed, although it's inconsistent with all the other functions which use `unsigned int` for the number of vectors.

vyasr: Yeah it's a silly problem, I don't think we care enough right now to go through the entire code base and make this consistent but we can revisit if we get around to exposing a C++ API. Honestly we probably should just rewrite these functions in terms of templated iterators, it's more C++ friendly code and would transparently enable more natural C++ APIs. Food for later thought.

cpp/box/Box.h
vyasr: Yeah, I don't know what the right answer is for this. Technically even `M_PI` isn't part of the standard, but I'm guessing nobody using freud has run into issues with it being undefined in their compiler. I'm fine leaving this for now, and we can revisit if we find that other similar constants are common and such a header file becomes more important.

cpp/box/Box.h
vyasr: I'm reading the lines under "Error handling" on that page. The second line states that "Domain error may occur if x and y are both 0". To see how to handle this properly, look at the documentation of `math_errhandling` [here](https://en.cppreference.com/w/cpp/numeric/math/math_errhandling). I totally appreciate that it looks gross, but this exact kind of problem was causing hard-to-find bugs for us in both Steinhardt and LocalDescriptors, so I'd like to nip that in the bud here rather than risk it cropping up in an unexpected manner later.

vyasr: Feel free to find me tomorrow or just Slack me if you want more specific guidance, I'm happy to explain in more detail.

bdice: (summary of Slack discussion) This is fine since we're assuming IEEE 754 compliance, which is very mainstream and not an edge case we want to worry about.

vyasr: After independent discussion, we concluded that:
1. Supporting non-IEEE 754 compliant systems is out of scope and unnecesary.
2. In the specific edge-cases that would be problematic (i.e. perfectly distributed systems where the values of `xi_mean` are 0), there is no "correct" answer for the center of mass, so rather than trying to impose our own convention we'll just accept whatever the standard library returns. This means that based on the exact floating point number provided, we could get cases like `std::arg(std::complex<float>(+0, -0))`, and that will be resolved according to the definition of `std::atan2`.

vyasr (approved):

PR #554, opened 2019-11-27
make_random_system no longer overwrites global np.random state

vyasr:
## Description
The `make_random_system` function no longer overwrites the overall NumPy random seed.

## Motivation and Context

There's no reason to overwrite the global seed and it's not constant with the `generate_system` behavior. This is *technically* a breaking change, not to the API but to the behavior of a function, but I'm OK with this since it's such a recent function that I doubt anyone is using it other than in our testing.

## Types of changes

## Checklist:

### DIFF
--- a/freud/data.py
+++ b/freud/data.py
@@ -197,9 +197,6 @@ class UnitCell(object):
 def make_random_system(box_size, num_points, is2D=False, seed=None):
     R"""Helper function to make random points with a cubic or square box.
 
-    This function has a side effect, by setting the random seed of numpy if a
-    seed is specified.
-
     Args:
         box_size (float): Size of box.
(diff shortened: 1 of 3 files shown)


freud/data.py
```diff
@@ -212,10 +209,14 @@ def make_random_system(box_size, num_points, is2D=False, seed=None):
             Generated box and points.
     """  # noqa: E501
     if seed is not None:
+        random_state = np.random.get_state()
```
bdice: Does there need to be a `try/finally` block to ensure that the state is reset in the event of failure mid-function?

bdice (changes requested):

vyasr:
LGTM thanks for the update.

PR #555, opened 2019-11-27
Fix/gsd 2d from system

vyasr:
## Description

Set z positions read from 2D GSD trajectories to 0.

## Motivation and Context

We had fixed the box reading (because the z dimension won't be 0), but sometimes the z position is also not quite 0. Having already made this change, I actually think a better solution might be to just add a threshold to the check in the constructor for `NeighborQuery` rather than just checking if it's equal to zero since this could be an issue with other formats as well. @bdice let me know your thoughts, I can change if you agree.

## Types of changes

## Checklist:


bdice (approved):
Discussed via Slack. It's fine to check if values are approximately zero (within some reasonable precision).

PR #556, opened 2019-11-27
Update msd.pyx to clairfy GPL license terms 

mikemhenry:
Not sure the exact way to re-phrase this but it is never an issue to mix GPL with another license unless you want to distribute it. The copyleft aspect is only "triggered" when distributing code. So you don't have to license your code GPL to _use_ some GPL code, you just can't later distribute it without following the GPL license. I hope that is clear, I'm not sure the changes in the doc address that perfectly, but it is not an issue to use the FFTW library in a non-GPL code. I don't want someone to think "welp I don't want to license my code GPL and I don't want to pay for the library" when they have no intention for the code to leave their laptop.

## Description
Changed a doc string to make things more clear.

## Motivation and Context
It is important to not spread FUD about the GPL license. I know that wasn't the intention of the doc string, but to the uninformed it makes it sound like the GPL license affects code use.

## How Has This Been Tested?
Spell Check

## Types of changes

## Checklist:

### DIFF
--- a/freud/msd.pyx
+++ b/freud/msd.pyx
@@ -106,9 +106,7 @@ cdef class MSD(_Compute):
 
           Note that while pyFFTW is released under the BSD 3-Clause license,
           the FFTW library is available under either GPL or a commercial
-          license. As a result, if you wish to distribute this module with
-          pyFFTW in code, your code must also be GPL licensed unless you
-          purchase a commercial license.
+          license.
 
     * :code:`'direct'`:
       Under some circumstances, however, we may be more interested in
--- a/freud/msd.pyx
+++ b/freud/msd.pyx
@@ -106,8 +106,8 @@ cdef class MSD(_Compute):
 
           Note that while pyFFTW is released under the BSD 3-Clause license,
           the FFTW library is available under either GPL or a commercial
-          license. As a result, if you wish to distribute this module with 
-          pyFFTW in code, your code must also be GPL licensed unless you 
+          license. As a result, if you wish to distribute this module with
+          pyFFTW in code, your code must also be GPL licensed unless you
           purchase a commercial license.
 
     * :code:`'direct'`:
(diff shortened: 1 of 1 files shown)


vyasr:
Thanks for the heads-up Mike! That's good to know, and I agree that the existing docstring is misworded. This clarification is helpful.

joaander:
The GPL doesn't clearly define "distribute" as a term. It does define "propagate" and "convey". It states "Each time you convey a covered work, the recipient automatically receives a license from the original licensors, to run, modify and propagate that work, subject to this License." which I think is what Mike is referring to.

The GPL is still very much in force even when you keep your modifications private. For example, it states "You may not propagate **or modify** a covered work except as expressly provided under this License." and "You may convey covered works to others for the sole purpose of having them make modifications exclusively for you, or provide you with facilities for running those works, provided that you comply with the terms of this License in conveying all material for which you do not control copyright. Those thus making or running the covered works for you must do so exclusively on your behalf, under your direction and control, on terms that prohibit them from making any copies of your copyrighted material outside their relationship with you."

In general the GPL is very complex and requires the advice of a legal professional to interpret properly. I do not think we should attempt to interpret the license for users. Maybe we should remove the "As a result...." and simply end with "the FFTW library is available under either the GPL or a commercial license."

mikemhenry:
I agree that we shouldn't try and interpret the GPL for users and I think making the license of the FFTW clear is the best approach. I will modify the PR to use that language.

Regarding keeping modifications private: 
https://www.gnu.org/licenses/gpl-faq.en.html#GPLRequireSourcePostedPublic

I had originally thought the warning was to let users know if they use the FFTW library, they have to release their code publicly under the GPL.

mikemhenry:
Also just for some context, I was looking at seeing if I could use freud to replace the deprecated MSD analyzer in hoomd and saw this note.

vyasr (approved):

vyasr:
@mikemhenry let me know if you need help doing that!

PR #579, opened 2020-01-24
Parallel reduction

### LINKED ISSUE #576: Parallelize reduction
Many features in freud perform highly parallel calculations that eventually require reduction of thread-local storage into a single primary data structure, typically a ManagedArray or a Histogram. Currently, the reduction process is always serialized, so it is a potential bottleneck when the data structures are large but the input data is small, e.g. accumulating a short trajectory into a PMFT with a large number of bins to achieve high spatial resolution.

We should parallelize our reduction scheme using [`tbb::parallel_reduce`](https://software.intel.com/en-us/node/506154). Using the imperative form of the function, we would define the Body struct to perform reduction of the thread-local storage objects. Since we use `tbb::enumerated_thread_specific` for this purpose, we would first need to iterate over this object serially once and generate an array of pointers to each of the thread-local variables. Then, the first time any Body's `operator()` method is called, we would store as a member variable the current thread index and use that variable to control which thread-local variable becomes the accumulation destination. Then, each time join is called, the values would be added to this bin. This method relies on the fact that `parallel_reduce` guarantees that reduction occurs left-to-right, but should not be making any other assumptions about the order of reduction.

Usage of `parallel_invoke` or `combinable` are also possible, but the latter is designed for serial reduction, and `parallel_invoke` requires a bit more work to use since it would require designing the appropriate recursive calls.

bdice:
## Description
~This pull request uses thread-parallel reduction when accumulating thread-local data from ThreadStorage and ThreadLocalHistogram classes.~ This pull request unifies the reduction logic scattered throughout freud into a couple locations.

## Motivation and Context
~I noticed that GaussianDensity was slow for large data sizes. This improves its performance.~ No significant performance gains were made. Instead, this PR is just a refactor.
Resolves: #576.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,12 @@ The format is based on
 and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## v2.2.0
+
+### Added
+* NeighborQuery objects can now create NeighborLists with neighbors sorted by bond distance.
+* LocalDescriptors `compute` takes an optional maximum number of neighbors to compute for each particle.
+
 ## v2.1.0 - 2019-12-19
 
 ### Added
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,6 +9,7 @@ and this project adheres to
 ### Added
 * NeighborQuery objects can now create NeighborLists with neighbors sorted by bond distance.
 * LocalDescriptors `compute` takes an optional maximum number of neighbors to compute for each particle.
+* Implemented parallel reduction for thread-local storage and histograms.
 
 ## v2.1.0 - 2019-12-19
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,7 +9,6 @@ and this project adheres to
 ### Added
 * NeighborQuery objects can now create NeighborLists with neighbors sorted by bond distance.
 * LocalDescriptors `compute` takes an optional maximum number of neighbors to compute for each particle.
-* Implemented parallel reduction for `ThreadStorage` and `Histogram` C++ classes.
 
 ### Fixed
 * Corrected calculation of neighbor distances in the Voronoi NeighborList.
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,7 +9,7 @@ and this project adheres to
 ### Added
 * NeighborQuery objects can now create NeighborLists with neighbors sorted by bond distance.
 * LocalDescriptors `compute` takes an optional maximum number of neighbors to compute for each particle.
-* Implemented parallel reduction for thread-local storage and histograms.
+* Implemented parallel reduction for `ThreadStorage` and `Histogram` C++ classes.
 
 ## v2.1.0 - 2019-12-19
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,9 @@ and this project adheres to
 * LocalDescriptors `compute` takes an optional maximum number of neighbors to compute for each particle.
 * Implemented parallel reduction for `ThreadStorage` and `Histogram` C++ classes.
 
+### Fixed
+* Corrected calculation of neighbor distances in the Voronoi NeighborList.
+
 ## v2.1.0 - 2019-12-19
 
 ### Added
--- a/cpp/util/Histogram.h
+++ b/cpp/util/Histogram.h
@@ -250,26 +250,19 @@ public:
             m_local_histograms.local().increment(value_bin, weight);
         }
 
+        // Reduce over histograms into the result array.
         void reduceInto(ManagedArray<T>& result)
         {
-            if (m_local_histograms.size() == 0)
-            {
-                // If no local histograms have been created, then no data can be reduced
-                // and an error will occur if we attempt to iterate over histograms.
-                // We simply reset the result array so it's all zeros.
-                result.reset();
-            }
-            else
-            {
-                // Reduce over histograms in parallel, into the result array.
-                // Iterate over the ThreadLocalHistogram and get pointers to each thread's data.
-                std::vector<util::ManagedArray<T>*> array_pointers;
-                for (auto hist = m_local_histograms.begin(); hist != m_local_histograms.end(); ++hist)
+            result.reset();
+            util::forLoopWrapper(0, result.size(), [=](size_t begin, size_t end) {
+                for (size_t i = begin; i < end; ++i)
                 {
-                    array_pointers.push_back(&(hist->getBinCounts()));
+                    for (auto hist = m_local_histograms.begin(); hist != m_local_histograms.end(); ++hist)
+                    {
+                        result[i] += hist->m_bin_counts[i];
+                    }
                 }
-                util::reduceManagedArrays(array_pointers, result);
-            }
+            });
         }
 
 
--- a/cpp/util/Histogram.h
+++ b/cpp/util/Histogram.h
@@ -250,6 +250,71 @@ public:
             m_local_histograms.local().increment(value_bin, weight);
         }
 
+        ManagedArray<T> reduce()
+        {
+            typedef std::vector<util::Histogram<T>*> ThreadStorageVector;
+            const unsigned int BODY_DEFAULT(0xffffffff);
+            struct reduceThreadStorage
+            {
+                const ThreadStorageVector& m_;
+                const std::vector<size_t> shape_;
+                const size_t size_;
+                unsigned int body_;
+
+                reduceThreadStorage(ThreadStorageVector& m) :
+                    m_(m), shape_(m_[0]->shape()), size_(m_[0]->size()), body_(BODY_DEFAULT) {}
+
+                // splitting constructor required by TBB
+                reduceThreadStorage(reduceThreadStorage& rm, tbb::split) :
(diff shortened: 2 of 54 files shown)


bdice:
I get a segfault locally on `test_PMFT.py` so further investigation / debugging is needed. It only fails on `test_shift_two_particles_dead_pixel`.

bdice:
Segfault fixed - it happened with empty histograms when no thread-local storages were initialized. This is ready for review.

cpp/util/Histogram.h
```diff
@@ -420,15 +495,10 @@ template<typename T> class Histogram
...
         util::forLoopWrapper(0, m_bin_counts.size(), [=](size_t begin, size_t end) {
             for (size_t i = begin; i < end; ++i)
             {
-                for (typename ThreadLocalHistogram::const_iterator local_bins = local_histograms.begin();
-                     local_bins != local_histograms.end(); ++local_bins)
-                {
-                    m_bin_counts[i] += (*local_bins).m_bin_counts[i];
-                }
-
                 cf(i);
```
bdice, line 443: ⚠️ This assumes that the order of reducing bin counts and calling cf doesn't matter.

cpp/util/Histogram.h
```diff
@@ -250,6 +250,81 @@ template<typename T> class Histogram
             m_local_histograms.local().increment(value_bin, weight);
         }
 
+        void reduceHistogramsInto(ManagedArray<T>& result)
```
bdice: This could be turned into a template function. One template is `util::Histogram<T>*` and the other is `util::ManagedArray<T>*`. See `HistogramVector` vs `ThreadStorageVector`. Where should it go?

cpp/util/Histogram.h
```diff
@@ -250,6 +250,81 @@ template<typename T> class Histogram
...
+                for (auto hist = m_local_histograms.begin(); hist != m_local_histograms.end(); ++hist)
+                {
+                    histogram_pointers.push_back(&(*hist));
+                }
+
+                reduceThreadLocalHistogram reducer(histogram_pointers);
+                tbb::parallel_reduce(tbb::blocked_range<unsigned int>(0, histogram_pointers.size()), reducer);
+
+                // Set result to a copy of the final reduced array
+                result = histogram_pointers[0]->getBinCounts().copy();
```
bdice: This line differs between the two implementations.

cpp/util/Histogram.h
```diff
@@ -250,6 +250,81 @@ template<typename T> class Histogram
...
+                // If no local histograms have been created, then no data can be reduced.
+                // We simply reset the result array so it's all zeros.
+                result.reset();
+            }
+            else
+            {
+                // Reduce over histograms in parallel, into the result array.
+                // Iterate over the ThreadLocalHistogram and get pointers to each thread's data.
+                HistogramVector histogram_pointers;
+                for (auto hist = m_local_histograms.begin(); hist != m_local_histograms.end(); ++hist)
```
bdice: This line differs between the two implementations.

bdice:
@vyasr This PR is ready for review. Thanks!

ChangeLog.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to
 ### Added
 * NeighborQuery objects can now create NeighborLists with neighbors sorted by bond distance.
 * LocalDescriptors `compute` takes an optional maximum number of neighbors to compute for each particle.
+* Implemented parallel reduction for thread-local storage and histograms.
```
vyasr: Use the class names directly here.

cpp/util/Histogram.h
vyasr: Actually I'd frame this as a design choice, not an assumption; the docstring for the function explicitly states that it's for post-processing. I can see how it might get misused if a user isn't careful, but I don't see any point in addressing the pre-processing case unless there's a need since this version covers all existing use-cases in freud.

cpp/util/Histogram.h
```diff
@@ -250,6 +250,81 @@ template<typename T> class Histogram
...
         }
 
+        void reduceHistogramsInto(ManagedArray<T>& result)
+        {
+            typedef std::vector<util::Histogram<T>*> HistogramVector;
+            const unsigned int BODY_DEFAULT(0xffffffff);
+            struct reduceThreadLocalHistogram
+            {
+                const HistogramVector& m_;
+                const std::vector<size_t> shape_;
```
vyasr: You only need `size_`, so I'd remove `shape_`. Do the same for the `ThreadStorage` version.

cpp/util/ThreadStorage.h
```diff
@@ -89,6 +89,80 @@ template<typename T> class ThreadStorage
...
     }
 
+    void reduceInto(ManagedArray<T>& result)
+    {
+        typedef std::vector<util::ManagedArray<T>*> ThreadStorageVector;
+        const unsigned int BODY_DEFAULT(0xffffffff);
+        struct reduceThreadStorage
+        {
+            const ThreadStorageVector& m_;
+            const std::vector<size_t> shape_;
```
vyasr: Remove (see above).

vyasr (changes requested):
For reference, could you post the output of running benchmarks in a couple cases? Mainly GaussianDensity and Nematic (especially since Nematic you also substantially reduced accumulation needs by adding to a local array in the loop).

I think the way to get around the small differences between the two implementations is to define a third function (not a method) that accepts a `ManagedArray` to reduce into and a vector of pointers to `ManagedArray`s and reduces into it. The `ThreadStorage` method could then just construct a list of `ManagedArray` pointers as it's doing now and then feed it to that function, and the `ThreadLocalHistogram` could make the list of pointers from the bin counts of each local histogram. The result would probably make sense to put directly into ManagedArray.h, although I could also see the argument for putting it into ThreadStorage.h.

Otherwise this looks great! The implementation is basically what we discussed aside from handling the empty array case and seems like it works nicely.

vyasr:
@bdice I addressed your comments as part of my overall review of the PR, so I'm going to resolve those conversations. Let me know if there's anything I didn't cover, though.

cpp/util/Histogram.h
bdice: Ok, I'll resolve this.

cpp/util/Histogram.h
bdice: Good catch -- originally I needed `shape_` but not anymore.

vyasr:
To document where this PR is currently stalled; updated benchmarks do not consistently show performance improvements from parallel reduction. We are awaiting more comprehensive performance testing to see if the additional complexity is merited.

@bdice let me know when you have benchmarks that you're confident in so that we can reevaluate this PR (no rush).

bdice:
@vyasr I tried to rewrite this to use serial reduction (but reduce repetition across the codebase, as we discussed). However, I can't get it to build and I don't understand why. It is acting like it can only find the `const` qualified `operator[]` for ManagedArrays, not the writable index. See snippet below.

```cpp
In file included from cpp/pmft/PMFTXYT.cc:4:
In file included from cpp/pmft/PMFTXYT.h:7:
In file included from cpp/pmft/PMFT.h:9:
In file included from cpp/locality/BondHistogramCompute.h:5:
cpp/util/Histogram.h:262:35: error: cannot assign to return value because function 'operator[]' returns a const value
                        result[i] += hist->m_bin_counts[i];
                        ~~~~~~~~~ ^
cpp/util/Histogram.h:445:26: note: in instantiation of member function 'freud::util::Histogram<unsigned int>::ThreadLocalHistogram::reduceInto' requested here
        local_histograms.reduceInto(m_bin_counts);
                         ^
cpp/pmft/PMFT.h:47:21: note: in instantiation of function template specialization 'freud::util::Histogram<unsigned int>::reduceOverThreadsPerBin<(lambda at cpp/pmft/PMFT.h:47:65)>' requested here
        m_histogram.reduceOverThreadsPerBin(m_local_histograms, [this, &prefactor, &jf](size_t i) {
                    ^
cpp/pmft/PMFTXYT.cc:47:11: note: in instantiation of function template specialization 'freud::pmft::PMFT::reduce<(lambda at cpp/pmft/PMFTXYT.cc:47:18)>' requested here
    PMFT::reduce([jacobian_factor](size_t i) { return jacobian_factor; });
          ^
cpp/util/ManagedArray.h:143:11: note: function 'operator[]' which returns const-qualified type 'const unsigned int &' declared here
    const T& operator[](size_t index) const
          ^~
```

vyasr:
I'm a bit swamped at the moment, I'll try and look at this in the next couple of days if possible.

bdice:
@vyasr No worries, thanks for taking a look when you're able. Please prioritize #586 over this one -- it shouldn't take very much brainpower to review that PR.

vyasr:
I've resolved the issues with this PR. The `reduceInto` functions were binding the `result` by value in the lambda, so it wasn't modifiable.

vyasr:
I had a few comments, but I decided to address them myself rather than asking you to take another pass. I did a bit of clean up to remove the functions that are no longer relevant (the managed array reduction, a non-const accessor to bin counts) and reverted the changelog entry since reduction is no longer parallel.

vyasr (approved):

Issue #580, opened 2020-01-24
Control NeighborList sort order


vyasr:
**Is your feature request related to a problem? Please describe.**
It should be possible to control the sort order of `NeighborList` objects. Specifically, while the first index should always the primary sort key, depending on the use-case it may be beneficial to have the secondary sort key be the distance instead of the second index.

**Describe the solution you'd like**
Adding this feature should simply require making a new function like `less_as_tuple` in NeighborBond.h that changes the sort order. Then, `NeighborList::compareNeighborBond` can choose which function to use based on an argument.

vyasr:
Resolved by #581

PR #581, opened 2020-01-24
Add option to limit number of neighbor bonds in LocalDescriptors

klarh:
## Description
This PR adds back the behavior described in the LocalDescriptors documentation:

> Spherical harmonic calculation can be restricted to some number of nearest neighbors through the num_neighbors argument; if a particle has more bonds than this number, the last one or more rows of bond spherical harmonics for each particle will not be set.

Simultaneously, the PR adds the ability to sort neighbor lists by distance, which is typically required for this limiting behavior to be useful.

## Motivation and Context
Related to #580 

## How Has This Been Tested?
I'm currently testing this with an updated version of pythia.

## Types of changes

## Checklist:

### DIFF
--- a/freud/locality.pyx
+++ b/freud/locality.pyx
@@ -206,8 +206,12 @@ cdef class NeighborQueryResult:
 
         raise StopIteration
 
-    def toNeighborList(self):
+    def toNeighborList(self, sort_by_distance=False):
         """Convert query result to a freud NeighborList.
+
+        Args:
+            sort_by_distance (bool):
+                If True, sort neighboring bonds by distance instead of index
 
         Returns:
             :class:`~NeighborList`: A :mod:`freud` :class:`~NeighborList`
@@ -222,7 +226,7 @@ cdef class NeighborQueryResult:
                 dereference(self.query_args.thisptr))
 
         cdef freud._locality.NeighborList *cnlist = dereference(
-            iterator).toNeighborList()
+            iterator).toNeighborList(sort_by_distance)
         cdef NeighborList nl = _nlist_from_cnlist(cnlist)
         # Explicitly manage a manually created nlist so that it will be
         # deleted when the Python object is.
--- a/freud/locality.pyx
+++ b/freud/locality.pyx
@@ -211,8 +211,9 @@ cdef class NeighborQueryResult:
 
         Args:
             sort_by_distance (bool):
-                If True, sort neighboring bonds by distance. By default, sort
-                neighboring bonds by point index.
+                If :code:`True`, sort neighboring bonds by distance.
+                If :code:`False`, sort neighboring bonds by point index
+                (Default value = :code:`False`).
 
         Returns:
             :class:`~NeighborList`: A :mod:`freud` :class:`~NeighborList`
--- a/freud/locality.pyx
+++ b/freud/locality.pyx
@@ -211,7 +211,8 @@ cdef class NeighborQueryResult:
 
         Args:
             sort_by_distance (bool):
-                If True, sort neighboring bonds by distance instead of index
+                If True, sort neighboring bonds by distance. By default, sort
+                neighboring bonds by point index.
 
         Returns:
             :class:`~NeighborList`: A :mod:`freud` :class:`~NeighborList`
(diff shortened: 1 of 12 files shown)


vyasr (changes requested):
This PR looks good. To avoid future changes that could cause problems, could you add tests to the NeighborList that ensure that the sorting behavior doesn't get changed again? Ideally one test with the sort_neighbors=False, and one where it's True.

vyasr:
Sorry, I didn't see that this was a WIP. Add tests whenever you think it's ready.

freud/locality.pyx
```diff
@@ -206,9 +206,13 @@ cdef class NeighborQueryResult:
 
         raise StopIteration
 
-    def toNeighborList(self):
+    def toNeighborList(self, sort_by_distance=False):
         """Convert query result to a freud NeighborList.
 
+        Args:
+            sort_by_distance (bool):
+                If True, sort neighboring bonds by distance instead of index
```
bdice: Clarify which index you mean (point, not query point). Maybe provide a sentence explaining the default behavior. End with a period.

tests/test_locality_NeighborList.py
```diff
@@ -228,6 +228,28 @@ def test_copy(self):
...
+    def test_ordering(self):
+        # default behavior sorts by (i, j, distance)
+        tuples = list(zip(self.nlist.query_point_indices,
+                          self.nlist.point_indices,
+                          self.nlist.distances))
+        sorted_tuples = list(sorted(tuples))
+
+        self.assertEqual(tuples, sorted_tuples)
+
+        # test sorting by (i, distance, j)
```
bdice: Could you re-use the data from setUp if you made this a separate test?

bdice (changes requested):
Overall looks good. A couple small change requests.

tests/test_locality_NeighborList.py
klarh: I don't immediately see a way to parameterize `setUp` in the unittest framework, did you have something in mind, or were you thinking to just always store two neighbor lists? Adding a parameter to it and just re-calling it would work, of course; let me know if you want it done that way.

bdice: Don’t parameterize, just let setUp do its thing and then use self.nq to create a sorted nlist. You just don’t have to create a new box, points, nq because those already exist.

klarh: Oh, of course! That was thoughtless of me.

tests/test_locality_NeighborList.py
```diff
@@ -15,9 +15,9 @@ def setUp(self):
                                exclude_ii=True)
 
         # Initialize Box and cell list
-        box, points = freud.data.make_random_system(self.L, self.N)
-        self.nq = freud.locality.AABBQuery(box, points)
-        self.nlist = self.nq.query(points,
+        box, self.points = freud.data.make_random_system(self.L, self.N)
```
bdice: @klarh Was this intentional or an accidental find and replace?

klarh: Intentional; do you prefer to save the query rather than the points?

bdice: The self.nq holds its points array. Just use self.nq.points in the test.

klarh: OK, fixed.

bdice (approved):
Thanks @klarh. I will let @vyasr take another pass of review before we finalize, but I am approving.

vyasr:
It looks good to me. I'm adding a bit more context to the docstring on _why_ this argument could be useful (NeighborList reuse) since it seems random without context. I'll commit after tests pass and then merge.

vyasr (approved):
All set

bdice:
@vyasr It looks like you force-pushed and took away the things I fixed in the docstring formatting... please make sure these changesets get integrated.
https://github.com/glotzerlab/freud/pull/581/commits/6482bd236934ab6d217ea32b63407df6acc7d17b
https://github.com/glotzerlab/freud/pull/581/commits/3b811d3e1e3bfcf433d2c4eabec6fe24f223b800

vyasr:
I did force push, but unless I'm missing something I didn't remove any of those changes: see [here](https://github.com/glotzerlab/freud/blob/master/freud/environment.pyx#L233), [here](https://github.com/glotzerlab/freud/blob/master/freud/environment.pyx#L235), and [here](https://github.com/glotzerlab/freud/blob/master/freud/locality.pyx#L214).

PR #585, opened 2020-02-11
Fix Voronoi distances

bdice:
## Description
It appears that distance calculations in the Voronoi neighborlist were not always correct. This PR rewrites that logic to compute distances based on the wrapped interparticle vector.

## Motivation and Context
Resolves an issue found by @klywang.

## How Has This Been Tested?
Added tests for Voronoi NeighborList distances.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/order/Cubatic.cc
+++ b/cpp/order/Cubatic.cc
@@ -171,7 +171,7 @@ template<typename T> quat<float> Cubatic::calcRandomQuaternion(T& dist, float an
     float theta = 2.0 * M_PI * dist();
     float phi = acos(2.0 * dist() - 1.0);
     vec3<float> axis = vec3<float>(cosf(theta) * sinf(phi), sinf(theta) * sinf(phi), cosf(phi));
-    float axis_norm = sqrt(dot(axis, axis));
+    float axis_norm = std::sqrt(dot(axis, axis));
     axis /= axis_norm;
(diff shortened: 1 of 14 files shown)


cpp/order/Cubatic.cc
```diff
@@ -171,7 +171,7 @@ template<typename T> quat<float> Cubatic::calcRandomQuaternion(T& dist, float an
     float theta = 2.0 * M_PI * dist();
     float phi = acos(2.0 * dist() - 1.0);
     vec3<float> axis = vec3<float>(cosf(theta) * sinf(phi), sinf(theta) * sinf(phi), cosf(phi));
-    float axis_norm = sqrt(dot(axis, axis));
+    float axis_norm = std::sqrt(dot(axis, axis));
```
klywang, line 174: Just ouf of curiousity, what library/namespace do these other math functions (ie, cosf, sinf) come from? - And why was it possible to use `sqrt` without referencing the `std` library before?

bdice: @klywang This may be hidden by an imported file that includes `using namespace std`, which removes the need for a namespace. The `cosf` and such should probably be replaced with `std::cos` but I might do that in a separate PR.

bdice (commented):
@vyasr approved this by reading over my shoulder.

PR #586, opened 2020-02-11
Minor refactors for v2.2.

bdice:
## Description
This is a minor cleanup of several parts of freud that I've put on the back burner for a while.

- All cmath functions are now called via `std::func()` instead of `func()` for clarity (and to ensure we get the right function).
- I added a `util::modulusPositive` method to reduce the many instances of modular division followed by conditionally adding the divisor back if the result is negative.
- Python boxes created from C++ objects will now retain their periodic flags. Even though those flags don't do anything yet, the flags were lost when passing through C++ and back to Python.
- Unused imports were removed from Cython files. This should speed up builds a bit, by cutting out unused interdependence between Cython modules.

## Motivation and Context
Cleanup of existing code.

## How Has This Been Tested?
Existing tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/LinkCell.h
+++ b/cpp/locality/LinkCell.h
@@ -242,9 +242,6 @@ public:
     querySingle(const vec3<float> query_point, unsigned int query_point_idx, QueryArgs args) const;
 
 private:
-    //! Rounding helper function.
-    static unsigned int roundDown(unsigned int v, unsigned int m);
-
     //! Helper function to compute cell neighbors
     const std::vector<unsigned int>& computeCellNeighbors(unsigned int cell) const;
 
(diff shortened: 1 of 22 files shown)


bdice:
@vyasr When you review this, please suggest an appropriate changelog line (if you think one is needed). I don't think any of these changes require a changelog line except potentially the bugfix for periodic box flags, but even that is not officially supported so it's hard to say that anything was really "fixed."

cpp/locality/LinkCell.h
```diff
@@ -159,7 +159,7 @@ class IteratorCellShell
 
  *  The cell coordinate (i,j,k) itself is computed like so:
  *  \code
- *  i = floorf((x + Lx/2) / w) % Nw
+ *  i = std::floor((x + Lx/2) / w) % Nw
```
vyasr, line 162: Lol the results of global find and replace.

vyasr (approved):

PR #588, opened 2020-02-19
Add a finite tolerance for 2D Voronoi computations.

### LINKED ISSUE #587: Incorrect Delauny triangulation returned for 2D boxes
**Describe the bug**
Voronoi diagrams calculated in 2D boxes return the expected results, however, the Delauny triangulation returned is incorrect. Expected results can be obatined using `scipy.spatial.Delauny` or with 3D structural equivelants.

**To Reproduce**
Using the following function:
```python
import freud
from matplotlib.collections import LineCollection
import gsd.hoomd
import numpy as np

def voro(filename, is2D=False):
    frame = gsd.hoomd.open(filename, mode='rb')[0]
    pos = np.copy(frame.particles.position)

    # Crop
    pos = pos[abs(pos[:, 0]) < 20]
    pos = pos[abs(pos[:, 1]) < 20]
    box = freud.Box(Lx=40, Ly=40, Lz=frame.configuration.box[2])

    # Calculate Voronoi diagram
    voro = freud.locality.Voronoi().compute((box,  pos))
    if is2D is False:
        # Removes bonds between layers
        voro.nlist.filter(voro.nlist.distances > 0.1)

    # Plot
    fig = plt.figure(figsize=(20, 20))
    ax = fig.gca()
    if is2D:
        voro.plot(ax=ax)
    line_data = np.asarray([[pos[i],
                             pos[i] + box.wrap(pos[j] - pos[i])]
                            for i, j in voro.nlist])[:, :, :2]
    line_collection = LineCollection(line_data, colors='black', alpha=0.2)
    ax.add_collection(line_collection)
    ax.set_ylim([-20, 20])
    ax.set_xlim([-20, 20])
    plt.show()
```
Where plots from the 2D case were created using the line `voro('2D_file.gsd', is2D=True` and the flattened 3D case were created using the line `voro('3D_file.gsd')`.

Scipy results were produced with the following code:

```python
import scipy.spatial
import numpy as np
import gsd.hoomd

def delauny(filename, is2D=False):
    plt.figure(figsize=(20, 20))
    frame = gsd.hoomd.open(filename, mode='rb')[0]
    pos = np.copy(frame.particles.position[:, 0:2])
    pos = pos[abs(pos[:, 0]) < 20]
    pos = pos[abs(pos[:, 1]) < 20]
    tri = scipy.spatial.Delaunay(pos)
    plt.triplot(pos[:, 0], pos[:, 1], tri.simplices.copy())
    plt.scatter(pos[:, 0], pos[:, 1], s=50, color='black')
    plt.ylim([-20, 20])
    plt.xlim([-20, 20])
    plt.show()
```
Where plots from the 2D case were created using the line `delauny('2D_file.gsd', is2D=True` and the flattened 3D case were created using the line `delauny('3D_file.gsd')`.

**Error output**
For the 2D case, the expected output, calculated from `scipy.spatial.Delauny` is:
image
Freud returns the following result:
image

In the 3D case, you get the expected output. `scipy.spatial.Delauny` returns:
image
Freud returns a similar results:
image

 - OS: macOS
 - Version of Python [e.g. 3.7] 3.8.1
 - Version of freud [e.g. 1.0] 2.1.0

bdice:
## Description
I examined issue #587. The issue was caused by a condition that is being used specifically in 2D systems to filter out unwanted bonds pointing in the z-direction, resulting from the hack used to enable 2D support. I believe this is the same underlying issue reported by Austin Dulaney on the forum: https://groups.google.com/forum/#!topic/freud-users/xZxG3KL3DIw

## Motivation and Context
Since voro++ doesn't natively support 2D systems, we build a planar system in a 3D box with a z-dimension of 1. When finding neighbors, we throw out the bonds that point perpendicular to the plane. Unfortunately this procedure was failing due to issues with floating-point precision. The check was effectively skipping if `abs(z) > 0`, which resulted in undesirable behavior when the z-component was slightly above or below zero due to numerical imprecision. I added a finite tolerance that should correct the problem. This should be totally safe because the Voronoi cell's normal vectors should always be confined to the plane or pointing perpendicular to the plane. The hack for 2D support creates Voronoi cells that are all prisms.

Resolves: #587.

## How Has This Been Tested?
Tests were added for symmetric neighbor lists with 2D systems. The tests failed before fixing the bug and succeeded after fixing the bug.

## Screenshots
Bonds in Delaunay triangulation now match the Voronoi diagram (up to the tiny facets resulting from numerical imprecision).
image
Histogram of neighbor counts: this now agrees with the number of vertices (equal to the number of edges).
image

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/Voronoi.cc
+++ b/cpp/locality/Voronoi.cc
@@ -129,6 +129,7 @@ void Voronoi::compute(const freud::locality::NeighborQuery* nq)
             // Save cell volume
             m_volumes[query_point_id] = cell.volume();
 
+            // Compute cell neighbors
             size_t neighbor_counter(0);
             for (auto neighbor_iterator = neighbors.begin(); neighbor_iterator != neighbors.end();
                  neighbor_iterator++, neighbor_counter++)
@@ -137,9 +138,13 @@ void Voronoi::compute(const freud::locality::NeighborQuery* nq)
                 const vec3<double> normal(normals[3 * neighbor_counter], normals[3 * neighbor_counter + 1],
                                           normals[3 * neighbor_counter + 2]);
 
-                // Ignore bonds in 2D systems that point up or down
-                if (box.is2D() && std::abs(normal.z) > 0)
+                // Ignore bonds in 2D systems that point up or down. This check
+                // should only be dealing with bonds whose normal vectors' z
+                // components are -1, 0, or +1 (within some tolerance).
+                if (box.is2D() && std::abs(normal.z) > 1e-5)
+                {
                     continue;
+                }
 
                 // Fetch neighbor information
                 const int point_id = *neighbor_iterator;
--- a/cpp/locality/Voronoi.cc
+++ b/cpp/locality/Voronoi.cc
@@ -141,7 +141,7 @@ void Voronoi::compute(const freud::locality::NeighborQuery* nq)
                 // Ignore bonds in 2D systems that point up or down. This check
                 // should only be dealing with bonds whose normal vectors' z
                 // components are -1, 0, or +1 (within some tolerance).
-                if (box.is2D() && std::abs(normal.z) > 1e-5)
+                if (box.is2D() && std::abs(normal.z) > 0.5)
                 {
                     continue;
                 }
--- a/freud/plot.py
+++ b/freud/plot.py
@@ -406,11 +406,18 @@ def voronoi_plot(box, polytopes, ax=None, color_by_sides=True, cmap=None):
 
     if color_by_sides:
         colors = np.array([len(poly) for poly in polytopes])
+        num_colors = np.ptp(colors) + 1
     else:
         colors = np.random.RandomState().permutation(np.arange(len(patches)))
+        num_colors = np.unique(colors).size
 
-    cmap = cm.get_cmap('Set1' if cmap is None else cmap,
-                       np.unique(colors).size)
+    # Ensure we have enough colors to uniquely identify the cells
+    if cmap is None:
+        if color_by_sides and num_colors <= 10:
+            cmap = 'tab10'
+        else:
+            cmap = 'tab20'
+    cmap = cm.get_cmap(cmap, num_colors)
     bounds = np.arange(np.min(colors), np.max(colors)+1)
 
     patch_collection.set_array(np.array(colors)-0.5)
--- a/freud/plot.py
+++ b/freud/plot.py
@@ -4,6 +4,7 @@
 import freud
 import io
 import numpy as np
+import warnings
 
 try:
     from matplotlib.figure import Figure
@@ -416,6 +417,10 @@ def voronoi_plot(box, polytopes, ax=None, color_by_sides=True, cmap=None):
         if color_by_sides and num_colors <= 10:
             cmap = 'tab10'
         else:
+            if num_colors > 20:
+                warnings.warn('More than 20 unique colors were requested. '
+                              'Consider providing a colormap to the cmap '
+                              'argument.', UserWarning)
             cmap = 'tab20'
     cmap = cm.get_cmap(cmap, num_colors)
     bounds = np.arange(np.min(colors), np.max(colors)+1)
--- a/tests/test_locality_Voronoi.py
+++ b/tests/test_locality_Voronoi.py
@@ -26,6 +26,19 @@ class TestVoronoi(unittest.TestCase):
             points[vor.nlist.query_point_indices]), axis=-1)
         npt.assert_allclose(wrapped_distances, vor.nlist.distances)
 
+        # Ensure every point has neighbors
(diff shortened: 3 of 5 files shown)


adulaney:
This fix appears to solve the issue I posted about on the forum:
https://groups.google.com/forum/#!topic/freud-users/xZxG3KL3DIw.

Thanks!

cpp/locality/Voronoi.cc
```diff
@@ -137,9 +138,13 @@ void Voronoi::compute(const freud::locality::NeighborQuery* nq)
                 const vec3<double> normal(normals[3 * neighbor_counter], normals[3 * neighbor_counter + 1],
                                           normals[3 * neighbor_counter + 2]);
 
-                // Ignore bonds in 2D systems that point up or down
-                if (box.is2D() && std::abs(normal.z) > 0)
+                // Ignore bonds in 2D systems that point up or down. This check
+                // should only be dealing with bonds whose normal vectors' z
+                // components are -1, 0, or +1 (within some tolerance).
+                if (box.is2D() && std::abs(normal.z) > 1e-5)
```
klywang: Are there any cases where the tolerance may differ? This seems like something that should be consistent but I'm curious if there are any special cases.

bdice:
I noticed an issue with the colormap in the screenshot above. The colormap was running out of colors to use. I also fixed that in this PR.
image

klywang (approved):
It appears to have solved the issue reported in #587 as well. 
image

cpp/locality/Voronoi.cc
bdice: I don’t think there are any special cases - the z component should always be -1, 0, or 1 as noted in the comment above. I looked at the normals from the data data you posted and it seemed to be a valid check.

vyasr: I'm confused about why this needs to be a tolerance. If we set the input positions to have all z positions = 0, shouldn't an explicit `!=0` check work here?

freud/plot.py
```diff
@@ -406,11 +406,18 @@ def voronoi_plot(box, polytopes, ax=None, color_by_sides=True, cmap=None):
...
 
-    cmap = cm.get_cmap('Set1' if cmap is None else cmap,
-                       np.unique(colors).size)
+    # Ensure we have enough colors to uniquely identify the cells
+    if cmap is None:
+        if color_by_sides and num_colors <= 10:
+            cmap = 'tab10'
+        else:
+            cmap = 'tab20'
+    cmap = cm.get_cmap(cmap, num_colors)
```
vyasr, line 425: Can you add a warning to the user if they try to plot with > 20? Just tell the user to provide a cmap in that case (no need to error though).

tests/test_locality_Voronoi.py
```diff
@@ -26,7 +26,22 @@ def test_basic_2d(self):
...
-    def test_basic_3d(self):
+        # Ensure every point has neighbors
+        assert np.all(vor.nlist.neighbor_counts > 0)
+
+        # Ensure neighbor list is symmetric
+        ijs = set(zip(vor.nlist.query_point_indices, vor.nlist.point_indices))
+        jis = set(zip(vor.nlist.point_indices, vor.nlist.query_point_indices))
+
+        # Every (i, j) pair should have a corresponding (j, i) pair
+        self.assertTrue(all((j, i) in jis for (i, j) in ijs))
```
vyasr, line 37: I think some of the logic in the last 4 lines (inside the assertion and the construction of ijs/jis) is also implemented `test_locality_NeighborQuery.py`, I'd try to steal that if you can (and modify if needed).

tests/test_locality_Voronoi.py
```diff
@@ -46,6 +61,16 @@ def test_basic_3d(self):
...
 
+        # Ensure every point has neighbors
+        assert np.all(vor.nlist.neighbor_counts > 0)
+
+        # Ensure neighbor list is symmetric
+        ijs = set(zip(vor.nlist.query_point_indices, vor.nlist.point_indices))
+        jis = set(zip(vor.nlist.point_indices, vor.nlist.query_point_indices))
+
+        # Every (i, j) pair should have a corresponding (j, i) pair
+        self.assertTrue(all((j, i) in jis for (i, j) in ijs))
```
vyasr, line 72: Same comment as above.

cpp/locality/Voronoi.cc
bdice: @vyasr and I discussed and preferred to use 0.5 as the tolerance, which ensures that we "split the difference" between 0 and 1. We hope it is a more robust choice.
```suggestion
                if (box.is2D() && std::abs(normal.z) > 0.5)
```

vyasr (changes requested):

vyasr (approved):

PR #590, opened 2020-02-25
Construct NeighborLists without distances, box functions to calculate distances from points

### LINKED ISSUE #368: NeighborLists need distances
We need to clarify the intended behavior of the `Steinhardt` class (and probably related classes) when using the `num_neighbors` option. If there are more than `num_neighbors` present, should the neighbors be sorted first by distance/weight?

Also neighbor list distances are required for the `Steinhardt` class to work properly, but manually-constructed neighbor lists currently can't specify distances (though the array is writable via code like `nlist.distances[:] = np.sqrt(2)`). (The LocalQl class computes its distances explicitly so this problem isn't seen there.)

klywang:
## Description
Added `freud.locality.NeighborList.from_points(box, point, query_points, query_point_indices, point_indices, weights)`. Allows for creation of custom `NeighborList`s without pre-computing distances. Distances are automatically calculated using the newly added `freud.box.Box.compute_distances(points, query_points)`, which calculated distances between two Nx3 arrays.

In addition, `freud.box.Box.compute_all_distances(point, query_points)` was added to calculate the distance between of a point, or a list of points, and all query points.

## Motivation and Context

Resolves: #368 

## How Has This Been Tested?

Added unit tests for `from_points`, `compute_distances`, and `compute_all_distances`. Functions are tested with points in a non-orthoganol box, near the boundary.

## Types of changes

## Checklist:


ChangeLog.md
```diff
@@ -4,6 +4,16 @@ The format is based on
 and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## v2.3.0
+
+### Added
+* Custom NeighborLists can be created from a set of points using `from_points`. 
+Distances will automatically be calculated using periodic boundary conditions.
```
bdice: In this file, write long lines instead of having a line break.

cpp/box/Box.h
```diff
@@ -381,6 +382,53 @@ class Box
...
+            for (size_t i = begin_n; i < end_n; ++i)
+            {
+                util::forLoopWrapper(0, Mvecs, [=](size_t begin_m, size_t end_m) {
+                    for (size_t j = begin_m; j < end_m; ++j)
+                    {
+                        dist[i*Mvecs + j] = computeDistance(points[i], query_points[j]);
+                    }
+                });
+            }
+        std::cout.flush();
```
bdice: You'll want to remove this line, which I assume was for testing. Also remove the `iostream` above.

cpp/box/Box.h
```diff
@@ -381,6 +382,53 @@ class Box
         });
     }
 
+    //! Calculate distance between two points using boundary conditions
+    /*! \param p_i Particle position
```
```suggestion
    /*! \param p_i Position of first point
```

cpp/box/Box.h
```diff
@@ -381,6 +382,53 @@ class Box
...
+        \param p_j Particle position to calculate distances from
```
```suggestion
        \param p_j Position of second point
```

cpp/box/Box.h
```diff
@@ -381,6 +382,53 @@ class Box
...
+            return std::sqrt(dot(v_ij, v_ij));
+    }
+
+    //! Calculate distance between a set of points and query points using boundary conditions
+    /*! \param points Particle positions
+        \param query_points Particle position to calculate distances from
+        \param dist Distances between points and query_points
+        \param Nvecs The number of points to calculate distances between
+    */
+    void computeDistances(vec3<float>* points, vec3<float>* query_points, float *dist, unsigned int Nvecs) const
```
bdice: I think I want to see `query_points` listed before `points`, everywhere in this pull request. This follows the convention that bonds point from query points to points. I would like @vyasr to confirm my preference before you rewrite all of the code, however.

cpp/box/Box.h
```diff
@@ -381,6 +382,53 @@ class Box
         });
     }
 
+    //! Calculate distance between two points using boundary conditions
+    /*! \param p_i Particle position
+        \param p_j Particle position to calculate distances from
+    */
+    float computeDistance(const vec3<float>& p_i, const vec3<float>& p_j) const
```
bdice: I would prefer `r_i` over `p_i` for naming consistency with the rest of freud's code base.

cpp/box/Box.h
```diff
@@ -381,6 +382,53 @@ class Box
...
+    {
+            vec3<float> v_ij = wrap(p_j - p_i);
```
bdice: Similarly, I would prefer `r_ij`.

cpp/box/Box.h
bdice: The input arrays can be `const`. Also I prefer `distances` over `dist` to make it clear that it's plural (array-like).
```suggestion
    void computeDistances(const vec3<float>* points, const vec3<float>* query_points, float* distances, const unsigned int Nvecs) const
```

cpp/box/Box.h
```diff
@@ -381,6 +382,53 @@ class Box
...
+            }
+        });
+    }
+
+    //! Calculate distance between a set of points and query points using boundary conditions
+    /*! \param points Particle positions
+        \param query_points Particle position to calculate distances from
+    */
+    void computeAllDistances(vec3<float>* points, vec3<float>* query_points,
+        float* dist, unsigned int Nvecs, unsigned int Mvecs) const
```
bdice: I would like `Nvecs` to be named so that it's clear it corresponds to either points or query points. e.g. `N_points` or `N_query_points`. Similarly, `Mvecs` should be the other one.

cpp/box/Box.h
```diff
@@ -381,6 +382,53 @@ class Box
...
+    {
+        util::forLoopWrapper(0, Nvecs, [=](size_t begin_n, size_t end_n) {
+            for (size_t i = begin_n; i < end_n; ++i)
+            {
+                util::forLoopWrapper(0, Mvecs, [=](size_t begin_m, size_t end_m) {
```
bdice: The `util::forLoopWrapper` is for parallelization (though the name isn't clear about that). I believe this nesting of the wrapper will trigger parallel execution inside of parallel execution, which sounds like a bad idea. I think the inner loop should be a plain `for`.

freud/_box.pxd
```diff
@@ -48,6 +48,11 @@ cdef extern from "Box.h" namespace "freud::box":
                     unsigned int) const
         vec3[float] centerOfMass(vec3[float]*, size_t, float*) const
         void center(vec3[float]*, size_t, float*) const
+        void computeDistances(vec3[float]* points, vec3[float]* query_points,
```
bdice: We haven't named the arguments elsewhere in this definition file, so I think you can remove the names `points`, etc. for consistency.

freud/box.pyx
```diff
@@ -422,6 +422,84 @@ cdef class Box:
         self.thisptr.center(<vec3[float]*> &l_points[0, 0], Np, l_masses_ptr)
         return vecs
 
+    def compute_distances(self, points, query_points):
+        R"""Calculate distances between two sets of points, using periodic boundaries.
+
+        Args:
+            points (:math:`\left(N, 3, \right)` :class:`numpy.ndarray`):
+                Array of points corresponding to points in a data structure.
```
bdice: The meaning of the "data structure" bit isn't clear to me.
```suggestion
                Array of points.
```

freud/box.pyx
```diff
@@ -422,6 +422,84 @@ cdef class Box:
...
+            query_points (:math:`\left(N, 3, \right)` :class:`numpy.ndarray`):
+                Array of query points that corresponds to a set of points.
```
bdice:
```suggestion
                Array of query points of the same length as points.
```
Or, if you switch the order of `points` and `query_points`, swap the definition (e.g. `Array of points of the same length as query_points.`)

freud/locality.pyx
```diff
@@ -524,6 +524,88 @@ cdef class NeighborList:
...
+        """  # noqa 501
+
+        query_point_indices = np.asarray(query_point_indices)
+        point_indices = np.asarray(point_indices)
+
+        if query_point_indices.shape != point_indices.shape:
+            raise ValueError(
+                "Points and query points indices have shapes {} and {}".format(
+                    points.shape, query_points.shape),
+                "The length of point index and query point index arrays must be equal."  # noqa 501
```
```suggestion
                "The length of point indices and query point indices must be equal."  # noqa 501
```

```suggestion
                "The length of point indices and query point indices must be equal."  # noqa 501
```

freud/locality.pyx
```diff
@@ -524,6 +524,88 @@ cdef class NeighborList:
...
+                Array of per-bond weights (if :code:`None` is given, use a
+                value of 1 for each weight) (Default value = :code:`None`).
+        """  # noqa 501
+
+        query_point_indices = np.asarray(query_point_indices)
+        point_indices = np.asarray(point_indices)
+
+        if query_point_indices.shape != point_indices.shape:
+            raise ValueError(
+                "Points and query points indices have shapes {} and {}".format(
```
bdice: You probably need to combine these strings -- it looks like you're passing the message as two separate arguments to `ValueError`.
```suggestion
                "Points and query points indices have shapes {} and {}. ".format(
```

freud/locality.pyx
```diff
@@ -524,6 +524,88 @@ cdef class NeighborList:
...
+                    points.shape, query_points.shape),
+                "The length of point index and query point index arrays must be equal."  # noqa 501
+            )
+
+        if np.amax(query_point_indices) >= query_points.shape[0]:
```
bdice: In general we prefer to do as much as possible (including error handling) in C++. I would like to consider moving this entire method into C++. Let me know if you want help with that.

tests/test_box_Box.py
```diff
@@ -439,6 +439,49 @@ def test_plot_2d(self):
...
+        box = freud.box.Box(2, 3, 4, 1, 0, 0)
+        testpoints = np.array([[-2.2, -1.3, 2], [0., 0., 1.], [0., 0., 0.1]])
+        testpoints_query = np.array(
+            [[-0.5, -1.3, 2.], [0.2, 0., 1.], [0., -0.4, 0.1]])
+        distance = box.compute_distances(testpoints, testpoints_query)
+        npt.assert_allclose(distance, [0.3, 0.2, 0.4], rtol=1e-6)
+
+    def test_compute_all_distances_2d(self):
+        box = freud.box.Box(2, 3, 0, 1, 0, 0, is2D=True)
+        testpoints = np.array([[0., 0., 0.]])
```
bdice: Just use the names `points` and `query_points` for these arrays.

tests/test_box_Box.py
```diff
@@ -439,6 +439,49 @@ def test_plot_2d(self):
...
+        testpoints_query = np.array(
+            [[0.2, 0., 0.], [0., -0.4, 0.], [1., 1., 0.]])
+        distance = box.compute_all_distances(testpoints, testpoints_query)
```
bdice: Pluralize array name.
```suggestion
        distances = box.compute_all_distances(testpoints, testpoints_query)
```

tests/test_locality_NeighborList.py
```diff
@@ -180,6 +180,62 @@ def test_from_arrays(self):
             nlist = freud.locality.NeighborList.from_arrays(
                 4, 4, query_point_indices, point_indices, distances, weights)
 
+    def test_from_points(self):
+        box = freud.box.Box(2, 3, 4, 1, 0, 0)
+        testpoints = np.array([[0, 0, 0], [-2.2, -1.3, 2]])
+        testpoints_query = np.array(
+            [[-0.5, -1.3, 2.], [0.5, 0, 0], [-2.2, -1.3, 2.], [0, 0, 0.2]])
+        testpoint_indices = np.array([1, 0, 1, 0])
+        testquery_point_indices = np.array([0, 1, 2, 3])
```
bdice: NeighborLists are always sorted by query point index. Do we enforce that the query point indices are sorted in the input, or do we perform that sorting? If perform that sorting on behalf of the user, I would like to be sure we test it by providing query point indices that are out of order. I can't recall.

bdice (changes requested):
Thanks for the PR! I appreciate this, it's very high-quality work. Most of my comments are about variables names - this is just a matter of consistency with the rest of freud and communicating expectations clearly. Give it a look and we'll move forward from there.

cpp/box/Box.h
vyasr: I'm fine with this change.

cpp/box/Box.h
vyasr: Actually nesting parallel loops like this is perfectly safe to do, especially since each iteration accesses a different index. In practice, we may not see much in the way of performance gains, but there's no harm in doing so. I could imagine use-cases where `Mvecs >> NVecs`, in which case this could be beneficial.

bdice: Okay, it seems like this is supported and would probably be good enough as it is. Ideally we would replace this with something like `tbb::blocked_range2d` for the best performance. See references:

https://stackoverflow.com/questions/29719787/parallel-more-than-one-nested-loops-with-tbb
https://software.intel.com/en-us/node/506147

vyasr: Yes, I thought about this as well, and I agree that `blocked_range2d` is potentially a more performant option. However, since it makes the code less readable I'd prefer not to prematurely optimize until we encounter meaningful performance hits with this version of the code.

bdice: I don't think it's any less readable and I don't think it's premature to change since it's a pretty simple alteration and it's an O(N^2) algorithm... but I won't hold up this PR for that.

vyasr: @klywang up to you if you would like to give this a shot. It shouldn't be too difficult to switch to using a `blocked_range2d`, but let me or Bradley know if you have trouble with it. I find it a bit less readable than for loops, but it should be faster in general so worth a shot if you're up for it.

tests/test_box_Box.py
klywang: I simplified the variable names in a new commit. I originally chose `testpoints` and `testpoints_query` for consistency with some previous tests, ie)
https://github.com/glotzerlab/freud/blob/1e99ce9e05113514725db1def1cecbfacea5cd60/tests/test_box_Box.py#L85-L92
I can change these test variables as well if this type of consistency is important.

tests/test_locality_NeighborList.py
klywang: We enforce that query point indices are sorted in the input. I can add something to test for the error.

bdice: Adding a test that raises an error for out-of-order query point indices in `test_locality_NeighborList.py` would be wonderful!

tests/test_box_Box.py
bdice: Overall I prefer `points` over `testpoints`. It's obviously a test. And `testpoints_query` is long and awka-awkward because they're normally called `query_points`, not `points_query`. You can rename just your new code and leave the old code as-is.

cpp/box/Box.h
```diff
@@ -396,18 +396,26 @@ class Box
         \param dist Distances between points and query_points
         \param Nvecs The number of points to calculate distances between
     */
+
```
bdice: Remove newline.

cpp/box/Box.h
```diff
@@ -396,18 +396,26 @@ class Box
...
-        const unsigned int* point_indices, float *dist, unsigned int Mvecs, unsigned int Nvecs) const
+        const unsigned int* point_indices, float *dist, unsigned int n_query_points, unsigned int n_points,
+        unsigned int n_query_indices, unsigned int n_point_indices) const
     {
-        if (Mvecs != Nvecs)
-            throw std::runtime_error("Inconsistent array sizes.");
-        util::forLoopWrapper(0, Nvecs, [=](size_t begin, size_t end) {
+        if (n_query_indices != n_point_indices)
+        {
+            throw std::runtime_error("Query point and point index array dimesnsions must match.");
```
```suggestion
            throw std::runtime_error("The number of query point indices and point indices must match.");
```

cpp/box/Box.h
```diff
@@ -396,18 +396,26 @@ class Box
...
+        util::forLoopWrapper(0, n_query_indices, [=](size_t begin, size_t end) {
             for (size_t i = begin; i < end; ++i)
             {
-                if (query_point_indices[i] > Mvecs)
-                    throw std::out_of_range("Index too big.");
-                if (point_indices[i] > Nvecs)
-                    throw std::out_of_range("Index too big.");
+                if (query_point_indices[i] >= n_query_points)
+                {
+                    throw std::out_of_range("Query point indices exceed array dimensions.");
```
bdice: Add `#include <sstream>` at the top of this file.
```suggestion
                    std::ostringstream msg;
                    msg << "Query point index " << query_point_indices[i] << " is out of range for " << n_query_points << " provided query points."
                    throw std::out_of_range(msg.str());
```

cpp/box/Box.h
```diff
@@ -396,18 +396,26 @@ class Box
...
+                }
+                if (point_indices[i] >= n_points)
+                {
+                    throw std::out_of_range("Point indices exceed array dimensions.");
```
```suggestion
                    std::ostringstream msg;
                    msg << "Point index " << point_indices[i] << " is out of range for " << n_points << " provided points."
                    throw std::out_of_range(msg.str());
```

cpp/box/Box.h
```diff
@@ -418,15 +426,15 @@ class Box
...
     void computeAllDistances(vec3<float>* query_points, vec3<float>* points,
-        float* dist, unsigned int Mvecs, unsigned int Nvecs) const
+        float* dist, unsigned int n_query_points, unsigned int n_points) const
     {
-        util::forLoopWrapper(0, Mvecs, [=](size_t begin_m, size_t end_m) {
+        util::forLoopWrapper(0, n_query_points, [=](size_t begin_m, size_t end_m) {
             for (size_t j = begin_m; j < end_m; ++j)
             {
-                util::forLoopWrapper(0, Nvecs, [=](size_t begin_n, size_t end_n) {
+                util::forLoopWrapper(0, n_points, [=](size_t begin_n, size_t end_n) {
```
bdice: I know we had a previous discussion on this, but before this PR is merged, I would like to run a benchmark on the vislab where we compute all distances between a large array of random points (maybe 10,000?) and use a normal (serial) for loop on the inside loop, outside loop, and both loops. I just want to confirm that TBB is parallelizing this reasonably.

freud/_box.pxd
```diff
@@ -49,12 +49,18 @@ cdef extern from "Box.h" namespace "freud::box":
         vec3[float] centerOfMass(vec3[float]*, size_t, float*) const
         void center(vec3[float]*, size_t, float*) const
         void computeDistances(vec3[float]* query_points, vec3[float]* points,
-                              unsigned int* query_point_indices, unsigned int* point_indices,
-                              float* dist, unsigned int Mvecs, unsigned int Nvecs
+                              unsigned int* query_point_indices,
+                              unsigned int* point_indices,
+                              float* dist, unsigned int n_query_points,
```
bdice:
```suggestion
                              float* dist,
                              unsigned int n_query_points,
```

freud/_box.pxd
```diff
@@ -49,12 +49,18 @@ cdef extern from "Box.h" namespace "freud::box":
...
+                              unsigned int n_points,
+                              unsigned int n_query_indices,
+                              unsigned int n_point_indices
                               ) except +
         void computeAllDistances(vec3[float]* query_points,
                                  vec3[float]* points, float* dist,
```
bdice:
```suggestion
                                 vec3[float]* points,
                                 float* dist,
```

freud/box.pyx
```diff
@@ -422,6 +422,100 @@ cdef class Box:
...
+            &l_query_point_indices[0], &l_point_indices[0],
+            <float *> &dist[0], n_query_points, n_points,
+            n_query_indices, n_point_indices
+        )
+
+        return np.squeeze(dist) if flatten else np.asarray(dist)
+
+    def compute_all_distances(self, query_points, points):
+        R"""Calculate distances between a set of points and all query points,
+        with periodic boundary conditions.
```
bdice: I'd like to avoid specific references to periodic boundary conditions if we decide to relax freud's current behavior where all boxes are periodic in every direction.
```suggestion
        R"""Calculate distances between a set of points and all query points.
```

freud/box.pyx
```diff
@@ -422,6 +422,100 @@ cdef class Box:
...
+
+        Args:
+            query_points (:math:`\left(N_{query_points}, 3 \right)` :class:`numpy.ndarray`):
```
bdice: You have to escape the second underscore so it prints as a literal underscore and doesn't subscript the p in "points".
```suggestion
            query_points (:math:`\left(N_{query\_points}, 3 \right)` :class:`numpy.ndarray`):
```

freud/box.pyx
```diff
@@ -422,6 +422,100 @@ cdef class Box:
...
+                Array of query points to calculate
+            points (:math:`\left(N_{points}, 3 \right)` :class:`numpy.ndarray`):
+                Array of points corresponding to points in a data structure.
+
+        Returns:
+            :math:`\left(N_{points}, N_{query_points}, \right)` :class:`numpy.ndarray`:
```
bdice:
```suggestion
            :math:`\left(N_{points}, N_{query\_points}, \right)` :class:`numpy.ndarray`:
```

cpp/box/Box.h
klywang: I made a new function(?) in `utils.h` called `ForLoopWrapper2D` using `blocked_range2d`. I can move it out of `utils.h` and implement it directly in `Box::computeAllDistances` if needed.

klywang:
@vyasr @bdice This should be ready for review again!

ChangeLog.md
```diff
@@ -4,6 +4,15 @@ The format is based on
 and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## v2.3.0 - 2020-xx-xx
+
+### Added
+* Custom NeighborLists can be created from a set of points using `from_points`.
+Distances will automatically be calculated using periodic boundary conditions.
+* The Box class has methods `compute_distances` and `compute_all_distances` to calculate distances from arrays of points and query points using periodic boundary conditions.
```
```suggestion
* The Box class has methods `compute_distances` and `compute_all_distances` to calculate distances between arrays of points and query points.
```

ChangeLog.md
```diff
@@ -4,6 +4,15 @@ The format is based on
 and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## v2.3.0 - 2020-xx-xx
+
+### Added
+* Custom NeighborLists can be created from a set of points using `from_points`.
+Distances will automatically be calculated using periodic boundary conditions.
```
bdice: Looks like you marked this as resolved without changing it.
```suggestion
* Custom NeighborLists can be created from a set of points using `from_points`. Distances will be calculated automatically.
```

cpp/box/Box.h
```diff
@@ -380,6 +380,75 @@ class Box
         });
     }
 
+    //! Calculate distance between two points using boundary conditions
+    /*! \param r_i Position of first point
+        \param r_j Position of second point
+    */
+    float computeDistance(const vec3<float>& r_j, const vec3<float>& r_i) const
```
bdice: Make this match the order of the docstring and also "from r_i to r_j" convention.
```suggestion
    float computeDistance(const vec3<float>& r_i, const vec3<float>& r_j) const
```

cpp/box/Box.h
```diff
@@ -380,6 +380,75 @@ class Box
...
+                {
+                    throw std::out_of_range("Point indices exceed number of points.");
+                }
+                distances[i] = computeDistance(query_points[query_point_indices[i]], points[point_indices[i]]);
+            }
+        });
+    }
+
+    //! Calculate distance between a set of points and query points using boundary conditions
+    /*! \param points Particle positions
```
bdice: Swap order of arguments in docstring to match method signature.

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
         self.thisptr.center(<vec3[float]*> &l_points[0, 0], Np, l_masses_ptr)
         return vecs
 
+    def compute_distances(
+            self, query_points, points, query_point_indices, point_indices):
+        R"""Calculate distances between two sets of points, using periodic boundaries.
+
+        Args:
+            query_points (:math:`\left(N, 3, \right)` :class:`numpy.ndarray`):
```
bdice:
```suggestion
            query_points (:math:`\left(N, 3\right)` :class:`numpy.ndarray`):
```

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+                Array of query points.
+            points (:math:`\left(N, 3, \right)` :class:`numpy.ndarray`):
```
bdice:
```suggestion
            points (:math:`\left(N, 3\right)` :class:`numpy.ndarray`):
```

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+                Array of points.
+            query_points_indices (:math:`\left(N, \right)` :class:`numpy.ndarray`):
+                Array of indices that correspond to a set of query points.
+            points_indices (:math:`\left(N, \right)` :class:`numpy.ndarray`):
+                Array of indices that correspond to a set of points, of same length and query_points_indices.
```
bdice:
```suggestion
                Array of indices that correspond to a set of points, of same length as ``query_points_indices``.
```

bdice: Do we need some mechanism for re-using `query_points` as `points` or `query_point_indices` as `point_indices` or vice versa? Just trying to make sure this is the right API design. @vyasr please weigh in.

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+
+        Returns:
+            :math:`\left(N, \right)` :class:`numpy.ndarray`:
+                Array of distances between query points and points.
+        """   # noqa: E501
+        query_points = np.asarray(query_points)
+        points = np.asarray(points)
+        query_point_indices = np.atleast_1d(query_point_indices)
```
bdice: Is there another function that uses `atleast_1d` that you copied this from? I think you need `asarray` here, and let `_convert_array` do the rest.

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+        query_points = np.atleast_2d(query_points)
+        points = np.atleast_2d(points)
+
+        query_points = freud.util._convert_array(query_points, shape=(None, 3))
+        points = freud.util._convert_array(points, shape=(None, 3))
+
+        query_point_indices = freud.util._convert_array(
+            query_point_indices, shape=(None,), dtype=np.uint32)
+        point_indices = freud.util._convert_array(
+            point_indices, shape=(None,), dtype=np.uint32)
```
bdice: Enforce the shape -- must match `query_point_indices`.
```suggestion
            point_indices, shape=query_point_indices.shape, dtype=np.uint32)
```

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+            const float[:, ::1] l_points = points
+            const unsigned int[::1] l_query_point_indices = \
+                query_point_indices
+            const unsigned int[::1] l_point_indices = point_indices
+            size_t n_query_points = query_points.shape[0]
+            size_t n_points = points.shape[0]
+            size_t n_query_indices = query_point_indices.shape[0]
+            size_t n_point_indices = point_indices.shape[0]
+            float[::1] dist = np.empty(
+                query_point_indices.shape[0], dtype=np.float32)
```
bdice:
```suggestion
                n_query_indices, dtype=np.float32)
```

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+
+        self.thisptr.computeDistances(
+            <vec3[float]*> &l_query_points[0, 0],
+            <vec3[float]*> &l_points[0, 0],
+            &l_query_point_indices[0], &l_point_indices[0],
+            <float *> &dist[0], n_query_points, n_points,
```
bdice: I don't think you need the cast to `float *` in this case.
```suggestion
            &dist[0], n_query_points, n_points,
```

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+        Returns:
+            :math:`\left(N, \right)` :class:`numpy.ndarray`:
+                Array of distances between query points and points.
+        """   # noqa: E501
+        query_points = np.asarray(query_points)
+        points = np.asarray(points)
+        query_point_indices = np.atleast_1d(query_point_indices)
+        point_indices = np.atleast_1d(point_indices)
+
+        flatten = query_point_indices.ndim == 1
```
bdice: Are you wanting to explicitly support multidimensional arrays for `query_point_indices`? I would prefer that we require 1D arrays / lists only. The `flatten` behavior for other functions is designed to allow things like `box.wrap([1, 1, 1])` to return `[1, 1, 1]` and `box.wrap([[1, 1, 1]])` to return `[[1, 1, 1]]`. That is, we allow single vectors (but treat them as an array of vectors) and match the return type by squeezing the array of 1 vector back to a single vector.

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+
+        self.thisptr.computeDistances(
+            <vec3[float]*> &l_query_points[0, 0],
+            <vec3[float]*> &l_points[0, 0],
+            &l_query_point_indices[0], &l_point_indices[0],
+            <float *> &dist[0], n_query_points, n_points,
+            n_query_indices, n_point_indices
+        )
+
+        return np.squeeze(dist) if flatten else np.asarray(dist)
```
bdice: If removing the `flatten` behavior:
```suggestion
        return np.asarray(dist)
```

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+            const float[:, ::1] l_points = points
+            size_t n_query_points = query_points.shape[0]
+            size_t n_points = points.shape[0]
+            float[:, ::1] dist = np.empty(
+                [query_points.shape[0], points.shape[0]], dtype=np.float32)
+
+        self.thisptr.computeAllDistances(
+            <vec3[float]*> &l_query_points[0, 0],
+            <vec3[float]*> &l_points[0, 0],
+            <float *> &dist[0, 0], n_query_points, n_points
```
bdice:
```suggestion
            &dist[0, 0], n_query_points, n_points
```

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+                Array of points with same length as query_points.
+
+        Returns:
+            :math:`\left(N_{points}, N_{query_points}, \right)` :class:`numpy.ndarray`:
+                Array of distances between corresponding query points and points.
+        """  # noqa: E501
+        query_points = np.asarray(query_points)
+        points = np.asarray(points)
+
+        flatten = query_points.ndim == 1
```
bdice: No need for `flatten` -- see conversations above.

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+        query_points = np.atleast_2d(query_points)
```
bdice: I think you can skip the step `query_points = np.asarray(query_points)` if you are calling `np.atleast_2d` on the same data.

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+        points = freud.util._convert_array(
+            points, shape=(None, 3))
+
+        cdef:
+            const float[:, ::1] l_query_points = query_points
+            const float[:, ::1] l_points = points
+            size_t n_query_points = query_points.shape[0]
+            size_t n_points = points.shape[0]
+            float[:, ::1] dist = np.empty(
+                [query_points.shape[0], points.shape[0]], dtype=np.float32)
```
bdice:
```suggestion
                [n_query_points, n_points], dtype=np.float32)
```

freud/box.pyx
```diff
@@ -422,6 +422,104 @@ cdef class Box:
...
+
+        self.thisptr.computeAllDistances(
+            <vec3[float]*> &l_query_points[0, 0],
+            <vec3[float]*> &l_points[0, 0],
+            <float *> &dist[0, 0], n_query_points, n_points
+        )
+        return np.squeeze(dist) if flatten else np.asarray(dist)
```
bdice:
```suggestion
        return np.asarray(dist)
```

freud/locality.pyx
```diff
@@ -524,6 +524,72 @@ cdef class NeighborList:
...
+            point_indices, weights=None
+    ):
+        R"""Create a NeighborList from a set of bond information arrays
+        when distances are not known.
+
+        Args:
+            box (:class:`~.box.Box`):
+                Simulation box.
+            query_points (:class:`np.ndarray`):
+                Array of floats corresponding to the positions of a set of query points.
```
bdice: Don't specify `float` -- copy a docstring from another method for an example. The important part is the shape, not the data type.

freud/locality.pyx
```diff
@@ -524,6 +524,72 @@ cdef class NeighborList:
...
+            points (:class:`np.ndarray`):
+                Array of floats corresponding to the positions of a set of points.
+            query_point_indices (:class:`np.ndarray`):
+                Array of integers corresponding to indices in the set of
+                query points.
```
```suggestion
                Array of indices corresponding to ``query_points``.
```

freud/locality.pyx
```diff
@@ -524,6 +524,72 @@ cdef class NeighborList:
...
+            point_indices (:class:`np.ndarray`):
+                Array of integers corresponding to indices in the set of
+                points.
```
```suggestion
                Array of indices corresponding to ``points``. Must match the
                length of ``query_point_indices``.
```

freud/locality.pyx
```diff
@@ -524,6 +524,72 @@ cdef class NeighborList:
...
+            weights (:class:`np.ndarray`, optional):
+                Array of per-bond weights (if :code:`None` is given, use a
+                value of 1 for each weight) (Default value = :code:`None`).
+        """  # noqa 501
+
+        query_point_indices = np.asarray(query_point_indices)
+        point_indices = np.asarray(point_indices)
+
+        distances = box.compute_distances(
```
bdice: It would be good to allow duck-typed boxes (e.g. matrices, tuples of `(Lx, Ly, Lz, ...)`) by using the `_convert_box` method before calling `compute_distances`. See elsewhere in `locality.pyx` for examples.

Also, move the `compute_distances` call lower in the function, so that all data conversions and validation checks occur first. Move it below `weights = freud.util._convert_array(...)`.

tests/test_box_Box.py
```diff
@@ -439,6 +439,67 @@ def test_plot_2d(self):
...
+        points = np.array([[0, 0, 0], [-2.2, -1.3, 0]])
+        query_points = np.array(
+            [[-0.5, -1.3, 0.], [0.5, 0, 0], [-2.2, -1.3, 0.], [0, 0.4, 0]])
+        point_indices = np.array([1, 0, 1, 0])
+        query_point_indices = np.array([0, 1, 2, 3])
+        distances = box.compute_distances(
+            query_points, points, query_point_indices, point_indices)
+        npt.assert_allclose(distances, [0.3, 0.5, 0.0, 0.4], rtol=1e-6)
+
+        # # 1 dimensional array
```
bdice:
```suggestion
        # 1 dimensional array
```

tests/test_box_Box.py
```diff
@@ -439,6 +439,67 @@ def test_plot_2d(self):
...
+
+        with self.assertRaises(IndexError):
+            box.compute_distances(
+                query_points, points, query_point_indices, point_indices + 1)
+        with self.assertRaises(IndexError):
+            box.compute_distances(
+                query_points, points, query_point_indices + 1, point_indices)
+        with self.assertRaises(RuntimeError):
+            box.compute_distances(
+                query_points, points, query_point_indices, point_indices[:-1])
```
bdice: Is this raising because the length of `query_point_indices` doesn't match the length of `point_indices`? If so, that should be a `ValueError`, not a `RuntimeError`. You might need to change the C++ exception that is thrown so that the Python exception corresponds to that.

tests/test_box_Box.py
```diff
@@ -439,6 +439,67 @@ def test_plot_2d(self):
...
+        query_points = np.array(
+            [[0.2, 0., 0.], [0., -0.4, 0.], [1., 1., 0.]])
+        distances = box.compute_all_distances(points, query_points)
+        npt.assert_allclose(distances, [[0.2, 0.4, np.sqrt(2)]], rtol=1e-6)
+
+        points = np.array([0., 0., 0.])
+        distances = box.compute_all_distances(points, query_points)
+        npt.assert_allclose(distances, [0.2, 0.4, np.sqrt(2)], rtol=1e-6)
+
+    def test_compute_all_distances_3d(self):
```
bdice, line 489: Be sure one of these tests computes the distance between points across the periodic boundaries.

bdice (changes requested):
Some additional comments. Please revert the changes to `box.cpp`. We re-generate those Cython C++ files when we make a new release, but we don't commit them for every PR because they're huge and would bloat the repository size.

tests/test_box_Box.py
klywang: `test_compute_distances_3d` computes distances across periodic boundaries on one of the tilted box sides.

klywang: One of the distances in `test_compute_all_distances_3d` is across the periodic boundary on a tilted face.

freud/box.pyx
klywang: This case is a bit different from `compute_distances` (where I agree `flatten` was unnecessary). If I provide a single query point (i.e. `[0, 0, 0]`) and a set of points (ie `[[0, 1, 0], [2, 0, 0]]`), I would expect something like `distances=[1, 2]`, rather than `distances=[[1, 2]]`. If it's best to maintain the 2D arrays in this case as well, I can remove `flatten` for `compute_all_distances` too.

freud/box.pyx
klywang: I did this in case someone tries to use specify indicies as an integer instead of an array. I will remove this if we want to discourage this kind of input.

freud/box.pyx
bdice: @klywang A couple considerations:
- The behavior should be symmetric (same with one query point and multiple points as with one point and multiple query points).
- What should happen with one point and one query point? A scalar? A 1D array with one element?

Requiring 2D input for now is fine with me and we could relax that later without breaking the API if users demand it.

cpp/box/Box.h
vyasr: I think this was a good move, let's move forward with that function for now.

cpp/box/Box.h
```diff
@@ -380,6 +380,75 @@ class Box
...
+        \param n_point_indices The number of points indices.
+    */
+    void computeDistances(const vec3<float>* query_points, const vec3<float>* points,
+        const unsigned int* query_point_indices, const unsigned int* point_indices,
+        float *distances, const unsigned int n_query_points, const unsigned int n_points,
+        const unsigned int n_query_indices, const unsigned int n_point_indices) const
+    {
+        if (n_query_indices != n_point_indices)
+        {
+            throw std::invalid_argument("The length of query point indices and point indices must match.");
```
vyasr:
```suggestion
            throw std::invalid_argument("The number of query point indices and point indices must match.");
```

cpp/box/Box.h
vyasr: @klywang please implement this change, it will make error messages much more useful.

cpp/box/Box.h
vyasr: @klywang also this change.

cpp/box/Box.h
vyasr: No longer relevant since we switched to a blocked range.

freud/box.pyx
vyasr: I was very involved with the original writing of this PR, but have not reviewed recently. However, my recollection was that from an API perspective I thought the cleanest solution would be to just accept `query_points` and `points` directly. It's trivial for a user to just provide the right subset using NumPy slicing, so I would prefer `compute_distances(query_points[query_point_indices], points[point_indices])` to  `compute_distances(query_points, points, query_point_indices, point_indices)`. I would do the same thing for the distance matrix calculation.

freud/box.pyx
vyasr: I think if we follow my API suggestion on the other comment (getting rid of indices), we should do an `atleast_2d` on the `points` and `query_points` to allow single point calculation.

freud/box.pyx
vyasr: Why not? Do you think it will infer this type?

freud/box.pyx
vyasr: Same question, why?

bdice: It's a cast to its own type. It's not necessary and this isn't done elsewhere in freud except where needed, e.g. for `<vec3[float]*>`.

freud/box.pyx
vyasr: I'd be OK with always returning a matrix even for (1xn), (mx1), or (1x1) outputs. However, I agree that if we force 2D now we can generalize later without breaking APIs, so if we can't agree now then it's better to require the stricter case.

freud/box.pyx
bdice: This is a good idea. So you would merge `compute_all_distances` with `compute_distances` and just have one function, right? I would prefer the name `compute_distances` over `compute_all_distances` in that case.

cpp/box/Box.h
```diff
@@ -380,6 +380,75 @@ class Box
...
+        \param distances Distances between points and query_points.
+        \param n_query_points The number of query points.
+        \param n_points The number of points.
+        \param n_query_indices The number of query points indices.
+        \param n_point_indices The number of points indices.
+    */
+    void computeDistances(const vec3<float>* query_points, const vec3<float>* points,
+        const unsigned int* query_point_indices, const unsigned int* point_indices,
+        float *distances, const unsigned int n_query_points, const unsigned int n_points,
+        const unsigned int n_query_indices, const unsigned int n_point_indices) const
```
vyasr: Minor nitpick, but we usually order this so that the array lengths go right after the arrays, e.g. `n_query_points` comes right after `query_points`.

cpp/box/Box.h
```diff
@@ -380,6 +380,75 @@ class Box
...
+                if (point_indices[i] >= n_points)
+                {
+                    throw std::out_of_range("Point indices exceed number of points.");
+                }
+                distances[i] = computeDistance(query_points[query_point_indices[i]], points[point_indices[i]]);
+            }
+        });
+    }
+
+    //! Calculate distance between a set of points and query points using boundary conditions
```
vyasr:
```suggestion
    //! Calculate distance between a set of points and query points using periodic boundary conditions
```

cpp/util/utils.h
```diff
@@ -49,6 +49,36 @@ inline void forLoopWrapper(size_t begin, size_t end, const Body& body, bool para
     }
 }
 
-}; }; // namespace freud::util
+//! Wrapper for 2 for-loops to allow the execution in parallel or not.
```
vyasr:
```suggestion
//! Wrapper for 2 for loops to allow the execution in parallel or not.
```

freud/_box.pxd
```diff
@@ -48,6 +48,13 @@ cdef extern from "Box.h" namespace "freud::box":
                     unsigned int) const
         vec3[float] centerOfMass(vec3[float]*, size_t, float*) const
         void center(vec3[float]*, size_t, float*) const
+        void computeDistances(vec3[float]*, vec3[float]*, unsigned int*,
+                              unsigned int*, float*, unsigned int,
+                              unsigned int, unsigned int, unsigned int
+                              ) except +
+        void computeAllDistances(vec3[float]*, vec3[float]*, float*,
+                                 unsigned int, unsigned int
+                                 ) except +
```
vyasr:
```suggestion
                                 )
```
This method doesn't throw any exceptions.

cpp/box/Box.h
```diff
@@ -380,6 +380,75 @@ class Box
         });
     }
 
+    //! Calculate distance between two points using boundary conditions
+    /*! \param r_i Position of first point
+        \param r_j Position of second point
+    */
+    float computeDistance(const vec3<float>& r_i, const vec3<float>& r_j) const
```
vyasr:
```suggestion
    inline float computeDistance(const vec3<float>& r_i, const vec3<float>& r_j) const
```

cpp/box/Box.h
```diff
@@ -380,6 +380,75 @@ class Box
...
+    {
+            vec3<float> r_ij = wrap(r_j - r_i);
```
vyasr:
```suggestion
            const vec3<float> r_ij = wrap(r_j - r_i);
```

freud/box.pyx
vyasr: If we change the API as I suggested, we will instead want to enforce this on points (not `point_indices`, which will be removed).

vyasr (changes requested):
It looks good! A few small comments, but the main thing we still need to resolve is the API. Then I think we're set!

vyasr:
I realized that my API recommendation might not have been clear from my review. Let me clarify here:
I still think we should have two separate functions, because `compute_all_distances` will often be doing way more work than is necessary. There are many use cases where you have a specific set of pairs that you want to calculated distances for. My recommendation was simply to change `compute_distances` to accept two sets of points, rather than use indices. My other recommendation would be to rename `compute_all_distances` to `compute_distance_matrix` or something similarly unambiguous; I thought I typed that, but I'm not finding it anywhere.
@bdice @klywang does that make sense?

freud/box.pyx
vyasr: Not quite, see my comment in the overall PR thread.

klywang: The decision to change the API was made offline. If I remember correctly, this was related to @bdice request to move error checking from cython to C++.
When constructing NeighborLists using `from_points`, the current API does error checking for index lengths in C++. If we revert `compute_distances(query_points, points)`, one of three things needs to happen with the NeighborList's `from_points` function:
1) Error checking happens in Cython before calling `compute_distances`
2) Let `compute_distances` throw it's error about `query_points` and `points` lengths not matching. The user can infer that the indicies for either do not match.
3) A new function is written in C++ to check this error.

vyasr:
For the record: @klywang, @bdice and I had a conversation about this PR last week to resolve some of the API issues we were discussing. Our final decision was to remove the new function in the `NeighborList` class and instead require users to precompute the distances. Now that the convenience function exists in the `Box` class this calculation is a one-liner, so we just need to document it in `NeighborList.from_arrays`. The API for the distance calculations will just be `compute(All)?Distances(query_points, points)`, with no indices as separate arguments; if users have indices, they can simply pass them appropriately using NumPy slicing (e.g. `computeDistances(query_points[query_point_indices], points[point_indices])`). The minor additional overhead of copying the enlarged list is preferable to complicating the API by requiring the user to provide the indices and then performing the indexing manually within `freud`.

klywang:
@vyasr @bdice I think it's ready for review. The CI/appevyor/pr check failed on a Steindhard order parameter test, which I have not moddified, so I think things should be fine.

Also, here are the benchmark results (100 points, 100 query points):
benchmark
I can attach code snippets for the benchmarking and the distance calculations if needed.

bdice:
@klywang Thanks for that benchmark!! That's really nice to know. I suspect the 1-thread performance is better due to improved cache utilization. I'll give it a look and may push minor changes.

bdice (approved):
I pushed some changes to this PR. I approve it in its current state. The one API break that I made was regarding the use of `np.squeeze`. Before my change, providing a `(3,)` array and an `(N, 3)` array would give a result of shape `(N,)` instead of `(1, N)` because it's "squeezed." However, we aren't distinguishing between a `(3,)` array and a `(1, 3)` array in the inputs. Either the dimensionality detection needs to be a lot smarter and handle cases like that consistently, or we should just always return 2D output. I decided to always return 2D output because that's also what the docstring says it does.

vyasr:
@klywang @bdice I think the benchmark is too small to really give you much insight into how the multithreading is behaving (only 100 points). You can see that the performance is basically saturating at 2 threads, and the question is how well these scale with the number of threads. The number of neighbors is probably small enough that the benchmark is mostly capturing things like TBB and the OS's overhead for thread handling. However, I'm pretty confident that the 2D blocked range solution is the right solution anyway, so let's move forward with the PR as is.

cpp/box/Box.h
```diff
@@ -380,6 +380,61 @@ class Box
...
+            const vec3<float> r_ij = wrap(r_j - r_i);
+            return std::sqrt(dot(r_ij, r_ij));
+    }
+
+    //! Calculate distances between a set of query points and points.
+    /*! \param query_points Query point positions.
+        \param n_query_points The number of query points.
+        \param points Point positions.
+        \param n_points The number of points.
+        \param distances Distances between points and query_points (length is n_query_points).
```
vyasr:
```suggestion
        \param distances Pointer to array of length n_query_points containing distances between each point and query_point (overwritten in place).
```

cpp/box/Box.h
```diff
@@ -380,6 +380,61 @@ class Box
...
+            }
+        });
+    }
+
+    //! Calculate all pairwise distances between a set of query points and points.
+    /*! \param query_points Query point positions.
+        \param n_query_points The number of query points.
+        \param points Point positions.
+        \param n_points The number of points.
+        \param distances Distances between points and query_points (length is n_query_points*n_points).
```
vyasr:
```suggestion
        \param distances Pointer to array of length n_query_points*n_points containing distances between points and query_points (overwritten in place).
```

cpp/box/Box.h
```diff
@@ -380,6 +380,61 @@ class Box
...
+        \param distances Distances between points and query_points (length is n_query_points).
+    */
+    void computeDistances(const vec3<float>* query_points, const unsigned int n_query_points,
+        const vec3<float>* points, const unsigned int n_points, float *distances) const
+    {
+        if (n_query_points != n_points)
+        {
+            throw std::invalid_argument("The number of query points and points must match.");
+        }
+        util::forLoopWrapper(0, n_query_points, [=](size_t begin, size_t end) {
```
vyasr:
```suggestion
        util::forLoopWrapper(0, n_query_points, [&](size_t begin, size_t end) {
```

This will capture all the inputs by reference. That does make it the inner function's responsibility not to change the values but in this case that looks safe and saves some copying.

cpp/box/Box.h
```diff
@@ -380,6 +380,61 @@ class Box
...
+    /*! \param query_points Query point positions.
+        \param n_query_points The number of query points.
+        \param points Point positions.
+        \param n_points The number of points.
+        \param distances Distances between points and query_points (length is n_query_points*n_points).
+    */
+    void computeAllDistances(const vec3<float>* query_points, const unsigned int n_query_points,
+        const vec3<float>* points, const unsigned int n_points, float* distances) const
+    {
+        util::forLoopWrapper2D(0, n_query_points, 0, n_points, [=](
```
vyasr:
```suggestion
        util::forLoopWrapper2D(0, n_query_points, 0, n_points, [&](
```

vyasr (commented):
I have a couple minor changes. I'm just going to apply them and I'll approve once tests pass. @klywang thanks for being a trooper! Sorry this PR has dragged so much.

vyasr (approved):

vyasr (approved):

vyasr:
For future reference, here's the performance as a function of the number of particles courtesy of @klywang. As expected, for larger numbers of particles the nested for loops become faster than a single for loop, but the blocked range is still the optimal solution in all cases.
benchmarks_all

PR #596, opened 2020-03-23
Add diffraction feature

### LINKED ISSUE #245: Diffraction module
Discussing with @mikemhenry, it would be nice to add a module for computing diffraction patterns in `freud`. There are some existing implementations that could be adapted:
- @cmelab [diffract.py](https://bitbucket.org/cmelab/cme_utils/src/e83c85aed54ee1de388e9c23ce496937da5f1214/cme_utils/analyze/diffract.py?at=master&fileviewer=file-view-default)
- platoviz [Diffraction.py](https://github.com/glotzerlab/platoviz/blob/master/platoviz/modules/Diffraction.py)

yzjin:
Add a new diffraction feature to freud
## Description
Compute diffraction patterns for crystals in colloidal systems. 
The proposed implementation right now is performing a 2D projection to square, computing Freud's gaussian density on 2D particle positions followed by a Fast Fourier Transform. We'll also take care of notes mentioned in issue #245.

## Motivation and Context
Resolves: #245 

## How Has This Been Tested?
Will add unit tests and compare the results with Injavis's diffraction diagrams.

## Types of changes

## Checklist:


freud/__init__.py
```diff
@@ -12,6 +12,7 @@
 from . import order
 from . import parallel
 from . import pmft
+from . import diffraction
```
bdice: Put this in alphabetical order (between density and environment).

freud/__init__.py
```diff
@@ -43,6 +44,7 @@
     'get_num_threads',
     'set_num_threads',
     'NumThreads',
+    'diffraction'
```
bdice: Alphabetize (see above).

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
+import numpy as np
```
bdice, line 17: Remove `diffraction.py` and put its functionality into `diffraction.pyx` or else the imports will get confused.

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+    def __init__(
+        self,
+        grid_size=512,
+        zoom=4,
+        peak_width=1,
+        length_scale=3.905,
+        bot=4e-6,
+        top=0.7
+    ):
+        """
```
bdice: Match the documentation style of other parts of freud.

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+        self.N = grid_size
+        self.zoom = zoom
+        self.peak_width = peak_width
+        self.bin_w = 2.0
+        self.length_scale = length_scale
+        self.bot = bot
+        self.top = top
+
+    def load(self, system):
```
bdice: Remove any methods that aren't used. I didn't see where this is used, at a quick glance.

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+                    )
+            self.orig = np.copy(xyz)
+            self.orig, self.image =
+                pbc.shift_pbc(xyz, np.array([L[0], L[1], L[2]]))
+        """
+        box, positions = system
+        self.box = box.to_matrix()
+        self.orig = np.copy(positions)
+
+    def pbc_2d(self, xy, N):
```
bdice: Any methods that are used internally (but shouldn't be part of the public API) should be made private:
```suggestion
    def _pbc_2d(self, xy, N):
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+        Note: orthorhombic boxes only
+        Parameters
+        ----------
+        rot : numpy.ndarray (3,3), rotation matrix
+        Returns
+        -------
+        numpy.ndarray (2,2), inverse shear matrix
+        """
+        # s = np.dot(rot.T, self.box)  # rotated box vectors
+        s = rowan.rotate(rot, self.box)
```
bdice: We'll need to update this to use box vectors (the matrix form of a box). Use `Box.to_matrix`: https://freud.readthedocs.io/en/stable/modules/box.html#freud.box.Box.to_matrix

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+        xy = np.absolute(s[0, 0] * s[1, 1] - s[0, 1] * s[1, 0])
```
bdice: These shear factors are given directly by the freud Box object. Verify the math and use those instead.

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+                [[s[0, 1], s[0, 2]], [s[1, 1], s[1, 2]]])
+        elif (zx >= xy) and (zx >= yz):
+            shear = np.array(
+                [[s[0, 2], s[0, 0]], [s[1, 2], s[1, 0]]])
+        else:
+            shear = np.array(
+                [[s[0, 0], s[0, 1]], [s[1, 0], s[1, 1]]])
+        s_det = np.linalg.det(shear)
+        if s_det == 0:
+            print("\nSingular rotation matrix. Bye Bye.")
```
bdice: This should raise a ValueError.

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+        Parameters
+        ----------
+        rot: numpy.ndarray (3, 3), rotation matrix
+        cutout: bool, return diffraction pattern with circle cutout
+        (default True)
+        Returns
+        -------
+        numpy.ndarray (N,N), diffraction pattern
+        """
+        box, positions = system
```
bdice: This will only work if the system is provided as a tuple. You should pre-process the system with the `_make_default_nq` function and then use the NeighborQuery properties for box and points.

https://github.com/glotzerlab/freud/blob/e4969d50e979ec1ec8d29af0fb62c2a48d93e83f/freud/locality.pyx#L707

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+        box = box.to_matrix()
```
bdice: Only use the matrix if necessary -- keep the freud Box object around and use its properties anywhere you can.

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+        NOTE: cut into a circle, not sure if needed-YJ
+        """
+        # if not cutout:
+        #     return dp
+
+        # idbig = self.circle_cutout(dp)
+        # dp[np.unravel_index(idbig, (self.N, self.N))] = np.log(self.bot)
+        return self.dp
+
+    def diffraction_pattern(self):
```
bdice: Make this a computed property (see decorators in MSD).

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+                :code:`None`, make a new figure and axis.
+                (Default value = :code:`None`)
+
+        Returns:
+            (:class:`matplotlib.axes.Axes`): Axis with the plot.
+        """
+        # import freud.plot
+
+
+def vector_projection(u, v):
```
bdice: Make the helper functions private (or remove if not used)

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+    u = unit_vector(u)
+    v = unit_vector(v)
+    angle = np.arccos(np.clip(np.dot(u, v), -1.0, 1.0))
+    if angle != angle:
+        # Catches nan values
+        return 0.0
+    return angle
+
+
+# def camera_to_rot(camera):
```
bdice: Delete these commented functions (we can always refer to them later if needed)

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
...
+        #     return dp
+
+        # idbig = self.circle_cutout(dp)
+        # dp[np.unravel_index(idbig, (self.N, self.N))] = np.log(self.bot)
+        return self.dp
+
+    def diffraction_pattern(self):
+        return self.dp
+
+    def plot(self, ax=None):
```
bdice: You'll want to define the plotting logic in `plot.py` and then import and call that method in `Diffraction.plot`. We don't require matplotlib as a hard dependency, and this structure allows us to use matplotlib as a "soft" dependency and raise errors if it's not available.

Also add methods for `__repr__` and `_repr_png_` like other compute objects in freud.

freud/diffraction.pyx
```diff
@@ -0,0 +1,348 @@
+import numpy as np
+from scipy import interpolate, ndimage
+import rowan
+
+# from cme_utils.manip import pbc
+# from cme_utils.manip.utilities import rotation_matrix_from_to as cme_rot
+
+
+class Diffraction:
```
bdice: See what MSD does -- you should probably inherit from `Compute`.

bdice (commented):
Here's some initial feedback.

freud/diffraction.pyx
```diff
@@ -0,0 +1,249 @@
+import numpy as np
+from scipy import interpolate, ndimage
```
bdice: Currently `scipy` is not a requirement for freud, so we must add it. I looked at the changelog and I think this code will require `scipy >=1.1`. We will need to add it to the following files (put it alphabetically, after rowan).
- `requirements.txt`
- `setup.py`, under the section `install_requires`

If we are able to use other functions in numpy instead of adding a dependency on scipy, that would be ideal, but I wouldn't worry too greatly about this.

freud/diffraction.pyx
```diff
@@ -0,0 +1,249 @@
...
+        Initialize the diffraction class
+        Parameters
+        ----------
+        grid_size : int, size of the diffraction grid (default 512)
+        zoom : (default 1)
+        peak_width : (default 1)
+        length_scale : (default 3.905)
+        bot : (default 4e-6)
+        top : (default 0.7)
+        """
```
bdice:
```suggestion
        R"""Computes a 2D diffraction pattern.
        
        The diffraction image represents the scattering of incident radiation,
        and is useful for identifying translational order present in the
        system. This class computes the static
        `structure factor <https://en.wikipedia.org/wiki/Structure_factor>`_
        :math:`S(\vec{q})` for a plane of wavevectors :math:`\vec{q}`
        orthogonal to a view plane. The view orientation :math:`(1, 0, 0, 0)`
        defaults to looking down the :math:`z` axis (at the :math:`xy` plane).
        The points in the system are converted to fractional coordinates, then
        binned into a grid whose resolution is given by ``grid_size``. The
        points are convolved with a Gaussian of width :math:`\sigma`, given by
        ``peak_width``. This convolution is performed as a multiplication in
        Fourier space.
        
        Args:
            grid_size (unsigned int):
                Size of the diffraction grid (Default value = 512).
            zoom (float):
                Scaling factor for incident wavevectors (Default value = 1).
            peak_width (float):
                Width of Gaussian convolved with points, in system length
                units (Default value = 1).
            length_scale (float):
                Not sure what this does (Default value = 3.905).
            bot (float):
                Plotting quantity -- should be removed (Default value
                = 4e-6).
            top (float):
                Plotting quantity -- should be removed (Default value
                = 0.7).
        """
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,249 @@
...
+        self.bin_w = 2.0
+        self.length_scale = length_scale
+        self.bot = bot
+        self.top = top
+
+    def pbc_2d(self, xy, N):
+        """
+        Reasonably fast periodic boundary conditions in two dimensions.
+        Normalizes xy coordinates to the grid size, N.
+        Parameters
```
bdice: This is using NumPy-style docstrings, while the rest of freud uses Google-style docstrings. Please rewrite all docstrings to use [Google style](http://google.github.io/styleguide/pyguide.html). Refer to other parts of freud and the class docstring I re-wrote above for examples.

freud/diffraction.pyx
```diff
@@ -0,0 +1,249 @@
...
+        img = ndimage.interpolation.affine_transform(img, A4, A5,
+                                                     mode="constant")
+        return img
+
+    def compute(self, system, rot, cutout=True):
+        """
+        2D FFT to get diffraction pattern from intensity matrix.
+        Parameters
+        ----------
+        rot: numpy.ndarray (3, 3), rotation matrix
```
bdice: Rename `rot` to `view_orientation` with a default value of `None` and write a docstring like:
```
            view_orientation ((:math:`4`) :class:`numpy.ndarray`, optional):
                View orientation. Uses :math:`(1, 0, 0, 0)` if not provided or :code:`None`
```

freud/plot.py
```diff
@@ -451,3 +451,29 @@ def voronoi_plot(box, polytopes, ax=None, color_by_sides=True, cmap=None):
         cb.set_label("Number of sides")
         cb.set_ticks(bounds)
     return ax
+
+
+def diffraction_plot(diffraction, ax=None, cmap='afmhot'):
+    """Helper function to plot diffraction pattern.
+
+    Args:
```
bdice, line 460: Update these argument docstrings.

freud/diffraction.pyx
```diff
@@ -0,0 +1,249 @@
...
+        Args:
+            ax (:class:`matplotlib.axes.Axes`, optional): Axis to plot on. If
+                :code:`None`, make a new figure and axis.
+                (Default value = :code:`None`)
+
+        Returns:
+            (:class:`matplotlib.axes.Axes`): Axis with the plot.
+        """
+        import freud.plot
+        return freud.plot.diffraction_plot(self.dp)
```
bdice: We want to use the "computed property" here instead of the internal quantity, so that it raises an error if the user requests a plot before the diffraction image has been computed.
```suggestion
        return freud.plot.diffraction_plot(self.diffraction)
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,249 @@
...
+        self.box, self.positions = system
+        self.box = self.box.to_matrix()
+        N = self.N / self.zoom
+        inv_shear = self.calc_proj(rot)
+        xy = np.copy(rowan.rotate(rot, self.orig)[:, 0:2])
+        xy = np.dot(xy, inv_shear.T)
+        xy = self.pbc_2d(xy, N)
+        im = self.bin(xy, N)
+
+        self.dp = np.fft.fft2(im)
```
bdice: Rename `self.dp` to `self._diffraction` everywhere.

freud/diffraction.pyx
```diff
@@ -0,0 +1,249 @@
...
+        Parameters
+        ----------
+        rot: numpy.ndarray (3, 3), rotation matrix
+        cutout: bool, return diffraction pattern with circle cutout
+        (default True)
+        Returns
+        -------
+        numpy.ndarray (N,N), diffraction pattern
+        """
+        self.box, self.positions = system
```
bdice: The first line of this method should pre-process the input `system` into a NeighborQuery. Also we should not use attributes of `self` to store `box` and `positions`. Just use `system.box` and `system.positions` where needed after pre-processing `system`.
```suggestion
        system = freud.locality._make_default_nq(system)
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,288 @@
...
+
+        # This matrix uses homogeneous coordinates. It is a 3x3 matrix that
+        # transforms 2D points and adds an offset.
+        inverse_transform = np.linalg.inv(
+            zoom_matrix @ shear_matrix @ shift_matrix)
+
+        img = scipy.ndimage.affine_transform(
+            input=img,
+            matrix=inverse_transform,
+            output_shape=(self.grid_size, self.grid_size),
```
bdice: We should add a new constructor argument `output_size` or something like that, to de-couple the final image size from the number of bins used for the grid and FFT. Then we could improve the quality of real-time visualization in OVITO.

bdice: The `k_values` should also be computed based on the `output_size`, not `grid_size`.

freud/diffraction.pyx
```diff
@@ -0,0 +1,288 @@
...
+                Inverse shear matrix.
+
+        Returns:
+            (:math:`N`, :math:`N`) :class:`numpy.ndarray`:
+                Transformed array of diffraction intensities.
+        """
+        if img.shape[0] % 2 == 0:
+            roll = img.shape[0] / 2 - 0.5
+        else:
+            roll = img.shape[0] / 2
```
bdice:
```suggestion
        roll = img.shape[0] / 2
        if img.shape[0] % 2 == 0:
            roll -= 0.5
```

freud/diffraction.pyx
yzjin: I've pushed a version that decouples the final img size from grid size. Currently it works well when the `output_size` is smaller than the `grid_size`. For the case where `output_size` is larger than `grid_size`, different zoom value will yield the same output size.

klywang:
https://github.com/glotzerlab/freud/blob/270ec42799a50e2392fc4c317a1a31abf14da07b/freud/diffraction.pyx#L46-L49

You may want to addres the differences between `grid_size` and `output_size` in the documentation. RIght now, the difference between the two parameters is unclear to users.

bdice:
@klywang I did some additional work on this, and I think it's ready for a final pass of review. I requested a review and you were selected by the algorithms-that-be. Thanks in advance! 👍 

cc: @tommy-waltmann since we've discussed this feature recently and @vyasr since I touched some stuff that I think you wrote in `util.pyx`, additional feedback is welcome so feel free to self-request a review if you are so inclined.

ChangeLog.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to
 * Custom NeighborLists can be created from a set of points using `from_points`. Distances will be calculated automatically.
 * The Box class has methods `compute_distances` and `compute_all_distances` to calculate distances between arrays of points and query points.
 * Hexatic can now compute 2D Minkowski Structure Metrics, using `weighted=True` along with a Voronoi NeighborList.
+* `freud.diffraction.DiffractionPattern` class (unstable) can be used to compute 2D diffraction patterns.
```
tommy-waltmann, line 17: Why are we referring to this as unstable?

tests/test_diffraction.py
```diff
@@ -0,0 +1,51 @@
+import freud
+import matplotlib
+import unittest
+import numpy.testing as npt
+matplotlib.use('agg')
+
+
+class TestDiffractionPattern(unittest.TestCase):
```
tommy-waltmann, line 10: We should add docstrings to the unit tests. @vyasr has mentioned to me that at some point he plans on having the CI style-check use more flake8 plugins, so to get out ahead of it complaining about the lack of docstrings in unit tests, we should plan on adding them now so we don't have to fix them later.

tommy-waltmann, line 10: Is there a limiting case of the diffraction pattern or some kind of pattern that we already know, so that we can have this compute it and compare answers?

tommy-waltmann (changes requested):
This is pretty close to merging, but I think it needs some kind of validation of the actual calculation, and should have some docstrings added to the unit tests.

ChangeLog.md
bdice: The API is not guaranteed to remain the same for all the freud 2.x line. In a later 2.y release, we can decide it is stable, but I am not confident enough to promise API stability at this time.

tests/test_diffraction.py
bdice: Consider zero particles, one particle, etc. The center value of the array is S(q=0) which should be the highest value in the image (for images with even side lengths, it may be a cluster of 4 pixels in the center).

bdice: @tommy-waltmann I'm not sure I know what you're wanting to write here. Feel free to make suggestions, but I haven't written docstrings for tests in any package before.

bdice: I won't be able to make these changes for a little while -- @tommy-waltmann, feel free to give it an attempt.

freud/diffraction.pyx
```diff
@@ -0,0 +1,316 @@
+# Copyright (c) 2010-2020 The Regents of the University of Michigan
+# This file is from the freud project, released under the BSD 3-Clause License.
+
+R"""
+The :class:`freud.diffraction` module provides functions for computing the
+diffraction pattern of particles in periodic systems.
```
klywang:
```suggestion
diffraction pattern of particles in systems with long range order.
```

Diffraction can be used to detect aperiodic, but highly ordered structures like quasicrystals as well.

freud/diffraction.pyx
```diff
@@ -0,0 +1,316 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+cdef class DiffractionPattern(_Compute):
+    R"""Computes a 2D diffraction pattern.
+
+    The diffraction image represents the scattering of incident radiation,
+    and is useful for identifying translational order present in the system.
```
klywang:
```suggestion
    and is useful for identifying translational and/or orientational order present in the system.
```
See comment about quasicrystals. Also, I believe you can use diffraction to detect orientational order in nematic crystals (lacks translational order but has orienational order).

I'm not sure if 'and' or 'or' is more appropriate here.

freud/diffraction.pyx
```diff
@@ -0,0 +1,316 @@
...
+
+        # Compute normals for each box face.
+        # The area of the face is the length of the vector.
+        box_face_normals = np.cross(
+            np.roll(box_matrix, 1, axis=-1),
+            np.roll(box_matrix, -1, axis=-1),
+            axis=0)
+
+        # Compute view axis projections.
+        projections = np.abs(box_face_normals.T @ np.array([0., 0., 1.]))
```
klywang, line 110: This is not a change request, but a question specifically related to @bdice's guidance on a PR I made -- is my understanding of this correct? Matrix operations are okay here because they maintain double float precision, while calculations written with single precision floats should be avoided due to potential rounding issues?

tests/test_diffraction.py
klywang: Another potential method is to convert the diffraction back into particle positions with iFFT + converting the everything back into discrete points. I've done this before while playing with the diffraction module if this is something we'd be interested in.

klywang (changes requested):
Just some minor changes to documentation.

tests/test_diffraction.py
klywang: You could do iFFT on the diffraction and then descritize those points and compare them to the simulation points? I'm not sure if that's an appropriate test to see if diffraction worked, but I've done that before with my own simulations while playing with the diffraction module

tommy-waltmann: Is this basically the inverse operation of the diffraction? Any cross-checks on these calculations are useful so feel free to add that to the tests.

klywang: Yup. I'll add it in

freud/diffraction.pyx
bdice: I'm not sure how we could detect orientational order when we don't use particles' orientations. I searched some literature to try and find diffraction images of nematic crystals. Some of the features that appear in those diffraction images _might_ come from the elliptical shape of the particles? I'm not 100% sure. Since this method treats all particles as point particles modeled by a Gaussian distribution, I don't want to make claims about detecting orientational/nematic order.

freud/diffraction.pyx
bdice: It's possible but unlikely that `box_face_normals` is in single (`float`) precision. Ultimately it doesn't matter because it will be upcast to `double` precision somewhere along the way, probably earlier in the calculation. The results of math operations are always upcast:
```python
>>> import numpy as np
>>> result = np.float(3.5) * np.double(7.2)
>>> type(result)  # This is a double because of upcasting
<class 'numpy.float64'>
```
If we were using something like Cython's `cdef float[:]` for data arrays, then it would be cast to single-precision when it is stored in such a variable.

tests/test_diffraction.py
bdice: @klywang Just to clarify, the user gets `S(q)` out of this calculation, not `I(q)`<sup>1</sup>. There is no complex phase associated with the resulting array that could be used to take the inverse FFT and recover raw point locations (unless there's some magic trick here that I don't know). Regardless, there are plenty of other helpful tests that can be written! 😄

<sup>1</sup> Definitions on Wikipedia: https://en.wikipedia.org/wiki/Structure_factor

klywang: Ohh. There was no magic -- I forgot that I did that while looking at the source code and had access to the imaginary portions.

freud/diffraction.pyx
klywang: When I added "orientational order", my intention was to include bond orientation than particle orientation. Maybe "rotational symmetry" would be more precise. 

I'm not too sure about nematic crystals either. I mostly wanted to include a less fringe case example than quasicrystals.

vyasr: Without incorporating a form factor I don't think you can say anything about orientations. If you do incorporate a form factor, then I think you may be able to see signal due to the correlations in nematic crystals, but I don't think this code should be attempting to support that right now.

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+    <https://en.wikipedia.org/wiki/Structure_factor>`_ :math:`S(\vec{q})` for
+    a plane of wavevectors :math:`\vec{q}` orthogonal to a view axis. The
+    view orientation :math:`(1, 0, 0, 0)` defaults to looking down the
+    :math:`z` axis (at the :math:`xy` plane). The points in the system are
+    converted to fractional coordinates, then binned into a grid whose
+    resolution is given by ``grid_size``. A higher ``grid_size`` will lead to
+    a higher resolution. The points are convolved with a Gaussian of width
+    :math:`\sigma`, given by ``peak_width``. This convolution is performed
+    as a multiplication in Fourier space. The computed diffraction pattern
+    is returned as a square array of shape ``(output_size, output_size)`.
```
bdice:
```suggestion
    is returned as a square array of shape ``(output_size, output_size)``.
```

bdice (commented):
I think the only thing this PR is lacking is better tests. Suggestions:

- Consider limiting cases zero particles, one particle, etc.
- The center value of the array is `S(q=0)` which should be the highest value in the image (for images with even side lengths, it may be in a cluster of 4 pixels in the center).

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+    converted to fractional coordinates, then binned into a grid whose
+    resolution is given by ``grid_size``. A higher ``grid_size`` will lead to
+    a higher resolution. The points are convolved with a Gaussian of width
+    :math:`\sigma`, given by ``peak_width``. This convolution is performed
+    as a multiplication in Fourier space. The computed diffraction pattern
+    is returned as a square array of shape ``(output_size, output_size)``.
+
+    Args:
+        grid_size (unsigned int):
+            Resolution of the diffraction grid. (Default value = 512).
```
bdice:
```suggestion
            Resolution of the diffraction grid (Default value = 512).
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+    <https://en.wikipedia.org/wiki/Structure_factor>`_ :math:`S(\vec{q})` for
+    a plane of wavevectors :math:`\vec{q}` orthogonal to a view axis. The
+    view orientation :math:`(1, 0, 0, 0)` defaults to looking down the
+    :math:`z` axis (at the :math:`xy` plane). The points in the system are
+    converted to fractional coordinates, then binned into a grid whose
+    resolution is given by ``grid_size``. A higher ``grid_size`` will lead to
+    a higher resolution. The points are convolved with a Gaussian of width
+    :math:`\sigma`, given by ``peak_width``. This convolution is performed
+    as a multiplication in Fourier space. The computed diffraction pattern
+    is returned as a square array of shape ``(output_size, output_size)``.
```
tommy-waltmann: Shouldn't this say grid size?
```suggestion
    is returned as a square array of shape ``(grid_size, grid_size)``.
```

tommy-waltmann: The diffraction pattern isn't really returned, you access it as a property
```suggestion
    can be accessed as a square array of shape ``(output_size, output_size)``.
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+    cdef int output_size
+    cdef double zoom
+    cdef double peak_width
+    cdef double[:] _k_values_orig
+    cdef double[:, :, :] _k_vectors_orig
+    cdef double[:] _k_values
+    cdef double[:, :, :] _k_vectors
+    cdef double[:, :] _diffraction
+
+    def __init__(self, grid_size=512, output_size=512, zoom=4, peak_width=1):
```
tommy-waltmann: `zoom` and `peak_width` can be moved to arguments of the `compute` method to enable faster repeated calculations of the same `grid_size`, but different `peak_width` and `zoom`. `output_size` could be moved as well given the state the code is in now, but I suspect that there are small problems with the current state of the code so it shouldn't be that way when we merge this PR.

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+        self._diffraction = np.real(
+            diffraction_fft * np.conjugate(diffraction_fft))
+
+        # Transform the image (scale, shear, zoom) and normalize S(q) by N^2
+        N = len(system.points)
+        self._diffraction = self._transform(
+            self._diffraction, system.box, inv_shear) / (N*N)
+
+        # Compute a cached array of k-vectors that can be rotated and scaled
+        if not self._called_compute:
```
tommy-waltmann, line 242: This variable only appears once. I'm guessing it should be first defined in the constructor and then set to `False`. It also needs to be set to `True` at the end of this block where the original kvalues and kvectors are computed.

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+            # Create a 1D axis of k-vector magnitudes
+            self._k_values_orig = np.fft.fftshift(np.fft.fftfreq(
+                n=self.output_size,
+                d=1/self.output_size))
+
+            # Create a 3D meshgrid of k-vectors with shape
+            # (output_size, output_size, 3)
```
tommy-waltmann, line 249:
```suggestion
            # (grid_size, grid_size, 3)
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+        # Compute the rotated and scaled k-values and k-vectors
+        self._k_values = self._k_values_orig / np.max(system.box.to_matrix())
+        self._k_vectors = rowan.rotate(view_orientation, self._k_vectors_orig)
+        self._k_vectors /= np.max(system.box.to_matrix())
+
+        return self
+
+    @_Compute._computed_property
+    def diffraction(self):
+        """:class:`numpy.ndarray`: diffraction pattern."""
```
tommy-waltmann: This should give the explicit dimensions of the returned array.
```suggestion
        """(:math:`grid_size`, :math:`grid_size`) :class:`numpy.ndarray`: diffraction pattern."""
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+        return self._diffraction
+
+    @_Compute._computed_property
+    def k_values(self):
+        """(:math:`N`, ) :class:`numpy.ndarray`: k-values."""
```
tommy-waltmann: What is `N`? Do we ever explicitly say what `N` is in the docs?

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+        return self._k_values
+
+    @_Compute._computed_property
+    def k_vectors(self):
+        """(:math:`N`, :math:`N`, 3) :class:`numpy.ndarray`: k-vectors."""
```
tommy-waltmann: Same as above, what is `N`?

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+        Args:
+            cmap (str):
+                Colormap name to use (Default value = :code:`'afmhot'`).
+            vmin (float):
+                Minimum of the color scale (Default value = 4e-6).
+            vmax (float):
+                Maximum of the color scale (Default value = 0.7).
+
+        Returns:
+            ((output_size, output_size, 4) :class:`numpy.ndarray`):
```
tommy-waltmann, line 321: Are we sure this returns something with dimensions of `output_size`? The `output_size` variable is never used in this method, and the diffraction pattern doesn't have dimensions of output size.

tommy-waltmann:
The biggest thing I have noticed so far is that there is a lot of confusion in the docs and code between `output_size` and `grid_size`. All the computed quantities are intially set to have dimensions of `grid_size`, but the docs say they are `output_size` in many places. We need a thorough review of which properties, values, plots, etc. should have which size. I suspect that some of the unit tests are only passing because the default values of `output_size` and `grid_size` are the same, and we should change tests to that they are done in cases where they are not the same.

bdice:
> The biggest thing I have noticed so far is that there is a lot of confusion in the docs and code between `output_size` and `grid_size`. All the computed quantities are intially set to have dimensions of `grid_size`, but the docs say they are `output_size` in many places. We need a thorough review of which properties, values, plots, etc. should have which size. I suspect that some of the unit tests are only passing because the default values of `output_size` and `grid_size` are the same, and we should change tests to that they are done in cases where they are not the same.

@tommy-waltmann The docs may need to be edited in a few places but the outputs appear to be using `output_size` as intended. The difference is that the `grid_size` is used internally and sets the spatial resolution of the FFT. This image is scaled/sheared/rotated, and `output_size` is the final width/height of the diffraction image. The `grid_size` parameter should not appear in the array shapes of any user-facing quantities -- all of those should use `output_size`.
https://github.com/glotzerlab/freud/blob/58ab65f3a20f001dc9da7b02376f7a9162a012d9/freud/diffraction.pyx#L169

freud/diffraction.pyx
bdice: @tommy-waltmann please validate this -- the `_transform` method should return `(output_size, output_size)` and that is carried through to this method, in my understanding. https://github.com/glotzerlab/freud/blob/58ab65f3a20f001dc9da7b02376f7a9162a012d9/freud/diffraction.pyx#L169

freud/diffraction.pyx
bdice: Using math here will make it render like grid<sub>s</sub>ize, and I recommend code font instead. Also it should be output_size.
```suggestion
        """(``output_size``, ``output_size``) :class:`numpy.ndarray`: diffraction pattern."""
```

freud/diffraction.pyx
bdice: I think this suggestion is incorrect. Please verify.

freud/diffraction.pyx
bdice: See https://github.com/glotzerlab/freud/blob/58ab65f3a20f001dc9da7b02376f7a9162a012d9/freud/util.pxd

We don't need to do anything differently with this variable. It is owned and managed by the parent class.

freud/diffraction.pyx
bdice: We cannot move `output_size` from the constructor to `compute` because the cached `_k_vectors_orig` and `_k_values_orig` rely on `output_size`. However, I think this is a good idea to move `zoom` and `peak_width` so they can be changed by the user interactively (which would also be useful if this feature were being used in a graphical interface).

freud/diffraction.pyx
bdice: See other comments - I think this is incorrect.

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+
+    def __init__(self, grid_size=512, output_size=512, zoom=4, peak_width=1):
+        self.grid_size = int(grid_size)
+        self.output_size = int(output_size)
+        self.zoom = np.double(zoom)
+        self.peak_width = np.double(peak_width)
+
+        # Cache these because they are system-independent.
+        self._k_values_orig = np.empty(grid_size)
+        self._k_vectors_orig = np.empty((grid_size, grid_size, 3))
```
bdice: This is being initialized incorrectly, I think:
```suggestion
        self._k_values_orig = np.empty(output_size)
        self._k_vectors_orig = np.empty((output_size, output_size, 3))
```

freud/diffraction.pyx
```diff
@@ -0,0 +1,313 @@
...
+
+        # Store these computed arrays which are exposed as properties.
+        self._k_values = np.empty_like(self._k_values_orig)
+        self._k_vectors = np.empty_like(self._k_vectors_orig)
+        self._diffraction = np.empty((grid_size, grid_size))
```
bdice:
```suggestion
        self._diffraction = np.empty((output_size, output_size))
```

freud/diffraction.pyx
tommy-waltmann: Yes, if I had known these arrays were initialized incorrectly, I would not have made this suggestion.

freud/diffraction.pyx
tommy-waltmann: This is another suggestion I would not have made had had I known the arrays in the constructor were initialized incorrectly.

tests/test_diffraction.py
tommy-waltmann: I've started work on these, and the tests for the max value at the center are failing. Also, is there a way to get freud to accept an empty system?

vyasr: Not that I'm aware of, the NeighborQuery construction will fail IIRC.

tommy-waltmann: So I've looked into why the tests for the center position are failing, and it is because the actual value at the center pixel is not the max, although the region in the center is always a maximum. I propose we alter the tests to test whether the center pixel has a value the same order of magnitude as the maximum value. Does that sound good to everyone else?

bdice: @tommy-waltmann So you're saying there's a higher peak at some k-vector that is not zero? This is plausible because the center pixels may share the intensity, e.g. the center 4 pixels might have `I(k=0)/4` each. It might be worth testing this hypothesis, you can sum over 3x3 regions (or maybe 2x2 regions) and verify if the highest sum is in the center. I _think_ that should always be true.

tommy-waltmann: Basically, there is a region in the center which is a maximum. Of the pixels in that region in the center, the central one is not quite as large as some nearby ones, but it is larger than areas not in the center.

tommy-waltmann: I've updated the tests to test for the existence of the peak at the center. We previously mentioned wanting to do some validation of a specific limiting case, like 1 particle as @bdice mentioned earlier. Is that something we could explicitly compute via another method by hand, or what is the best way to validate that case?

freud/diffraction.pyx
```diff
@@ -0,0 +1,314 @@
...
+
+        return self
+
+    @_Compute._computed_property
+    def diffraction(self):
+        """
+        (``output_size``, ``output_size``) :class:`numpy.ndarray`:
+            diffraction pattern.
+        """
+        return self._diffraction
```
tommy-waltmann: We want to be returning numpy arrays here and for the other properties, but as I've written the unit tests, I've noticed that this is not a numpy array. It is a `view.MemoryView` object. Normally to get the numpy arrays from lower level array constructs, we use `freud.util.make_managed_numpy_array`, but since this is not something taken from cpp, we may need a different way to convert it. I have used `np.asarray()` to convert in the unit tests, if that makes a copy of the underlying array data, then I think that is the best way to do the conversion.

tommy-waltmann (changes requested):
One last thing about the types we are returning ...

freud/diffraction.pyx
bdice: Good catch, @tommy-waltmann. [According to the Cython docs](https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html#coercion-to-numpy), coercion of the Cython memory view into a NumPy array with `np.asarray` should not require a copy. Can you make this change for all the properties where we return array-like quantities? Thanks!
```suggestion
        return np.asarray(self._diffraction)
```

tests/test_diffraction.py
tommy-waltmann: I've added a test for a single particle system that mimics a 0 particle system. There is a single point at the origin, and a large box of length 100.

tommy-waltmann (approved):

bdice:
I am thoroughly reviewing this code right now. I will fix any issues I find, write examples for freud-examples, and merge this PR.

A few notes worth sharing here:

- I found a small bug: the histogram was using `np.linspace(0, 1, grid_size)` which left us with `grid_size-1` bins instead of `grid_size` bins as intended. I'm going to need to do some additional testing, since this messes up the array roll-shifts that put the peak exactly the in the center for odd/even numbers of bins.
- @tommy-waltmann You can make a system with zero particles, but currently you can't use it in a NeighborQuery (this is an artificial requirement that could be removed if we tested the NeighborQuery code with zero particles and fixed any issues that arise). This doesn't affect the diffraction code, which doesn't use NeighborQuery. One way to make a system with zero particles is `freud.data.make_random_system(box_size=10, num_points=0)`. Another solution is to use a NumPy array with shape `(0, 3)`, which can't be constructed directly from a list like `[]`. You can explicitly construct such an array like this:
```python
>>> zero_points = np.zeros((0, 3))
>>> zero_points
array([], shape=(0, 3), dtype=float64)
```

I will add/fix tests for:
- k=0 peak centering issues related to the `grid_size` bug
- ensure the result array shapes are correct (`output_size`, not `output_size-1` as it was due to the bug mentioned above)
- tests for 0 particles

bdice:
In dd14133, I fixed the `roll` and `roll_shift` behavior so that the `k=(0, 0, 0)` peak is always in the center of the diffraction pattern at index `(dp.output_size//2, dp.output_size//2)`, regardless of `grid_size`, `output_size`, `view_orientation`, or `zoom`.

This is not just an aesthetic consideration, despite its subtlety. It is helpful to ensure that the diffraction pattern's center behaves smoothly in a real-time visualization context (OVITO), but this also ensures that the values of the `k_vectors` property agree with the image and that `k=(0, 0, 0)` aligns with the exact peak location.

Here's the script I used, if this problem should ever arise again. It is somewhat hard to debug this in a script. Visual outputs in a Jupyter Notebook are very helpful.

```python
import numpy as np
import freud
import matplotlib
import rowan

box, points = freud.data.UnitCell.fcc().generate_system(num_replicas=10, sigma_noise=0.05)
q = rowan.random.rand()
grid_size = np.random.randint(255, 513)
output_size = np.random.randint(255, 513)
dp = freud.diffraction.DiffractionPattern(grid_size=grid_size, output_size=output_size)
print(q, grid_size, output_size)
dp.compute((box, points), view_orientation=q, zoom=4)
print(np.unravel_index(np.argmax(dp.diffraction), dp.diffraction.shape))
print(dp.diffraction.shape)
c = dp.output_size//2
w = 8
s = w//4

# Plot the center of the diffraction image with a crosshair
plt.matshow(dp.diffraction[c-w:c+w+1, c-w:c+w+1])
plt.hlines(w, -0.5, 2*w+0.5)
plt.vlines(w, -0.5, 2*w+0.5)
plt.xticks(np.arange(0, 2*w, s), np.arange(-w, w, s))
plt.yticks(np.arange(0, 2*w, s), np.arange(-w, w, s))
plt.show()

# Plot a matrix showing how q rotates the x, y, and z axes
plt.matshow(rowan.rotate(q, np.eye(3)), vmin=-1, vmax=1, cmap='RdBu')
plt.show()
```
Sample output:
image

bdice (approved):
I have thoroughly revised this PR and I think it is ready to merge. I've stared this code for too long today, so I am going to merge it tomorrow after one final look.

tests/test_diffraction.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def test_k_values_and_k_vectors(self):
+        dp = freud.diffraction.DiffractionPattern()
+        box, positions = freud.data.UnitCell.fcc().generate_system(4)
+        dp.compute((box, positions))
+
+        output_size = dp.output_size
+        npt.assert_allclose(dp.k_values[output_size//2], 0)
+        center_index = (output_size//2, output_size//2)
+        npt.assert_allclose(dp.k_vectors[center_index], [0, 0, 0])
```
tommy-waltmann: Perhaps we could run this test using many random systems of different sizes instead of a single fcc system, so the `k_values` and `k_vectors` will have different values across different systems. I think adding this would be a more robust test that the middle `k_vector`/`k_value` is what we want it to be.

tommy-waltmann (commented):
After looking at Bradley's changes from the weekend, I have one more suggestion.

tests/test_diffraction.py
vyasr: I am always strongly in favor of adding more tests, particularly of non-toy systems.

bdice (approved):
Approving and merging.

PR #597, opened 2020-03-23
provide per particle msd in MSD method

zhou-pj:
## Description
provide user access to per particle based msd in MSD method

## Types of changes

## Checklist:


freud/msd.pyx
```diff
@@ -238,7 +238,14 @@ cdef class MSD(_Compute):
...
         """:math:`\\left(N_{frames}, \\right)` :class:`numpy.ndarray`: The mean
         squared displacement."""
-        return np.concatenate(self.particle_msd, axis=1).mean(axis=-1)
+        return np.concatenate(self._particle_msd, axis=1).mean(axis=-1)
+
+    @_Compute._computed_property
+    def particle_msd(self):
+        """:math:`\\left(N_{frames}, N_{particles} \\right)`
+        :class:`numpy.ndarray`: The per particle based mean squared
+        displacement."""
```
bdice: You should check how the docs are rendering this and consider the following suggestion. I don't think you can have a line break in the first section with the math shape and ndarray type. However, if you remove the line break, then flake8 will yell at you unless you add the appropriate `noqa`. This construction ought to work:
```suggestion
        """:math:`\\left(N_{frames}, N_{particles} \\right)` :class:`numpy.ndarray`:
        The per particle based mean squared displacement.
        """  # noqa: E501
```

bdice (changes requested):
I'd like to get a second review from @vyasr and will add him as a reviewer. This needs tests, but otherwise I think it's close to done. Consider adding tests where you have one particle move and another particle that is stationary, or something like that. Also update changelog/credits.

freud/msd.pyx
zhou-pj: I am having trouble make the doc and got a couple no module found error, this is the last one(and it doesn't go away when i try install sphinxcontrib): `Could not import extension sphinxcontrib.bibtex (exception: No module named 'sphinxcontrib.bibtex')`. but `make html` works normal on my signac repo. Any idea how I can fix it?

zhou-pj:
there is a existing test on msd accuracy so I just added the comparison on a per particle base as well in that test.

freud/msd.pyx
bdice: @zhou-pj You need to install the sphinx extension: `pip install sphinxcontrib-bibtex`. We should probably add this as a "development requirement" but we don't have a `requirements-dev.txt` or similar file right now. Separate issue created: #598.

zhou-pj: thanks! I still have trouble checking the html doc locally though. No matter what I change in the docstring, the html would not change accordingly. Is there another docstring place it is pulling from? Other than that I think the rest is good.

vyasr: Sphinx works by importing the Python module that you're working with and then inspecting the imported object to get docstrings. For pure Python packages (e.g. signac), this means that even if you update the source you have to make sure your PYTHONPATH is pointing to the development version. For freud, this is a little more annoying because pyx files cannot be imported directly, they must be compiled into shared libraries before Python can import them. I'm assuming the reason your documentation is not changing is because you actually need to rerun setup.py to rebuild freud to make the new docstrings accessible to Sphinx. Usually, to avoid continually installing freud into the user site directory I get around this by running `python setup.py build_ext --inplace`. This command will compile the Cython into C++, and compile the resulting C++, but it won't copy freud into your global site-packages folder. This way, you won't be able to import this version of freud from other directories, but you will be able to import it from the root of the repository (`import freud` will see the nested freud folder containing the Cython source files, and all the compiled modules will be placed inside that folder so the import will work). You can then add this folder to your PYTHONPATH. This is very similar to how running HOOMD out of the build directory works.

zhou-pj: I checked the doc it looks good now

bdice (approved):
Looks good.

vyasr (approved):
LGTM

PR #573, opened 2020-01-21
Define wrapping behavior for aperiodic boxes (for aperiodic GaussianDensity).

bdice:
## Description
I am doing some analyses where I need to have an aperiodic box. I experimented with some behaviors and decided on this:
- If a box dimension is configured to be aperiodic, then it is treated as if the box is infinite in that direction when wrapping a set of input vectors.
- For a box that is aperiodic in all 3 directions, `vec = wrap(vec)` is an identity mapping. That is, particles that are outside the box won't be affected by wrapping if the box is aperiodic.
- Corollary: there is no "maximum distance" that particles can have between them if the box is aperiodic.

I have not investigated how neighbor finding is affected, since I only used this feature to enable aperiodic behavior in GaussianDensity. This PR should probably not be merged until we think about all the implications and add tests of features with aperiodic systems (or just throw runtime errors in each feature until it is officially supported with aperiodic boxes)!

## Motivation and Context
I needed an aperiodic GaussianDensity feature.

## How Has This Been Tested?
Need to write tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds or improves functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [ ] I have updated the documentation (if relevant).
- [ ] I have added tests that cover my changes (if relevant).
- [ ] All new and existing tests passed.
- [ ] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst).
- [ ] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

PR #605, opened 2020-04-28
Remove various unnecessary imports (mostly in pxd files).

vyasr:
## Description

Removes unused imports from pxd files. Also avoids including all of tbb.h everywhere tbb is used.

## Motivation and Context

I assume due to copy-pasting during the 2.0 rewrite many imports (like vector, complex, and shared_ptr) were added to all files unnecessarily. The tbb includes were causing warnings due to deprecated functionality that exists in some of the headers that it includes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] I have updated the documentation (if relevant).
- [x] I have added tests that cover my changes (if relevant).
- [x] All new and existing tests passed.
- [ ] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst).
- [ ] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

Issue #464, opened 2019-09-25
Double precision


vyasr:
**Is your feature request related to a problem? Please describe.**
Some calculations (like Steinhardt order parameters) can be highly sensitive to small changes in specific cases (such as near-ideal crystals). In those cases, using double- rather than single-precision arithmetic would help. In addition most users working with numpy are creating double precision arrays by default, so there's a lot of unnecessary copying of data happening every time they use freud.

**Describe the solution you'd like**
We should define a simple `Scalar` typedef in `VectorMath` or `HOOMDMath` that can be used everywhere in place of float. We would then add a preprocessor flag that allows us to decide whether to compile in single or double precision. We would also need to create a function exposed to Python so that our Cython code in common.pyx can determine what NumPy types to use for its arrays.

**Describe alternatives you've considered**
If this ends up being too much of a pain, we could just make freud a double precision library always. There would be performance implications but in general I'd expect it to be different by a factor of two or so in any case.

joaander:
The original motivation for single precision is that file formats like DCD and GSD store values in single precision. Thus a single-precision freud *avoids* the need to make copies and convert data types in those use-cases. Yes, data inputs passed from default-constructed numpy-arrays are double - but users can be taught to specify `dtype=numpy.float32` where performance matters.

For the use-case you mention, what is the instability? Do the particle coordinates really need to be in double? Or is it just one particular step in the Steinhardt calculation (i.e. the computation and/or averaging of spherical harmonics)? If the latter, one option is to use doubles where they are necessary in intermediate calculations but still accept inputs as floats.

Preprocessor directives don't work well with pip/conda packages. To enable full single and full double selectable at run time, consider instead templating all operations on Real and instantiate them with both single and double. 

As you say, all this is contingent on how much performance loss there is by switching to full double. We should evaluate that quantitatively to inform the path forward. We should include a measurement of the performance overhead of converting float input to doubles to simulate the performance loss when performing interactive analysis on gsd trajectories.

vyasr:
While I still think this feature would be nice, at the moment freud is explicitly using double precision wherever necessary and that seems good enough. We can revisit this if someone starts finding significant issues in the future, but since we're not doing anything that requires e.g. strict conserved quantities I don't think this is critical.

PR #606, opened 2020-05-11
Sphere Voxelization

### LINKED ISSUE #209: Improve support for aperiodic boxes / boundary conditions
Almost all systems that we analyze are periodic in all of their 2 or 3 dimensions. However, not all methods in `freud` support boxes that are aperiodic in one or more dimensions.

- [ ] Survey modules and determine which ones do/don't fully support aperiodicity
- [ ] Write tests to check for aperiodic behavior
- [ ] Fix failing tests by adding support for aperiodic behavior

### LINKED ISSUE #610: Ambiguous widths/dimensionality in GaussianDensity
In `GaussianDensity` and `SphereVoxelization`, the `width` variable in C++ is a copy of `m_width` but with z=1 if a 2D box is provided. This allows users to specify an integer for all dimensions, but it makes the behavior of `width` a bit ambiguous. If the user inputs `width=100` in the constructor, the `width` property returns `[100, 100, 100]` regardless of the dimensionality of the input system's box.

### LINKED ISSUE #611: GaussianDensity does not normalize correctly in 2D.
**Describe the bug**
The gaussian normalization factor for 2D systems is incorrect in GaussianDensity. It uses the normalization that would be appropriate for 3D systems. The formula should be:
fcab90af48acc1ff0f5076815cfb8022

This will be resolved by #606 and is now tested for validity, this issue exists to make a note for posterity.

bdice:
## Description
This PR adds a new feature for "voxelization," which is similar to GaussianDensity in that it produces a 2D or 3D "image" from a set of points. A user provides a set of points, an output size, and a radius, which is used to create a voxel grid where each voxel is a 1 if it is within `radius` of one or more of the provided points, and 0 otherwise.

This PR also extends the periodicity logic in freud to allow for aperiodic calculations in GaussianDensity and Voxelization. e.g. If the box is aperiodic, then a point on the left edge of the box will not wrap and fill voxels on the right edge.

**This PR supercedes #573.**

## Motivation and Context
This feature is useful for producing image-like data from point configurations, which I plan to use for input to a machine learning algorithm.

## How Has This Been Tested?
New tests have been added.

## Types of changes

## Checklist:


tommy-waltmann (approved):
I think this is good to go.

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2010-2019 The Regents of the University of Michigan
+// This file is from the freud project, released under the BSD 3-Clause License.
+
+#include <cmath>
+#include <stdexcept>
+#include <tbb/tbb.h>
```
bdice: Need to replace with specific tbb headers (see #605).

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,142 @@
...
+//! internal
+/*! \brief Function to compute the voxels array
+ */
+void SphereVoxelization::compute(const freud::locality::NeighborQuery* nq)
+{
+    auto box = nq->getBox();
+    auto n_points = nq->getNPoints();
+    m_box = box;
+
+    vec3<unsigned int> width(m_width);
```
bdice: Could this variable be removed and replaced by `m_width` everywhere?

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,142 @@
...
+}
+
+//! internal
+/*! \brief Function to compute the voxels array
+ */
+void SphereVoxelization::compute(const freud::locality::NeighborQuery* nq)
+{
+    auto box = nq->getBox();
+    auto n_points = nq->getNPoints();
+    m_box = box;
```
bdice: Replace uses of `box` with `m_box`.
```suggestion
    m_box = nq->getBox();
    auto n_points = nq->getNPoints();
```

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,142 @@
...
+            const int bin_x = int((point.x + lx / 2.0f) / grid_size_x);
+            const int bin_y = int((point.y + ly / 2.0f) / grid_size_y);
+            // In 2D, only loop over the z=0 plane
+            const int bin_z = m_box.is2D() ? 0 : int((point.z + lz / 2.0f) / grid_size_z);
+
+            // Reject bins that are outside the box in aperiodic directions
+            // Only evaluate over bins that are within the cutoff
+            for (int k = bin_z - bin_cut_z; k <= bin_z + bin_cut_z; k++)
+            {
+                if (!periodic.z && (k < 0 || k >= int(m_width.z)))
```
bdice, line 93: I would like to see this conditional logic structure replicated in `GaussianDensity`. It should be faster to evaluate and reject the inner loops in advance.

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,142 @@
...
+                        // Check to see if this distance is within the specified r_max
+                        if (r_sq < r_max_sq)
+                        {
+                            // Assure that out of range indices are corrected for storage
+                            // in the array i.e. bin -1 is actually bin 29 for nbins = 30
+                            const unsigned int ni = (i + m_width.x) % m_width.x;
+                            const unsigned int nj = (j + m_width.y) % m_width.y;
+                            const unsigned int nk = (k + m_width.z) % m_width.z;
+
+                            m_voxels_array(ni, nj, nk) = 1;
```
bdice, line 131:
```suggestion
                            // This array value could be written by multiple threads in parallel.
                            // This is only safe because all threads are writing the same value (1).
                            m_voxels_array(ni, nj, nk) = 1;
```

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,142 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    });
+
+    // Now reduce all the arrays into one.
```
bdice: Remove this commented section.

cpp/density/SphereVoxelization.h
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2010-2019 The Regents of the University of Michigan
```
bdice: Make sure all the copyrights are updated (see #602).
```suggestion
// Copyright (c) 2010-2020 The Regents of the University of Michigan
```

freud/density.pyx
```diff
@@ -301,6 +301,117 @@ cdef class GaussianDensity(_Compute):
             return None
 
 
+cdef class SphereVoxelization(_Compute):
+    R"""Computes the integer density of a system on a grid.
```
bdice:
```suggestion
    R"""Computes a grid of voxels occupied by spheres placed on a set of points.
```

freud/density.pyx
```diff
@@ -301,6 +301,117 @@ cdef class GaussianDensity(_Compute):
...
+
+    Replaces particle positions with a sphere of fixed radius and calculates the
+    contribution from each to the proscribed grid based upon the distance of
+    the grid cell from the center of the sphere. The resulting data is a
+    regular grid of particle densities with an integer for each grid cell that
+    indicates the number of particles whose spheres overlap that cell. The
```
bdice: This isn't true -- it's not a number of particles, it's always a 0 or 1 state.

freud/density.pyx
```diff
@@ -301,6 +301,117 @@ cdef class GaussianDensity(_Compute):
...
+    dimensions of the image (grid) are set in the constructor, and can either
+    be set equally for all dimensions or for each dimension independently.
+
+    Args:
+        width (int or list or tuple):
+            The number of bins to make the image in each direction (identical
+            in all dimensions if a single integer value is provided).
```
bdice: Also make these changes for GaussianDensity.
```suggestion
        width (int or Sequence[int]):
            The number of bins to make the image in each dimension (identical
            in all dimensions if a single integer value is provided).
```

freud/density.pyx
```diff
@@ -301,6 +301,117 @@ cdef class GaussianDensity(_Compute):
...
+    @property
+    def r_max(self):
+        """float: Sphere radius used for voxelization."""
+        return self.thisptr.getRMax()
+
+    @property
+    def width(self):
+        """int or list or tuple: The number of bins to make the image in each
+        direction (identical in all dimensions if a single integer value is
+        provided)."""
```
bdice: Also apply this fix to GaussianDensity.
```suggestion
        """tuple[int]: The number of bins in the image in each dimension
        (identical in all dimensions if a single integer value is
        provided)."""
```

bdice (commented):
@tommy-waltmann I have some comments. Would you be able to take a look at these?

cpp/density/SphereVoxelization.cc
tommy-waltmann: There is no explicit tbb in this file, so if I understand correctly, these need to be the tbb headers required for forLoopWrapper, right?

cpp/density/SphereVoxelization.cc
tommy-waltmann: I think so. It looks like later in the method we go back to using `m_width` anyway. Do you remember what the original intent of defining the `width` variable was?

tommy-waltmann: Looks like its from `GaussianDensity`, do you remember the original intent of defining things this way?

cpp/density/SphereVoxelization.cc
tommy-waltmann: I concur

cpp/density/SphereVoxelization.cc
tommy-waltmann: It turns out it is necessary, I'm not exactly sure why, other than in the 2D case, `m_voxels_array` ends up being 3 dimensional instead of 2

bdice: This currently isn't a great design. The `width` variable is a copy of `m_width` but with z=1 if a 2D box is provided. This allows users to specify an integer for all dimensions, but it makes the behavior a bit ambiguous. If the user inputs `width=100` in the constructor, the `width` property returns `[100, 100, 100]` regardless of what dimensionality the box has... I'm not sure what to do about this but we can raise an issue.

bdice: See #610. Closing this conversation for now -- this can be resolved later in a way that is consistent for GaussianDensity and SphereVoxelization.

cpp/density/GaussianDensity.cc
```diff
@@ -81,44 +81,51 @@ void GaussianDensity::compute(const freud::locality::NeighborQuery* nq)
...
                         {
-                            // Evaluate the gaussian ...
-                            const float x_gaussian
-                                = A * std::exp((-1.0f) * (delta.x * delta.x) / (2.0f * sigmasq));
-                            const float y_gaussian
-                                = A * std::exp((-1.0f) * (delta.y * delta.y) / (2.0f * sigmasq));
-                            const float z_gaussian
-                                = A * std::exp((-1.0f) * (delta.z * delta.z) / (2.0f * sigmasq));
+                            // Evaluate the gaussian
+                            const float gaussian = A * std::exp(-r_sq / (2 * sigmasq));
```
bdice: There is a bug in the Gaussian normalization in the old code for 2D systems. I am adding a separate changelog line to address that.

vyasr:
I see that I'm assigned on here, I won't be able to review until mid-week but will plan to review around Wednesday.

benchmarks/benchmark_density_GaussianDensity.py
```diff
@@ -27,7 +27,7 @@ def bench_run(self, N):
 def run():
     Ns = [1000, 10000]
     width = 100
-    r_max = 10
+    r_max = 1
```
vyasr, line 30: Why were these changes made? They seem unrelated to everything else, is there a compelling reason to use different parameters?

cpp/box/Box.h
```diff
@@ -295,10 +295,25 @@ class Box
...
+            v_frac.x = util::modulusPositive(v_frac.x, 1.0f);
+        }
+        if (m_periodic.y)
+        {
+            v_frac.y = util::modulusPositive(v_frac.y, 1.0f);
+        }
+        if (m_periodic.z)
+        {
+            v_frac.z = util::modulusPositive(v_frac.z, 1.0f);
+        }
```
vyasr, line 316: While this wrapping logic is correct, the discussion in #209 suggests that there are cases where it won't work. However, on further consideration I'm not sure that this is the case; do we ever wrap positions? AFAIK we only ever need to wrap vectors between two points, not vectors representing a point. @bdice let's make sure this is true before we merge this change since it will immediately make it possible for classes to operate on aperiodic boxes even if the behavior isn't quite correct. Also, are there any cases where we would need to disable that behavior?

cpp/density/GaussianDensity.cc
```diff
@@ -31,94 +31,107 @@ vec3<unsigned int> GaussianDensity::getWidth()
...
 
-//! internal
-/*! \brief Function to compute the density array
- */
+//! Compute the density array.
 void GaussianDensity::compute(const freud::locality::NeighborQuery* nq)
 {
-    auto box = nq->getBox();
+    // Don't allow 3D boxes after computing in 2D (m_width is altered)
+    if (!nq->getBox().is2D() && m_box.is2D())
```
vyasr: Should we also forbid the other direction, i.e. 
```suggestion
    if (nq->getBox().is2D() != m_box.is2D())
```
?

cpp/box/Box.h
vyasr: To be clear, while I think making the rest of freud work for aperiodic systems is out of scope for this PR, if we're introducing this change I think we do at least need to add error messages everywhere that aperiodic boxes could break the logic before this PR can be merged.

cpp/density/GaussianDensity.cc
```diff
@@ -31,94 +31,107 @@ vec3<unsigned int> GaussianDensity::getWidth()
...
                         {
-                            // Evaluate the gaussian ...
-                            const float x_gaussian
-                                = A * std::exp((-1.0f) * (delta.x * delta.x) / (2.0f * sigmasq));
-                            const float y_gaussian
-                                = A * std::exp((-1.0f) * (delta.y * delta.y) / (2.0f * sigmasq));
-                            const float z_gaussian
-                                = A * std::exp((-1.0f) * (delta.z * delta.z) / (2.0f * sigmasq));
+                            // Evaluate the gaussian
+                            const float gaussian = normalization * std::exp(-r_sq / (2 * sigmasq));
```
vyasr:
```suggestion
                            const float gaussian = normalization * std::exp(-r_sq / (float(2.0) * sigmasq));
```

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,131 @@
...
+namespace freud { namespace density {
+
+SphereVoxelization::SphereVoxelization(vec3<unsigned int> width, float r_max)
+    : m_box(box::Box()), m_width(width), m_r_max(r_max)
+{
+    if (r_max <= 0.0f)
+        throw std::invalid_argument("SphereVoxelization requires r_max to be positive.");
+}
+
+//! Get a reference to the last computed voxels
```
vyasr:
```suggestion
//! Get a reference to the last computed voxels.
```

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,131 @@
...
+        {
+            const vec3<float> point = (*nq)[idx];
+            // Find which bin the particle is in
+            const int bin_x = int((point.x + Lx / 2.0f) / grid_size_x);
+            const int bin_y = int((point.y + Ly / 2.0f) / grid_size_y);
+            // In 2D, only loop over the z=0 plane
+            const int bin_z = m_box.is2D() ? 0 : int((point.z + Lz / 2.0f) / grid_size_z);
+
+            // Reject bins that are outside the box in aperiodic directions
+            // Only evaluate over bins that are within the cutoff
```
vyasr:
```suggestion
            // Only evaluate over bins that are within the cutoff, rejecting bins
            // that are outside the box in aperiodic directions.
```

cpp/util/utils.h
```diff
@@ -47,34 +48,59 @@ inline void forLoopWrapper(size_t begin, size_t end, const Body& body, bool para
...
+            [&body](const tbb::blocked_range3d<size_t>& r) {
+                body(r.pages().begin(), r.pages().end(), r.rows().begin(), r.rows().end(), r.cols().begin(),
+                     r.cols().end());
+            });
+    }
+    else
+    {
+        body(begin_page, end_page, begin_row, end_row, begin_col, end_col);
+    }
+}
```
vyasr: Let's remove this function. We can add it at a later date if we have a need for it, but we have enough experience with how long cruft lies around in freud that I don't want to carry around unused functions without a demonstrated need. Also remove the corresponding blocked_range3d include at the top.

doc/source/reference/credits.rst
```diff
@@ -101,6 +101,9 @@ Bradley Dice - **Lead developer**
 * Added finite tolerance to ensure stability of 2D Voronoi NeighborList computations.
 * Improved stability of Histogram bin calculations.
 * Improved error handling of Cubatic input parameters.
+* Added SphereVoxelization feature.
```
vyasr:
```suggestion
* Worked with Tommy Waltmann to add the SphereVoxelization feature.
```

doc/source/reference/credits.rst
```diff
@@ -285,6 +288,10 @@ Mike Henry
 * Fixed syntax in freud-examples notebooks for v2.0.
 * Updated documentation links
 
+Tommy Waltmann
+
+* Added SphereVoxelization feature.
```
vyasr:
```suggestion
* Worked with Bradley Dice to add the SphereVoxelization feature.
```

freud/_density.pxd
```diff
@@ -57,3 +57,13 @@ cdef extern from "RDF.h" namespace "freud::density":
...
         const freud.util.ManagedArray[float] &getNr()
+
+cdef extern from "SphereVoxelization.h" namespace "freud::density":
+    cdef cppclass SphereVoxelization:
+        SphereVoxelization(vec3[unsigned int], float) except +
+        const freud._box.Box & getBox() const
+        void reset()
+        void compute(const freud._locality.NeighborQuery*) except +
+        const freud.util.ManagedArray[unsigned int] &getVoxels() const
+        vec3[unsigned int] getWidth() const
```
vyasr, line 68:
```suggestion
        vec3[unsigned int] getWidth() const
```

cpp/density/SphereVoxelization.h
```diff
@@ -0,0 +1,65 @@
...
+        return m_r_max;
+    }
+
+    //! Compute the voxelization.
+    void compute(const freud::locality::NeighborQuery* nq);
+
+    //! Get a reference to the last computed voxels.
+    const util::ManagedArray<unsigned int>& getVoxels() const;
+
+    vec3<unsigned int> getWidth();
```
vyasr:
```suggestion
    vec3<unsigned int> getWidth() const;
```

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,131 @@
...
+}
+
+//! Get a reference to the last computed voxels
+const util::ManagedArray<unsigned int>& SphereVoxelization::getVoxels() const
+{
+    return m_voxels_array;
+}
+
+//! Get width.
+vec3<unsigned int> SphereVoxelization::getWidth()
```
vyasr:
```suggestion
vec3<unsigned int> SphereVoxelization::getWidth() const
```

vyasr (changes requested):
This looks pretty good, I have some inline requests but mostly nothing too major, the one possible exception being the consideration of how to handle aperiodic boxes now that they've been introduced. We also need to add tests of the box wrapping logic now that it's supporting aperiodic cases. I'd also like to see at least one test of the new voxelization that actually tests its behavior, either something like the PMFT tests where the bin counts are validated using an equivalent Python method or at the very least a known system where the counts can be stored, perhaps something generated with `freud.data.make_alternating_lattice` or similar.

cpp/density/SphereVoxelization.cc
tommy-waltmann: What is the need for the first `const` here? This returns a copy of `m_width`, meaning the user can't modify the data in the in the `SphereVoxelization` class, so why do we care whether or not the user can change the returned copy?

tommy-waltmann:
> 
> 
> This looks pretty good, I have some inline requests but mostly nothing too major, the one possible exception being the consideration of how to handle aperiodic boxes now that they've been introduced. We also need to add tests of the box wrapping logic now that it's supporting aperiodic cases. I'd also like to see at least one test of the new voxelization that actually tests its behavior, either something like the PMFT tests where the bin counts are validated using an equivalent Python method or at the very least a known system where the counts can be stored, perhaps something generated with `freud.data.make_alternating_lattice` or similar.

I implemented an fft version of the aperiodic voxelization for my photonics project with Bradley (we found that it was too slow and aren't going to use it), so I can add test case(s) that will serve as a cross-check for the actual behavior of SphereVoxelization.

cpp/density/GaussianDensity.cc
tommy-waltmann: The variable `sigmasq` is a float, so what is the need for casting here? Also, isn't casting in cpp done as `(float) 2.0`, not `float(2.0)`?

vyasr: You're mostly right about these two things. The `(float) 2.0` style is actually C-style casting, not C++ style. The function style cast `float(2.0)` was added in C++, but it behaves exactly like C-style casts except in cases where there are multiple arguments, in which case it is capable of calling more complex constructors.

The preferred method of casting in C++ is actually by explicitly specifying the type of the cast, in this case `static_cast` (https://en.cppreference.com/w/cpp/language/static_cast). 
C-style casts are pretty unsafe in general since they will even try to cast things that aren't safe to cast to each other. However, in cases like this where the object is a float literal I don't personally worry about that. If we wanted to *strictly* adhere to best practices we would want to go through the entire code base and replace most such casts with the `static_cast` variant, but there's not really a good reason to.

You're also right that casting isn't strictly necessary here. The reason I made the change is in anticipation of a future release of freud that supports variable/mixed precision (see #464). The easiest way to do that is something like what HOOMD does, where there's a global typedef of the desired type (`Scalar` in HOOMD) and all literals are typecast throughout the code base. This of course relies on compilers to optimize away the type conversion since otherwise you don't really get the same performance tradeoffs, but that's a pretty safe assumption in such simple cases AFAIK.

cpp/density/SphereVoxelization.cc
vyasr: You are correct, this is a mistake on my part. Ignore this, as well as the corresponding suggestions on the pxd file.

vyasr: You are correct. For future reference, I had accidentally added a const qualifier to the return type, which is meaningless here. In the interest of being able to batch commit these suggestions, I've updated the suggestion here, in the header, and in the corresponding pxd file.

vyasr:
Thanks Tommy. I fixed a couple errors you caught in my review, and the test you added generally looks good. I'll give this another pass once Bradley applies my changes, then I think we should be good to go.

cpp/density/GaussianDensity.cc
tommy-waltmann: We should have each instance of `SphereVoxelization` and `GaussianDensity` only perform the calculation in a single number of dimensions. It removes the ambiguity of what is and isn't allowed in terms of going between numbers of dimensions between calculations and prevents any unseen errors associated with that that we might be missing.

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,145 @@
...
+*/
+
+namespace freud { namespace density {
+
+SphereVoxelization::SphereVoxelization(vec3<unsigned int> width, float r_max)
+    : m_width(width), m_r_max(r_max)
+{
+    // dummy initial box, dimensionality of the calculation will be defined by
+    // the first neighbor query passed to the compute method
+    m_box = NULL;
```
bdice: I support the idea of preventing dimensionality changes in either direction, but I am not sure `NULL` is the right strategy here. Honestly I have no idea how `m_box` would act after initializing it to `NULL`.

`NULL` is not recommended for use in C++11 or newer. https://stackoverflow.com/questions/55520687/is-null-a-pointer

[`nullptr`](https://en.cppreference.com/w/cpp/language/nullptr) is the preferred way to use null pointers, but `m_box` _isn't a pointer_. 

I would rather keep a flag like `bool m_has_computed` that is used when checking the "new" box dimensionality against the stored box dimensionality.

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,145 @@
...
+
+    // Don't allow changes in the number of dimensions for the calculation
+    // after its been set.
+    if (nq->getBox().is2D() != m_box.is2D())
+    {
+        throw std::invalid_argument(
+            "Each SphereVoxelization instance should be used to do calculations "
+            "in a fixed number of dimensions. You have changed the number of "
+            "dimensions used going from one calculation to another.");
+    }
```
bdice: With my comment above, this would become the following. Note that error messages should avoid using second-person "you" in their text. It sounds less accusatory that way. 😉
```suggestion
    // Set the number of dimensions for the calculation the first time it is done
    if (!m_has_computed || nq->getBox().is2D() != m_box.is2D())
    {
        m_box = nq->getBox();
        m_has_computed = true;
    } else {
        throw std::invalid_argument(
            "The dimensionality of the box passed to SphereVoxelization has changed. "
            "A new instance must be created to handle a different dimensionality.");
    }
```

bdice: And remove the second assignment of `m_box` below, as needed. And make GaussianDensity match.

cpp/density/SphereVoxelization.cc
tommy-waltmann: I agree with switching this to `nullptr`, but the class is designed such that nothing can happen before `compute()` is called, so I don't think we have to worry about how `m_box` acts after setting it to `nullptr` because the first thing that happens in `compute()` is that we set `m_box` to something that isn't `nullptr`.

bdice: To clarify, I wasn’t recommending to use nullptr for this case. You should not assign a pointer type to m_box because it isn’t a pointer. I would like to be explicit with the m_has_computed flag. The m_box should be default-constructed and the flag should track its assignment. That’s a much safer approach.

cpp/density/SphereVoxelization.cc
tommy-waltmann: I think the boolean condition in the if statement here should read `!m_has_computed || nq->getBox().is2D() == m_box.is2D()`

bdice: Right. Good catch.

cpp/util/utils.h
tommy-waltmann: I don't think this is within scope of the PR, I agree that the change should be made, but let's do this somewhere else.

vyasr: @tommy-waltmann I'm not sure I follow your logic. This function is being added by this PR, so choosing _not_ to add it definitely seems in scope for this PR.

cpp/box/Box.h
tommy-waltmann: I'm unfamiliar with many of the calculations in freud right now, did you have any in mind specifically that could benefit from this type of error message?

cpp/density/GaussianDensity.cc
```diff
@@ -31,94 +31,116 @@ vec3<unsigned int> GaussianDensity::getWidth()
...
+    if (!m_has_computed || nq->getBox().is2D() == m_box.is2D())
+    {
+        m_box = nq->getBox();
+        m_has_computed = true;
+    }
+    else
+    {
+        throw std::invalid_argument(
+            "The dimensionality of the box passed to GaussianDensity has "
+            "changed. A ne instance must be created to handle a different "
```
vyasr:
```suggestion
            "changed. A new instance must be created to handle a different "
```

cpp/box/Box.h
vyasr: Offhand I'd say anything that touches neighbor finding, you wouldn't want to find periodic neighbors in an aperiodic box. So I would probably put errors in the `query` method of `NeighborQuery` (you might have to do it in the subclasses, I honestly can't remember if I made that method is pure virtual or not). I _think_ that all non-pair computations other than `GaussianDensity` (and now `SphereVoxelization`) should be safe because they just interact with the point positions or orientations. These two are special because the smearing requires accounting for periodicity, but I don't think any of the others do.

tommy-waltmann: `query()` is not pure virtual, so I'll add an error there. That should be the only place where we have not already accounted for box periodicity right? In terms of docs updates for this, should a notice go in every pair calculation, or is there a separate place for this to go?

ChangeLog.md
```diff
@@ -11,15 +11,18 @@ and this project adheres to
...
 ### Changed
 * Cython is now a required dependency (not optional). Cythonized `.cpp` files have been removed.
+* An instance of GaussianDensity cannot compute 3D systems if it has been previously computed 2D systems.
 
 ### Fixed
 * Histogram bin locations are computed in a more numerically stable way.
 * Improved error handling of Cubatic input parameters.
 * PMFTs are now properly normalized such that the pair correlation function tends to unity for an ideal gas.
 * PMFTXYT uses the correct orientations when points and query\_points differ.
+* GaussianDensity gaussian normalization in 2D systems has been corrected.
```
vyasr:
```suggestion
* GaussianDensity Gaussian normalization in 2D systems has been corrected.
```

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,140 @@
...
+//! Get width.
+vec3<unsigned int> SphereVoxelization::getWidth() const
+{
+    return m_width;
+}
+
+//! Compute the voxels array.
+void SphereVoxelization::compute(const freud::locality::NeighborQuery* nq)
+{
+    // set the number of dimensions for the calculation the first time it is done
```
vyasr:
```suggestion
    // Set the number of dimensions for the calculation the first time it is done.
```

cpp/util/utils.h
```diff
@@ -3,9 +3,10 @@
 
 #include <algorithm>
 #include <cmath>
-#include <tbb/parallel_for.h>
 #include <tbb/blocked_range.h>
 #include <tbb/blocked_range2d.h>
+#include <tbb/blocked_range3d.h>
+#include <tbb/parallel_for.h>
```
vyasr:
```suggestion
#include <tbb/parallel_for.h>
```

freud/density.pyx
```diff
@@ -301,6 +300,115 @@ cdef class GaussianDensity(_Compute):
...
+    either be set equally for all dimensions or for each dimension
+    independently.
+
+    Args:
+        width (int or Sequence[int]):
+            The number of bins to make the grid in each dimension (identical
+            in all dimensions if a single integer value is provided).
+        r_max (float):
+            Sphere radius.
+    """  # noqa: E501
```
vyasr: I don't think there's any reason to request longer lines here, usually we only do this in the docstrings for classes where the argument type involves a long `:class:numpy.ndarray...` type (or similar) that overflows on a single line.
```suggestion
    """
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+            True if the box should be considered periodic
+    """
+    eff_rad = r_max / box_size * width
+
+    # enlarge the box for the fft by adding more segments of the same length
+    # we will cut the extra off later so the fft will be aperiodic.
+    if periodic:
+        buf_size = 0
+    else:
+        buf_size = int(round(eff_rad + 1))
```
vyasr:
```suggestion
    buf_size = 0 if periodic else int(round(eff_rad + 1))
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
```
vyasr:
```suggestion
    buf_size = 0 if periodic else int(round(eff_rad + 1))
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+    # set the overlaps to 1, instead of larger integers
+    np.clip(image, 0, 1, out=image)
+    return image
+
+
+def _put_points_on_grid(points, new_width, box_size, width, buf_size, ndim):
+    """
+    Creates a grid where the voxels are 1 if there is a point there (else 0)
+    """
+    d = tuple(new_width for _ in range(ndim))
```
vyasr:
```suggestion
    d = (new_width, )*ndim
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    Does voxelization by doing an aperiodic fft of the sphere over the
+    points on the grid in 3 dimensions
+
+    Args:
+        box_size (float):
+            Length of the (assuemd cubic) box for the calculation.
+        width (int):
+            Number of grid spaces in each direction of the box
+        points (:np.ndarray: (N,3)):
```
vyasr:
```suggestion
        points (:np.ndarray: (N, 3)):
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
```
vyasr:
```suggestion
        points (:np.ndarray: (N, 3)):
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+    # make the grid with the points on it
+    arr = _put_points_on_grid(points, new_width, box_size,
+                              width, buf_size, ndim=3)
+
+    # make the sphere
+    sphere = _make_sphere_3d(new_width, eff_rad)
+
+    # do the ffts
+    fft_arr = np.fft.fftn(arr)
+    fft_arr *= np.fft.fftn(sphere)
```
vyasr:
```suggestion
    fft_arr = np.fft.fftn(arr) * np.fft.fftn(sphere)
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+    # make the grid with the points on it
+    arr = _put_points_on_grid(points, new_width, box_size,
+                              width, buf_size, ndim=2)
+
+    # make the sphere
+    sphere = _make_sphere_2d(new_width, eff_rad)
+
+    # do the ffts
+    fft_arr = np.fft.fft2(arr)
+    fft_arr *= np.fft.fft2(sphere)
```
vyasr:
```suggestion
    fft_arr = np.fft.fft2(arr) * np.fft.fft2(sphere)
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+                      buf_size:-buf_size]
+
+    # set the overlaps to 1, instead of larger integers
+    np.clip(image, 0, 1, out=image)
+    return image
+
+
+def _put_points_on_grid(points, new_width, box_size, width, buf_size, ndim):
+    """
+    Creates a grid where the voxels are 1 if there is a point there (else 0)
```
vyasr:
```suggestion
    Creates a grid where the voxels are 1 if there is a point there and 0 if not.
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    d = tuple(new_width for _ in range(ndim))
+    arr = np.zeros(d)
+    img_points = points / (box_size / width)  # points in units of grid spacing
+    for pt in img_points:
+        shifted_pt = list()
+        for i in range(ndim):
+            shifted_pt.append(int(round(pt[i])))
+        arr[tuple(shifted_pt)] = 1
```
vyasr:
```suggestion
        shifted_pt = tuple(int(round(pt[i])) for i in range(ndim))
        arr[shifted_pt] = 1
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    eff_rad = r_max / box_size * width
+
+    # enlarge the box for the fft by adding more segments of the same length
+    # we will cut the extra off later so the fft will be aperiodic.
+    if periodic:
+        buf_size = 0
+    else:
+        buf_size = int(round(eff_rad + 1))
+    new_width = 2*buf_size + width
```
vyasr:
```suggestion
    new_width = 2 * buf_size + width
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+        shifted_pt = list()
+        for i in range(ndim):
+            shifted_pt.append(int(round(pt[i])))
+        arr[tuple(shifted_pt)] = 1
+    return arr
+
+
+def _make_sphere_3d(new_width, eff_rad):
+    """makes a grid in 3D with voxels that are within eff_rad of the center
+    having value 1 (else 0)"""
```
vyasr:
```suggestion
    """Makes a grid in 3D with voxels that are within ``eff_rad`` of the
    center having value 1 and other voxels having value 0."""
```

tests/SphereVoxelization_fft.py
```diff
@@ -0,0 +1,144 @@
...
+    r_rad = int(round(eff_rad))
+    ctr = new_width // 2
+    arr = np.zeros((new_width, new_width, new_width))
+
+    for i in range(-r_rad, r_rad):
+        for j in range(-r_rad, r_rad):
+            for k in range(-r_rad, r_rad):
+                if np.linalg.norm([i, j, k]) < eff_rad:
+                    arr[ctr+i, ctr+j, ctr+k] = 1
```
vyasr, line 120: Does this need to be modded by `new_width` like in `_make_sphere_2d` below?

cpp/density/SphereVoxelization.cc
```diff
@@ -0,0 +1,140 @@
...
+#include "SphereVoxelization.h"
+
+/*! \file SphereVoxelization.cc
+    \brief Routines for computing voxelized densities from spheres centered at points.
+*/
+
+namespace freud { namespace density {
+
+SphereVoxelization::SphereVoxelization(vec3<unsigned int> width, float r_max)
+    : m_box(box::Box()), m_width(width), m_r_max(r_max), m_has_computed(false)
```
vyasr:
```suggestion
    : m_box(), m_width(width), m_r_max(r_max), m_has_computed(false)
```

cpp/density/GaussianDensity.cc
```diff
@@ -13,7 +13,7 @@
 namespace freud { namespace density {
 
 GaussianDensity::GaussianDensity(vec3<unsigned int> width, float r_max, float sigma)
-    : m_box(box::Box()), m_width(width), m_r_max(r_max), m_sigma(sigma)
+    : m_box(box::Box()), m_width(width), m_r_max(r_max), m_sigma(sigma), m_has_computed(false)
```
vyasr:
```suggestion
    : m_box(), m_width(width), m_r_max(r_max), m_sigma(sigma), m_has_computed(false)
```

vyasr (changes requested):
@tommy-waltmann this is almost ready. I have a number of comments, but I think all but one of them can be applied directly (assuming you agree with them). The one open question I had is I think in the validation testing of SphereVoxelization, specifically in your reference implementation.

tests/SphereVoxelization_fft.py
tommy-waltmann: Under the assumption that the radius is smaller than half the size of the grid, the mod isn't really necessary, and if that assumption doesn't hold, the calculation doesn't even make much sense. I don't remember why I thought it would be necessary to put the mod in `_make_sphere_2d`, but upon closer inspection it shouldn't be necessary and might hide errors if it is left there, so I'm going to remove it.

bdice: The original motivation was that I was experimenting with centering the sphere at `(0, 0)` instead of `(width//2, width//2)`. It's safe to remove.

benchmarks/benchmark_density_GaussianDensity.py
tommy-waltmann: @bdice I don't remember why these were changed, do you?

bdice: There are two ratios, `r_max/L` and `sigma/r_max` that were somewhat unrealistic for practical use, AND the benchmark was taking too long.

Previously, `sigma=0.1` and `r_max=10` meant that `sigma/r_max` was very small, and the contributions near `r=r_max` were at the 100-sigma limit of a normal distribution (effectively zero). Changing that ratio to `sigma/r_max=10` is much more realistic. In practice it's fine to use an even lower value like 3, just depends on how precisely the integral over all grid spaces needs to get to 1. The tests in this PR explicitly check this normalization, which is related to issue #611.

The other relevant ratio is `r_max/L`, which was previously `3.1`. This meant that every point's Gaussian is contributing to (and being computed for) an extremely large proportion of grid cells. Like `4/3 * pi * (1/3.1)**3 = 0.1406`. That's not really how I've seen GaussianDensity used. I assumed a system of 10x10x10 particles in a simple cubic lattice, where every particle has a diameter of 1 (radius 0.5, diameter is equivalent to 10 bin widths since `width=100`).

vyasr (approved):
LGTM!

PR #608, opened 2020-05-13
Fix parameter checking for Cubatic.

bdice:
## Description
I noticed some issues in Cubatic while looking at some unrelated code. I have resolved the following issues:
- `n_replicates` was not accessible via a C++ getter, and was being saved as a Python property, contrary to how all other parameters are handled.
- `seed` was `cdef`'d and conflicted with the property `seed`. This led to inconsistent behavior for `cubatic.seed` depending on whether a seed was provided or left as the default.
- The error handling for input parameters `t_initial`, `t_final`, and `scale` was duplicated between Python and C++ and it wasn't even consistent between the two. All error handling is now done in C++.
- The `seed` value could be provided as any type, and non-integer types would throw a warning and use the system time instead. Invalid seeds _should_ raise an error, rather than using system time and being non-reproducible for invalid-but-consistent seeds (e.g. seeds provided as a string). That is now enforced through Cython's knowledge of parameter types.
- I also did some minor re-ordering so that the pxd and C++ files were more in line with each other.

## Motivation and Context
Improves usability and testability of Cubatic.

## How Has This Been Tested?
Expanded testing of error handling in Cubatic.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_order_Cubatic.py
+++ b/tests/test_order_Cubatic.py
@@ -96,6 +96,31 @@ class TestCubatic(unittest.TestCase):
             op_max, 0.2,
             err_msg="per particle order parameter value is too high")
 
+    def test_valid_inputs(self):
+        with self.assertRaises(ValueError):
+            # t_initial must be greater than t_final
+            freud.order.Cubatic(
+                t_initial=0.001,
+                t_final=5.0,
+                scale=0.95,
+                n_replicates=10)
+
+        with self.assertRaises(ValueError):
+            # t_final must be greater than 1e-6
+            freud.order.Cubatic(
+                t_initial=5.0,
+                t_final=1e-7,
+                scale=0.95,
+                n_replicates=10)
+
+        with self.assertRaises(ValueError):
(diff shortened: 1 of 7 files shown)


tests/test_order_Cubatic.py
```diff
@@ -96,6 +96,31 @@ def test_disordered(self):
...
+                n_replicates=10)
+
+        with self.assertRaises(ValueError):
+            # scale must be less than 1
+            freud.order.Cubatic(
+                t_initial=5.0,
+                t_final=0.001,
+                scale=1.1,
+                n_replicates=10)
+
```
b-butler, line 131: Since we are testing all error conditions here we might as well have one for scale less than zero.

b-butler (approved):
I left an additional comment on a possible test case, but it LGTM.

tests/test_order_Cubatic.py
bdice: Great idea! Fixed.

PR #613, opened 2020-05-18
Refactor/pmft tests

vyasr:
## Description

I consolidated most of the common testing for PMFTs into a single class.

## Motivation and Context

There was a lot of duplication in the PMFT tests that made it difficult to validate that all methods are behaving correctly without manually checking each one. This cleanup helps with that. There were also some errors with the way that bins were computed in a few places that are not actually being tested right now because most orientations in the test are being set to 0. For the most part, the underlying classes are being tested, only the methods defined in the tests themselves had problems. There are a few exceptions, for which I will make additional PRs.

## How Has This Been Tested?

Tests still pass.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [ ] I have updated the documentation (if relevant).
- [ ] I have added tests that cover my changes (if relevant).
- [x] All new and existing tests passed.
- [x] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst).
- [ ] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

PR #614, opened 2020-05-18
Fix/pmft normalization

vyasr:
## Description

This PR should be merged after #613. It fixes and adds tests for the normalization in the PMFT.

## Motivation and Context

This PR is a follow-up that fixes problems with #523, which attempted to resolve problems with PMFT normalization. While #523 correctly identified that normalization was inconsistent among the different PMFT classes, it applied the incorrect solution. That PR modified the Jacobian in PMFTXY and PMFTXYZ to include an additional factor corresponding to the angular degrees of freedom that were implicitly integrated out. However, that part of the volume element was also ignored in the volume normalization applied in PMFT::reduce, so the net result was previously correct for those classes but incorrect for PMFTXYT and PMFTR12. This PR rectifies that problem by removing the added normalizations from #523 from PMFTXY and PMFTXYZ, as well as removing the unnecessary factors of 2*PI from PMFTXYT and PMFTR12 that previously caused disagreement between classes.

## How Has This Been Tested?

Tests for disagreement between different PMFTs were added in #523 and still pass. A new test was added to validate the expecting limiting behavior of the pair correlation function for an ideal gas of particles.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -10,6 +10,7 @@ and this project adheres to
 * Support for garnett 0.7.
 * Custom NeighborLists can be created from a set of points using `from_points`. Distances will be calculated automatically.
 * The Box class has methods `compute_distances` and `compute_all_distances` to calculate distances between arrays of points and query points.
+* Hexatic can now compute 2D Minkowski Structure Metrics, using `weighted=True` along with a Voronoi NeighborList.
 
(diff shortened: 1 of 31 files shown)


vyasr:
@klywang approved these changes as part of #615.

PR #615, opened 2020-05-18
Fix/pmftxyt orientation

vyasr:
## Description

Fixes uses of orientations and query_orientations in PMFTXYT. Should be merged after #614.

## Motivation and Context

This class is currently using the wrong set of orientations to rotate the local frame.

In debugging this, I realized that the current coordinate system makes some awkward choices, namely the theta bins are computed relative to the vector pointing from the point to the query_point. This means that if the point is oriented such that the vector corresponding to the default orientation is aligned with the r_ij vector (in the direction pointing towards the query_point), then theta is defined as 0. Not only is this choice backwards from the `bondVector` calculation in freud's `NeighborComputeFunctional`, it completely neglects the orientation of the query_point. However, this is just a convention choice, so there is no correct answer. Since this behavior has been consistent for a long time (this has not changes since pre-2.0 versions of freud), I'm inclined to just add some documentation of the definition of the angle and move forward with it as is. Does that sound right? This also affects PMFTR12, which makes some arbitrary choice about what theta1 and theta2 are, so I'll update those docstrings too if appropriate.

## How Has This Been Tested?

New test was added to validate the expected behavior.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -18,6 +18,7 @@ and this project adheres to
 * Histogram bin locations are computed in a more numerically stable way.
 * Improved error handling of Cubatic input parameters.
 * PMFT's are now properly normalized such that the pair correlation function tends to unity for an ideal ga.
+* PMFTXYT uses the correct orientations when points and query\_points differ.
 
 ## v2.2.0 - 2020-02-24
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -17,6 +17,7 @@ and this project adheres to
 ### Fixed
 * Histogram bin locations are computed in a more numerically stable way.
 * Improved error handling of Cubatic input parameters.
+* PMFT's are now properly normalized such that the pair correlation function tends to unity for an ideal ga.
 
 ## v2.2.0 - 2020-02-24
 
--- a/cpp/pmft/PMFT.h
+++ b/cpp/pmft/PMFT.h
@@ -60,7 +60,7 @@ protected:
      *  least one angular term, but that term should not contain a factor of
      *  2*PI since that factor is effectively divided out of the volume here.
      *
-     *  \param JacobFactor A function with one parameter (the histogram bind index) that returns the volume of the element in the histogram bin corresponding to the index.
+     *  \param JacobFactor A function with one parameter (the histogram bin index) that returns the volume of the element in the histogram bin corresponding to the index.
      */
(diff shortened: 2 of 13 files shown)


ChangeLog.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to
 ### Fixed
 * Histogram bin locations are computed in a more numerically stable way.
 * Improved error handling of Cubatic input parameters.
+* PMFT's are now properly normalized such that the pair correlation function tends to unity for an ideal ga.
```
bdice, line 21:
```suggestion
* PMFTs are now properly normalized such that the pair correlation function tends to unity for an ideal gas.
```

cpp/pmft/PMFT.h
```diff
@@ -4,6 +4,8 @@
 #ifndef PMFT_H
 #define PMFT_H
 
+#include <tbb/tbb.h>
```
bdice, line 7: We recently replaced `tbb.h` with specific headers. #605

cpp/pmft/PMFT.h
```diff
@@ -33,6 +35,33 @@ class PMFT : public locality::BondHistogramCompute
...
+     *
+     *  **IMPORTANT NOTE**: The inv_num_dens factor in the calculation in this
+     *  function is just volume / Np, so it does not include volume elements in
+     *  the orientational degrees of freedom. This means that the corresponding
+     *  normalization factors *should not* be applied to the Jacobian factor
+     *  argument. For instance, any full-dimensional PMFT in 2D must contain at
+     *  least one angular term, but that term should not contain a factor of
+     *  2*PI since that factor is effectively divided out of the volume here.
+     *
+     *  \param JacobFactor A function with one parameter (the histogram bind index) that returns the volume of the element in the histogram bin corresponding to the index.
```
bdice:
```suggestion
     *  \param JacobFactor A function with one parameter (the histogram bin index) that returns the volume of the element in the histogram bin corresponding to the index.
```

ChangeLog.md
vyasr: Lol thanks.

cpp/pmft/PMFT.h
vyasr: I'm honestly not even sure how this got added here. Removed in #614.

klywang (approved):
This looks good!

PR #618, opened 2020-05-18
Add weighted feature to Hexatic.

### LINKED ISSUE #533: Plots should  use `plt.subplots`
In settings like JupyterLab there are occasionally issues with plots displaying when no axis is provided. The reason is likely because we are fully using the object-oriented API of `matplotlib`. While this usage pattern is generally the best way to use the code now, the one exception is that instead of instantiating figures and axes in that manner, we should still use `pyplot.subplots` to create the objects. This method will ensure sufficient global state for proper displaying when the backends are all set correctly.

### LINKED ISSUE #617: Weighted Hexatic (Minkowski Structure Metrics in 2D systems)

@gabriele16 requested the ability to calculate the order parameter psi_k by weighting the angle theta_ij between two neighbors by the length of their Voronoi side, similarly to what done in these two papers:

https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.114.035702
https://aip.scitation.org/doi/10.1063/1.4774084

Add a `weighted` argument to enable the Minkowski Structure Metric variant of the Hexatic order parameter.

**Additional context**
This tool has an interactive example: https://morphometry.org/

bdice:
## Description
Adds support for weighted bonds in the Hexatic order parameter, enabling Minkowski Structure Metric computations in 2D systems. Requested by user @gabriele16.

## Motivation and Context
Resolves: #617.

## How Has This Been Tested?
(Not yet tested.)

## Screenshots
index

## Types of changes

## Checklist:

### DIFF
--- a/cpp/order/HexaticTranslational.cc
+++ b/cpp/order/HexaticTranslational.cc
@@ -53,8 +53,8 @@ void Hexatic::compute(const freud::locality::NeighborList* nlist,
 {
     computeGeneral(
         [this](const vec3<float>& delta) {
-            const float psi_ij = std::atan2(delta.y, delta.x);
-            return std::exp(std::complex<float>(0, m_k * psi_ij));
+            const float theta_ij = std::atan2(delta.y, delta.x);
+            return std::exp(std::complex<float>(0, m_k * theta_ij));
         },
         nlist, points, qargs);
 }
--- a/cpp/order/HexaticTranslational.cc
+++ b/cpp/order/HexaticTranslational.cc
@@ -22,22 +22,29 @@ void HexaticTranslational<T>::computeGeneral(Func func, const freud::locality::N
     freud::locality::loopOverNeighborsIterator(
         points, points->getPoints(), Np, qargs, nlist,
         [=](size_t i, std::shared_ptr<freud::locality::NeighborPerPointIterator> ppiter) {
-            const vec3<float> ref = (*points)[i];
+            float total_weight(0);
+            const vec3<float> ref((*points)[i]);
 
             for (freud::locality::NeighborBond nb = ppiter->next(); !ppiter->end(); nb = ppiter->next())
             {
                 // Compute vector from query_point to point
                 const vec3<float> delta = box.wrap((*points)[nb.point_idx] - ref);
+                const float weight(m_weighted ? nb.weight : 1.0);
 
                 // Compute psi for this vector
-                m_psi_array[i] += func(delta);
+                m_psi_array[i] += weight * func(delta);
+                total_weight += weight;
             }
-
-            m_psi_array[i] /= std::complex<float>(m_k);
+            if (m_weighted)
+            {
+                m_psi_array[i] /= std::complex<float>(total_weight);
+            } else {
+                m_psi_array[i] /= std::complex<float>(m_k);
+            }
         });
 }
 
-Hexatic::Hexatic(unsigned int k) : HexaticTranslational<unsigned int>(k) {}
+Hexatic::Hexatic(unsigned int k, bool weighted) : HexaticTranslational<unsigned int>(k, weighted) {}
 
 Hexatic::~Hexatic() {}
 
@@ -52,7 +59,7 @@ void Hexatic::compute(const freud::locality::NeighborList* nlist,
         nlist, points, qargs);
 }
 
-Translational::Translational(float k) : HexaticTranslational<float>(k) {}
+Translational::Translational(float k, bool weighted) : HexaticTranslational<float>(k, weighted) {}
 
 Translational::~Translational() {}
 
--- a/cpp/order/HexaticTranslational.h
+++ b/cpp/order/HexaticTranslational.h
@@ -26,7 +26,7 @@ template<typename T> class HexaticTranslational
 {
 public:
     //! Constructor
-    HexaticTranslational(T k) : m_k(k) {}
+    HexaticTranslational(T k, bool weighted = false) : m_k(k), m_weighted(weighted) {}
 
     //! Destructor
     virtual ~HexaticTranslational() {}
@@ -34,6 +34,11 @@ public:
     T getK() const
     {
         return m_k;
+    }
+
+    bool isWeighted() const
+    {
+        return m_weighted;
     }
 
     //! Get a reference to the order parameter array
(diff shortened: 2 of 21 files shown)


bdice:
@gabriele16 can you comment on this PR? I'd like to add you as a reviewer.

gabriele16:
please add me as a reviewer

cpp/order/HexaticTranslational.cc
```diff
@@ -22,22 +22,29 @@ void HexaticTranslational<T>::computeGeneral(Func func, const freud::locality::N
...
 
             for (freud::locality::NeighborBond nb = ppiter->next(); !ppiter->end(); nb = ppiter->next())
             {
                 // Compute vector from query_point to point
                 const vec3<float> delta = box.wrap((*points)[nb.point_idx] - ref);
+                const float weight(m_weighted ? nb.weight : 1.0);
 
                 // Compute psi for this vector
-                m_psi_array[i] += func(delta);
+                m_psi_array[i] += weight * func(delta);
```
gabriele16, line 35: The summation over the neighbours to calculate sum_ij weight*exp( k*i*theta_ij) with the weights is correct.

cpp/order/HexaticTranslational.cc
```diff
@@ -22,22 +22,29 @@ void HexaticTranslational<T>::computeGeneral(Func func, const freud::locality::N
...
+                total_weight += weight;
+            }
+            if (m_weighted)
+            {
+                m_psi_array[i] /= std::complex<float>(total_weight);
```
gabriele16, line 40: normalization also is correct.

bdice:
@gabriele16 Here is a Jupyter notebook that produces results similar to my expectations.
[order.Hexatic-Minkowski.ipynb.zip](https://github.com/glotzerlab/freud/files/4645161/order.Hexatic-Minkowski.ipynb.zip)

cpp/order/HexaticTranslational.h
```diff
@@ -48,40 +53,41 @@ template<typename T> class HexaticTranslational
...
                         const freud::locality::NeighborQuery* points, freud::locality::QueryArgs qargs);
 
-    const T m_k;
+    const T m_k;     //!< The symmetry order for Hexatic, or normalization for Translational
+    const bool m_weighted; //!< Whether to use neighbor weights in computing the order parameter (default false)
     util::ManagedArray<std::complex<float>> m_psi_array; //!< psi array computed
 };
 
-//! Compute the translational order parameter for a set of points
+//! Compute the hexatic order parameter for a set of points
```
bdice, line 61: I swapped the order of the `Hexatic` and `Translational` classes so that it mirrors the order in the rest of the code.

freud/order.pyx
```diff
@@ -255,11 +255,16 @@ cdef class Hexatic(_PairCompute):
     Args:
         k (unsigned int, optional):
             Symmetry of order parameter. (Default value = :code:`6`).
+        weighted (bool, optional):
+            Determines whether to use neighbor weights in the computation of
+            spherical harmonics over neighbors. If enabled and used with a
+            Voronoi neighbor list, this results in the Minkowski Structure
+            Metrics :math:`q'_l`. (Default value = :code:`False`)
```
bdice: @gabriele16 What's the best symbol to use here? I typically use `\psi` to refer to hexatic / 2D and `q` to refer to Steinhardt-like 3D parameters, but I see some literature uses `q` for the 2D case.

freud/plot.py
```diff
@@ -7,7 +7,7 @@
 import warnings
 
 try:
-    from matplotlib.figure import Figure
+    import matplotlib.pyplot as plt
```
bdice, line 10: @vyasr FYI, I fixed this issue we discussed offline a while back. Using `pyplot` ensures that the figure is "managed" and can be displayed.

bdice (commented):
Some comments to help reviewers.

freud/order.pyx
gabriele16: Using _psi_k_ for the hexatic order throughout might be better and is consistent with the rest of the code, also in the examples. I have not seen using _q_ for 2D systems.

bdice: I decided to use `\psi'_k` for the 2D Minkowski Structure Metric, to match what we've done with Steinhardt, where `q'_l` is the 3D Minkowski Structure Metric.

gabriele16:
@bdice Thanks for sharing the notebook with the example and for the implementation. The calculation seems correct based on the example provided. However, I tested this using a simple function that takes in the voronoi weights and computes the hexatic order parameter explicitly and I find that the values of psi_k obtained in the two different ways disagree. The notebook is attached. The function is get_weighted_psi. I believe there should be a bug in get_weighted_psi but if I set the weights to 1 I get the same results. Could @bdice  or @vyasr please comment on this, so if we find that the two versions agree I can proceed with the review? Thanks.
[order.Hexatic-Minkowski.ipynb.zip](https://github.com/glotzerlab/freud/files/4649259/order.Hexatic-Minkowski.ipynb.zip)

gabriele16:
I have tested the weights feature on the shared notebook and on a different system and it gives the correct results I believe.  However, using a simple function to explicitly compute psi_k along with voronoi.nlist.weights, I get different results. Somehow, voro.nlist.weights might be different from the weights computed with freud.order.Hexatic(k=k, weighted=True). The function works correctly if the weights are set to 1 as I get the same results as in freud.order.Hexatic(k=k, weighted=False). I have attached the notebook, which is edited from the original one. If @bdice or @vyasr  have an idea where this went wrong, it would help to proceed with the review. Thanks.
[order.Hexatic-Minkowski.ipynb 2.zip](https://github.com/glotzerlab/freud/files/4650947/order.Hexatic-Minkowski.ipynb.2.zip)

bdice:
@gabriele16 Can you share the function you are using to compute the comparison data?

gabriele16:
The notebook with the comparison is attached here.
[order.Hexatic-Minkowski.exp.ipynb.zip](https://github.com/glotzerlab/freud/files/4651299/order.Hexatic-Minkowski.exp.ipynb.zip)

bdice:
@gabriele16 I think the inconsistency is caused by (1) differences in the NeighborLists and (2) an index error in the code.

(1) The notebook code is combining some data from an AABBQuery "nearest neighbors" query with data from a Voronoi neighbor search. However, the two NeighborLists aren't consistent with each other. The Voronoi neighbors aren't the same as the 6 nearest neighbors for many particles (and that's also true for some particles with 6 neighbors, not just the defects with 5 or 7). The assertion below fails.
```python
if weighted == True:
        voro = freud.locality.Voronoi()
        voro.compute((box, positions))
        voro_len = voro.nlist.weights
        assert (voro.nlist.query_point_indices == nlist.query_point_indices).all()
    else:
        voro_len = np.ones(len(nlist.point_indices))
```

2. It looks like that notebook used the wrong neighbor weights. It used the _point index_ `j`, rather than the _neighbor index_ corresponding to a given neighbor bond in the list.

I have made some slight modifications and the explicit Python calculation now matches the freud calculation within single-precision. See attached.
[order.Hexatic-Minkowski.exp2.zip](https://github.com/glotzerlab/freud/files/4652190/order.Hexatic-Minkowski.exp2.zip)

gabriele16 (approved):

gabriele16:
All looks correct to me. The test notebook shows that the explicit python calculation gives the same results as the implemented freud feature within single precision, so I have closed the review. Thanks @bdice for the implementation and spotting the inconsistency in my previous notebook.

vyasr (approved):
This looks good, thanks for making the change so quickly Bradley.

PR #619, opened 2020-05-18
Fix/pmftxyz eq ors normalization

### LINKED ISSUE #616: Fix normalization of accumulated quantities (originally: PMFTXYZ is missing normalization by equiv orientations)
If `equiv_orientations` are provided to `PMFTXYZ`, it currently will not divide by the number of equivalent orientations, leading to overcounting. It looks like this issue was introduced in #300 when the redundant logic between various PMFT classes was consolidated. There is unfortunately an API question here, not something that can be immediately fixed, because right now the equivalent orientations are provided in the call to `compute` and therefore could change between calls. If this is desirable, we will need to use a histogram that support non-integral bins so that we can perform a normalization during accumulation. If it's not desirable, then we should probably accept those as constructor arguments instead. For the 2.x line, I think the most sensible solution is to simply throw an error in `compute` if these differ in subsequent calls where reset is not called.

**Update**: The issue originally raised here is resolved by #619. However, the broader problems along this line are discussed in @bdice's comment below and remain unresolved.

vyasr:
## Description

Store and use the last number of equivalent orientations to normalize the PMFT. This PR should be merged after #615 and only following discussion regarding whether this solution is appropriate on #616.

## Motivation and Context

Resolves: #616.

## How Has This Been Tested?

Updated existing test to include checking the behavior when multiple equivalent quaternions are specified.

## Types of changes

## Checklist:


bdice (changes requested):
I'd like to have some discussion on #619 before we accept this solution.

vyasr:
> I'd like to have some discussion on #619 before we accept this solution.

I assume you mean #616 as mentioned in the PR.

bdice (approved):
LGTM, unless there's further discussion on #616.

PR #622, opened 2020-05-18
Updated docs for Cubatic order parameter

patricklawton:
## Description
Updated the documentation of the cubatic order parameter to reflect the theory, terminology and math used in the cited paper. 

## Motivation and Context
Helpful to users for understanding the order parameter being used and whether or not it is a relevant parameter for their system, while giving users the minimum amount of information to follow the logic of the source paper. 

## How Has This Been Tested?
Tested documentation with sphinx.

## Types of changes

## Checklist:


freud/order.pyx
```diff
@@ -32,9 +32,33 @@ logger = logging.getLogger(__name__)
...
+    :math:`m(\Omega)=\langle M(H_4) \rangle_0`. Explicitly: 
+
+    .. math::
+      \min_{\Omega}\overline{M} \circ m_{\Omega}
+
+    Here, :math:`H_4` is the so called strong orientational coordinate of a cube, 
+    expressed in terms of Cartesian basis vectors. :math:`H_4` is a rank 4 
+    homogeneous tensor which is invariant under orthogonal transformations of 
+    the basis vectors, and which maps directly onto a given set of vectors 
+    describing cubes orientation. :math:`\Omega` defines the geometric features
```
```suggestion
    describing a cube's orientation. :math:`\Omega` defines the geometric features
```

freud/order.pyx
```diff
@@ -32,9 +32,33 @@ logger = logging.getLogger(__name__)
...
+    which characterize the cubatic phase, and is comprised of the vectors
+    :math:`M`, :math:`\overline{M}`, :math:`m`. A distribution function 
+    describing the the system's orientational phase, :math:`p_0(H_4;\Omega)`,
+    can then be constructed using this orientational coordinate. Thus, solving 
+    the optimization problem indicated above yields a scalar value which 
+    quantifies the degree of cubatic ordering for the system. Freud solves this
```
bdice: "This optimization problem is solved..." (Avoid starting the sentence with the word "freud" -- we avoid that because we don't want to capitalize it and confuse this package with the person Sigmund Freud.)

freud/order.pyx
```diff
@@ -32,9 +32,33 @@ logger = logging.getLogger(__name__)
...
+    the basis vectors, and which maps directly onto a given set of vectors 
+    describing cubes orientation. :math:`\Omega` defines the geometric features
+    which characterize the cubatic phase, and is comprised of the vectors
+    :math:`M`, :math:`\overline{M}`, :math:`m`. A distribution function 
+    describing the the system's orientational phase, :math:`p_0(H_4;\Omega)`,
+    can then be constructed using this orientational coordinate. Thus, solving 
+    the optimization problem indicated above yields a scalar value which 
+    quantifies the degree of cubatic ordering for the system. Freud solves this 
+    optimization problem using simulated annealing as opposed to the 
+    Newton-Raphson root finding method.
```
```suggestion
    Newton-Raphson root finding method described in :cite:`Haji_Akbari_2015`.
```
(Why do we prefer simulated annealing? I don't immediately know the answer to this but it would be good to explain.)

bdice (commented):
Left some suggestions for improvement. @patricklawton this is a good overview of the theory. It would be helpful if you could add comments that explain how the various computed properties like `order` or `particle_order` map to the mathematical terms shown in this docstring.

freud/order.pyx
bdice: I saw a comment in the C++ code:
```cpp
    // The paper recommends using a Newton-Raphson scheme to optimize the order
    // parameter, but in practice we find that simulated annealing performs
    // much better, so we perform replicates of the process and choose the best
    // one.
```

bdice:
I gave this a brief review while considering whether it should be merged for the 2.3.0 release. I think there is some further work to be done. This is a good step in the right direction but leaves out some details. The math symbols used here don't quite match the meaning of the symbols in the original paper and need a bit more clarification so the reader knows why these expressions are being shown (and not the others in the paper).

I think we'll need to re-evaluate what level of detail is appropriate, and consider a higher-level approach that explains the intentions of the method without spelling it all out mathematically. I suggest that we refer readers to the original paper for equations, since the equations aren't exactly straightforward to parse without the context of the early parts of the paper. For some methods, the equation *is* what readers really want, because it's concise and hopefully self-explanatory -- that was my original expectation when I suggested this docstring improvement. However, this method may be an exception because of the complexity of defining the mathematical objects being used in a short space like a docstring. I'm open to others' thoughts on this.

vyasr:
@bdice I was not involved with your discussions with @patricklawton about this pull request, so I'm playing a little catch-up. My concern is aligned with what you mention in your last comment: this method may be too complex to document the math sufficiently within freud. For context, prior to my thorough rewrite of the C++ in #427 (which took the better part of a day of reading the paper), would you have felt comfortable that you actually understood the method even as someone who has probably read the C++ code before? For that matter, even after my rewrite do you think the method could be comprehended by reading the code? I think there's no shame in the answer being a resounding no, because in my experience talking to other members of the group most people who have tried to read the paper don't understand it. IIRC even Harper was only able to write the original version of this code with help from Jim to actually work through the math. The high level of mathematical sophistication and abstraction makes me think that what we really need is for someone in the group to write a more approachable summary of the paper itself, which is outside the scope of freud's docs. I'd be happy to help with that once I'm closer to done with thesis-writing 🙃  but I'm not sure what the appropriate goal posts are for the documentation here.

bdice:
@vyasr I concur with your thoughts on this. @patricklawton and I were aiming to add more math to the documentation _in general_ and Cubatic was a somewhat arbitrary choice of starting point that turns out to be _very hard to summarize well_. I am still in favor of adding more math to docstrings in general, but Cubatic may not be an appropriate target for the scope of complexity that I had in mind. I would recommend closing this PR, making an issue (currently on mobile or else I’d do it now - though I thought one existed and I could not find it), and describing the scope of work with specific methods that would benefit from an equation or two that might be almost self-explanatory but make freud’s conventions unambiguous.

vyasr:
Done, see #649.

PR #624, opened 2020-05-27
Fixed Steinhardt for particles with no neighbors

mmartirossyan:
## Description

Particles with 0 neighbors were have NaN order parameter values because there was a division by 0. This prevents the division by 0.

## Motivation and Context

Resolves: #???

It is useful for cases where particles have no neighbors (ie other particles are further apart than the r_max defined).

## How Has This Been Tested?

There has been a test_no_neighbors function added to the test_order_Steinhardt.py file which initializes a single particle in a box and computes its Steinhardt order parameter, and makes sure that it is 0.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,7 @@ and this project adheres to
 * Improved error handling of Cubatic input parameters.
 * PMFTs are now properly normalized such that the pair correlation function tends to unity for an ideal gas.
 * PMFTXYT uses the correct orientations when points and query\_points differ.
+* Fixed Steinhardt for particles without neighbors.
 
 ## v2.2.0 - 2020-02-24
 
--- a/cpp/order/Steinhardt.cc
+++ b/cpp/order/Steinhardt.cc
@@ -135,7 +135,8 @@ void Steinhardt::baseCompute(const freud::locality::NeighborList* nlist,
             {
                 // Cache the index for efficiency.
                 const unsigned int index = m_qlmi.getIndex({static_cast<unsigned int>(i), k});
-                m_qlmi[index] /= total_weight;
+                if (total_weight > 0)
+                    m_qlmi[index] /= total_weight;
                 // Add the norm, which is the (complex) squared magnitude
                 m_qli[i] += norm(m_qlmi[index]);
                 // This array gets populated by computeAve in the averaging case.
--- a/doc/source/reference/credits.rst
+++ b/doc/source/reference/credits.rst
@@ -287,6 +287,10 @@ Mike Henry
 * Fixed syntax in freud-examples notebooks for v2.0.
 * Updated documentation links
 
+Maya Martirossyan
+
+* Fixed Steinhardt for particles without neighbors.
+
 Source code
 -----------
 
--- a/tests/test_order_Steinhardt.py
+++ b/tests/test_order_Steinhardt.py
@@ -309,11 +309,12 @@ class TestSteinhardt(unittest.TestCase):
 
     def test_no_neighbors(self):
         box = freud.box.Box.cube(10)
-        positions = [(0,0,0)]
+        positions = [(0, 0, 0)]
         comp = freud.order.Steinhardt(6)
-        comp.compute((box, positions), neighbors={'r_max':1.25})
-        
+        comp.compute((box, positions), neighbors={'r_max': 1.25})
+
         npt.assert_allclose(comp.particle_order, [0])
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/tests/test_order_Steinhardt.py
+++ b/tests/test_order_Steinhardt.py
@@ -307,6 +307,13 @@ class TestSteinhardt(unittest.TestCase):
 
         st._repr_png_()
 
+    def test_no_neighbors(self):
+        box = freud.box.Box.cube(10)
+        positions = [(0,0,0)]
+        comp = freud.order.Steinhardt(6)
+        comp.compute((box, positions), neighbors={'r_max':1.25})
+        
+        npt.assert_allclose(comp.particle_order, [0])
 
 if __name__ == '__main__':
     unittest.main()


vyasr:
Hi @mmartirossyan, thanks for the pull request! It looks like your code needs a few style fixes to pass our tests, which you can see by clicking on the "Details" link next to the 'ci/circleci' check below. You can also run these locally if you install flake8 (`pip install flake8` would work). If you plan on making further changes, it can be helpful to install this as a pre-commit hook (see [here](https://freud.readthedocs.io/en/latest/reference/development/howtoadd.html#python) for instructions on how to do that).

vyasr:
After discussion with @bdice, we actually think the existing behavior is probably more appropriate. Since the order parameter is defined over bonds, it really isn't well-defined for particles with no bonds at all; we could just as easily choose to define this case as infinity or 1. As a result, we would prefer not to modify this behavior. If the NaNs are causing issues for your use case, would a simple replacement with 0 satisfy your needs? Something like

```
op[numpy.isnan[op]] = 0
```

klarh:
@bdice @vyasr I want to put my two cents in and advocate for classes in `freud.order` to return 0 when calculating averages for particles with no neighbors. I can see the mathematical argument behind having these return nans in that case, but from a practical standpoint I think it simply adds a burden to the user for an atypical (but not impossible) case. Assuming for the sake of discussion that the two alternatives are:

1. Set the value for particles with no neighbors to 0; postprocessing steps that need to know whether the value would be exactly 0 or nan with the current setup have to check the number of neighbors that those particles had
2. Set the value for particles with no neighbors to nan; postprocessing steps that are expecting a real value have to filter out nans before continuing

I think it is more useful to go with option 1, since in that way the exceptional behavior is less problematic for the typical types of analysis we do.

A more physicsy perspective might say that the things in `freud.order` are supposed to be order parameters, which should be indicator functions for a phase—taking a value of 0 in one phase, and some other value in the other phase. In the case of identifying solids with the Steinhardt order parameter (with an appropriately-set cutoff radius), low-density particles with no bonds are not part of the solid phase, and should have a low value of the order parameter. @mmartirossyan also made a good point in a private discussion that order parameters should be robust to edge cases, and I would argue that returning a nan decreases that robustness.

bdice:
@klarh I am confused by your comment. You said at the top that you want to return zero, but you prefer option 2 which returns NaN? I read it a few times and didn't quite follow your logic.

I think NaN is the correct return value because the request to "sum spherical harmonics in a rotationally invariant way over this particle's neighbors" is ill defined when there are no neighbors to iterate over. For users that want this result to be zero (or one, or infinity, which all seem equally possible given the ill-defined nature of the calculation), I would recommend the code snippet suggested by @vyasr. I think it's a stretch for freud (as a library that generally avoids making assumptions about its input data) to take a specific stance on the "physicality" of an mathematically undefined quantity that would lead it to being zero.

Regarding robustness: Users seeking a "robust" order parameter should refer to https://aip.scitation.org/doi/10.1063/1.4774084 and use Minkowski Structure Metrics, via computing Voronoi neighbors and computing Steinhardt with that neighbor list and `weighted=True`. (See #618 for examples of this in 2D systems.) "Robustness" is a well known issue for many forms of the Steinhardt order parameter and it has been addressed in the literature in several ways (see also https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.75.2714).

@mmartirossyan Your pull request is excellent overall, and I sincerely appreciate your willingness to contribute even though I'm not sure it's the right change to make in this case. We have some options that could move us forward. Please let us know what you think.

Potential next steps:
1. Use @vyasr's code snippet to get the results you want for now.
2. Document the expected behavior of NaN in the case of particles having no neighbors.
3. Make sure this behavior of "returning NaN for order parameters computed over neighbors when no neighbors exist" is consistent across the various options for Steinhardt (e.g. third-order invariants Wl, second-shell-averaged variants, etc.) and other order parameters (e.g. Hexatic, maybe SolidLiquid?) that do similar math.

klarh:
@bdice (responding in pieces)

> @klarh I am confused by your comment. You said at the top that you want to return zero, but you prefer option 2 which returns NaN? I read it a few times and didn't quite follow your logic.

Ah, yes, of course I meant 1. I'll edit my comment.

> I think NaN is the correct return value because the request to "sum spherical harmonics in a rotationally invariant way over this particle's neighbors" is ill defined when there are no neighbors to iterate over. For users that want this result to be zero (or one, or infinity, which all seem equally possible given the ill-defined nature of the calculation), I would recommend the code snippet suggested by @vyasr. I think it's a stretch for freud (as a library that generally avoids making assumptions about its input data) to take a specific stance on the "physicality" of an mathematically undefined quantity that would lead it to being zero.

I would say _summing_ zero complex or real numbers is perfectly well-defined to be the additive identity of 0, but I think you must mean average; there I agree that it is not well-defined, but for a package that is computing order parameters, I think we can come to the conclusion that the value should be 0 based on which use case is more useful: producing nans (which requires a layer of boilerplate to mask or remove the nans for most or all of the places where people are using freud's order parameter code) or producing zeros (which don't percolate destructively through the most straightforward statistical analysis that users will nearly always be performing on results computed by freud). I think that producing zeros is further reinforced by considering an approximate sense of monotonicity or smoothness for order parameters that are intended to indicate solidity in some way—like Steinhardt—where they would decrease in magnitude as the density decreases to 0.

Overall, I think that it is simply a practically-useful choice to have the value for particles with no neighbors be 0; if it is ill-defined, and we have to pick something to return, I say we might as well pick a value that doesn't cause errors when visualizing results and performing statistical analysis of data. Why force users to coat their interactions with freud in a layer of boilerplate that won't be needed the majority of the time?

vyasr:
@bdice @klarh to be honest I don't have a strong opinion on what the right behavior is and I see the merits of both arguments. For what it's worth, the scientific Python ecosystem seems pretty consistent on its treatment of averages of empty sets (largely by virtue of matching NumPy I suspect). For instance, here's a random sampling:

```
>>> import pandas as pd
>>> import numpy as np
>>> import tensorflow as tf
>>> np.mean([])
/Users/vramasub/miniconda3/envs/main/lib/python3.7/site-packages/numpy/core/fromnumeric.py:3335: RuntimeWarning: Mean of empty slice.
  out=out, **kwargs)
/Users/vramasub/miniconda3/envs/main/lib/python3.7/site-packages/numpy/core/_methods.py:161: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
nan
>>> pd.Series(dtype=float).mean()
nan
>>> pd.DataFrame().mean().mean()
nan
>>> tf.math.reduce_mean(tf.constant([]))
... # Bunch of tensorflow output
<tf.Tensor: shape=(), dtype=float32, numpy=nan>
```

The issue really arises due to the approximation of building neighbor lists with finite size. You can guarantee nonzero neighbors if you use a k-nearest neighbors search rather than a purely distance-based cutoff, which I would argue is always a more robust choice for something like Steinhardt anyway. The one exception in the case where you have a single particle, in which case NaN is unequivocally the correct choice because (even ignoring details about how large N really needs to be for the thermodynamic limit to apply) a single particle does not constitute a phase. My main concern with automatically masking out NaNs is that it could prevent users from recognizing that they have made a poor choice of neighbor querying parameters, so if we want to go the automatic masking route, I think we at least need to warn loudly that this occurred rather than silently provide users possibly undesirable output.

mmartirossyan:
@vyasr @bdice 

I appreciate you all adding your comments and thinking through the specifics of this pull request, and I'm going to jump in and add my two cents here.

I will re-state the earlier point that @klarh made, which is that an order parameter should be robust to all edge cases. This should include, for example, crystal structures that include particles that, by definition, have 0 neighbors (ex: the central particle in a clathrate cage structure). For that structure, if NaN's are thrown and the user is forced to modify the Steinhardt neighbor parameters like you suggest (by say, increasing the r_max), they would end up misusing the parameter by forcing that particle to have neighbors.

I think that the user can still realize they have made a poor choice of neighbor parameters if you don't throw NaN's- if there are an abundance of particles that were assigned 0 as their Steinhardt parameter it would send the same message. It's my understanding that it's fairly impossible for there to even be a configuration where 0 is the Steinhardt parameter for a particle with neighbors anyway.

vyasr:
Sorry for the long delay, I've been pretty swamped lately. Summarizing the arguments here, I think the two sides to this are 1) mathematically the order parameter is not well-defined for 0 neighbors, and 2) the order parameter is a physical quantity that should have some meaningful interpretation. Looking at this for the first time in months, I'm coming around to the argument for (2): the order parameter _should_ be a physical quantity. However, since this is purely a question of convention and we've had some debate, I'd like to get some additional opinions. @xiyudu you've probably been the heaviest user of this feature throughout freud's history and you wrote the original version, what are your thoughts?

bdice:
1. @vyasr I don’t think this is “purely a question of convention.” If it were, it would be possible to switch between conventions (like converting degrees to radians). NaN values can be converted to 0 by the user in a single line of code. However, it is not possible to guess which 0 values should have been NaN values without additional information about neighbor counts (which may have been computed on the fly and thus would not exist in an accessible object).
2. I would accept a PR that introduces a flag to switch between behavior, leaving NaN as the default and 0 as an opt-in, or even better, a PR that leaves the class as-is and includes an example of the one line of code needed to convert NaNs to 0. I note that the same changes would also need to be applied to Hexatic and SolidLiquid order parameters.
3. If the consensus is that the convention must be changed, it should wait until a major version.
4. Some of my code is dependent on the current behavior, which leads me to suspect others may rely on it as well.

I am also curious to hear @xiyudu’s thoughts.

vyasr:
> 1. @vyasr I don’t think this is “purely a question of convention.” If it were, it would be possible to switch between conventions (like converting degrees to radians). NaN values can be converted to 0 by the user in a single line of code. However, it is not possible to guess which 0 values should have been NaN values without additional information about neighbor counts (which may have been computed on the fly and thus would not exist in an accessible object).

I agree, and that is one of the reason I had reservations about such a change. Maya makes a good point, though, that in practice getting a true 0 is basically impossible. Unless you place your points on precise lattice points and choose `l` and `m` such that the associated Legendre polynomial includes a `sin(0)` or `cos(pi/2)` term you will get nonzero values. Leaving aside the semantics of the word "convention", my point was that whether or not freud should make such decisions is a question of design philosophy. In general we err against making such decisions, but I'm a bit more easily swayed on this particular topic because I don't use Steinhardt OPs, but I know that they are commonly used features, so I don't want to stand in the way of a more useful feature. I'd rather get opinions from people with more practical usage experience than I have. Feel free to tag anyone else who you think would have a useful opinion there. @b-butler might also have thoughts since he helped rewrite it and I think has used it some for descriptors for his information theory project. Also perhaps @j-proc.

> 2. I would accept a PR that introduces a flag to switch between behavior, leaving NaN as the default and 0 as an opt-in, or even better, a PR that leaves the class as-is and includes an example of the one line of code needed to convert NaNs to 0. I note that the same changes would also need to be applied to Hexatic and SolidLiquid order parameters.

I considered both of these approaches as well. I would be completely fine with including the one-liner, I think that's the easiest solution. I'd prefer not to use the flag, one way or another I think we should make a decision and let freud produce consistent output.

> 3. If the consensus is that the convention must be changed, it should wait until a major version.

100% agree, even if we decide to move forward with this change we can't release it until freud 3.0.

> 4. Some of my code is dependent on the current behavior, which leads me to suspect others may rely on it as well.
> 
> I am also curious to hear @xiyudu’s thoughts.

xiyudu:
Sorry to only be able to reply right now!
I have hesitation in converting NaN to 0 as a default convention, as 0 does have physical meanings for Steinhardt order parameters.  It might be a rare case in the Qls, but are quite common in the Wl cases or the SolidLiquid order parameter when you are computing the dot products between two nearest neighbors.  And I would argue that having a dot product of 0 and NaN for lack of neighbors are two very different things.  
I like the idea of including an optional one-liner, where users can use to automatically mask NaN to 0s, and I think it is really on the users side to make that conscious choice, instead of Freud making it for them.

vyasr:
Thanks @xiyudu that is helpful information. If 0 is not unambiguously interpretable as having no neighbors for different Steinhardt modes, then I would not want to lose that information when users could apply the proposed mask. In that case I think our best course of action is adding the masking line to our documentation. If anyone is so inclined, we could add the flag to the class, but I don't see much benefit to passing a flag over masking after the fact.

bdice:
#MaskUpNaNs 😷 👍

vyasr:
@mmartirossyan in light of the discussions we've had on this thread, I'm going to close this PR, but please don't be discouraged! We really appreciate the contribution and it's helped spark some useful debate here. If you have other thoughts on how to improve freud (new features, bug fixes, better docs), please don't hesitate to reach out or make further contributions. Thanks for your work!

Sorry it's taken so long to get things wrapped up. I've opened #672 to address the proposed changes here.

Issue #623, opened 2020-05-26
Adding short examples to the following API: Cluster, Density, Environment, Locality, Order


mstryk:
**Is your feature request related to a problem? Please describe.**
API could use more short examples in the documentation. The examples page of Freud includes too much depth for simple referencing purposes when using Freud.

**Describe the solution you'd like**
Commonly used modules should include 3-4 short examples. These examples should be less than 6 lines and be similar to those already in the box module.

**Additional context**
Trying to make Freud more user friendly for beginners.

klywang:
I think this would be very helpful!

PR #625, opened 2020-06-02
Doc/add examples

mstryk:
Added Short examples under the Density, Cluster, Environment, and Order modules. Added doctest. 
## Description
Added 6 short examples to commonly used modules. These modules are commented and include all assumptions. These are shorter versions of the examples under https://github.com/glotzerlab/ovito-scripts. All examples have been configured to work with doctest  and will only show up if they are up to date with the API (thank you @bdice ). 

## Motivation and Context
I was tasked with helping to make the Freud website to be more user friendly towards beginners. I realized that the longer and more complex examples page was not ideal for fulfilling the simple need of referencing a computation. Thus, I suggested the addition of shortened examples so that a beginner does not have to switch to the examples page to find a couple lines of code. This is meant to be similar to how a student might search up a function on stack exchange just to see how a function is used. 

## How Has This Been Tested?
The examples were run through and passed all styling tests and were made to resemble previous examples under the Box Module.  Additionally, doctest ensures that examples are up to date with the modules in the event of an API update, or else they will simply not appear.
## Types of changes

## Checklist:

### DIFF
--- a/doc/source/reference/credits.rst
+++ b/doc/source/reference/credits.rst
@@ -290,7 +290,6 @@ Mike Henry
 Michael Stryk
 
 * Added short examples into Cluster, Density, Environment, and Order Modules.
-* Updated old API in freud ovito-scripts.
 
 Source code
 -----------
--- a/doc/source/reference/credits.rst
+++ b/doc/source/reference/credits.rst
@@ -287,6 +287,11 @@ Mike Henry
 * Fixed syntax in freud-examples notebooks for v2.0.
 * Updated documentation links
 
+Michael Stryk
+
+* Added short examples into Cluster, Density, Environment, and Order Modules.
+* Updated old API in freud ovito-scripts.
+
 Source code
 -----------
 
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -185,13 +185,14 @@ cdef class ClusterProperties(_Compute):
 
         Example::
 
-            >>> # Assuming box, positions, and nlist have already been defined
+          
             >>> # Compute clusters using box, positions, and nlist data
+            >>> box, points = freud.data.make_random_system(10, 100, seed=0)
             >>> cl = freud.cluster.Cluster()
-            >>> cl.compute(box, positions, neighbors=nlist)
+            >>> cl.compute(box, points, neighbors={'r_max': 1.0}))
             >>> # Compute cluster properties based on identified clusters
             >>> cl_props = freud.cluster.ClusterProperties()
-            >>> cl_props.compute(box, positions, cl.cluster_idx)
+            >>> cl_props.compute(box, points, cl.cluster_idx)
 
         Args:
             system:
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -187,11 +187,9 @@ cdef class ClusterProperties(_Compute):
 
             >>> # Assuming box, positions, and nlist have already been defined
             >>> # Compute clusters using box, positions, and nlist data
-            >>> import freud
-            >>> import numpy as np
             >>> cl = freud.cluster.Cluster()
             >>> cl.compute(box, positions, neighbors=nlist)
-            >>> # Compute Cluster properties (size, position, center, etc)
+            >>> # Compute cluster properties based on identified clusters
             >>> cl_props = freud.cluster.ClusterProperties()
             >>> cl_props.compute(box, positions, cl.cluster_idx)
 
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -184,7 +184,6 @@ cdef class ClusterProperties(_Compute):
         :code:`centers` and :code:`gyrations` attributes.
 
         Example::
-
             >>> # Compute clusters using box, positions, and nlist data
             >>> box, points = freud.data.make_random_system(10, 100, seed=0)
             >>> cl = freud.cluster.Cluster()
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -185,16 +185,15 @@ cdef class ClusterProperties(_Compute):
 
         Example::
             >>> import freud
-
             >>> # Compute clusters using box, positions, and nlist data
-
             >>> box, points = freud.data.make_random_system(10, 100)
-
-            >>> cl = freud.cluster.Cluster().compute((box, points), neighbors={'r_max': 1.0})
-
+            >>> cl = freud.cluster.Cluster()
+            >>> cl.compute((box, points), neighbors={'r_max': 1.0})
+            freud.cluster.Cluster()
             >>> # Compute cluster properties based on identified clusters
-
-            >>> cl_props = freud.cluster.ClusterProperties().compute((box, points), cl.cluster_idx)
+            >>> cl_props = freud.cluster.ClusterProperties()
+            >>> cl_props.compute((box, points), cl.cluster_idx)
+            freud.cluster.ClusterProperties()
 
         Args:
             system:
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -183,6 +183,19 @@ cdef class ClusterProperties(_Compute):
         this method, these properties can be accessed with the
         :code:`centers` and :code:`gyrations` attributes.
 
+        Example::
+
+            >>> #Assuming box, positions, and nlist have already been defined
+            >>> #Compute clusters using box, positions, and nlist data
+            >>> import freud
+            >>> import numpy as np
+            >>> cl = freud.cluster.Cluster()
+            >>> cl.compute(box, positions, neighbors=nlist)
+            >>> # Compute Cluster properties (size, position, center, etc)
+            >>> cl_props = freud.cluster.ClusterProperties()
+            >>> cl_props.compute(box, positions, cl.cluster_idx)
+
+
         Args:
             system:
                 Any object that is a valid argument to
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -185,8 +185,8 @@ cdef class ClusterProperties(_Compute):
 
         Example::
 
-            >>> #Assuming box, positions, and nlist have already been defined
-            >>> #Compute clusters using box, positions, and nlist data
+            >>> # Assuming box, positions, and nlist have already been defined
+            >>> # Compute clusters using box, positions, and nlist data
             >>> import freud
             >>> import numpy as np
             >>> cl = freud.cluster.Cluster()
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -184,13 +184,17 @@ cdef class ClusterProperties(_Compute):
         :code:`centers` and :code:`gyrations` attributes.
 
         Example::
+            >>> import freud
+
             >>> # Compute clusters using box, positions, and nlist data
-            >>> box, points = freud.data.make_random_system(10, 100, seed=0)
-            >>> cl = freud.cluster.Cluster()
-            >>> cl.compute(box, points, neighbors={'r_max': 1.0}))
+
+            >>> box, points = freud.data.make_random_system(10, 100)
+
+            >>> cl = freud.cluster.Cluster().compute((box, points), neighbors={'r_max': 1.0})
+
             >>> # Compute cluster properties based on identified clusters
-            >>> cl_props = freud.cluster.ClusterProperties()
-            >>> cl_props.compute(box, points, cl.cluster_idx)
+
+            >>> cl_props = freud.cluster.ClusterProperties().compute((box, points), cl.cluster_idx)
 
         Args:
             system:
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -195,7 +195,6 @@ cdef class ClusterProperties(_Compute):
             >>> cl_props = freud.cluster.ClusterProperties()
             >>> cl_props.compute(box, positions, cl.cluster_idx)
 
-
         Args:
             system:
                 Any object that is a valid argument to
--- a/freud/cluster.pyx
+++ b/freud/cluster.pyx
@@ -185,7 +185,6 @@ cdef class ClusterProperties(_Compute):
 
         Example::
 
-          
             >>> # Compute clusters using box, positions, and nlist data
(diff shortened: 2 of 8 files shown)


freud/cluster.pyx
```diff
@@ -183,6 +183,18 @@ cdef class ClusterProperties(_Compute):
...
         :code:`centers` and :code:`gyrations` attributes.
 
+        Example::
+
+            >>> # Assuming box, positions, and nlist have already been defined
+            >>> # Compute clusters using box, positions, and nlist data
+            >>> import freud
+            >>> import numpy as np
+            >>> cl = freud.cluster.Cluster()
+            >>> cl.compute(box, positions, neighbors=nlist)
```
klywang, line 193: You may want to include a comment here that explains (briefly) why you need to calculate this first.
```suggestion
            >>> # Determine clusters
            >>> cl = freud.cluster.Cluster()
            >>> cl.compute(box, positions, neighbors=nlist)
```

freud/cluster.pyx
```diff
@@ -183,6 +183,18 @@ cdef class ClusterProperties(_Compute):
         this method, these properties can be accessed with the
         :code:`centers` and :code:`gyrations` attributes.
 
+        Example::
+
+            >>> # Assuming box, positions, and nlist have already been defined
+            >>> # Compute clusters using box, positions, and nlist data
+            >>> import freud
+            >>> import numpy as np
```
klywang: These can be excluded in the short examples
```suggestion

```

freud/cluster.pyx
```diff
@@ -183,6 +183,18 @@ cdef class ClusterProperties(_Compute):
...
+            >>> cl = freud.cluster.Cluster()
+            >>> cl.compute(box, positions, neighbors=nlist)
+            >>> # Compute Cluster properties (size, position, center, etc)
```
klywang:
```suggestion
            >>> # Compute cluster properties based on identified clusters
```
We may want to be a little more descriptive here. Talk about why

freud/cluster.pyx
```diff
@@ -183,6 +183,18 @@ cdef class ClusterProperties(_Compute):
...
+            >>> cl_props = freud.cluster.ClusterProperties()
+            >>> cl_props.compute(box, positions, cl.cluster_idx)
```
klywang, line 196: Use `clp` instead of `cl_props` for consistensy with the **Examples** section.

freud/density.pyx
```diff
@@ -360,6 +360,15 @@ cdef class LocalDensity(_PairCompute):
     def compute(self, system, query_points=None, neighbors=None):
         R"""Calculates the local density for the specified points.
 
+        Example::
+
+            >>> # Assuming box, positions, and system (box, points) have already been defined
+            >>> # Compute Local Density
+            >>> # r_max and diameter can be changed as needed
```
klywang: We can remove this line for the minimal example
```suggestion
            >>> # r_max and diameter can be changed as needed
```

freud/density.pyx
```diff
@@ -360,6 +360,15 @@ cdef class LocalDensity(_PairCompute):
...
+            >>> import freud
```
klywang, line 364: Remove this as well

freud/density.pyx
```diff
@@ -360,6 +360,15 @@ cdef class LocalDensity(_PairCompute):
...
+            >>> ld = freud.density.LocalDensity(r_max=3, diameter=0.05)
+            >>> ld.compute(system)
```
klywang, line 370: Use `density` instead of `ld` here to be consistent with the **Examples** section.

freud/environment.pyx
```diff
@@ -534,6 +534,18 @@ cdef class EnvironmentCluster(_MatchEnv):
...
+
+            >>> # Recommend setting registration and global_search = False
+            >>> # Assuming box, points/positions, system (box, points),
+            >>> # r_max, num_neighbors and threshold have already been defined
+            >>> # Compute clusters of particle with matching environments
+            >>> import freud
+            >>> env_cluster = freud.environment.EnvironmentCluster()
+            >>> env_cluster.compute(system, threshold=threshold,
+            >>> neighbors={'num_neighbors': num_neighbors},
+            >>> registration=registration, global_search=global_search)
```
klywang: Pick realistic values for variables that should be a scalar or a boolean (True/False). You can just copy whatever they used from the **Example** page. Replace `system` with `(box, positions)`.

freud/order.pyx
```diff
@@ -175,10 +175,24 @@ cdef class Nematic(_Compute):
     def compute(self, orientations):
         R"""Calculates the per-particle and global order parameter.
 
+        Example::
+
+            >>> # Assuming user has set director = np.array[x,y,z]
+            >>> # Assuming box, points/positions, and system (box, points)
+            >>> # have already been defined
+            >>> # Recommend visualizing the data with the particle_property below
+            >>> import freud
```
klywang, line 184: Remove this too

freud/order.pyx
```diff
@@ -175,10 +175,24 @@ cdef class Nematic(_Compute):
...
+            >>> from ovito.data import *
+            >>> import numpy as np
```
klywang: Remove these import statements.

freud/order.pyx
```diff
@@ -175,10 +175,24 @@ cdef class Nematic(_Compute):
     def compute(self, orientations):
         R"""Calculates the per-particle and global order parameter.
 
+        Example::
+
+            >>> # Assuming user has set director = np.array[x,y,z]
```
klywang: The synatx for `np.array[x, y, z]` should be `np.array([x, y, z])` -- but judging from the examples for the `Nematic`  order parameter, you can just use a list like `[1, 0, 0]` for the example. 

I think it would be best, though, for this example, to remove this comment and copy the value from the example for the `director` variable below.

freud/order.pyx
```diff
@@ -175,10 +175,24 @@ cdef class Nematic(_Compute):
...
+            >>> # Assuming box, points/positions, and system (box, points)
+            >>> # have already been defined
+            >>> # Recommend visualizing the data with the particle_property below
+            >>> import freud
+            >>> from ovito.data import *
+            >>> import numpy as np
+            >>> order_param = freud.order.Nematic(director)
+            >>> order_param.compute(points.orientations)
```
klywang: `points` should be an `Nx3` array in these examples so they should not have an `orientation` attribute. You may want to assign fake orientations in this example or include this in your comments as a preassigned value.

freud/order.pyx
```diff
@@ -175,10 +175,24 @@ cdef class Nematic(_Compute):
...
+            >>> data.create_user_particle_property(name='NematicOrderParameter',
+            >>> data_type=float, data=order_param.particle_tensor
```
klywang: Remove these. These are for `Ovito`

freud/order.pyx
```diff
@@ -272,6 +286,22 @@ cdef class Hexatic(_PairCompute):
...
 
+        Example::
+
+            >>> # Assuming user has defined k and box, points/positions,
+            >>> # and system(box, points) have already been defined
+            >>> # Compute the K-atic order for the 2 D system
+            >>> # Recommend visualizing data with the particle_property below
+            >>> import freud
+            >>> from ovito.data import *
+            >>> import numpy as np
```
klywang: Remove these import statements as well

freud/order.pyx
```diff
@@ -272,6 +286,22 @@ cdef class Hexatic(_PairCompute):
...
+            >>> order_param = freud.order.Hexatic(k)
+            >>> order_param.compute(system)
+            >>> psi_k = np.copy(order_param.particle_order)
+            >>> psi_k -= np.mean(psi_k)
+            >>> data.create_user_particle_property(name='HexaticOrderParameter',
+            >>> data_type=float, data=np.angle(psi_k, deg=True)/order_param.k)
```
klywang: Remove this

freud/order.pyx
```diff
@@ -272,6 +286,22 @@ cdef class Hexatic(_PairCompute):
...
+            >>> # and system(box, points) have already been defined
+            >>> # Compute the K-atic order for the 2 D system
+            >>> # Recommend visualizing data with the particle_property below
+            >>> import freud
+            >>> from ovito.data import *
+            >>> import numpy as np
+            >>> order_param = freud.order.Hexatic(k)
+            >>> order_param.compute(system)
+            >>> psi_k = np.copy(order_param.particle_order)
+            >>> psi_k -= np.mean(psi_k)
```
klywang: What is the purpose of including these lines in the example?
If we think it would be beneficial to show people how to access the final order parameters, it may be more useful just to print the values, like in these `hoomd` examples [[x](https://hoomd-blue.readthedocs.io/en/stable/module-hoomd-data.html)]:
```python
>>> print(order_param.particle_order)
array([ 0.2119954 -0.5681961j ,  0.18827692-0.4118587j ,
       -0.13218744-0.6692502j , ..., -0.4630672 -0.15254533j,
        0.36158592+0.08465159j,  0.15578252-0.44301566j], dtype=complex64)
```
Otherwise, it may be best just to remove these.

freud/order.pyx
```diff
@@ -543,6 +573,19 @@ cdef class Steinhardt(_PairCompute):
     def compute(self, system, neighbors=None):
         R"""Compute the order parameter.
 
+        Example::
+
+            >>> # Assuming box, points, and system (box, points) are defined
+            >>> # Recommend visualizing data with the particle_property below
+            >>> # r_max and 1 can be changed as desired
```
klywang: You can remove this

freud/order.pyx
```diff
@@ -543,6 +573,19 @@ cdef class Steinhardt(_PairCompute):
...
+            >>> import freud
+            >>> from ovito.data import *
+            >>> import numpy as np
```
klywang: Remove these import statements as well

freud/order.pyx
```diff
@@ -543,6 +573,19 @@ cdef class Steinhardt(_PairCompute):
...
+            >>> ql = freud.order.Steinhardt(l=6)
+            >>> ql.compute(system, {'r_max': 3})
+            >>> ql_property = data.create_user_particle_property(name='Ql',
+            >>> data_type=float, data=ql.particle_order)
```
klywang: Remove anything related to `ovito` here

klywang (changes requested):
Thank you for doing this! Just a few changes, and then we'll be good to go:
(1) Remove import statements from the examples. If you end up using `numpy`, use the full name, rather than `np` so it's explicit. We should avoid using extra packages unless necessary for the example, however.
(2) Remove anything related to `ovito` here.
(3) Try to use realistic numbers are inputs whenever possible. You can assume the user has information like `box` or `points` already, but we want to avoid introducing too many assumed variables. It makes it harder to read. Plus, sometimes the examples are useful because they provide dummy values that give you a good idea of what that parameter should look like.

freud/order.pyx
```diff
@@ -175,10 +175,24 @@ cdef class Nematic(_Compute):
...
+        Example::
+
+            >>> # Assuming user has set director = np.array[x,y,z]
+            >>> # Assuming box, points/positions, and system (box, points)
+            >>> # have already been defined
+            >>> # Recommend visualizing the data with the particle_property below
+            >>> import freud
+            >>> from ovito.data import *
+            >>> import numpy as np
+            >>> order_param = freud.order.Nematic(director)
```
klywang: We may want to pick a more descriptive name for this variable, such as `nematic`? -- or `nop` from the **Examples** for consistency?

freud/order.pyx
```diff
@@ -272,6 +286,22 @@ cdef class Hexatic(_PairCompute):
...
+        Example::
+
+            >>> # Assuming user has defined k and box, points/positions,
+            >>> # and system(box, points) have already been defined
+            >>> # Compute the K-atic order for the 2 D system
+            >>> # Recommend visualizing data with the particle_property below
+            >>> import freud
+            >>> from ovito.data import *
+            >>> import numpy as np
+            >>> order_param = freud.order.Hexatic(k)
```
klywang: We may want to pick a more descriptive name here as well. Something like `hex_order` if we choose `k=6` or `k_order` for more generality.

bdice:
1. Can we try to adopt doctest or a similar tool to ensure that these examples work as a part of freud's continuous integration? Otherwise API changes can break the examples and that's really bad for users. https://docs.python.org/3/library/doctest.html

2. I would like the examples to be runnable if possible. I would avoid assuming that `system` already exists. Use the `freud.data` module to generate systems as needed (it's as short as one line to generate a random system).

bdice:
@mstryk @klywang Really nice work so far. I wasn't sure how to use `doctest` myself so I spent a half hour learning it -- I didn't want to encourage you to pursue something that was unnecessarily difficult. I pushed some commits that enable `doctest` as a part of continuous integration, and I fixed the currently failing doctests. I think it works pretty well. I hope these updates enable you to write more docstrings, now that you have working and validated examples! 👍

mstryk:
@bdice Thank you for adding adding doctests to the examples! I had never considered that such a thing could exist or what happens when example are outdated by the modules. I will open up the pull request from a draft to a full request. I will try to integrate doctest into any further examples I may add for other modules using your examples.

klywang (approved):
@bdice Thank you for implementing `doctest` in CI and fixing the examples! @mstryk Everything looks good now.

klywang:
Resolves issue #623

doc/source/reference/credits.rst
```diff
@@ -287,6 +287,11 @@ Mike Henry
 * Fixed syntax in freud-examples notebooks for v2.0.
 * Updated documentation links
 
+Michael Stryk
+
+* Added short examples into Cluster, Density, Environment, and Order Modules.
+* Updated old API in freud ovito-scripts.
```
vyasr: I'd remove this line (and maybe add a credits file if one doesn't exists to the ovito-scripts repo so we can keep track of who did what there).

vyasr:
This is good stuff, thanks @mstryk! @bdice thanks for getting doctest working, I think we've discussed before but nice to see it in action. I think this PR is ready for a final pass from you if you'd like, otherwise we can call it ready.

PR #628, opened 2020-06-10
Add cropping feature

### LINKED ISSUE #583: Add cropping feature
There is currently no easy way to crop simulation boxes in freud. Manual cropping can result in user error (see #582) if done incorrectly.

A feature which allows users to crop existing systems. 
`points_inside = points[box.is_inside(points)]`
As suggested by @bdice

klywang:
## Description
Added a new function to crop points from larger boxes to smaller boxes. Returns a mask (`True` for points in a box and `False` for features points of a box).

## Motivation and Context
This can be helpful in situations where a user intends to analyse a portion of data from a simulation or wishes to create a new system from a portion of existing data. A mask, rather than positions, are returned in case other per particle values need to be cropped.
Resolves: #583 

## How Has This Been Tested?
A test with a triclinic box was constructed to test analysis. Points were placed in clusters just outside the corner of the box to ensure points which may be considered edge cases were addressed.

## Types of changes

## Checklist:


cpp/box/Box.h
```diff
@@ -435,6 +435,46 @@ class Box
         );
     }
 
+    //! TODO: Title
+    /*! \param
+    *   \param
+    *   TODO: Fill this out
+    */
+    inline unsigned int isInBox(const vec3<float>& point) const
```
bdice: **tl;dr:** I think you can be more consistent with floating-point-behavior near the boundaries and reduce this function to 1-2 lines by checking if the point's periodic image is equal to `(0, 0, 0)`.

**long form:** In general, freud should implement explicit triclinic box math operations as few times as possible, because otherwise we can have combinations of operations that are mutually inconsistent about how points near the boundary are treated with floating precision issues. For example, you could have a point very close to the upper x-limit of the box that this function thinks is in the box, but its periodic image could be `(1, 0, 0)`. Thus we should try to express box operations like this one in terms of other, more basic box operations (e.g. converting between fractional/absolute coordinates, getting a periodic image).

klywang: I moved the per particle calculations for `getImage` to a seperate function and used it to determine particle membership in the box for consistency as suggested.

ChangeLog.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to
 * Custom NeighborLists can be created from a set of points using `from_points`. Distances will be calculated automatically.
 * The Box class has methods `compute_distances` and `compute_all_distances` to calculate distances between arrays of points and query points.
 * Hexatic can now compute 2D Minkowski Structure Metrics, using `weighted=True` along with a Voronoi NeighborList.
+* The Box class has a method `crop` to determine particle membership in a box.
```
bdice: What do you think about renaming this method to "contains"? I dislike "crop" because it implies to me that the output is an array of positions (given an array of input positions), but I agree with your reasoning for returning a mask of indices instead.

cpp/box/Box.h
```diff
@@ -435,6 +445,31 @@ class Box
...
+        \param cropped_mask Mask of points inside the box.
+    */
+    void crop(const vec3<float>* points, const unsigned int n_points,
+        unsigned char* cropped_mask) const
+    {
+        util::forLoopWrapper(0, n_points, [&](size_t begin, size_t end) {
+            for (size_t i = 0; i < n_points; ++i)
+            {
+                vec3<int> image = getImage(points[i]);
+                unsigned char is_in_box = 0;
```
bdice: Unused variable?

cpp/box/Box.h
```diff
@@ -435,6 +445,31 @@ class Box
...
+        \param n_points The number of points.
+        \param cropped_mask Mask of points inside the box.
+    */
+    void crop(const vec3<float>* points, const unsigned int n_points,
+        unsigned char* cropped_mask) const
+    {
+        util::forLoopWrapper(0, n_points, [&](size_t begin, size_t end) {
+            for (size_t i = 0; i < n_points; ++i)
+            {
+                vec3<int> image = getImage(points[i]);
```
```suggestion
                const vec3<int> image = getImage(points[i]);
```

cpp/box/Box.h
```diff
@@ -435,6 +445,31 @@ class Box
...
+    //! Get mask of points that fit inside the box.
+    /*! \param points Point positions.
+        \param n_points The number of points.
+        \param cropped_mask Mask of points inside the box.
+    */
+    void crop(const vec3<float>* points, const unsigned int n_points,
+        unsigned char* cropped_mask) const
+    {
+        util::forLoopWrapper(0, n_points, [&](size_t begin, size_t end) {
+            for (size_t i = 0; i < n_points; ++i)
```
bdice: For bonus points you could write this entire `for` loop as a single `std::transform` with a lambda function.

bdice: Add `#include <algorithm>` at the top.
```cpp
std::transform(&points[begin], &points[end], &cropped_mask[begin],
    [this](vec3<float> point) -> unsigned char {
        return getImage(point) == vec3<int>(0, 0, 0);
    });
```
Reference: https://en.cppreference.com/w/cpp/algorithm/transform

bdice: TBB alert. ⚠️ Need to be sure to use `begin` and `end` (not `0` to `n_points`) or else you'll have every parallel thread compute the entire range instead of just their chunk. 
```suggestion
            for (size_t i = begin; i < end; ++i)
```

cpp/box/Box.h
```diff
@@ -435,6 +445,31 @@ class Box
         );
     }
 
+    //! Get mask of points that fit inside the box.
+    /*! \param points Point positions.
+        \param n_points The number of points.
+        \param cropped_mask Mask of points inside the box.
+    */
+    void crop(const vec3<float>* points, const unsigned int n_points,
+        unsigned char* cropped_mask) const
```
bdice: Is there a reason we prefer `unsigned char*` over `bool*`?

doc/source/reference/credits.rst
```diff
@@ -259,7 +259,7 @@ Matthew Palathingal
 Kelly Wang
 
 * Enabled NeighborList indexing.
-* Added methods ``compute_distances`` and ``compute_all_distances`` to Box.
+* Added methods ``compute_distances``, ``compute_all_distances``, and ``crop`` to Box.
```
bdice: Give yourself a new line since this is a separate PR. 😄
```suggestion
* Added methods ``compute_distances`` and ``compute_all_distances`` to Box.
* Added ``crop`` feature to Box.
```

bdice (commented):
@klywang I know this is still a draft but I had an idea that you might like, using `std::transform`. 😄

ChangeLog.md
klywang: That makes sense. I'll change the function name. I do want the word "crop" to appear somewhere in the docstring so users can find this function by searching the page. I'll add an example in the docs to make it clear.

cpp/box/Box.h
klywang: Extra support is required for boolean arrays. I wasn't sure if it was worth including an additional import line (`from libcpp cimport bool`) to create a mask .
I tried to use `bint`, but I couldn't get those results to appear correctly in numpy arrays so I settled on the smallest valid data type in both cython and C++ to represent 0/1 and converted it to `True`/`False` with python.

cpp/box/Box.h
klywang: If I understand correctly, would this code using `std::transform` make the operation occur over a single thread rather than in parallel?

bdice: With TBB, the full range of indices is split into sub-ranges for each thread. The outer part here:
`util::forLoopWrapper(0, n_points, [&](size_t begin, size_t end) {`
is defining that you want to go from `0` to `n_points`. Then, TBB chunks that range up into sub-ranges and sends them to each thread, which will execute from `begin` to `end`. So if you had two threads, thread 1 might go from `begin=0` to `end=n_points/2` and thread 2 might go from `begin=n_points/2` to `end=n_points`.

The "TBB Alert" I included as a separate comment was marked as resolved but this appears to still have the bug. Hopefully the explanation above helps.
> Need to be sure to use begin and end (not 0 to n_points) or else you'll have every parallel thread compute the entire range instead of just their chunk.

The `transform` I am proposing is an exact replacement for the thread-local `for` loop from `begin` to `end`.

Replace this:

```cpp
for (size_t i = 0; i < n_points; ++i)  // This uses the wrong indices! Should be `i = begin; i < end`
{
	const vec3<int> image = getImage(points[i]);
	unsigned char is_in_box = 0;
	if (image == vec3<int>(0, 0, 0))
	{
		cropped_mask[i] = 1;
	}
	else
	{
		cropped_mask[i] = 0;
	}
}
```

with this:

```cpp
std::transform(&points[begin], &points[end], &cropped_mask[begin],
    [this](vec3<float> point) -> unsigned char {
        return getImage(point) == vec3<int>(0, 0, 0);
    });
```

klywang: Ah that makes sense. 

Also I marked it as resolved because I fixed it locally but I hadn't pushed the changes yet! I wanted to see what your response was here before I pushed the changes.

freud/box.pyx
```diff
@@ -494,6 +495,37 @@ cdef class Box:
 
         return np.asarray(distances)
 
+    def crop(self, points):
+        R"""Returns mask corresponding to particle membership in a box.
```
tommy-waltmann: Is calling this a mask common terminology? I was initially confused as to what this calculation was doing until I read through the implementation.

freud/box.pyx
```diff
@@ -494,6 +495,37 @@ cdef class Box:
...
+
+        Args:
+            points (:math:`\left(N, 3\right)` :class:`numpy.ndarray`):
+                Array of points.
+
+        Returns:
+            :math:`\left(N, \right)` :class:`numpy.ndarray`:
+                Array of booleans, where `True` corresponds to points within the box,
```
tommy-waltmann, line 525: Using the terminology of "inside" vs "outside" the box is a little ambiguous to me, because a simulation happens entirely within a box, and I think this would be confusing to people reading the documentation. Can we instead say something along the lines of "particles which are/aren't taking advantage of the box's periodic boundary conditions"?

ChangeLog.md
tommy-waltmann: For internal consistency, make sure the cpp method in the box class gets renamed as well.

tommy-waltmann (commented):
This is in good shape, just a couple of suggestions that might make the documentation clearer.

freud/box.pyx
bdice: @tommy-waltmann Yes, this is a common term. The suggestion below might help clarify. https://numpy.org/doc/stable/reference/maskedarray.generic.html
```suggestion
        R"""Returns boolean array (mask) corresponding to particle membership in a box.
```

freud/box.pyx
bdice: I think @klywang's chosen language (inside/outside) is unambiguous here. The box has finite volume and defined boundaries, even though periodic boundary conditions are intended to behave as though the box is infinite. @klywang @tommy-waltmann I have some background info below on topics related to this that may be of interest to you and future developers.

freud assumes periodic boundary conditions in virtually all methods. This is an assumption that we often want to relax, in cases like @klywang has here, creating a virtual box that is a subset of a larger box and can be used to "crop." We have long-standing issues (#209, #573) from users in our group and external users where explicit support for non-periodic boxes would be helpful. We have flags for tracking box periodicity that are almost never used internally, and this is not the desired behavior.

To Tommy's question: whether a set of input points "obeys periodic boundary conditions" is not always well-defined, and it depends largely on what users are trying to achieve. Consider cases like polymers where polymer beads may be wrapped such that the molecule is contiguous and its center of mass is inside the periodic box, even though some of its components are outside the box. In such a case, does the molecule "obey" periodic boundary conditions?

Another subtlety here that @klywang got right, but I wanted to echo the motivation -- we try to use "points" instead of "particles" in all docstrings (wherever appropriate) because many tools in the freud library are fairly generic and have some general applicability outside the space of nanoparticle self-assembly. Particles could be atoms (much smaller scale), but they could also be points in a geometry sense, or even data points in a 2D/3D vector space that obeys periodic boundary conditions (I have used freud for such a thing). According to a post on our forum from a while back, we have even had users leverage the library for analyzing crowd dynamics.

cpp/box/Box.h
```diff
@@ -271,20 +272,30 @@ class Box
         });
     }
 
+    //! Get periodic image of a vector.
+    /*! \param v The vector to check.
+    */
+    inline vec3<int> getImage(const vec3<float>& v) const
```
vyasr:
```suggestion
    inline void getImage(const vec3<float>& v, vec3<int> &image) const
```

cpp/box/Box.h
```diff
@@ -271,20 +272,30 @@ class Box
...
+    {
+            vec3<float> f = makeFractional(v) - vec3<float>(0.5, 0.5, 0.5);
+            vec3<int> image(0, 0, 0);
+            image.x = (int) ((f.x >= 0.0f) ? f.x + 0.5f : f.x - 0.5f);
```
vyasr:
```suggestion
            image.x = (int) ((f.x >= 0.0f) ? f.x + 0.5f : f.x - 0.5f);
```

cpp/box/Box.h
```diff
@@ -271,20 +272,30 @@ class Box
...
+            image.y = (int) ((f.y >= 0.0f) ? f.y + 0.5f : f.y - 0.5f);
+            image.z = (int) ((f.z >= 0.0f) ? f.z + 0.5f : f.z - 0.5f);
+            return image;
+    }
```
vyasr:
```suggestion
    }
```

cpp/box/Box.h
```diff
@@ -271,20 +272,30 @@ class Box
...
+    void getImages(vec3<float>* vecs, unsigned int Nvecs, vec3<int>* res) const
     {
         util::forLoopWrapper(0, Nvecs, [=](size_t begin, size_t end) {
             for (size_t i = begin; i < end; ++i)
             {
-                vec3<float> f = makeFractional(vecs[i]) - vec3<float>(0.5, 0.5, 0.5);
-                res[i].x = (int) ((f.x >= 0.0f) ? f.x + 0.5f : f.x - 0.5f);
-                res[i].y = (int) ((f.y >= 0.0f) ? f.y + 0.5f : f.y - 0.5f);
-                res[i].z = (int) ((f.z >= 0.0f) ? f.z + 0.5f : f.z - 0.5f);
+                res[i] = getImage(vecs[i]);
```
vyasr:
```suggestion
                getImage(vecs[i], res[i]);
```
The other suggestions above are also made to enable this. The reason I'm suggesting it is because with this many levels of indirection (the array being allocated in Cython-generated C, then returning the image vector from the internal C++ function) I don't trust compilers to optimize out the copies.

cpp/box/Box.h
```diff
@@ -435,6 +446,25 @@ class Box
...
+        \param cropped_mask Mask of points inside the box.
+    */
+    void crop(const vec3<float>* points, const unsigned int n_points,
+        bool* cropped_mask) const
+    {
+        util::forLoopWrapper(0, n_points, [&](size_t begin, size_t end) {
+            for (size_t i = begin; i < n_points; ++i)
+            {
+                std::transform(&points[begin], &points[end], &cropped_mask[begin], 
+                [this](vec3<float> point) -> bool {
```
vyasr:
```suggestion
                [this](const vec3<float> &point) -> bool {
```

ChangeLog.md
vyasr: @klywang looks like this hasn't actually been addressed yet, and I agree with @bdice that I find the current naming confusing as well. I expect `crop` to actually return the reduced set of positions given a mask.

freud/box.pyx
```diff
@@ -494,6 +495,37 @@ cdef class Box:
...
+        """  # noqa: E501
+        points = freud.util._convert_array(
+            np.atleast_2d(points), shape=(None, 3))
+
+        cdef:
+            const float[:, ::1] l_points = points
+            size_t n_all_points = points.shape[0]
+
+        # cropped_mask = freud.util._convert_array(
+        #    np.ones(n_all_points), dtype=np.uint8)
```
vyasr:
```suggestion

```

freud/box.pyx
klywang: @bdice In regards to your "points" vs "particles" discussion -- is there also some ambiguity associated with "particles" when working with polygons/polyhedra? My understanding is that some analysis uses center of mass, while others use particle vertices for analysis.

vyasr: All of this is good background, but I don't think it necessarily addresses Tommy's core concern that users accustomed to periodic boundaries may be confused by this. The core problem is that freud's convention for what is "inside" is documented in the tutorials I wrote [here](https://freud.readthedocs.io/en/latest/gettingstarted/examples/module_intros/box.Box.html?highlight=origin#Using-boxes) and [here](https://freud.readthedocs.io/en/latest/gettingstarted/tutorial/periodic.html?highlight=origin#periodic-boundary-conditions), but not in the Box class itself. The latter link is referenced in the Box docstring, but you have to follow the link and it's still kind of hidden in a note at the bottom.

I suggest we move that note to the docstring for the Box class so that we clearly indicate the convention that boxes go from -L/2 to L/2, then reference that here (something like "where inside is defined by the conventions defined by :class:`~.Box`, ignoring periodicity"). Then we have something unambiguous that we can point to when users are confused as to why (0.9, 0, 0) is not inside a box of size (1, 1, 1).

vyasr (commented):
Almost ready!

cpp/box/Box.h
klywang: It looks like `std::transform` requires an output. For a binary operations, it should take the form:
```cpp
transform(Iterator inputBegin1, Iterator inputEnd1, 
         Iterator inputBegin2, Iterator OutputBegin, 
         binary_operation) 
```
I'm not sure what to use as an output iterator when the function is void. I saw something about `void` pointers but I'm not sure that's relevant to this problem. There are some iterators like `std::for_each` but that only take unary operators so I'm not sure they're an appropriate replacement for `std::transform` in this case. Are you aware of any alternate solutions?

vyasr: I didn't change the output of the function, I just suggested passing the input by reference to avoid making a copy of it. Also I'm not sure what you mean about the binary operator, this is a unary operator (in terms of number of iterators, see the first signature on [this page I assume you are referring to](https://en.cppreference.com/w/cpp/algorithm/transform)) since you're just comparing everything to 0.

klywang: Actually... Just ignore my comment. I was just confused.

cpp/box/Box.h
tommy-waltmann: This is more of a cpp question, but if `getImage` returned `vec3<int>&`, that would be another way to avoid the copies right? Would the compiler would be more or less likely to realize it can inline `getImage` if it was done that way?

freud/box.pyx
bdice: @klywang points vs. particles: I think most users put the center of mass at (0, 0, 0) in the particle frame. We don't have any methods in freud that use polytope vertices.

klywang:
This should be ready for review now.

doc/source/reference/credits.rst
```diff
@@ -263,7 +263,8 @@ Matthew Palathingal
 Kelly Wang
 
 * Enabled NeighborList indexing.
-* Added methods ``compute_distances`` and ``compute_all_distances`` to Box.
+* Added methods ``compute_distances``, ``compute_all_distances`` to Box.
```
tommy-waltmann: If I'm not mistaken, using "and" instead of a comma is correct grammar. Why was this change made?

tommy-waltmann (commented):
This PR looks to be almost ready. I just had one thing to ask about, and wanted to remind about @vyasr  comment clarifying inside vs. outside in the docs for the Box class. Also, you may want to mark resolved conversations on comments which you feel you have addressed, it makes review much easier.

doc/source/reference/credits.rst
klywang: Oh that was a typo. My bad.

klywang:
>  wanted to remind about @vyasr comment clarifying inside vs. outside in the docs for the Box class.

So from @vyasr's comment:
>I suggest we move that note to the docstring for the Box class so that we clearly indicate the convention that boxes go from -L/2 to L/2, then reference that here (something like "where inside is defined by the conventions defined by :class:~.Box, ignoring periodicity"). Then we have something unambiguous that we can point to when users are confused as to why (0.9, 0, 0) is not inside a box of size (1, 1, 1).

I linked the docs where box conventions are discussed here:
https://github.com/glotzerlab/freud/blob/132ee36f4e6d1bf7e852e837d215817121bead68/freud/box.pyx#L501-L505

I think that should be fine, but I can be more explicit if needed.

vyasr:
> > wanted to remind about @vyasr comment clarifying inside vs. outside in the docs for the Box class.
> 
> So from @vyasr's comment:
> 
> > I suggest we move that note to the docstring for the Box class so that we clearly indicate the convention that boxes go from -L/2 to L/2, then reference that here (something like "where inside is defined by the conventions defined by :class:~.Box, ignoring periodicity"). Then we have something unambiguous that we can point to when users are confused as to why (0.9, 0, 0) is not inside a box of size (1, 1, 1).
> 
> I linked the docs where box conventions are discussed here:
> https://github.com/glotzerlab/freud/blob/132ee36f4e6d1bf7e852e837d215817121bead68/freud/box.pyx#L501-L505
> 
> I think that should be fine, but I can be more explicit if needed.

I think that's helpful, but I don't think it goes far enough. I think the basic conventions for freud boxes should be directly embedded in the class docstring. As we've seen with signac, people often don't really go searching for information if it's not in the most obvious place. This problem is not just about this PR and the `contains` method, that's just what's making us think about the issue. I think the primary source for this information needs to be moved so that it's in the first place where people will look, which is the box class.

klywang:
@tommy-waltmann @vyasr 

Would something like this be better?
https://github.com/glotzerlab/freud/blob/6a030627f0f8b3a24fa1e667cd809ace67d3d4fa/freud/box.pyx#L498-L508

tommy-waltmann (approved):

vyasr:
> @tommy-waltmann @vyasr
> 
> Would something like this be better?
> https://github.com/glotzerlab/freud/blob/6a030627f0f8b3a24fa1e667cd809ace67d3d4fa/freud/box.pyx#L498-L508

The extra info is helpful, but what I was requesting is to add this information to the docstring for the box class itself. The conventions for a box (centered at the origin, with the extent in each dimension described by the half-closed interval `[-L/2, L/2)`) should be immediately available when you do `help(freud.Box)`. Right now that docstring links to the information on PBCs and the box on RTD (although the bounds are actually a little bit hidden there too), but I think we need this to be immediately visible.

I propose the following changes, then I'll be happy :) :
1. Add a sentence or two the box docstring that explains that by a convention a freud Box is centered at the origin and the extent in each dimension is described by the half-closed interval `[-L/2, L/2)`. This sentence can go before the sentence about "For more information, see ...".
2. On the documentation page that both `contains` and `Box` are linking to (https://freud.readthedocs.io/en/stable/gettingstarted/tutorial/periodic.html), move the note at the bottom (which describes the bounds of the box) up so that it's right next to the box matrix on that page (right after the line that says "where each column is one of the box vectors").
3. In the docstring you've written for `contains`, remove this part of the first sentence: ", where the origin, ``[0, 0, 0]``, is defined as the center of the box." Once you do (1) and (2), this information will be in the docstring for the Box class, which is where people should look for information on the conventions defining a box.

freud/box.pyx
```diff
@@ -29,8 +29,10 @@ np.import_array()
 cdef class Box:
     R"""The freud Box class for simulation boxes.
 
-    This class defines an arbitrary triclinic geometry within which all points
-    are confined. For more information, see the `documentation
+    This class defines an arbitrary triclinic geometry within which all points are confined.
+    By convention, the freud Box is centered at the origin (``[0, 0, 0]``),
+    with the extent in each dimension described by the half-closed interval ``[-L/2, L/2]``.
```
vyasr:
```suggestion
    with the extent in each dimension described by the half-open interval ``[-L/2, L/2)``.
```

vyasr (approved):
We're done 🎉

PR #634, opened 2020-07-27
Delete extra nlist. Resolves #626.

### LINKED ISSUE #626: Memory leak in neighborlists from preprocess_arguments
**Describe the bug**
As reported in [this mailing list post](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/freud-users/q0LP_J_veLI), there appears to be a memory leak when `compute` is called for `SolidLiquid` without an explicit neighbor list, but instead with a set of query arguments. This issue is most likely arising due to some undesirable references floating around after creation in `_PairCompute._preprocess_arguments`, and merits further investigation.

**To Reproduce**
As in the mentioned post:

```
solidCalc = freud.order.SolidLiquid(normalize_q=True, l=6, q_threshold=0.7, solid_threshold=7)
nbDict = dict(mode='ball',r_max=1.4)
def orderParameter(snapshot):
    solidCalc.compute(snapshot, neighbors=nbDict)
    return solidCalc.largest_cluster_size
```

**Error output**
Memory usage grows each time the function is called.

vyasr:
## Description

`makeDefaultNList` currently generates a NeighborList object on the heap, then makes a local copy on the stack to avoid needing to worry about managing the memory of that object. However, the original heap object is never deleted. This PR deletes it.

## Motivation and Context

Resolves: #626 

## Types of changes

## Checklist:


cpp/locality/NeighborComputeFunctional.cc
```diff
@@ -15,8 +15,14 @@ NeighborList makeDefaultNlist(const NeighborQuery* nq, const NeighborList* nlist
         auto nqiter(nq->query(query_points, num_query_points, qargs));
         nlist = nqiter->toNeighborList();
     }
+    // Ideally we wouldn't allocate a new NeighborList at all, but we don't want to force
+    // calling code to manage the memory of a raw pointer, so we prefer to return by value
+    // here, and the toNeighborList function has to return a raw pointer to work with the
+    // Cython interface of freud. Therefore, for now we're sticking with this suboptimal
+    // solution until we can refactor these internalsa little more cleanly.
```
bdice:
```suggestion
    // solution until we can refactor these internals a little more cleanly.
```

bdice: @vyasr What is the "until we can refactor" here? Is it a matter of developer time/effort or some other technical constraint? What would need to be done?

vyasr:
@tommy-waltmann you can see the error I referred to in the failing tests. My guess is that something in the NeighborList doesn't actually get copied by the copy constructor, but I looked at this for a grand total of 10 minutes so don't take my word for it. I probably won't look at this again until after my data meeting, so feel free to modify/push changes here if you figure out what's wrong.

cpp/locality/NeighborComputeFunctional.cc
```diff
@@ -15,8 +15,14 @@ NeighborList makeDefaultNlist(const NeighborQuery* nq, const NeighborList* nlist
...
     locality::NeighborList new_nlist = NeighborList(*nlist);
     new_nlist.validate(num_query_points, nq->getNPoints());
+    delete nlist;
```
bdice: This is only safe if the `nlist` is on the heap, right? Is that guaranteed to be true for all cases where the user is passing in `NeighborList* nlist`? This seems like it would be unsafe if someone did the following, where the `NeighborList*` points to a `NeighborList` on the stack:
```c++
auto nq = ...;
auto my_nlist = NeighborList(...);
freud::locality::makeDefaultNlist(nq, &my_nlist, ...);
```
Examples of this behavior: https://www.geeksforgeeks.org/g-fact-30/

bdice (changes requested):

vyasr:
@bdice sorry I marked this PR as a draft after I created it, maybe you saw it before I did that and reviewed. @tommy-waltmann started looking into the problem so I just pushed my 5 min fix from when this issue was originally reported. I know that it doesn't work, he was going to investigate why. No need for review at this time.

vyasr:
OK well when I looked at the questions in the review I spotted the problem, so I think this should work now... sorry for the runaround. This is no longer a draft lol.

cpp/locality/NeighborComputeFunctional.cc
vyasr: You're right that this usage would cause problems with that implementation. In practice that isn't how the function gets used, though. However, this comment made me realize the core problem, which is that we should *never* be deleting a user provided nlist anyway. We only want to delete the nlist we make ourselves here. That should resolve the problems.

tommy-waltmann:
This fixes the case where the nlist argument to `makeDefaultNList` is `nullptr`, but breaks the case where it is a real nlist managed by a cython object. In that case, deleting the nlist in `makeDefaultNList` creates a double free when the cython object later tries to clean it up. I will push a commit soon which fixes this in both cases.

cpp/locality/NeighborComputeFunctional.cc
vyasr: The technical constraint is that `toNeighborList` needs to return a raw pointer for Cython, while `makeDefaultNlist` promises a `NeighborList`,  not a pointer. The latter choice simplifies code in all the compute classes that use this function since they can just have a NeighborList object and not think about deleting this pointer, otherwise they'd have to delete it. We can't use a shared pointer directly without making a copy in the case where the user provides a `NeighborList`, which is counterintuitive since that's supposed to be the efficient case.

vyasr:
@tommy-waltmann the last commit I pushed should fix that.

cpp/locality/NeighborComputeFunctional.cc
tommy-waltmann: We could overload the `toNeighborList` function with an implementation that returns a `NeighborList`, not a pointer to `NeighborList`. It would avoid an extra copy and an awkward delete in `makeDefaultNList`, but it would add an extra copy when `toNeighborList` returns, and create a lot of almost duplicate code compared to the old `toNeighborList`, so I'm not sure if its really beneficial or not.

tommy-waltmann (approved):

cpp/locality/NeighborComputeFunctional.cc
vyasr: You would need to define a separate function for it. Changing the return value alone is not sufficient for the compiler to differentiate two function signatures.

You could reasonably write an internal function (say `populateNeighborList`) that contains almost all the logic of `toNeighborList` that accepts a pointer to a `NeighborList` and simply writes the data into that `NeighborList`. The existing `toNeighborList` function would make a `new NeighborList()` and pass that directly through to `populateNeighborList`. You could then write a new function that creates a `nl = NeighborList()` on the stack, then calls `populateNeighborList(*nl)` to update this object directly, then return it by value. However, in this case the new function would still be returning by value, so you'd have to look at the precise rules for copy elision in C++ to verify that you can depend on compilers to optimize out the copy operation; otherwise you'd be back where you started, except that the copy operation would be implicit on the function return rather than explicit inside `makeDefaultNlist`.

vyasr:
@bdice this PR should be ready to merge when you're happy with it.

bdice (approved):
This solves the problem. Eventually I think this would be a good candidate for an API change, if we could work out the issues with Cython around smart pointers and memory management.

PR #647, opened 2020-08-01
Update CI

bdice:
Updates flake8 configuration and gets Windows dependencies from PyPI instead of conda-forge. I will merge this when CI passes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,8 +55,9 @@ references:
         conda update -q conda certifi
         conda install python pip tbb tbb-devel
         # Install from PyPI instead of conda-forge for consistency with Linux.
-        # For some reason, the SSL certificates used by pip aren't correct and must be manually set to use certifi's certificate.
-        python${PYVER} -m pip install --cert $(python${PYVER} -m certifi) --progress-bar off -U -r requirements-testing.txt
+        # For some reason, the SSL certificates used by pip aren't correct, even with certifi.
+        curl -O https://curl.haxx.se/ca/cacert.pem
+        python${PYVER} -m pip install --cert cacert.pem --progress-bar off -U -r requirements-testing.txt
 
   get_style_requirements: &get_style_requirements
     run:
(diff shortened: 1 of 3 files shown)


bdice:
I finally figured out how to solve the issue with certificates. This is the sequence of events (and what I learned):
1. pip would try to install all packages in `requirements-testing.txt`. One of these packages is MDAnalysis.
2. MDAnalysis doesn't have wheel builds on PyPI, so pip fetched the source `.tar.gz` and started to build it.
3. While fetching dependencies for MDAnalysis, pip would attempt to download NumPy (note that this is building with PEP 517 enabled).
4. For some unknown reason, the SSL certificates can't be found while fetching NumPy! This triggers an SSL error similar to this one: https://stackoverflow.com/questions/49324802/pip-always-fails-ssl-verification
  - This answer in particular was helpful because it gave a lot of details: https://stackoverflow.com/a/49991986
5. I debugged the CI over SSH to discover that it didn't have `cacert.pem` in the location I expected (`C:\tools\miniconda3\lib\site-packages\pip\_vendor\requests` or something like that). Apparently pip [can be shipped debundled from its certificates](https://github.com/pypa/pip/blob/master/src/pip/_vendor/README.rst).
6. I searched the pip source code for `cacert` and discovered that calling `python -m certifi` will output a location to `certifi`'s copy of `cacert.pem`.
7. I tried every version of `cacert.pem` that I could find locally, installing with:
```bash
python${PYVER} -m pip install --cert /path/to/cacert.pem --progress-bar off -U -r requirements-testing.txt

# with each of these cert paths:
$(python -m certifi)
/c/tools/miniconda3/Lib/site-packages/certifi/cacert.pem
/c/tools/miniconda3/Lib/site-packages/pip/_vendor/certifi/cacert.pem
/c/tools/miniconda3/Library/ssl/cacert.pem
```
8. I tried downloading cURL's copy of `cacert.pem` from https://curl.haxx.se/docs/caextract.html
9. Installation appears to work with that copy of `cacert.pem`.
10. CI passes. 🎉

PR #654, opened 2020-09-09
Unpin ReadTheDocs conda environment to prevent broken package failures

bdice:
This PR un-pins our conda environment used on ReadTheDocs. We have had between 5 and 7 failures on ReadTheDocs caused by pinned versions no longer existing (packages are marked as broken and removed).<sup>1</sup> **The purpose of this PR is to reduce maintenance burdens.** I think these problems might have been avoided if we just unpinned the conda environment and let the conda solver pull the latest versions of all dependencies. I would not expect any other problems to be caused by this change, but we can revert to a fully-pinned environment if I'm wrong and this somehow increases the maintenance cost.

<sup>1</sup> Looking at the history of our environment yml file, `jupyter_client`, `nbconvert`, `jupyter_sphinx`, `ld_impl_linux-64`, and `libgfortran` have all had pinned versions that were marked as broken and removed, and potentially two other times that I updated packages (unclear if it was to fix version problems or for some other reason)


bdice:
I accidentally pushed to `master` instead of this branch. The commit I meant to make on this branch is 3f44951a7eb8d6670664fc1dac22c451f17278a1.

PR #657, opened 2020-09-14
Always normalize Hexatic order parameters by the number of neighbors.

### LINKED ISSUE #655: Hexatic normalization in docs doesn't match implementation
@Plastikschuessel noted that the documentation for the Hexatic order parameter doesn't match the implementation. The [docs say](https://freud.readthedocs.io/en/stable/modules/order.html#freud.order.Hexatic) that the formula normalizes by `1/n`, but the implemented code normalizes by `1/k`. In many systems (particularly hexagonal and square crystals), `k=n`, but in some cases (e.g. 2D quasicrystals), the order is 12-fold even though any given particle has fewer than 12 neighbors.

https://github.com/glotzerlab/freud/blob/3f44951a7eb8d6670664fc1dac22c451f17278a1/freud/order.pyx#L250-L251

https://github.com/glotzerlab/freud/blob/3f44951a7eb8d6670664fc1dac22c451f17278a1/cpp/order/HexaticTranslational.cc#L38-L45

A temporary workaround to achieve the desired behavior is to enabled `weighted=True` in the constructor. In ball queries and nearest neighbor queries (but not Voronoi queries), the weight for each bond defaults to 1. If `weighted=True`, the normalization will divide by the number of neighbor bonds (a sum with 1 for each bond).

To resolve this issue, we should verify the literature to see which convention is more common and consider changing to normalization by `1/n` instead of `1/k`.

bdice:
## Description
As noted by @Plastikschuessel and documented in #655, the hexatic order parameter normalization does not match its documentation. This changes the behavior to always normalize by `1/n` (the number of bonds, equal to the "total weight" for a non-weighted neighbor list) instead of `1/k` (the order parameter symmetry).

### To-do:
- [x] Fix documentation as described in #655 
- [x] Add some validation tests for correctness of Hexatic order parameter
- [x] Update credits/changelog

## Motivation and Context
Resolves: #655 and #665.

## How Has This Been Tested?
No tests failed in the original implementation. This indicates that new tests for correctness should be added. I added new normalization tests (they failed on the old code and passed on the new code).

## Types of changes

## Checklist:


freud/box.pyx
```diff
@@ -505,8 +505,8 @@ cdef class Box:
         ``[-L/2, L/2]``.
         Particles laying at a coordinate such as ``[0, L, 0]`` would be considered outside the box.
         More information about coordinate conventions can be found `here
-        <https://freud.readthedocs.io/en/latest/gettingstarted/examples/module_intros/box.Box.html?highlight=origin#Using-boxes>`_
-        and `here <https://freud.readthedocs.io/en/latest/gettingstarted/tutorial/periodic.html?highlight=origin#periodic-boundary-conditions>`_.
+        <https://freud.readthedocs.io/en/latest/gettingstarted/examples/module_intros/box.Box.html?highlight=origin#Using-boxes>`__
+        and `here <https://freud.readthedocs.io/en/latest/gettingstarted/tutorial/periodic.html?highlight=origin#periodic-boundary-conditions>`__.
```
bdice, line 509: I fixed a Sphinx error that I noticed while rendering the docs. The word "here" was being treated as a named link target, because it had one underscore. The target name "here" was being used twice, causing an error. I changed both instances of "here" to an anonymous link target (two underscores) to fix the problem.

freud/order.pyx
```diff
@@ -249,25 +249,44 @@ cdef class Hexatic(_PairCompute):
     :math:`n` neighbors :math:`j` is given by:
 
     :math:`\psi_k \left( i \right) = \frac{1}{n}
-    \sum_j^n e^{k i \phi_{ij}}`
+    \sum_j^n e^{i k \phi_{ij}}`
```
tommy-waltmann, line 252: Is using `i` as the argument to the function the best idea given the imaginary number `i` is also in the same formula? I know that its pretty easy to tell which is which and most people can tell the difference without any problem. For extra clarity though, its something to consider changing.

doc/source/reference/freud.bib
```diff
@@ -1,183 +1,208 @@
...
-    Month = {Dec},
-    Number = {24},
-    Pages = {244115},
-    Publisher = {AIP Publishing},
-    Title = {Crystal nucleation of hard spheres using molecular dynamics, umbrella sampling, and forward flux sampling: A comparison of simulation techniques},
-    Url = {http://dx.doi.org/10.1063/1.3506838},
-    Volume = {133},
-    Year = {2010},
-    Bdsk-Url-1 = {http://dx.doi.org/10.1063/1.3506838}}
+  author        = {Filion, L. and Hermes, M. and Ni, R. and Dijkstra, M.},
```
tommy-waltmann, line 2: Are all the changes to the citation style and indenting here within PR scope?

freud/order.pyx
```diff
@@ -249,25 +249,44 @@ cdef class Hexatic(_PairCompute):
...
     The quantity :math:`\phi_{ij}` is the angle between the
     vector :math:`r_{ij}` and :math:`\left(1, 0\right)`.
 
+    If the weighted mode is enabled, contributions of each neighbor are
+    weighted. Neighbor weights :math:`w_j` default to 1 but are defined for a
+    :class:`freud.locality.NeighborList` from :class:`freud.locality.Voronoi`
+    or one with user-provided weights. The formula is modified as follows:
+
+    :math:`\psi'_k \left( i \right) = \frac{1}{\sum_j^n w_j}
+    \sum_j^n w_j e^{i k \phi_{ij}}`
```
tommy-waltmann, line 265: Same as above.

freud/order.pyx
```diff
@@ -377,13 +396,25 @@ cdef class Hexatic(_PairCompute):
...
 
+    The translational order parameter is used to measure order in the bonds
+    of 2D systems. The translational order parameter for a particle :math:`i`
+    and its :math:`n` neighbors :math:`j` is given by a sum over the
+    neighbors, treating the 2D vectors between each pair of particles as a
+    complex number with real part corresponding to the x-component of the
+    vector and imaginary part corresponding to the y-component of the vector,
+    divided by a normalization constant :math:`k`:
+
+    :math:`\psi\left( i \right) = \frac{1}{k} \sum_j^n x_{ij} + y_{ij} i`
```
tommy-waltmann, line 407: Same as above.

tests/test_order_Hexatic.py
```diff
@@ -137,6 +137,43 @@ def test_weighted_hex(self):
...
+        rijs = box.wrap(points[point_indices] -
+                        points[query_point_indices])
+        distances = np.linalg.norm(rijs, axis=-1)
+        weights = np.array([1, 0.7, 0.3, 0])
+        nlist = freud.NeighborList.from_arrays(
+            len(points), len(points),
+            point_indices, query_point_indices,
+            distances, weights)
+        rijs = box.wrap(points[nlist.point_indices] -
+                        points[nlist.query_point_indices])
```
tommy-waltmann: Is this definition of `rijs` the same as the one above?

tommy-waltmann (commented):
Just a few small things to consider before merging.

freud/order.pyx
bdice: I swapped the order on purpose from `ki` to `ik` so that the `i` is more obviously an imaginary number in the context of `e^i`. I recognize this issue -- I tried to think of other options but `j, k` don't work because `k` is the hexatic order ... we could use `\alpha, \beta` to denote particles like some literature but I find that confusing sometimes as well. Possibly `m, n` if that dodges other uses of quantum numbers. I think I'm going to leave this as-is for now. I would recommend fixing this with a pull request that addresses this globally and is consistent across all modules.

doc/source/reference/freud.bib
bdice: It happened automatically on save (in VSCode with some LaTeX extension) and I preferred the style that it adopted, so I decided to just commit it.

tests/test_order_Hexatic.py
bdice: Good catch, I think this can be removed. I'll double check it locally.

freud/order.pyx
tommy-waltmann: Yeah, most of the conventional letters are already used and if the `i,j` indices are consistently used throughout the documentation, then it could be confusing to switch in just one place.

tests/test_order_Hexatic.py
bdice: I found an issue (seemingly numerical precision) in the Voronoi NeighborList calculation while fixing this. I pushed changes to fix this in eade73a and created a corresponding issue #665 to document the problem. The issue originally mentioned in this conversation is fixed in c197ce8.

freud/order.pyx
```diff
@@ -377,13 +396,25 @@ cdef class Hexatic(_PairCompute):
...
+    The translational order parameter as written above is **complex-valued**.
+
     .. note::
         **2D:** :class:`freud.order.Translational` is only defined for 2D
         systems. The points must be passed in as :code:`[x, y, 0]`.
 
     Args:
         k (float, optional):
-            Symmetry of order parameter. (Default value = :code:`6.0`).
+            Normalization of order parameter (Default value = :code:`6.0`).
```
tommy-waltmann, line 417: This parameter is still related the the symmetry of the translational order right? Ideally the description in the docs should say something about both.

tommy-waltmann (commented):
One last thing before this is ready for merging.

freud/order.pyx
bdice: Since this value is explicitly used as a normalization (the final result is divided by `k`), and it's not necessarily indicative of any kind of `k`-fold symmetry in the particle's neighbors, I think this change clarifies how the value is used. Also we don't really know how the Translational order parameter has been used in practice before and we're probably going to deprecate and remove it in freud 3.0, as I discussed with @vyasr in #655.

vyasr: Since this PR contains a bunch of changes to Translational that are unrelated to the original goal of the PR, we should just apply the deprecation here as well.

bdice: @vyasr I already created an issue #659 and will make sure that gets done before the next release -- but I would like to get this PR reviewed and merged with its current scope, rather than stall it further. The existing changes to `Translational` in this PR just aimed to improve its docs, because of the specific change request in #655. 😕

tommy-waltmann (approved):

freud/order.pyx
vyasr: Fair enough, the changes here are close enough that I'm fine with leaving them here and deprecating separately.

PR #660, opened 2020-09-29
Static structure factor S(q)

### LINKED ISSUE #652: 1-d structure factor calculation
Right now freud does not compute 1-d structure factor. It would be nice to have that feature. 
The equation of the 1-d structure factor is given here : http://isaacs.sourceforge.net/phys/scatt.html (equation 1).

Desired solution: A function that takes an mdtraj trajectory and gives out the structure factor (s(q) and q).

**Additional context**
This repository (MIT License) : https://github.com/mattwthompson/scattering/pulls has some tools to calculate scattering properties of materials. If needed, hints can be taken from this repo.

bdice:
## Description
A commonly desired quantity from simulations is S(q), the static structure factor. The recently introduced `diffraction` module (#596) is an ideal location for this feature. In my understanding, the structure factor can be calculated two different ways: directly (which is expensive, `O(N^2)` for a system of `N` particles), or indirectly via a Fourier transform of a radial distribution function. I have heard there is significant controversy over the choice of method and the regimes in which each is correct. I hope to offer both methods in this pull request, as well as some clarity in the documentation for when each might be (in)appropriate for use.

The second thing to outline is the _scope_ of this pull request. **This is a first-pass, and will only support systems with a single species, and only one set of particle positions (i.e. `points == query_points`).** This PR is intentionally held to a very narrow scope, to reduce the complexity of the initial implementation and to solidify testing requirements for the "base case" upon which further features may someday be added, such as:
- multi-species structure factors (via `points` and `query_points` as well as hints on how to normalize the result correctly)
- accumulation over multiple frames (using freud's standard `reset=False` approach -- but may require additional normalizations!)
- particle form factors (e.g. via a secondary array of values)
- bonded contributions e.g. from polymers (I think there might be another term for this, but not sure)

## Motivation and Context
Discussed with @ramanishsingh and also desired for my own research.

The reason to implement this feature in freud (rather than point users to another existing package) is that it complements and can leverage the fast neighbor-finding and other features of freud, the feature itself can be implemented in parallelized C++, and it fits in the scope of colloidal-scale simulation analysis that freud emphasizes. 

Resolves: #652 

## TODO (help welcome)

- [x] Consolidate implementation code from Cython into C++
- [x] Validate results approximately match against a known reference
- [ ] Validate FFT-RDF method against direct method for *q* values greater than 4pi/L (or something like that) where they should agree
- [ ] Write docs and seek expert knowledge (literature) about when each method is valid
- [ ] Write tests to ensure behavior is not broken if/when new features are added

## How Has This Been Tested?
I plan to compare this code against a few existing implementations to verify its accuracy.

Validate against:
- https://github.com/mattwthompson/scattering/
- other implementations from Glotzer group members

## Types of changes

## Checklist:


vyasr:
Paraphrasing a discussion I had offline with @bdice, I'm a little confused at how the RDF-based calculation actually has better scaling behavior than the "direct" calculation. The RDF is also an `O(n^2)` calculation, the only reason it's typically faster is because people use short cutoffs to reduce the number of interactions. If you use the maximum cutoff that is currently being used, I would expect that in a uniformly distributed system the RDF would be faster by the ratio of the volume of the ball of radius L/2 to the volume of the box, since the RDF will ignore some neighbors for that reason. Otherwise, the two should have the same scaling behavior. Before we finish this off I would like to see benchmarks to validate that and to consider how much benefit we get from providing both implementations. Not saying it's not worthwhile, just want to be sure.

bdice:
Following up from discussion with @vyasr, here's some timing as of commit 3e9066b. This isn't quite enough data to assess "scaling behavior" but it's what I could quickly gather.

| Time (s)      | N=1000 | N=2000 | Notes                                                                        |
|---------------|--------|--------|------------------------------------------------------------------------------|
| Python RDF    | 0.361  | 0.369  | Uses freud (C++) for RDF, cost of Python FFT/integration is probably much greater. |
| Python Direct | 7.30   | 39.0   |                                                                                |
| C++ RDF       | 0.0056 | 0.0162 |                                                                                |
| C++ Direct    | 1.25   | 5.09   |                                                                                |

I'm not sure what the breakdown of time spent looks like, but this addresses the general question of "is RDF _faster_" with a clear yes. However, based on this small amount of data, the scaling might be something like O(N^2) for both methods as @vyasr had hypothesized, since the RDF is long-ranged.

vyasr:
I just took a brief look at the code and I don't have any immediate thoughts on why either method as implemented here would be so much faster than the other. My intuition would be that if you looked at the code before the loop over k in both `accumulateDirect` and `accumulateRDF` the performance would be comparable up to some small scaling factor; both `RDF::compute` and `box.computeAllDistances` are `O(N^2)` calculations, with the primary difference that `RDF::compute` has a larger prefactor because it also performs binning of the distances (meaning that the RDF code prior to the k loop should be _slower_). AFAICT the code in the k loop is also nearly identical; they're two nested for loops (with the inner loop hidden either by `std::for_each` or `util::simpson_integrate`) over the exact same variables. There must be some operation in the direct case that's more expensive, but it's not apparent to me what it is.

This seems like a fun problem to dig into, but unfortunately not one I have time for at the moment. I'll keep an eye on this though and maybe I'll get a chance before you finish up the PR :)

bdice:
@DomFijan I finished most of the tests that I wanted to write. There is a failing test that we need to figure out how to fix. I am having difficulty with memory management in freud and I can't figure out whether I'm breaking a pattern or not. In the failing test, I'm assigning an array like `S_k = sf.S_k` to save a computed structure factor array, but then calling `compute` and accessing the `sf.S_k` attribute again causes the previous variable `S_k` to change, which should not happen. freud's ManagedArray is supposed to provide zero-copy access (for performance). The ManagedArray should detect that it needs to re-allocate a new array when `compute` is called the second time. It appears there is something going wrong (perhaps a reference is being held someplace that shouldn't be) but I can't see where the issue is right now. I might consult with @vyasr, who designed the memory management system.

freud/diffraction.pyx
```diff
@@ -21,13 +21,225 @@ import scipy.ndimage
...
+    as a sum of the partial scattering functions as:
+
+    .. math::
+
+        S(k) - 1 = \sum_{\alpha}\sum_{\beta} \frac{N_{\alpha}
+        N_{\beta}}{N_{total}^2} \left(S_{\alpha \beta}(k) - 1\right)
+
+    This class is based on the MIT licensed `scattering library
+    <https://github.com/mattwthompson/scattering/>`__ and literature references
+    :cite:`Liu2016`.
```
bdice: Additional suggested citations for Faber-Ziman:
- https://openscholarship.wustl.edu/etd/1358/
- http://isaacs.sourceforge.net/manual/page26_mn.html
- https://www.tandfonline.com/doi/abs/10.1080/14786436508211931

bdice: We can include a formula for converting to the Ashcroft-Langreth formalism if desired.
https://iopscience.iop.org/article/10.1088/0034-4885/69/1/R05/pdf

bdice: @DomFijan suggests adding code examples, especially for cross-term scattering -- we can use the tests as a reference for this. These could also be added to https://github.com/glotzerlab/freud-examples/pull/24

vyasr:
I'm going to take a look at the failing test now and see what I come up with.

vyasr:
The reduction was missing a `prepare` call to the structure factor ManagedArray, so there was no check being done for whether or not reallocation was necessary. It looks like this bug was hiding a couple of other issues in the tests that are now showing up. The symmetry test needs a slightly looser tolerance but otherwise looks fine, so I just bumped the tolerance. The partial sum normalization test, however, looks to have a real bug that was previously not caught because I suspect that the formula's symmetry guarantees the tested equivalence if `S_AA == S_AB == S_BA == S_BB`, which was previously the case. That issue will need to be addressed before this PR is ready.

bdice:
I'm copying the current to-do list discussed with @DomFijan into this PR:

- [ ] Convert freud-like class into an actual freud _Compute class
- [ ] Docs
- [ ] Tests

As for dynasor / dependency management:
- [ ] First pass: Leave hard dependency on dynasor
- [ ] Second pass: Copy C code into a C++ module in freud (eliminate calc_rho_k)
- [ ] Third pass: Copy reciprocal_isotropic into freud (eliminate dynasor dependency)
- [ ] Fourth pass: Move as much logic as possible into C++

DomFijan:
CI is currently erroring out because our numpy requirement is  ==1.14.0, whereas ASE version that has Debye implementation used for a test requires numpy =>1.15.0. Would upping the numpy version requirement be alright in this case?

freud/diffraction.pyx
```diff
@@ -21,13 +21,220 @@ import scipy.ndimage
...
+
+    .. note::
+        This code assumes all particles have a form factor :math:`f` of 1.
+
+    Partial structure factors can be computed by providing ``query_points`` to the
+    :py:meth:`compute` method. When computing a partial structure factor, the
+    total number of points in the system must be specified. The normalization
+    criterion is based on the Faber-Ziman formalism. For particle types
+    :math:`\alpha` and :math:`\beta`, we compute the total scattering function
+    as a sum of the partial scattering functions as:
```
tommy-waltmann: According to the description in the PR, `query_points` has to be the same as `points`. Is that outdated now?

cpp/diffraction/StaticStructureFactorDebye.h
```diff
@@ -0,0 +1,94 @@
...
+    Computes structure factors from the Fourier transform of a radial distribution function (RDF).
+    This method is not capable of resolving k-vectors smaller than the magnitude 4 * pi / L, where L is the
+    smallest side length of the system's periodic box.
+*/
+
+namespace freud { namespace diffraction {
+
+class StaticStructureFactorDebye
+{
+    using StaticStructureFactorDebyeHistogram = util::Histogram<float>;
```
tommy-waltmann: Doesn't this using statement just make everything unnecessarily verbose? I would suggest shortening the name somewhat.

cpp/diffraction/StaticStructureFactorDebye.h
```diff
@@ -0,0 +1,94 @@
...
+
+public:
+    //! Constructor
+    StaticStructureFactorDebye(unsigned int bins, float k_max, float k_min = 0);
+
+    //! Destructor
+    virtual ~StaticStructureFactorDebye() = default;
```
tommy-waltmann: Do we plan on having classes derived from this one? Right now there isn't a need to make this virtual.

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -0,0 +1,104 @@
...
+                auto const distance = distances[distance_index];
+                S_k += util::sinc(k * distance);
+            }
+            S_k /= static_cast<double>(n_total);
+            m_local_histograms.increment(k_index, S_k);
+        };
+    });
+    m_frame_counter++;
+    m_reduce = true;
+}
```
tommy-waltmann: There's a lot of `auto` variables here. I think some of them are fine to be left as `auto`, but many like in lines 56, 57, and 72 should be declared as their actual type to prevent confusion.

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -0,0 +1,104 @@
...
+        = box.is2D() ? std::min(box_L.x, box_L.y) : std::min(box_L.x, std::min(box_L.y, box_L.z));
+    auto const r_max = std::nextafter(float(0.5) * min_box_length, float(0));
+    auto const points= neighbor_query->getPoints();
+    auto const n_points= neighbor_query->getNPoints();
+
+    // The minimum k value of validity is 4 * pi / L, where L is the smallest side length.
+    // This is equal to 2 * pi / r_max.
+    m_min_valid_k = std::min(m_min_valid_k, freud::constants::TWO_PI / r_max);
+
+    auto distances = std::vector<float>(n_points * n_query_points);
```
tommy-waltmann:
```suggestion
    std::vector<float> distances(n_points * n_query_points);
```

cpp/util/Histogram.h
```diff
@@ -72,13 +72,13 @@ class Axis
     virtual size_t bin(const float& value) const = 0;
 
     //! Return the boundaries of bins.
-    std::vector<float> getBinEdges() const
+    const std::vector<float> getBinEdges() const
```
tommy-waltmann: This comment goes for every new `const` return type declaration in this file: Either return the `std::vector` and allow a copy to be made or return by `const std::vector<>&`

setup.py
```diff
@@ -90,6 +90,7 @@ def setup(*args, **kwargs):
         "scipy>=1.1",
     ],
     tests_require=[
+        "ase==3.22.0",
```
tommy-waltmann: Given the issues that CI is having with ASE, is there a lower ASE version we can use here?

tommy-waltmann (changes requested):

setup.py
DomFijan: Lowered the version of ASE to 3.21.1 which seems to have worked out fine!

freud/diffraction.pyx
DomFijan: This is correct. Multi-species structure factors (or partial structure factors) were originally to be added later, but were added as implementation wasn't to hard. The docs are correct.

DomFijan: We have implemented the support for "multi-species structure factors" (partial structure factors). This can be done by providing different set of `query_points` and `points` along with `N_total`. The original PR description is outdated.

cpp/util/Histogram.h
DomFijan: Would there be any preference in one over the other in this case?

cpp/diffraction/StaticStructureFactorDebye.h
DomFijan: I've changed the name to S_kHistogram.

cpp/diffraction/StaticStructureFactorDebye.h
DomFijan: I don't think so? @bdice

cpp/diffraction/StaticStructureFactorDebye.cc
DomFijan: I've changed some auto variables. Let me know if you're happy with changes.

cpp/util/Histogram.h
tommy-waltmann: The decision would have to be made on a case-by-case basis. Anytime you are returning a local variable you should return by `std::vector<>`, most other cases it would probably be better to return by `const std::vector<>&`. If you want more advice with this we should talk in person.

cpp/diffraction/StaticStructureFactorDebye.cc
bdice: @tommy-waltmann @DomFijan Many C++ experts consider it good style to adopt "Almost Always Auto." References:
- https://www.fluentcpp.com/2020/12/04/auto-for-types-but-not-for-concepts/
- https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/

I see no advantage to dropping `auto` in the cases @tommy-waltmann pointed out and would advocate to revert commit 0b3aeba3e7359066c783ccab545514b5aa37575e.

cpp/util/Histogram.h
bdice: The difference has to do with ownership and mutability. If the object is returning one of its members, it can return by constant reference (`const std::vector<T>&`) to avoid returning a new copy of the data, while also ensuring that the returned reference is immutable (constant) so that the calling code cannot mutate this object's internal state. If the vector being returned is not a member (for example, if it is allocated and filled by the function being called) then you must return `std::vector<T>` -- note that this cannot be constant because it is a returning by copy (or move if the copy is elided), and it cannot be a reference because the data's lifetime is limited to the function body and the reference would be invalid.

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -0,0 +1,104 @@
...
+StaticStructureFactorDebye::StaticStructureFactorDebye(unsigned int bins, float k_max, float k_min)
+{
+    if (bins == 0)
+    {
+        throw std::invalid_argument("StaticStructureFactorDebye requires a nonzero number of bins.");
+    }
+    if (k_max <= 0)
+    {
+        throw std::invalid_argument("StaticStructureFactorDebye requires k_max to be positive.");
+    }
```
bdice, line 27: Add a check for `k_min >= 0`.
```suggestion
    if (k_max <= 0)
    {
        throw std::invalid_argument("StaticStructureFactorDebye requires k_max to be positive.");
    }
    if (k_min < 0)
    {
        throw std::invalid_argument("StaticStructureFactorDebye requires k_min to be non-negative.");
    }
```

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -0,0 +1,104 @@
...
+                                             const vec3<float>* query_points, unsigned int n_query_points, unsigned int n_total)
+{
+    auto const& box = neighbor_query->getBox();
+
+    // The r_max should be just less than half of the smallest side length of the box
+    auto const box_L = box.getL();
+    auto const min_box_length
+        = box.is2D() ? std::min(box_L.x, box_L.y) : std::min(box_L.x, std::min(box_L.y, box_L.z));
+    float const r_max = std::nextafter(float(0.5) * min_box_length, float(0));
+    const vec3<float>* points= neighbor_query->getPoints();
```
bdice: Separate from the conversation about why we should prefer `auto` (see other thread), we should be consistent between "east const" and "west const": https://hackingcpp.com/cpp/design/east_vs_west_const.html
```suggestion
    vec3<float>* const points = neighbor_query->getPoints();
```

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -0,0 +1,104 @@
...
+    std::vector<float> distances(n_points * n_query_points);
+    box.computeAllDistances(points, n_points, query_points, n_query_points, distances.data());
+
+    auto const k_bin_centers = m_histogram.getBinCenters()[0];
+
+    util::forLoopWrapper(0, m_histogram.getAxisSizes()[0], [&](size_t begin_k_index, size_t end_k_index) {
+        for (size_t k_index = begin_k_index; k_index < end_k_index; ++k_index)
+        {
+            float const k = k_bin_centers[k_index];
+            double S_k = 0.0;
```
bdice, line 73: I'm unsure why I chose to use `double` instead of `float` here. Please try switching to `float` and then we can determine if the difference matters, now that we have some tests for numerical accuracy.

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -0,0 +1,104 @@
...
+
+void StaticStructureFactorDebye::reduce()
+{
+    m_structure_factor.prepare(m_histogram.shape());
+    m_local_histograms.reduceInto(m_structure_factor);
+
+    // Normalize by the frame count if necessary
+    if (m_frame_counter > 1)
+    {
+        util::forLoopWrapper(0, m_structure_factor.size(), [this](size_t begin, size_t end) {
```
bdice: For consistency with #837, we can just use `&` instead of specifying `this`.
```suggestion
        util::forLoopWrapper(0, m_structure_factor.size(), [&](size_t begin, size_t end) {
```

cpp/diffraction/StaticStructureFactorDebye.h
bdice: It is probably unnecessary to define the destructor. Try deleting it. I think this class can adhere to the "Rule of zero": https://en.cppreference.com/w/cpp/language/rule_of_three

cpp/util/Histogram.h
bdice: Reading this again, I am not sure what encouraged me to change this file at all. I think I was trying to fix warnings from VSCode but I'm not sure if this change actually does anything. Maybe let's just revert the changes in this file, or perhaps only make changes that return class members by constant reference (where applicable) and revert the rest.

setup.py
```diff
@@ -90,6 +90,7 @@ def setup(*args, **kwargs):
         "scipy>=1.1",
     ],
     tests_require=[
+        "ase==3.21.1",
```
bdice: Test pinnings in `setup.py` should be a minimum version (they do not affect our CI). If a version older than 3.21.1 works, then change the number.
```suggestion
        "ase>=3.21.1",
```

requirements/requirements-test.txt
```diff
@@ -1,3 +1,4 @@
+ase==3.21.1
```
bdice: `requirements-test.txt` should pin to the _newest_ available version of a package.
```suggestion
ase==3.22.0
```

tests/test_diffraction_StaticStructureFactorDebye.py
```diff
@@ -0,0 +1,215 @@
...
+
+        atoms = ase.Atoms(
+            positions=points, pbc=True, cell=box.L, numbers=np.ones(len(points))
+        )
+        xrd = XrDebye(atoms=atoms, wavelength=1.0, method="??", damping=0)
+        S_ase = []
+        for i in sf.bin_centers / (2 * np.pi):
+            S_ase.append(xrd.get(i))
+        S_ase = np.asarray(S_ase) / len(points)
+        npt.assert_allclose(sf.S_k, S_ase, rtol=1e-5, atol=1e-5)
```
bdice: Make this test have a soft dependency on ase if possible. See examples:
- Optional module: https://github.com/glotzerlab/freud/blob/ea8c460ddc81b153616d2695fa68627d653a49b3/tests/integration/test_reader_integrations.py#L34
- Optional function/class: https://github.com/glotzerlab/freud/blob/ea8c460ddc81b153616d2695fa68627d653a49b3/tests/test_environment_LocalDescriptors.py#L336


```suggestion
    def test_debye_ase(self):
        """Validate Debye method agains ASE implementation"""
        ase = pytest.importorskip("ase")
        xrdebye = pytest.importorskip("ase.utils.xrdebye")
        bins = 1000
        k_max = 100
        k_min = 0
        box, points = freud.data.UnitCell.fcc().generate_system(4, sigma_noise=0.01)
        box.periodic = False
        print(box)
        system = freud.locality.NeighborQuery.from_system((box, points))
        sf = freud.diffraction.StaticStructureFactorDebye(bins, k_max, k_min)
        sf.compute(system)
        atoms = ase.Atoms(
            positions=points, pbc=True, cell=box.L, numbers=np.ones(len(points))
        )
        xrd = xrdebye.XrDebye(atoms=atoms, wavelength=1.0, method="??", damping=0)
        S_ase = []
        for i in sf.bin_centers / (2 * np.pi):
            S_ase.append(xrd.get(i))
        S_ase = np.asarray(S_ase) / len(points)
        npt.assert_allclose(sf.S_k, S_ase, rtol=1e-5, atol=1e-5)
```

bdice: I think this is generally on the right track with most parameters should set to 1 or 0, e.g. wavelength 1, damping 0, method `None` (not the default of Iwasa). I would also try `alpha=0` so that there is no angular damping of scattering intensity?

You can probably use `calc_pattern` with SAXS and provide an array of `q` values instead of repeatedly calling `get(x)`. SAXS lets you specify the values of `q` directly. The XRD mode requires you to specify the angles, which is not what we want ([see description here](https://wiki.fysik.dtu.dk/ase/ase/xrdebye.html#ase.utils.xrdebye.XrDebye.calc_pattern)).

bdice (commented):
@DomFijan I have a few suggestions below.

cpp/diffraction/StaticStructureFactorDebye.cc
tommy-waltmann: Okay, I'll buy the arguments from the articles. We can revert those lines.

cpp/diffraction/StaticStructureFactorDebye.cc
tommy-waltmann: Since the rest of freud uses "west const", as the article calls it, I would be in favor of updating all files in this PR to use west const instead of east to stay consistent.

requirements/requirements-test.txt
tommy-waltmann: the newest version of this package caused testing issues, there was some discussion about it in a resolved thread above. Not sure if the version in this file needs to match what's in `setup.py` or not.

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -0,0 +1,104 @@
...
+    // Construct the Histogram object that will be used to track the structure factor
+    auto axes
+        = S_kHistogram::Axes {std::make_shared<util::RegularAxis>(bins, k_min, k_max)};
+    m_histogram = S_kHistogram(axes);
+    m_local_histograms = S_kHistogram::ThreadLocalHistogram(m_histogram);
+    m_min_valid_k = std::numeric_limits<float>::infinity();
+    m_structure_factor.prepare(bins);
+}
+
+void StaticStructureFactorDebye::accumulate(const freud::locality::NeighborQuery* neighbor_query,
```
tommy-waltmann, line 45: there are some places in this function where we use `float` and other places where we use `double`. @bdice What are freud's rules about when to use single vs double precision?

requirements/requirements-test.txt
DomFijan: Higher versions then 3.21.1 cause pip to freak out in the CI because of the numpy dependency version conflict. 
see
https://github.com/glotzerlab/freud/pull/660#issuecomment-915550017

cpp/diffraction/StaticStructureFactorDebye.cc
DomFijan: Upon changing `double` to `float` certain test fail on about 10% of values checked. It seems we should leave double here.

setup.py
DomFijan: Wouldn't this cause pip to fail to install when installing with tests due to https://github.com/glotzerlab/freud/pull/660#issuecomment-915550017  ?

cpp/diffraction/StaticStructureFactorDebye.h
DomFijan: Deleted.

cpp/diffraction/StaticStructureFactorDebye.cc
DomFijan: The commit was reverted.

tests/test_diffraction_StaticStructureFactorDebye.py
DomFijan: Agreed on parameters. I updated `method` to `None` and added `alpha=0` . The reason I opted out from using `calc_pattern` was because that function does division of q by 2 \pi under the hood which might made it hard to track 2 \pi 's. It also uses `get(x)` under the hood anyways. I updated the test so that it uses `calc_pattern` now with some extra comments. It does look cleaner though IMO.

cpp/util/Histogram.h
DomFijan: I tried to do the latter suggestion from @bdice last comment. Let me know if I was successful.

cpp/diffraction/StaticStructureFactorDebye.cc
DomFijan: Updated to "west const".

bdice:
@DomFijan Please check on the failing C++ style checks: https://app.circleci.com/pipelines/github/glotzerlab/freud/3142/workflows/91c53b9e-78e3-4efc-aefe-aa42b6828e1d/jobs/12256

DomFijan:
@bdice I fixed some style issues, but CI is now failing with following message:
`Too long with no output (exceeded 10m0s): context deadline exceeded`

bdice:
> @bdice I fixed some style issues, but CI is now failing with following message:
> `Too long with no output (exceeded 10m0s): context deadline exceeded`

No worries. It fails with that message sometimes. Re-running tests may solve it. I think clang-tidy might take a lot of resources and hangs the CI servers sometimes. Either way, I’ll take a look and review soon. Thanks!

cpp/diffraction/StaticStructureFactorDebye.cc
bdice: @tommy-waltmann The rules are generally "use `float` where possible, and `double` where precision requires it." This is discussed further in #464. Also see the comment thread below: https://github.com/glotzerlab/freud/pull/660/files#r707543882

Resolving this.

cpp/util/utils.h
```diff
@@ -21,11 +22,64 @@ inline float clamp(float v, float lo, float hi)
...
+
+    for (size_t bin_index = 0; bin_index < num_bins; bin_index++)
+    {
+        integral += simpson_prefactor(bin_index) * integrand(bin_index);
+    }
+
+    integral *= dx / 3.0;
+    return integral;
+}
+
```
bdice: It looks like the Simpson integration method is also unused. I'll remove it.

cpp/locality/NeighborQuery.h
```diff
@@ -51,6 +51,20 @@ struct QueryArgs
...
+          scale(scale), exclude_ii(exclude_ii)
+    {}
+
+    //! Ball query factory
+    static QueryArgs make_ball(float r_max, float r_min = 0, bool exclude_ii = DEFAULT_EXCLUDE_II)
+    {
+        return QueryArgs(QueryType::ball, DEFAULT_NUM_NEIGHBORS, r_max, r_min, DEFAULT_R_GUESS, DEFAULT_SCALE,
+                         exclude_ii);
+    }
+
```
bdice: I think these changes can be reverted because they're no longer being used.

bdice:
@DomFijan You refactored the calculation to include all point-point distances in commit 03dd664df2f88d88dd41adfe47c2a88c4fc51b07, including those greater than the largest `r_max = L_min/2` permitted by the box. Can you comment on how you justified that change? Theory? Literature? Matching to other expected results? We should add a comment explaining that in the C++ code and/or Python docstrings. That's the only comment I have left on this PR and it can be discussed after merging, so I'm going to review #820 and then merge these PRs tomorrow.

DomFijan:
> @DomFijan You refactored the calculation to include all point-point distances in commit [03dd664](https://github.com/glotzerlab/freud/commit/03dd664df2f88d88dd41adfe47c2a88c4fc51b07), including those greater than the largest `r_max = L_min/2` permitted by the box. Can you comment on how you justified that change? Theory? Literature? Matching to other expected results? We should add a comment explaining that in the C++ code and/or Python docstrings. That's the only comment I have left on this PR and it can be discussed after merging, so I'm going to review #820 and then merge these PRs tomorrow.

The results from the calculation weren't matching the "direct" implementation results which looked very sane and were matching Dynasor. Once I discovered that ASE implements Debye formula scattering I made a test that would directly compare the two Debye implementations. They weren't matching as one would expect, but ASE was matching the "direct" implementation pretty well (not perfectly because low k-values). So the first thing that I thought off was to revert back the change we made couple of months ago to go to ball query from whole system (if memory serves me correctly). And that fixed it. Note that ASE calculates the scattering WITHOUT PBC exclusively as far as I could figure out, while our implementation allows PBCs if box has PBCs turned on. Turning on PBCs gives worse results at low k-values then without PBCs (that would make sense), but the results are very similar in general.

As for why this is correct physically or mathematically and why ball query doesn't seem to be working well I can only speculate. The amount of distance pairs when going from L/2 ball query to whole system is considerable (same order of magnitude - I did not expect this), and these are almost exclusively mid to low k-values that get affected. It could be that the whole nature of the sinc function makes the problem of low amount of sampling worse (I think this is the most likely explanation given how Debye compares to "direct" method and how Debye with and without PBCs look comparatively). 

Do you think we should state in the docs that our implementation is supposed to match ASE's? On another note I believe there is a rework for ASE's Debye implementation on their gitlab in works.

freud/diffraction.pyx
```diff
@@ -23,11 +23,248 @@ import freud.locality
...
+    .. math::
+
+        S(k) = \frac{1}{N} \sum_{i=0}^{N} \sum_{j=0}^{N} \text{sinc}(k r_{ij})
+
+    where :math:`N` is the number of particles, :math:`\text{sinc}` function is
+    defined as :math:`\sin x / x` (no factor of :math:`\pi` as in some
+    conventions). For more information see `this Wikipedia article
+    <https://en.wikipedia.org/wiki/Structure_factor>`__. The equation 4 from
+    the link can be obtained by replacing :math:`\frac{\sin(k r)}{kr}` with
+    :math:`\text{sinc}(k r)`. For a full derivation see :cite:`Farrow2009`.
```
bdice: I'm not sure if I would reference a specific equation on Wikipedia -- this article could easily change, invalidating our documentation. Moreover, we have already defined the `sinc` function. I will remove this sentence.

freud/diffraction.pyx
```diff
@@ -23,11 +23,248 @@ import freud.locality
...
+            ...     query_points=B_points,
+            ...     N_total=N_particles
+            ... )
+
+        Args:
+            system:
+                Any object that is a valid argument to
+                :class:`freud.locality.NeighborQuery.from_system`.
+                Note that box is allowed to change when calculating trajectory
+                average static structure factor.
```
bdice, line 123: I'm unsure about this comment:
> Note that box is allowed to change when calculating trajectory average static structure factor.

I'm leaving it in but I think it's a little awkward. We discuss accumulating and resetting, but usually not "trajectory average." The accumulation/reset doesn't necessarily have to be over an MD trajectory (could be independent replicas, or Monte Carlo which isn't really a trajectory, etc.). It also doesn't fit with the rest of freud. I'm still not sure if changing boxes produce a valid result... we know that NVT (constant box) averages are fine but I can't say for sure if NPT averages, for instance, would be valid. Especially without periodic boundaries.

requirements/requirements-test.txt
bdice: @tommy-waltmann @DomFijan I think something got lost in translation here. We have two sets of testing requirements. One set (in `.circleci/ci-oldest-reqs.txt`) is for Python 3.6 and tests our _oldest_ supported dependencies. The other set of dependencies (in `requirements-test.txt`) is for Python 3.7, 3.8, and 3.9 and tests against the _newest_ released versions. Dependabot keeps the "newest" pinnings updated so we know if new package versions break our tests.

The confusion over NumPy versions shouldn't apply here. In the oldest requirements, I pinned to `ase==3.16.0`, which is a fairly old version where the xrdebye code was updated (according to the ase changelog). That version of `ase` doesn't require any particular version of NumPy and appears to work just fine in my local testing with the "oldest" environment.

In the newest dependencies, I pinned to `ase==3.22.0`. This requires `numpy>=1.15.0`. However, we are testing with NumPy 1.21.2, which means this should be fine.

setup.py
bdice: We don't install our testing dependencies in CI from `setup.py`. These values are just for user reference (or if the user requests to install the testing dependencies using some `python setup.py ...` command). That's also why these are all very broad pinnings like `>=MAJOR.0`. I identified that `3.16.0` was the lowest version we'll support, so I'm pinning this to `>=3.16`.

bdice:
@DomFijan Thanks for the info in your comment above. Your explanation and extensive testing are very helpful. I don't think we need to do anything else.

I made a final pass and fixed a bunch of small things. I'll merge once tests pass. Then we can move on to #820, and then finalize the example notebooks.

tommy-waltmann:
Why was this PR merged before getting an approving review?

bdice:
@tommy-waltmann I have been discussing with @DomFijan via Slack in the `#structure-factor` channel and we agreed to merge this and make further improvements in a new PR. I feel it meets the quality standards of the package and left comments above to indicate my approval of @DomFijan’s hard work, but I was not able to “approve” the PR since I was the initial author. Your suggestions for improvement would be very welcome, and we can apply them in a new PR. This just needed to move forward (after a year of development) so that we can finalize #820. This feature is also “unstable” (as is DiffractionPattern) so we can adapt its API and behavior freely.

freud/diffraction.pyx
DomFijan: I agree that this doesn't seem to fit the general language of freud. I will change this in the new PR. 

I can't see why NPT simulations would produce incorrect results. Especially in this implementation where there is no underlying k-space grid but you histogram distances directly which (aside from applying PBCs correctly) shouldn't care about box properties. The box properties are supposed to fluctuate around some equilibrium value anyways if a proper ensamble is sampled so this shouldn't be a problem. Is there anything that bothers you in particular here? We could also test this - by calculating S(q) for a system at same thermodynamic state point using NPT and NVT and comparing. But if proper equilibrium is reached in both there is no way that results would differ. This would be equivalent to some other structural property being different between NPT and NVT simulation under same conditions which shouldn't be the case(?).

As for calculations without PBCs, these probably don't make much sense, aside from possibly having a HUGE experimental sample of couple of millions ++ of particles  where one would want to calculate scattering numerically from positions of particles.

bdice: I guess I don’t have specific objections aside from inconsistent language. I know our accumulation code for RDF and several other classes has to normalize by the box volume, and that means that we require constant box volume for correct results (we could improve this behavior by tracking more info about each frame’s box and perhaps reducing data over threads for each frame?).

DomFijan: I see. That makes sense in those cases. S(k) doesn't care about volume (right?), so it shouldn't be an issue here. 

As for other functions, perhaps a good approach would be to track if the box changes in the compute function and then reduce only when box change occurs? Would such an approach make more sense? Then box wouldn't have to be stored for each step or reduction happens only when it is actually needed?

bdice: Yes, I think this might be fine since it doesn't normalize by box volumes. I think it is possible to make changing boxes work for RDF and other functions using a strategy like you describe, but I think it would require careful attention. See also #616 for previous conversations on this topic.

PR #661, opened 2020-09-29
Enable CMake builds of C++ library.

bdice:
## Description
This is an initial sketch of a `CMakeLists.txt` build file for freud's C++ components.

## Motivation and Context
It might be advantageous to use CMake with a system like [scikit-build](https://scikit-build.readthedocs.io/en/latest/) to build freud, because the C++ library could be built separately from the Cython modules. That would offer substantial performance improvements for build times. Furthermore, having a proper CMake build system allows for IDE integrations to analyze the library and make suggestions while developing the C++ code. That ease of development is the primary reason I decided to start this PR. Someday it may be worth exploring having a true C++ API for freud, but that's not yet something I am willing to consider stable.

## How Has This Been Tested?
*This will need to be thoroughly tested on multiple platforms before deployment to PyPI or conda-forge.*

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds or improves functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation improvement (updates to user guides, docstrings, or developer docs)

## Checklist:
- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.
- [ ] I have updated the documentation (if relevant).
- [ ] I have added tests that cover my changes (if relevant).
- [ ] All new and existing tests passed.
- [ ] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst).
- [ ] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

Issue #629, opened 2020-06-12
Introduce pipeline for undefined behavior in C++


klywang:
**Is your feature request related to a problem? Please describe.**
Opening up discussion to appropriate tools to detect bugs such as memory leaks and other bugs the C++ compilter may miss.

**Describe the solution you'd like**
A static analyzer that can inspect python/cython code without throwing false positives.

**Describe alternatives you've considered**
Potential debuggers include compiler sanitizers, **cppcheck**, and **clang-tidy**.

**Additional context**
Discussed in issue #626. Creating a seperate issue for further evaulation of viable tools that may exist now or emerge in the future.

PR #668, opened 2020-10-10
Feature/enable skbuild

vyasr:
## Description

Use scikit-build and CMake to build the C++ and Cython components of freud.

## Motivation and Context

This pull request dramatically speeds up builds of freud (takes < 30 seconds on my Mac using Ninja). Changing to a proper build system should also dramatically simplify addressing issues like #464 and #629. It will also enable easier integration with IDEs for the C++ code, and long term it will enable exposing a C++ API for freud.

Outstanding tasks:

- [x] Enable automatic download of submodules.
- [x] Contact voro++ maintainer to discuss modifications that would allow using it as a library (rather than having to handle compiling its sources ourselves).
- [x] Update CI scripts to work with new build system.
- [x] Update and test deployment (to Test PyPI).
- [x] Update development requirements.
- [x] Update documentation on building freud.

**Update** -- The comments below are mostly outdated now due to later changes to the CMake configuration, but I'll leave them here so that we can track the discussions in the future if needed.

This PR supersedes #661. I glanced through that PR, and I note that this implementation is different: rather than compiling a single shared object for the entire C++ library, I am instead following a more distributed model where each Cython module is compiled against the corresponding C++ code. We can change this in the future if we move towards a more thorough separation of the C++ code and exposing those APIs, but for now I'd prefer to retain this model since it's cleaner and generates leaner builds.

@joaander as a more experienced CMake user, some specific questions for you (in addition to any review you can provide).
- What is a reasonable minimum version of CMake to use? I've looked through the changelog etc briefly, but I'd prefer a more informed opinion.
- If you look at `freud/CMakelists.txt`, I'm forced to link the `_util` Object library directly to a couple of Cython modules rather than to the corresponding C++ Object libraries (i.e. `_environment` instead of `environment`). I _think_ that the reason is that the relevant C++ util code (the `diagonalize.cc` source file) is only included by the source files of the corresponding C++ modules, not the headers. As a result, I've marked the corresponding target includes (for instance, in `cpp/environment/CMakeLists.txt`) as PRIVATE; however, I note that changing this to PUBLIC doesn't fix the issue, so I'm not certain this is correct. It may also be some limitation with Object library includes propagation that I'm not finding in the documentation. Any thoughts?
- I wrote a simple FindTBB.cmake script, but I know that we discussed the more sophisticated version along with the config file that you have on HOOMD's next branch. Would you recommend I try to copy in that file and the corresponding macro?

## How Has This Been Tested?

Existing builds on my machine are functional. Additional testing will come in the form of the to-do list above.

**Update** - Building and testing on CI now works on all systems. The wheel building process has also been manually verified via push to test PyPI.

## Types of changes

## Checklist:


CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+cmake_minimum_required(VERSION 3.10)
```
bdice: I've seen recommendations to use at least 3.14 for "modern" features. I used 3.16 in #661.

cpp/environment/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    AngularSeparation.h
+    AngularSeparation.cc
+    BondOrder.h
+    BondOrder.cc
+    LocalBondProjection.h
+    LocalBondProjection.cc
+    LocalDescriptors.h
+    LocalDescriptors.cc
+    MatchEnv.h
+    MatchEnv.cc
```
bdice: Missing `Registration.h`? Or is that not necessary?
```suggestion
    MatchEnv.cc
    Registration.h
```

cpp/locality/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+    NeighborPerPointIterator.h
+    NeighborQuery.cc
+    NeighborQuery.h
+    PeriodicBuffer.cc
+    PeriodicBuffer.h
+    RawPoints.h
+    Voronoi.cc
+    Voronoi.h
+
+    # For now, compile voro++ object in directly.
```
bdice: There are CMake configuration files for voro++. Do they need to be edited, or just called correctly?

setup.py
```diff
@@ -419,100 +13,53 @@ def find_tbb(tbb_root=None, tbb_include=None, tbb_link=None):
...
+    url='https://github.com/glotzerlab/freud',
+    download_url='https://pypi.org/project/freud-analysis/',
+    project_urls={
+        'Homepage': 'https://github.com/glotzerlab/freud',
+        'Documentation': 'https://freud.readthedocs.io/',
+        'Source Code': 'https://github.com/glotzerlab/freud',
+        'Issue Tracker': 'https://github.com/glotzerlab/freud/issues',
+    },
+    python_requires='>=3.6',
+    install_requires=[
```
bdice, line 52: Does this need to list scikit-build and cmake? Not sure how these interact with pyproject.toml.

bdice (commented):
Cool, @vyasr. This will help the library a lot. I have a few questions you might be able to answer.

bdice:
@vyasr Also I didn't understand your reasoning for the distributed approach: "it's cleaner and generates leaner builds" -- how so? I would have thought the other approach which compiles one shared object for the whole C++ library and then just links it to the Cython modules would have been cleaner/leaner. Happy to discuss on Slack/call if it's a long explanation.

cpp/environment/CMakeLists.txt
vyasr: Good catch. It's not strictly necessary since it's header only and the whole cpp/environment is included, but better to be explicit here.

cpp/locality/CMakeLists.txt
vyasr: They need to be edited. The current voro++ CMake files are not configured properly for building as a library in another code base, see the to-do list in the PR description.

setup.py
vyasr: No, the `install_requires` is to determine dependencies of the code, whereas pyproject.toml is for dependencies of the build process. The requirements in pyproject.toml indicate what needs to be installed to even run setup.py.

CMakeLists.txt
vyasr: Right, on the flip side I don't know how onerous that is. 3.14 did only come out last year, I assume that for instance Ubuntu or Arch package managers for anything other than the newest OS versions might not have a sufficiently new CMake. Just want to make sure we're not causing problems for installation on e.g. any of our clusters like Comet.

vyasr: To be clear: you are absolutely right that I am using features that require 3.14 (all the `target_includes` that I'm doing don't exist in 3.10 AFAIK). I just want to make sure that won't be a problem for our HPC user base. HOOMD is still requiring only 3.9, although I don't know if that's just a legacy thing or planned to change for the v3 release.

vyasr:
Regarding the leaner/cleaner argument, I should clarify. My primary objection is to having one monolithic CMakeLists file. That's poor CMake style, harder to read, and more difficult to integrate into complex builds. It also makes it less likely that you'll properly isolate build requirements for different pieces of the library. It's much easier to maintain if we have separate CMakeLists for each module.

As far as the result of separate libraries being leaner, it's primarily a product of the fact that building the underscore Cython modules requires pulling a lot of the cpp header files and results in duplication. Also, it adds an extra file that needs to be installed with the build (the libfreud.so library). My usage of object libraries in CMake means that the only actual library files created are the ones required for Cython. However, I don't think either any of this really matters: the difference in total filesize is very small, the compile times will be almost identical (all the compilation is identical), and the extra file is necessary anyway if we ever want to expose a libfreud module. That being the case, I can go ahead update `cpp/CMakeLists` to build a freud C++ library now. The object file methods make it easy to mostly get the best of both worlds (separate CMakeLists files while generating a single large library).

bdice:
@vyasr That makes a lot more sense. I support your argument for many CMake list files. I do think separating “libfreud” as a C++ library would be a positive change. That should also simplify TBB linkage, I think? (The Cython modules should just link to libfreud, which links to TBB.)

vyasr:
No, unless you mean in the context of the CMake configuration it doesn't really change linking. You'll still need to link all the Cython libraries against TBB as well, shared library linking isn't transitive in that way. CMake handles this appropriately though.

CMakeLists.txt
joaander: HOOMD v2.x builds with CMake 2.8.10.1 or newer. HOOMD v3.x will bump that requirement to 3.9. 3.9 was the first version to have a functioning findMPI that uses modern practices. I suggest supporting the oldest version you can without needing to implement excessive workarounds. This makes the software accessible to more users, many of whom do not update their software stacks that regularly.

CMake/UpdateSubmodules.cmake
```diff
@@ -0,0 +1,14 @@
+find_package(Git)
+if (GIT_FOUND)
+	option(UPDATE_SUBMODULES "Download missing submodules during build" ON)
+	if (UPDATE_SUBMODULES )
+		message("Updating submodules.")
+		execute_process(COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
```
joaander: I removed this from HOOMD v3. I found it caused more problems than it solved. Users and developers should manage their git repositories themselves.

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+# Set a default build type if none was specified
+if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+    message(STATUS "Setting build type to '${DEFAULT_BUILD_TYPE}' since none was specified.")
+    set(CMAKE_BUILD_TYPE "${DEFAULT_BUILD_TYPE}" CACHE
+        STRING "Choose the type of build." FORCE)
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+endif()
+
+set(CMAKE_CXX_STANDARD 11)
```
joaander: Can you bump this to 14? TBB will be requiring C++14 soon.

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
joaander, line 29: This shouldn't be necessary. CMake should use the right compile flags depending on the type of library being built.

cpp/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+target_link_libraries(
+    libfreud
+    PUBLIC TBB
+    )
+
+# Set the name of the library to just freud.
+set_target_properties(
+    libfreud
+    PROPERTIES OUTPUT_NAME freud
```
joaander: Why rename this. It doesn't appear to be an importable python module.

joaander (commented):
Looking good. I left a few comments and questions.

Regarding `_util` I'm not sure what the problem is. As you say, maybe it is some limitation with object libraries. Since you are building everything into one `libfreud` target, you could instead make the child CMakeList.txt files set a list of source files in a variable and then have the `add_library` compile all those source files together. That would work around any limitations with `OBJECT` and wouldn't result in any files being compiled more than once.

Regarding FindTBB, feel free to use the solution from HOOMD to support finding modern TBB's by their embedded config file (the `find_package_config_first` macro) - that way you don't need to rely on your own `FindTBB.cmake` except on systems like Ubuntu which doesn't provide the TBB config or on platforms with older TBB installations. The `FindTBB` from HOOMD determines the `TBB_VERSION` and implements a few compiler workarounds that may not be needed for freud. The only other main difference is that the target is `TBB::tbb` for compatibility with the TBB provided config files.

vyasr:
@jglaser might also be interested in this.

CMakeLists.txt
bdice: CMake version pi.0 (3.14.0) was [announced on Pi Day 2019](https://blog.kitware.com/cmake-3-14-0-available-for-download/). I'm fine with matching HOOMD's requirement of CMake 3.9 as long as it doesn't require significant workarounds. We provide conda-forge builds and/or wheels for freud on multiple platforms and architectures, so far fewer users will build from source compared to HOOMD. Also since CMake is installable with pip or conda, I think it would be acceptable to have a version requirement 3.14 that is now >18 months old.

CMake/UpdateSubmodules.cmake
vyasr: Yes I recall you doing this. I'll probably replace this with an immediate error so that it's more obvious to the user why it's failing (rather than throwing a compile error that a file is missing). Could you elaborate on what issues you ran into with this approach?

CMakeLists.txt
vyasr: I don't see any reason that we can't. @bdice I assume you would be happy with these changes? That would also allow us to use at least some of the newer algorithms in the standard library :)

CMakeLists.txt
vyasr: Got it, I had issues with it early on but it might have been because I was misusing OBJECT and SHARED in my early iterations. It does indeed seem to work correctly on shared libraries without this.

CMake/UpdateSubmodules.cmake
bdice: I agree with @joaander - it gets clunky when changing submodules.

CMakeLists.txt
bdice:
```suggestion
set(CMAKE_CXX_STANDARD 14)
```

cpp/CMakeLists.txt
vyasr: It's not, however because it's a shared library and not a Python extension module CMake automatically prepends `lib` to the name. A `liblibfreud.so` seems very strange, and will also not be what people expect if we get around to exposing our C++ API. I don't want to name the actual library (in CMake) freud, though, because that conflicts with the project name and is confusing (not to mention I'm not 100% sure that CMake will always disambiguate correctly).

bdice: @vyasr I would recommend clarifying this in the comment. I found that out earlier while playing with CMake and forgot that would happen. Maybe update the comment to say:
```
# CMake prepends 'lib' to the shared library name.
# This sets the name of the library to libfreud instead of liblibfreud.
```

.gitignore
```diff
@@ -14,6 +14,9 @@ doc/build/*
 doc/_build/*
 freud/*.so
 freud/*.cpp
+freud/*.cxx
+freud/libfreud.dylib
```
bdice: Would prefer to match the catch-all style. We shouldn't ever commit a dylib file.
```suggestion
freud/*.dylib
```

CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+  set(Red         "${Esc}[31m")
+  set(DefaultColor "${Esc}[m")
+endif()
+set(submodules
+    Eigen
+    fsph
+    voro++)
+foreach (submodule ${submodules})
+    if( NOT EXISTS "${PROJECT_SOURCE_DIR}/extern/${submodule}/.git" )
+        message( FATAL_ERROR "${Red}The git submodules are not available. Please run git submodule update --init --recursive.${DefaultColor}"
```
bdice:
```suggestion
        message( FATAL_ERROR "${Red}The git submodule ${submodule} is not available. Please run git submodule update --init --recursive.${DefaultColor}"
```

freud/CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+        # Avoid deprecation warnings for unsupported NumPy API versions.
+        # See https://numpy.org/doc/1.19/reference/c-api/deprecations.html
+        PRIVATE "NPY_NO_DEPRECATED_API=NPY_1_10_API_VERSION"
+        # Default voro++ verbosity is high.
+        PRIVATE "VOROPP_VERBOSE=1"
+        )
+    if (COVERAGE)
+        target_compile_definitions(
+            ${cython_module}
+            #  Enable line tracing for coverage purposes if requested.
```
bdice:
```suggestion
            # Enable line tracing for coverage purposes if requested.
```

freud/CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+            PRIVATE ${PROJECT_SOURCE_DIR}/cpp/${cython_module}
+            )
+    endif()
+
+    install(TARGETS ${cython_module} DESTINATION freud)
+
+    # Coverage requires the Cython-compiled C++ files for line coverage.
+    if (COVERAGE)
+        install(FILES ${${cython_module}} DESTINATION freud)
+	endif()
```
bdice: Tabs/spaces?
```suggestion
    endif()
```

bdice (commented):
We will also need to update our installation guide and developer documentation.

CMakeLists.txt
vyasr: The loop will terminate as soon as any one submodule fails, so specifying names will be misleading.

cpp/CMakeLists.txt
vyasr: Good idea, I added more explicit documentation of this.

vyasr:
> We will also need to update our installation guide and developer documentation.

Yes, this is noted in the PR to-do list.

CMakeLists.txt
vyasr: With minor modifications we're able to support CMake >= 3.3. I modified CI to test against 3.6.3, which is the oldest version available on PyPI.

vyasr:
The Windows CI is currently failing because it appears that the available generators are not suitable for compiling the code. The scikit-build docs indicate that you should use the Visual Studio version used to build the Python distribution, but I'm not sure if it's actually that restrictive. @bdice I'd appreciate if you could give that a look, I have no experience with the Windows CI setup.

There is one significant (but IMO not disqualifying) regression currently associated with this PR. Due to some subtle issues with how these tools work together, Cython code coverage does not work with the new builds. I noticed the issue this morning and was able to identify the underlying issue by running the compilation manually and comparing Cython-generated C++ files, but I don't know how viable implementing a fix will be. I've posted a question to the Cython-users mailing list, we'll see what comes out of it.

@chr1shr indicated that he would be open to a PR improving the CMake configuration of voro++. In the interest of pushing progress forward on this PR, I would prefer to get this wrapped up and then submit a subsequent PR to change how we compile voro++. The current method is perfectly workable.

CMakeLists.txt
bdice: But isn't it useful to know _which_ submodule raised the error? 🤨 The same solution applies regardless, but it seems more helpful than not identifying the submodule name.

vyasr:
> Regarding `_util` I'm not sure what the problem is. As you say, maybe it is some limitation with object libraries. Since you are building everything into one `libfreud` target, you could instead make the child CMakeList.txt files set a list of source files in a variable and then have the `add_library` compile all those source files together. That would work around any limitations with `OBJECT` and wouldn't result in any files being compiled more than once.

Yes, this question has become less relevant since I originally created this PR due to the switch to compiling a single shared library for the C++ API. Additionally, the modifications I made to allow this compilation to work with older versions of CMake forced me to get rid of specifying libraries to link to Object libraries anyway, so most of the linking is now performed on the `libfreud` target anyway. 

> Regarding FindTBB, feel free to use the solution from HOOMD to support finding modern TBB's by their embedded config file (the `find_package_config_first` macro) - that way you don't need to rely on your own `FindTBB.cmake` except on systems like Ubuntu which doesn't provide the TBB config or on platforms with older TBB installations. The `FindTBB` from HOOMD determines the `TBB_VERSION` and implements a few compiler workarounds that may not be needed for freud. The only other main difference is that the target is `TBB::tbb` for compatibility with the TBB provided config files.

Sounds good, I'll take a look at that config and apply the relevant changes.

CMakeLists.txt
vyasr: I disagree, because almost nobody who's going to run into this issue in practice runs `git submodule update` on specific submodules. The most common error is not initializing submodules at all, and the people making that mistake are also the most likely to see this. Under those circumstances I think it's more confusing to see a single name, especially since the recommended solution doesn't specify a submodule. If you were to error on a specific submodule, the recommended solution should also be to update only that submodule.

bdice: We disagree but I don't feel strongly, so you win. 😉

CMake/UpdateSubmodules.cmake
joaander: As a developer, it is frustrating when you make a change to submodule, compile, and then that change is gone. It is also frustrating that the very slow `git submodule update --init` process occurs every time you compile. As a user who builds the code only once, it is a minor teaching moment to learn to use `git clone --recursive` or `git submodule update --init`.

cpp/CMakeLists.txt
joaander: Most old-style CMake scripts name the package in all caps to avoid the name conflict. Most new-style scripts use namespaces. You can name the target `freud::freud` and then the library should be named `libfreud` without needing `set_target_properties`

CMake/UpdateSubmodules.cmake
vyasr: Makes sense. The current build script just raises errors to indicate when submodules are not present, which I think is a useful compromise.

cpp/CMakeLists.txt
vyasr: It looks like that namespacing is not supported for targets that must be built, only for imported ones: https://gitlab.kitware.com/cmake/cmake/-/issues/20444. So it works for TBB, but not for freud.

cpp/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+add_subdirectory(density)
+add_subdirectory(environment)
+add_subdirectory(locality)
+add_subdirectory(order)
+add_subdirectory(parallel)
+add_subdirectory(pmft)
+add_subdirectory(util)
+
+add_library(
+    libfreud
```
tommy-waltmann: I believe this will create a library called liblibfreud.so, was the variable named knowing the library would be named like this?

cpp/cluster/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_library(
+    _cluster
+    OBJECT
+    Cluster.h
+    Cluster.cc
+    ClusterProperties.h
+    ClusterProperties.cc
```
tommy-waltmann: I don't think the headers are necessary here, or in other places like this.

cpp/CMakeLists.txt
bdice: @tommy-waltmann This is intentional, it's for making it clear that this is the C++ library. The name is fixed later, so it ends up as "freud": https://github.com/glotzerlab/freud/blob/0618c150441e27a7dfe4cf4cb37174d9da188e45/cpp/CMakeLists.txt#L47

cpp/cluster/CMakeLists.txt
bdice: I didn't think so either, but I wasn't sure. @vyasr can you explain?

joaander: Adding headers to a target is a convenience for IDE project targets in CMake - it makes the headers appear in the project. I don't know if it also factors into the compile_command.json output.

vyasr: Yes exactly, some IDEs consume the CMakeLists.txt files directly. I'm not sure about compile_commands, but I think it's basically an issue of inconsistency between development environments where some will read the cmake file whereas others will look at JSON. I _think_ the compile_commands.json is necessary for semantic completion in certain contexts.

vyasr:
I think this PR should be good to go now. I merged in the PR adding Windows support and augmented the docs a little bit. @bdice to address one of the issues that you ran into when enabling Windows support (requiring newer CMake versions) I added a note to our installation documentation indicating that this might be the case and pointing to the relevant CMake generator documentation.

bdice:
For some reason, the ReadTheDocs build is failing because it's uninstalling scikit-build.
https://readthedocs.org/projects/freud/builds/12216572/

```
$ conda install --yes --quiet --name 668 mock pillow sphinx sphinx_rtd_theme

[...] truncated [...]

The following packages will be REMOVED:

  jupyter_sphinx-0.3.1-py39hde42818_1
  pybtex-0.23.0-py39hde42818_0
  pybtex-docutils-0.2.2-py39hde42818_2
  python_abi-3.9-1_cp39
  scikit-build-0.11.1-py39hf149a3a_2
  sphinxcontrib-bibtex-1.0.0-py_0
```

bdice:
I fixed the first problem with ReadTheDocs builds by swapping channel priority (`conda-forge` is now secondary). Now it's unable to find TBB. https://readthedocs.org/projects/freud/builds/12217822/
```
[42/79] Linking CXX shared library cpp/libfreud.so
FAILED: cpp/libfreud.so 
: && /usr/bin/c++ -fPIC -O3 -DNDEBUG   -shared -Wl,-soname,libfreud.so -o cpp/libfreud.so cpp/cluster/CMakeFiles/_cluster.dir/Cluster.cc.o cpp/cluster/CMakeFiles/_cluster.dir/ClusterProperties.cc.o cpp/density/CMakeFiles/_density.dir/CorrelationFunction.cc.o cpp/density/CMakeFiles/_density.dir/GaussianDensity.cc.o cpp/density/CMakeFiles/_density.dir/LocalDensity.cc.o cpp/density/CMakeFiles/_density.dir/RDF.cc.o cpp/density/CMakeFiles/_density.dir/SphereVoxelization.cc.o cpp/environment/CMakeFiles/_environment.dir/AngularSeparation.cc.o cpp/environment/CMakeFiles/_environment.dir/BondOrder.cc.o cpp/environment/CMakeFiles/_environment.dir/LocalBondProjection.cc.o cpp/environment/CMakeFiles/_environment.dir/LocalDescriptors.cc.o cpp/environment/CMakeFiles/_environment.dir/MatchEnv.cc.o cpp/locality/CMakeFiles/_locality.dir/AABBQuery.cc.o cpp/locality/CMakeFiles/_locality.dir/LinkCell.cc.o cpp/locality/CMakeFiles/_locality.dir/NeighborComputeFunctional.cc.o cpp/locality/CMakeFiles/_locality.dir/NeighborList.cc.o cpp/locality/CMakeFiles/_locality.dir/PeriodicBuffer.cc.o cpp/locality/CMakeFiles/_locality.dir/Voronoi.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/cell.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/common.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/container.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/unitcell.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/v_compute.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/c_loops.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/v_base.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/wall.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/pre_container.cc.o cpp/locality/CMakeFiles/_locality.dir/__/__/extern/voro++/src/container_prd.cc.o cpp/order/CMakeFiles/_order.dir/Cubatic.cc.o cpp/order/CMakeFiles/_order.dir/HexaticTranslational.cc.o cpp/order/CMakeFiles/_order.dir/Nematic.cc.o cpp/order/CMakeFiles/_order.dir/RotationalAutocorrelation.cc.o cpp/order/CMakeFiles/_order.dir/SolidLiquid.cc.o cpp/order/CMakeFiles/_order.dir/Steinhardt.cc.o cpp/order/CMakeFiles/_order.dir/Wigner3j.cc.o cpp/parallel/CMakeFiles/_parallel.dir/tbb_config.cc.o cpp/pmft/CMakeFiles/_pmft.dir/PMFTR12.cc.o cpp/pmft/CMakeFiles/_pmft.dir/PMFTXY.cc.o cpp/pmft/CMakeFiles/_pmft.dir/PMFTXYT.cc.o cpp/pmft/CMakeFiles/_pmft.dir/PMFTXYZ.cc.o cpp/util/CMakeFiles/_util.dir/diagonalize.cc.o  -Wl,-rpath,/home/docs/checkouts/readthedocs.org/user_builds/freud/conda/668/lib:  /home/docs/checkouts/readthedocs.org/user_builds/freud/conda/668/lib/libtbb.so && :
/usr/bin/ld: cannot find libtbb.so.2
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
  File "/home/docs/checkouts/readthedocs.org/user_builds/freud/conda/668/lib/python3.8/site-packages/skbuild/setuptools_wrap.py", line 589, in setup
    cmkr.make(make_args, env=env)
  File "/home/docs/checkouts/readthedocs.org/user_builds/freud/conda/668/lib/python3.8/site-packages/skbuild/cmaker.py", line 496, in make
    raise SKBuildError(
```

vyasr:
I don't know offhand what's going on there. It _is_ finding TBB, CMake prints the location and it's being provided as an additional target for the compiler on that invocation of `c++`. What it's not finding is `libtbb.so.2`, which I'm assuming is an artifact of this particular build of TBB provided by conda where `libtbb.so` needs `libtbb.so.2`. It seems like the way that RTD uses the conda environment is to manually provide all the absolute paths rather than to activate the environment (based on looking at the build link you pasted I never see a `conda activate`) so things that aren't explicitly added aren't found. I assume we don't have any control at that level of the build?

Actually I take that back. It looks like the conda library directory is being added to the rpath. Then I'm not sure why it wouldn't find this file, unless it's actually not present.

tests/integration/test_reader_integrations.py
```diff
@@ -5,6 +5,12 @@
 import sys
 import unittest
 
+try:
+    import MDAnalysis
+    MDANALYSIS = True
+except ImportError:
+    MDANALYSIS = False
+
```
tommy-waltmann, line 14: Why is this, and the other logic changes in this file, a part of this PR?

ChangeLog.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to
 
 ### Changed
 * NeighborList raises a `ValueError` instead of a `RuntimeError` if provided invalid constructor arguments.
+* freud now builds using scikit-build.
```
```suggestion
* freud now builds using scikit-build (requires CMake).
```

tests/integration/test_reader_integrations.py
vyasr: MDAnalysis does not provide wheels, and it's not compatible with the newest clang versions, so compiling will fail outright. This causes problems for our wheel builds because we run tests on CI that depend on MDAnalysis (the integration tests). The newest clang releases happened after our last release of freud, and we only build with clang when we build OS X wheels at release, so we simply hadn't run into this problem yet. Since I identified the issues when modifying the wheel building process to work with scikit-build, I had to make these changes to test the whole thing. Ideally it wouldn't be here, but even if I did it in a separate PR it would immediately need to be merged here, so I just left it as part of this PR.

doc/source/gettingstarted/installation.rst
```diff
@@ -64,47 +81,37 @@ You can also build **freud** in place so that you can run from within the folder
 Building **freud** in place has certain advantages, since it does not affect your Python behavior except within the **freud** directory itself (where **freud** can be imported after building).
 Additionally, due to limitations inherent to the distutils/setuptools infrastructure, building extension modules can only be parallelized using the build_ext subcommand of setup.py, not with install.
 As a result, it will be faster to manually run build_ext and then install (which normally calls build_ext under the hood anyway) the built packages.
-In general, the following options are available for setup.py in addition to the standard setuptools options (notes are included to indicate which options are only available for specific subcommands such as build_ext):
+
+CMake Options
++++++++++++++
+
+The scikit-built tool allows setup.py to accept three different sets of options separated by ``--``, where each set is provided directly to scikit-build, to CMake, or to the code generator of choice, respectively.
```
bdice:
```suggestion
The scikit-build tool allows setup.py to accept three different sets of options separated by ``--``, where each set is provided directly to scikit-build, to CMake, or to the code generator of choice, respectively.
```

doc/source/gettingstarted/installation.rst
```diff
@@ -64,47 +81,37 @@ You can also build **freud** in place so that you can run from within the folder
...
+For example, the command ``python setup.py build_ext --inplace -- -DCOVERAGE=ON -G Ninja -- -j 4`` tell scikit-build to perform an in-place build, it tells CMake to turn on the ``COVERAGE`` option and use Ninja for compilation, and it tells Ninja to compile with 4 parallel threads.
+For more information on these options, see the `scikit-build docs <scikit-build.readthedocs.io/>`__.
+
+In addition to standard CMake flags, the following CMake options are available for **freud**:
 
 .. glossary::
 
-    -\\-PRINT-WARNINGS
-      Specify whether or not to print compilation warnings resulting from the build even if the build succeeds with no errors.
+    --COVERAGE
```
bdice: This may need a `\\`, I noticed this pattern below.
```suggestion
    -\\-COVERAGE
```

doc/source/reference/development/howtoadd.rst
```diff
@@ -54,7 +54,7 @@ To keep modules well-organized, **freud** implements the following structure:
 - Python code is stored in the ``freud`` folder at the root of the repository.
 - C++ code is exposed to Python using Cython code contained in pxd files with the following convention: ``freud/_MODULENAME.pxd`` (note the preceding underscore).
 - The core Cython code for modules is contained in ``freud/MODULENAME.pyx`` (no underscore).
-- Generated Cython C++ code (e.g. ``freud/MODULENAME.cpp``) should not be committed during development. These files are updated only at release time.
+- Generated Cython C++ code (e.g. ``freud/MODULENAME.cpp``) should not be committed during development. These files are generated using Cython when building from source, and are unnecessary when installing compiled binaries.
```
bdice: Are these actually named `*.cxx` now? I noticed that in the `.gitignore`.
```suggestion
- Generated Cython C++ code (e.g. ``freud/MODULENAME.cxx``) should not be committed during development. These files are generated using Cython when building from source, and are unnecessary when installing compiled binaries.
```

doc/source/reference/development/howtoadd.rst
```diff
@@ -84,8 +84,13 @@ Steps for Adding New Code
 Once you've determined to add new code to **freud**, the first step is to create a new branch off of :code:`master`.
 The process of adding code differs based on whether or not you are editing an existing module in **freud**.
 Adding new methods to an existing module in **freud** requires creating the new C++ files in the ``cpp`` directory, modifying the corresponding ``_MODULENAME.pxd`` file in the ``freud`` directory, and creating a wrapper class in ``freud/MODULENAME.pyx``.
-If the new methods belong in a new module, you must create the corresponding ``pxd`` and ``pyx`` files accordingly.
-In addition, you will need to import the new module in ``freud/__init__.py`` by adding :code:`from . import MODULENAME` so that your module is imported by default.
+If the new methods belong in a new module, you must create the corresponding ``cpp`` directory and the ``pxd`` and ``pyx`` files accordingly.
+
+In order for code to compile, it must be added to the relevant ``CMakeLists.txt`` file.
+New C++ files for existing modules must be added to the corresponding ``cpp/MODULENAME/CmakeLists.txt`` file.
```
bdice:
```suggestion
New C++ files for existing modules must be added to the corresponding ``cpp/MODULENAME/CMakeLists.txt`` file.
```

doc/source/reference/development/howtoadd.rst
```diff
@@ -84,8 +84,13 @@ Steps for Adding New Code
...
+For new modules, a ``cpp/NEWMODULENAME/CMakeLists.txt`` file must be created, and in addition the new module must be added to the ``cpp/CMakeLists`` file in the form of both an ``add_subdirectory`` command and addition to the ``libfreud`` library in the form of an additional source in the ``add_library`` command.
```
bdice:
```suggestion
For new modules, a ``cpp/NEWMODULENAME/CMakeLists.txt`` file must be created, and in addition the new module must be added to the ``cpp/CMakeLists.txt`` file in the form of both an ``add_subdirectory`` command and addition to the ``libfreud`` library in the form of an additional source in the ``add_library`` command.
```

vyasr:
OK I'm very confused now. As far as I can tell the tbb libraries on conda-forge right now are missing libtbb.so.2, but libtbb.so always points to it. I'm not sure how this works at all, but I'm triggering a rebuild of master on RTD to make sure this isn't some recent change in the tbb libraries. It doesn't seem like it (looking backward to older tbb versions), but I would like to verify.

freud/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+    else()
+        set_target_properties(${cython_module} PROPERTIES INSTALL_RPATH "$ORIGIN")
+    endif()
+endforeach(cython_module)
+
+# The SolidLiquid class has an instance of cluster::Cluster as a member, so
+# including the header requires the Cluster.h header. Would prefer to inherit
+# this information from the _order library, but that's not possible since we're
+# linking to libfreud.
+target_include_directories(
```
bdice: Is there a way to make each module include the "include directories" in `${PROJECT_SOURCE_DIR}/cpp/${cython_module}/CMakeLists.txt`?

bdice (changes requested):
Looks good overall! We just need to fix a few minor issues and make ReadTheDocs work.

vyasr:
Ah OK I found it. The real library is installed by the `tbb` package on conda-forge. My guess is that the old Cython install was finding all of this, but the new install finds `libtbb.so` and thinks it's good enough. I'll need to improve the TBB finding so that it uses the config when available, that probably will fix it. I'll try to get around to this later today.

doc/source/gettingstarted/installation.rst
vyasr: Actually that pattern is wrong, I'm not sure why it got introduced. It renders a backslash in between the two dashes, [see the docs right now](https://freud.readthedocs.io/en/stable/gettingstarted/installation.

freud/CMakeLists.txt
vyasr: Since we're linking against libfreud, I think any (not super convoluted) attempt to use the CMakeLists files in the cpp folders will lead to multiple builds of the same objects and linker errors. I don't think you can selectively include parts of another CMakeLists.txt file.

doc/source/reference/development/howtoadd.rst
vyasr: Yep that's the default output now.

freud/CMakeLists.txt
bdice: That sounds fine to me, then. Thanks for the answer.

vyasr:
Actually hold off on calling this completely done. I had intended to format the cmake files using a standard linter and add that to CI. I had found something decent and will apply it before merging this pr.

doc/source/reference/development/howtoadd.rst
bdice: @vyasr Did you mean to accept my suggestion? You seemed to agree with my comment, which I think would indicate that we need to change `cpp` to `cxx` in this line about generated code, but you resolved the thread without applying the suggestion.

doc/source/gettingstarted/installation.rst
bdice: I fixed it.

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+# This setting is required when building shared libraries from object
+# libraries. CMake intentionally chooses not to infer such information, so we
+# should expect to specify this.
+# https://stackoverflow.com/questions/50600708/combining-cmake-object-libraries-with-shared-libraries
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/CMake")
+include(${CMAKE_CURRENT_LIST_DIR}/CMake/freud-macros.cmake)
```
tommy-waltmann: I think it would be more modular to call `add_subdirectory(CMake)` here, and then in that directory's CMakeLists have these two lines.

vyasr: Good call, I'll make that change.

doc/source/reference/development/howtoadd.rst
vyasr: Yes, I must have missed clicking it when I went through to batch the suggestions.

vyasr:
I applied a reformatting. I'll make a separate PR to make this part of CI etc so that we can get this merged and in production.

vyasr:
RTD builds are back to failing, but now it's because the build has upgraded to using Python 3.9, scipy wheels aren't available yet, and compiling scipy fails. I think we can probably just go ahead and merge this PR at this point? This problem won't be resolved until scipy 3.9 wheels are available.

doc/source/gettingstarted/installation.rst
bdice: For posterity, this is the escaping behavior for smart double-dashes in docutils. https://docutils.sourceforge.io/docs/user/smartquotes.html#id5

bdice (approved):
Looks good to me. Merge when you're ready. Thanks for the scikit- _boooo_ ild
spooky parrot

vyasr:
@tommy-waltmann in the interest of keeping things moving along, I went ahead and merged. If you have further comments though, please do open another PR (or just an issue) with proposed changes!

PR #669, opened 2020-10-13
Enable scikit-build on Windows

bdice:
## Description
Windows compatibility for scikit-build. Follows up on #668.
- Uses HOOMD's `FindTBB.cmake`.
- Handles MSVC symbol exports. MSVC requires explicit declarations of symbols that need to be exported, with `__declspec(dllexport)`. When we compile `libfreud`, we can tell CMake to automatically export all symbols. This lifts most of the burden for us.
- Fixes some problems where CMake wasn't working as expected. In particular, `target_link_libraries` didn't recognize `TBB` and had to be given `${TBB_LIBRARY}` instead.
- Some symbols, specifically static class members like `NeighborQuery::DEFAULT_NUM_NEIGHBORS` or `NeighborQueryIterator::ITERATOR_TERMINATOR` are not exported by CMake's setting and we must do that manually. The solution I found was to define `__declspec(dllexport)` for all pertinent classes and separate all declarations into `*.h` and implementations into `*.cc`.
- The `msd` module must be linked against `libfreud`, since it uses the `parallel` feature to determine how many threads to use. This seems similar to the configuration for `interface` requiring `locality`. _Update from discussion with @vyasr: All modules need to be linked against libfreud._
- Installs CMake 3.14 on Windows only, since it's the oldest version that supports Visual Studio 2019 (which is installed on this CI image and is used for builds).

## Motivation and Context
Windows builds are useful for integration with OVITO. Also it's our second most popular build after Linux:
image
(data collected for freud 2.3.0 only, on Oct. 13, 2020)

## How Has This Been Tested?
Builds eventually passed, after strenuous efforts.

## Types of changes

## Checklist:

### DIFF
--- a/CMake/FindTBB.cmake
+++ b/CMake/FindTBB.cmake
@@ -1,10 +1,27 @@
-find_path(TBB_INCLUDE_DIR tbb/tbb.h)
+find_library(TBB_LIBRARY tbb
+             HINTS ENV TBB_LINK)
 
-find_library(
-    TBB_LIBRARY tbb
-    HINTS ${TBB_INCLUDE_DIR}/../lib )
+get_filename_component(_tbb_lib_dir ${TBB_LIBRARY} DIRECTORY)
 
-add_library(TBB SHARED IMPORTED)
-set_target_properties(TBB PROPERTIES
-	IMPORTED_LOCATION "${TBB_LIBRARY}"
-	INTERFACE_INCLUDE_DIRECTORIES "${TBB_INCLUDE_DIR}")
+find_path(TBB_INCLUDE_DIR tbb/tbb.h
+          HINTS ENV TBB_INC
+          HINTS ${_tbb_lib_dir}/../include)
+
+if(TBB_INCLUDE_DIR AND EXISTS "${TBB_INCLUDE_DIR}/tbb/tbb_stddef.h")
+    file(STRINGS "${TBB_INCLUDE_DIR}/tbb/tbb_stddef.h" TBB_H REGEX "^#define TBB_VERSION_.*$")
+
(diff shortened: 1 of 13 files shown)


freud/locality.pyx
```diff
@@ -183,6 +182,8 @@ cdef class NeighborQueryResult:
 
     def __iter__(self):
         cdef freud._locality.NeighborBond npoint
+        cdef freud._locality.NeighborBond ITERATOR_TERMINATOR
+        ITERATOR_TERMINATOR = freud._locality.NeighborBond(-1, -1, 0.0, 1.0)
```
bdice, line 186: I couldn't figure out how to export the static class member and import it in Cython. Windows rules around linking and exported symbols make this really tough. This is probably too complex for us to worry about, since the workaround is simple and it's only needed in one place in the Cython code.

CMake/FindTBB.cmake
```diff
@@ -1,10 +1,27 @@
-find_path(TBB_INCLUDE_DIR tbb/tbb.h)
```
bdice, line 1: This is a direct copy of HOOMD's `FindTBB.cmake`. I wasn't able to make the previous version work.

CMakeLists.txt
```diff
@@ -41,6 +41,17 @@ foreach (submodule ${submodules})
     endif()
 endforeach(submodule)
 
+# Define preprocessor directives for Windows
+if (WIN32)
+    # Export all symbols (forces creation of .lib file)
```
bdice:
```suggestion
    # Export all symbols (forces creation of .def file)
```

bdice:
Replaced by #676.

PR #671, opened 2020-10-15
Feature/expose steinhardt quantities

### LINKED ISSUE #627: Expose more Steinhardt quantities
A [discussion on the mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/freud-users/q0LP_J_veLI) suggested that the qlm array computed in SolidLiquid (and consequently, in Steinhardt) could be useful to expose to the Python API.

We need to decide which internal arrays could be useful. Once we've made that decision, since all the arrays are already using `ManagedArray` internally and have getter functions in C++ it should be trivial to update the Python API for Steinhardt. We will need to add getters to SolidLiquid so that it can retrieve the underlying Steinhardt arrays.

vyasr:
## Description

Expose the raw qlmi arrays in Steinhardt and SolidLiquid.

## Motivation and Context

Resolves: #627 

## How Has This Been Tested?

Added some new simple tests of base cases. More tests could be added, but I think the existing tests for the derived Steinhardt quantities are sufficient.

## Types of changes

## Checklist:


freud/order.pyx
```diff
@@ -611,6 +611,14 @@ cdef class Steinhardt(_PairCompute):
             &self.thisptr.getQl(),
             freud.util.arr_type_t.FLOAT)
 
+    @_Compute._computed_property
+    def particle_harmonics(self):
+        """:math:`\\left(N_{particles}, 2*l+1\\right)` :class:`numpy.ndarray`:
+        The raw array of \\overline{q}_{lm}(i)."""
```
bdice: This docstring needs to indicate the indexing convention. Is `m=0` at the beginning or the center?

freud/order.pyx
```diff
@@ -801,6 +809,14 @@ cdef class SolidLiquid(_PairCompute):
             &self.thisptr.getQlij(),
             freud.util.arr_type_t.FLOAT)
 
+    @_Compute._computed_property
+    def particle_harmonics(self):
+        """:math:`\\left(N_{particles}, 2*l+1\\right)` :class:`numpy.ndarray`:
+        The raw array of \\overline{q}_{lm}(i)."""
```
bdice: Indicate indexing convention.

tests/test_order_SolidLiquid.py
```diff
@@ -49,6 +50,30 @@ def test_identical_environments(self):
                 self.assertEqual(comp.cluster_sizes[0], len(positions))
                 npt.assert_array_equal(comp.num_connections, 12)
 
+    def test_l_zero(self):
```
bdice: Tests need to be added for one or more nontrivial cases that enforce the conventions of which angles are phi/theta and m-indexing. Especially tests that leverage both even and odd parity. I arbitrarily suggest something like `l=6, m=2,3` with 1 neighbor that is at a non-zero value of phi, theta.

Edit: these tests probably fit better in `Steinhardt` but could go in `SolidLiquid` too.

bdice (changes requested):
Requesting docstring clarification and more tests (they can be regression tests).

b-butler (commented):
I don't have any additional comments besides what bdice has already mentioned.

tests/test_order_SolidLiquid.py
vyasr: I just wrote a general test that validates the entire calculation. Seems more reliable.

vyasr:
Finally got around to finishing this up. Assuming tests pass this should be good to go.

@bdice should wheel building be triggered for _every_ PR now? It's no longer just on deployment? I didn't keep close track of the `cibuildwheel` changes, so I'm not sure, but something seems off here.

freud/order.pyx
```diff
@@ -625,6 +625,15 @@ cdef class Steinhardt(_PairCompute):
             &self.thisptr.getQl(),
             freud.util.arr_type_t.FLOAT)
 
+    @_Compute._computed_property
+    def particle_harmonics(self):
+        """:math:`\\left(N_{particles}, 2*l+1\\right)` :class:`numpy.ndarray`:
+        The raw array of \\overline{q}_{lm}(i). The array is provided in the
+        order given by fsph: :math:`0, 1, ..., l, -1, ..., -l`."""
```
```suggestion
        The raw array of \\overline{q}_{lm}(i). The array is provided in the
        order given by fsph: :math:`m = 0, 1, ..., l, -1, ..., -l`."""
```

freud/order.pyx
```diff
@@ -815,6 +824,15 @@ cdef class SolidLiquid(_PairCompute):
             &self.thisptr.getQlij(),
             freud.util.arr_type_t.FLOAT)
 
+    @_Compute._computed_property
+    def particle_harmonics(self):
+        """:math:`\\left(N_{particles}, 2*l+1\\right)` :class:`numpy.ndarray`:
+        The raw array of \\overline{q}_{lm}(i). The array is provided in the
+        order given by fsph: :math:`0, 1, ..., l, -1, ..., -l`."""
```
```suggestion
        The raw array of \\overline{q}_{lm}(i). The array is provided in the
        order given by fsph: :math:`m = 0, 1, ..., l, -1, ..., -l`."""
```

tests/test_order_Steinhardt.py
```diff
@@ -27,17 +27,49 @@ def test_shape(self):
...
-        N = 1000
-        L = 10
-
-        box, positions = freud.data.make_random_system(L, N)
-
-        comp = freud.order.Steinhardt(0)
-        comp.compute((box, positions), neighbors={"r_max": 1.5})
+    @pytest.mark.parametrize("sph_l", range(3, 8))
+    def test_l_nonzero(self, sph_l):
+        """Test th calculated qlmi."""
```
bdice:
```suggestion
        """Test the calculated qlmi."""
```

bdice (approved):
Why do we need to expose qlmi for SolidLiquid again? Approved pending some small changes.

bdice:
> @bdice should wheel building be triggered for every PR now? It's no longer just on deployment? I didn't keep close track of the cibuildwheel changes, so I'm not sure, but something seems off here.

@vyasr We do run cibuildwheel for everything now -- we may eventually remove our CircleCI tests and just run cibuildwheel as a means of testing. It's basically doing the same thing. Feel free to ping me if you have other thoughts; let's keep this PR discussion on topic so it can be merged.

tests/test_order_SolidLiquid.py
```diff
@@ -51,6 +52,36 @@ def test_identical_environments(self):
                 assert comp.cluster_sizes[0] == len(positions)
                 npt.assert_array_equal(comp.num_connections, 12)
 
+    def test_l_zero(self):
+        from scipy.special import sph_harm
```
b-butler: In another test function (for Steinhardt) we use `pytest.importorskip` but here we just import. Is there a reason we switch between the two.

tests/test_order_SolidLiquid.py
```diff
@@ -51,6 +52,36 @@ def test_identical_environments(self):
                 assert comp.cluster_sizes[0] == len(positions)
                 npt.assert_array_equal(comp.num_connections, 12)
 
+    def test_l_zero(self):
```
b-butler: Why is this here rather than the `Steinhardt` class test of spherical harmonic accuracy? I know they use the same underlying code, but I think having them in the same file would make sense from a like tests go together perspective.

b-butler (commented):
I only have a couple comment to address/reply to.

tests/test_order_SolidLiquid.py
vyasr: @b-butler I originally wrote the test in SolidLiquid and only tested the `l = 0` case. When Bradley requested a regression test for more realistic `l`, I copied the `l_zero` test you asked about into the Steinhardt tests and started writing the regression tests. However, I ended up instead just writing a more rigorous test actually recalculating the values using `scipy` for arbitrary `l`, which then superseded the `l = 0` test, so I removed that from Steinhardt and simply forgot to remove it from SolidLiquid. That's now removed, thanks.

tests/test_order_SolidLiquid.py
vyasr: No reason, this was older and is now removed.

vyasr:
@bdice we expose the `qlmi` from `SolidLiquid` because that's specifically what was originally requested (see [this thread on our Google group](https://groups.google.com/g/freud-users/c/q0LP_J_veLI/m/dImXMd8-AgAJ)). 

I didn't intend to discuss `cibuildwheel` any further on this thread, as long as what's currently happening is the expected behavior I'm fine moving forward for now.

PR #672, opened 2020-10-15
Document how to mask NaNs in Steinhardt.

vyasr:
## Description

Document how to handle NaNs in Steinhardt.

## Motivation and Context

Replaces #624

## How Has This Been Tested?

N/A

## Types of changes

## Checklist:

### DIFF
--- a/freud/order.pyx
+++ b/freud/order.pyx
@@ -531,6 +531,16 @@ cdef class Steinhardt(_PairCompute):
     :math:`q_{lm}` values over all particles before computing the order
     parameter of choice.
 
+    .. note::
+        The value of per-particle order parameter will be set to NaN for
+        particles with no neighbors. We choose this value rather than setting
+        the order parameter to 0 because in more complex order parameter
+        calculations (such as when computing the :math:`w_l`), it is possible
+        to observe a value of 0 for the per-particle order parameter even with
+        a finite number of neighbors. If you would like to ignore this
+        distinction, you can mask the output order parameter values using
+        NumPy: :code:`particle_order[numpy.isnan[particle_order]] = 0`.
+
     Args:
         l (unsigned int):
             Spherical harmonic quantum number l.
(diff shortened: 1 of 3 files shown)


freud/order.pyx
```diff
@@ -531,6 +531,16 @@ cdef class Steinhardt(_PairCompute):
...
 
+    .. note::
+        The value of per-particle order parameter will be set to NaN for
+        particles with no neighbors. We choose this value rather than setting
+        the order parameter to 0 because in more complex order parameter
+        calculations (such as when computing the :math:`w_l`), it is possible
+        to observe a value of 0 for the per-particle order parameter even with
+        a finite number of neighbors. If you would like to ignore this
+        distinction, you can mask the output order parameter values using
+        NumPy: :code:`particle_order[numpy.isnan[particle_order]] = 0`.
```
bdice: I recommend https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html:
```suggestion
        NumPy: :code:`numpy.nan_to_num(particle_order)`.
```

bdice (changes requested):
We also need to check if this affects Hexatic, Translational, SolidLiquid, LocalDescriptors, ...

freud/order.pyx
vyasr: Nice, I wasn't aware of this function. Just to be sure, there isn't any way that we could generate actual infinities, right? Otherwise we'd want to set those args appropriately as well.

bdice: @vyasr I don't think you can get infinity here. Regardless, I hope that pointing to the function is probably enough info for users to figure out the rest. I would like to link to the function in NumPy's docs instead of (or in addition to) providing a code snippet.

vyasr:
@bdice I agree that would be a good idea. Would you be OK with me opening a separate issue to track that and merging this PR as is? I've cherry-picked @mmartirossyan's test from her original PR to this one (with minor modifications), so the follow-up PR would basically be to review freud for other methods where this problem could arise and copy over the test to make sure the same behavior is consistent, as well as updating docs with the note and snippet added here.

bdice (approved):
Looks good! I created the follow-up issue.

Issue #424, opened 2019-09-04
Systematically test read-only arrays


vyasr:
**Is your feature request related to a problem? Please describe.**
In the past, read-only input arrays have caused failures in freud. We have fixed this issue, and have some tests randomly checking for this.

**Describe the solution you'd like**
We should add a systematic method for testing read-only arrays.

vyasr:
This hasn't been a problem in a long time. We can revisit adding tests if it appears again.

PR #674, opened 2020-10-16
Add inplace argument

### LINKED ISSUE #578: Reduce copying of input data
In certain cases box wrapping in freud may copy input data twice. This occurs when the call to `_convert_array` makes a copy because it has to convert dtypes, but then the array is immediately copied again afterward.

We should introduce an option to `_convert_array` that ensures copying no matter what, so that calling code never has to worry about this and the fewest number of copies possible are made.

Charlottez112:
Inplace argument added in box.wrap

## Description

Added inplace argument in both box.wrap and util._convert_array

## Motivation and Context

Resolves: Reduce copying of input data and add an option to operate on the data directly.

## How Has This Been Tested?

Added some unit tests in ```test_box_Box.py``` and ```test_util.py```, and passed these tests.

## Types of changes

## Checklist:


freud/box.pyx
```diff
@@ -301,9 +302,13 @@ cdef class Box:
         vecs = np.atleast_2d(vecs)
         vecs = freud.util._convert_array(vecs, shape=(None, 3)).copy()
```
vyasr: Notice that we're copying the input vectors here. That means that any change you make is still going to be to a copy of the input arrays. You need to change whether or not a copy is made based on the `inplace` argument.

Furthermore, the `_convert_array` function (which you can find in `freud/util.pyx`) may internally copy the array if it decides it has to (look at the call to `numpy.require` in that function). That means you also need to modify `_convert_array` to basically throw an error in the `inplace` case if the requirements aren't satisfied. I'm happy to give you more information on that offline if you'd like, or you can try to read the docs for these functions and try to understand them a bit more and come back with specific questions, up to you!

freud/box.pyx
```diff
@@ -301,9 +302,13 @@ cdef class Box:
...
-        cdef const float[:, ::1] l_points = vecs
-        cdef unsigned int Np = l_points.shape[0]
-        self.thisptr.wrap(<vec3[float]*> &l_points[0, 0], Np)
+        if inplace is False:
+            cdef const float[:, ::1] l_points = vecs
+            cdef unsigned int Np = l_points.shape[0]
+            self.thisptr.wrap(<vec3[float]*> &l_points[0, 0], Np)
+        else:
+            cdef unsigned int Np = vecs.shape[0]
+            self.thisptr.wrap(<vec3[float]*> &vecs[0, 0], Np)
```
bdice: 1. Write the "true" part first. Don't check `is False`, just let the value be interpreted as truthy/falsy.
2. I don't think you can declare `cdef const float[:, ::1]` conditionally. If I recall, Cython requires that all variable `cdef` declarations are executed regardless of the code path. This probably introduces the copy operation you were trying to avoid so I separated declaration and assignment.
```suggestion
        cdef const float[:, ::1] l_points
        if inplace:
            cdef unsigned int Np = vecs.shape[0]
            self.thisptr.wrap(<vec3[float]*> &vecs[0, 0], Np)
        else:
            l_points = vecs  # Need to make sure this copies exactly once
            cdef unsigned int Np = l_points.shape[0]
            self.thisptr.wrap(<vec3[float]*> &l_points[0, 0], Np)
```

freud/box.pyx
```diff
@@ -291,6 +291,7 @@ cdef class Box:
         Args:
             vecs (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`):
                 Unwrapped vector(s).
+            inplace
```
vyasr: You'll need to write a docstring describing this argument. If you're unsure how to document something that isn't a numpy array, have a look at the `is2D` argument to the box constructor.

bdice: Add parameter docs.
```suggestion
            inplace (bool):
                Whether to operate on vectors in place and overwrite input
                data (Default value: False).
```

freud/box.pyx
```diff
@@ -301,9 +302,13 @@ cdef class Box:
...
-        self.thisptr.wrap(<vec3[float]*> &l_points[0, 0], Np)
+        if inplace is False:
+            cdef const float[:, ::1] l_points = vecs
+            cdef unsigned int Np = l_points.shape[0]
+            self.thisptr.wrap(<vec3[float]*> &l_points[0, 0], Np)
+        else:
+            cdef unsigned int Np = vecs.shape[0]
+            self.thisptr.wrap(<vec3[float]*> &vecs[0, 0], Np)
 
         return np.squeeze(vecs) if flatten else vecs
```
vyasr: If the calculation is done `inplace`, then you don't need to return anything; the user's expectation then is to use the array provided as an argument.

freud/box.pyx
Charlottez112: Yeah that makes sense. I didn't read that carefully and thought vecs = freud.util._convert_array(). Let me try to fix that

bdice (changes requested):
Thanks for the PR @Charlottez112! I added some comments.

We supported "in place" operations in a previous version of freud, but I am hesitant to re-introduce in-place operations. I think parameters like this introduce API confusion and inconsistency with _every other method_ that could possibly operate in place but hasn't been enabled. It means we have to double (or at least augment) our testing code, worry about whether data is read-only (we just closed #424), keep more documentation up to date, etc. and the benefits are small (avoiding one extra copy of data). I'm open to others' thoughts on this but as a maintainer I am unsure about reneging on our changes in the freud v2 design.

vyasr (changes requested):
Thanks for the contribution @Charlottez112! You'll want to link this PR to #578, although depending on exactly what functions eventually get modified by this PR it may not be sufficient to finish all of #578.

The reason for the current failure of the tests is an unfortunate idiosyncrasy of Cython. You have put a `cdef` inside an `if` statement, which is not allowed in Cython (happy to chat more about why offline if you'd like an extended explanation). The appropriate fix is to replace code that looks like this:

```
if CONDITION:
    cdef int x = 1
else:
    cdef int x = 2
```
with something like this:
```
cdef int x
if CONDITION:
    x = 1
else:
    x = 2
```

You should also write new tests for the inplace behavior. Have a look at some of the existing tests for box wrapping in `tests/test_box_Box.py` and try to modify the box wrapping one to use your inplace argument. Your test should verify that arrays are actually modified in place.

vyasr:
@bdice for context, these changes are requested in #578. Moreover, this is _not_ a reversion to 2.x behavior, where 1) freud was inconsistent with which operations were in place and which made copies (and inplace operations returned the object even when they were done in place), and 2) the API gave little indication about when operations were in place and when they weren't. The API introduced here avoids any ambiguity about whether or not operations are in place. While uniformity would be nice, as long as requesting an in place operation is clearly performed by specifying an argument I don't see a problem with not supporting it in every method. All aspects of this mirror pandas, for instance, which does not support inplace operations for everything that could, but has the same type of API. I don't think the maintenance burden is as onerous as you fear; the requisite changes to `_convert_array` to enable inplace operation in the first place should also proof us against the issues in #424, so I don't think there's too much extra to be done. If anything, the tests for this will effectively test #424 for the functions where it's enabled. Other than an extra argument I don't think there's much extra to document either.

bdice:
@vyasr I could see an argument for its inclusion (performance), I just want to be sure we're not adding debt to maintainability/uniformity that someone will have to pay off. The most convincing piece of evidence for the inclusion of this feature would be a benchmark, wrapping a system of N vectors in-place and out-of-place, for a range of N values between 100 and 100,000 (or whatever it takes to make it run for around 10 seconds). @Charlottez112 Could you write a benchmark? We have a `benchmarks` folder with examples. If the benchmark is compelling and we decide to move forward, I would like to see this enabled for all of the applicable box methods (so they all match and are tested rigorously). e.g. converting to real/fractional coordinates, etc.

vyasr:
@bdice I agree that a benchmark is critical. For reference, I enabled in place operations on my local branch and tested using the following script:
```
import freud
from timeit import timeit

num_repeats = 10

for num_points in (int(1e6), int(1e7), int(1e8)):
    box, points = freud.data.make_random_system(10, num_points)
    print(f"Created system with {num_points} points")

    time_copy = timeit("box.wrap(points)",
                       globals={"box": box, "points": points},
                       number=num_repeats)
    time_inplace = timeit("box.wrap(points, inplace=True)",
                          globals={"box": box, "points": points},
                          number=num_repeats)
    print(f"The speedup for {num_points} points is {time_copy/time_inplace}")
```
and here are the results:
```
Created system with 1000000 points
The speedup for 1000000 points is 1.527889194475669
Created system with 10000000 points
The speedup for 10000000 points is 1.6664398113928647
Created system with 100000000 points
The speedup for 100000000 points is 2.306493797046124
```

I agree that we should at least try to implement this for all the relevant box methods if we plan to do it. However, not worth moving to other methods until we've evaluated this one more thoroughly, both in terms of performance and the maintenance burden.

vyasr:
We had some discussion of this feature on Slack, and the general consensus is that adding this feature is worthwhile. One useful suggestion that came up was generalizing the idea of inplace operations to providing an `out` argument that is the new array to be populated. This allows the user to provide a preallocated array (possibly the input array). The default behavior of `out=None` will result in a new array being returned.

@Charlottez112 let's try to get the box wrapping working the way that we want it, then we can move on to other methods. Can you try and address some of the comments currently left on this PR? We can start from there.

freud/util.pyx
```diff
@@ -195,7 +195,7 @@ cdef class _Compute(object):
         return repr(self)
 
 
-def _convert_array(array, shape=None, dtype=np.float32):
+def _convert_array(array, shape=None, dtype=np.float32, inplace=False):
```
vyasr: I think we should make this function capable of requiring a copy. Right now box wrapping (and possibly a few other places) contain extra internal logic to either make an extra copy or not, and adding in place operations makes that logic more complicated than simply adding a `.copy()` to the output of this function. I would suggest adding a new argument `copy` that has the following options:

- `default` (better name welcome): This would match the current behavior: copy if needed. Errors if any of the shape/dimensionality checks fail, or the `np.require` call fails. Whether or not a copy is made is entirely determined by whether or not `np.require` makes a copy internally.
- `inplace`: Never makes a copy. Errors if any of the shape/dimensionality checks fail. Also errors if the input array does not have the correct flags. This can either be implemented by checking the flags directly (which this PR currently does), or by calling `np.require` and raising an Exception if the output is a copy. This method would also remove the initial call to `np.asarray` and replace it with a type check
- `copy`: Always returns a copy of the input array. Errors if any of the shape/dimensionality checks fail. Replace the call to `np.require` with a call to `np.array(array, dtype=dtype, order='C')`. 

We will want to make sure that copies are minimized. That also means that the initial `np.asarray` call should probably pass the `dtype` and `order` arguments through in the `default` and `copy` cases, if we're going to retain the call. Moreover, it's redundant to call both `np.asarray` and `np.require` the way that we do now, so we can simplify that if we figure out the exact right numpy function that makes copies in the situations we want.

freud/box.pyx
```diff
@@ -291,6 +291,9 @@ cdef class Box:
         Args:
             vecs (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`):
                 Unwrapped vector(s).
+            out (:class:`numpy.ndarray` or :code:`None`):
+                If None, a new allocation is used for the returned array.
+                The array provided must have the same shape as the input array.
```
vyasr:
```suggestion
            out (:class:`numpy.ndarray` or :code:`None`):
                The array in which to place the wrapped vectors. It must be of
                dtype `np.float32` or larger. If ``None``, this function will
                return a newly allocated array (Default value = None).
```

freud/util.pyx
```diff
@@ -204,17 +204,25 @@ def _convert_array(array, shape=None, dtype=np.float32):
         shape: (tuple of int and :code:`None`): Expected shape of the array.
             Only the dimensions that are not :code:`None` are checked.
             (Default value = :code:`None`).
+        out (:class:`numpy.ndarray` or :code:`None`):
+            If None, a new allocation is used for the returned array.
+            The array provided must have the same shape as the input array.
+
```
vyasr:
```suggestion
            out (:class:`numpy.ndarray` or :code:`None`):
                The array in which to place the wrapped vectors. The dtype
                of the array must be at least as large as the provided ``dtype``
                argument. If ``None``, this function will return a newly
                allocated array (Default value = None).
```

vyasr (changes requested):
@Charlottez112 thanks for the changes. The code looks pretty good, but I think that there are some cases that are probably not covered at the moment (for example, what happens if both `out` and `dtype` are provided to `_convert_array` and they don't match?). I think what would help a lot with this PR is if you could make a list of possible code paths so that we can really identify all the ways this might get called and what the expected behavior should be in this case. For example, something like:
> box.wrap called with vecs of dtype np.float and out=None: input array must be copied since `box.wrap` shouldn't happen in place, the copied array must have the right shape
Don't worry about what numpy functions exist for things etc, first figure out how things _should_ work. Start by just focusing on how `box.wrap` should behave, then once we iron out those details we can figure out what `_convert_array` should handle in each of these cases.

Regarding copying, we can figure out how to minimize copies, and decide whether there are cases where we can guarantee that no copies will be made (even internally) and whether there should be a way to trigger errors if a copy is necessary.

Charlottez112:
Hey @vyasr I'm not sure I completely follow you. The way I think about this is that we can start with adding the ```out``` option for ```box.wrap``` first and modify ```_convert_array``` accordingly. Then the ```dtype``` and ```shape``` of  ```_convert_array```  are just the default values. The situation where the ```out``` and ```dtype``` in ```_convert_array```  don't match is tested, and I added tests for cases where the out array provided does not have the same shape and dtype as the input array of ```box.wrap```. Situations like this are handled by the internal ```numpy.copyto``` function, which raises errors accordingly. 
I think the ```out``` option makes our inital plan of minimizing copies a little more complicated to realize, and I'm not sure how exactly it can help us do that. I guess if the input array is given to ```out```, the user is suggesting that they don't want to make copies? For other cases like the out array is a new array but has the same shape and dtype as the input array, it can get a little complicated to decide and control when copies should be made. Athough ```numpy.copyto``` may still make unnecessary copies, I think it reduces the number of copies compared to the original code.
@bdice I have finished the tests I wanted to add, and I think the ```out``` option added in ```_convert_array``` and ```box.wrap``` should work, but I feel like we still have design questions about minimizing copies. I'm not sure if you guys are satisfied with what we have now. Could you let me know what you think?

vyasr:
@Charlottez112 yes I agree that there are more complexities when `out` is provided, and that we cannot guarantee 0-copy conversion in all cases. I think this would be easier to sort out in a call rather than over Github; the long gaps between periods of work on this PR have made it too asynchronous to easily work on. I'll ping you on Slack to set up a call.

vyasr:
Just had a chat with @Charlottez112 to go through this thoroughly and iron out any questions. I think I probably caused some confusion with some of my earlier comments on this thread, and our switch from `inplace` to `out` was also not super well motivated, so we went back to basics and talked through why we're making these changes and how the code should look. Here's a minimal, complete set of changes that I think need to happen (note that these changes are all relative to master, not the current state of this PR, so this can be used as a checklist when reviewing the final diff):

1. In `cpp/box/Box.h`, make the following changes:
    1. add an extra argument `vec3<float> *out = nullptr` to `Box::wrap`.
    2. Inside `Box::wrap`, add `if (out == nullptr) out = vecs`.
    3. In the loop in `Box::wrap`, change `vecs[i] = wrap(vecs[i])` to `out[i] = wrap(vecs[i])`
2. In `freud/_box.pxd` modify the signature of `wrap` accordingly: `void wrap(vec3[float]* vs, unsigned int Nv, vec3[float]* out) const`
3. In `freud/util.pyx`, at the beginning of `_convert_array` add something like `if array is None: return np.empty(shape, dtype)`. I think `empty` is definitely preferable here since any array created via this API should always be getting overwritten.
4. In `freud/box.pyx`, make the following changes:
    1. Change `Box.wrap` to accept a parameter `out=None`. 
    2. Near the beginning of the method add something like `return_value = out is None`
    3. Remove the `.copy()` from the `_convert_array` call for vecs.
    4. Just below that line, add `out = freud.util._convert_array(out, shape=(None, 3))`.
    5. Add a new cdef `cdef const float[:, ::1] l_out = out`.
    6. Change the C++ function call to `self.thisptr.wrap(<vec3[float]*> &l_points[0, 0], Np, <vec3[float]*> &l_out[0, 0])`
    7. Change the return to only return `if return_value`

Once that's done, we can reevaluate whether there are other copies that can be reduce.

Charlottez112:
> 3. In `freud/util.pyx`, at the beginning of `_convert_array` add something like `if array is None: return np.empty(shape, dtype)`. I think `empty` is definitely preferable here since any array created via this API should always be getting overwritten.
> 4. In `freud/box.pyx`, make the following changes:
>    
>    1. Change `Box.wrap` to accept a parameter `out=None`.
>    2. Near the beginning of the method add something like `return_value = out is None`
>    3. Remove the `.copy()` from the `_convert_array` call for vecs.
>    4. Just below that line, add `out = freud.util._convert_array(out, shape=(None, 3))`.
>    5. Add a new cdef `cdef const float[:, ::1] l_out = out`.
>    6. Change the C++ function call to `self.thisptr.wrap(<vec3[float]*> &l_points[0, 0], Np, <vec3[float]*> &l_out[0, 0])`
>    7. Change the return to only return `if return_value`

Thank you @vyasr for the suggestions. What I did was a little different. ```numpy.empty``` cannot take ```(None, 3)``` as ```shape```, so I used ```out = freud.util._convert_array(out, shape=vecs.shape)``` instead. Since we are always storing the modified array in the ```out``` array in ```Box.h```, I didn't use ```return_value``` but made it always return ```out``` instead.

vyasr:
> Thank you @vyasr for the suggestions. What I did was a little different. `numpy.empty` cannot take `(None, 3)` as `shape`, so I used `out = freud.util._convert_array(out, shape=vecs.shape)` instead. Since we are always storing the modified array in the `out` array in `Box.h`, I didn't use `return_value` but made it always return `out` instead.

@Charlottez112 perfect! Re: the shape, that's my mistake, you did the right thing. For the return, I just checked how numpy treats this, and if `out` is provided it is also returned:
```
>>> x = np.ones(10)
>>> y = np.empty(10)
>>> tmp = np.add(x, x, y)
>>> tmp is y
True
```
so your decision to also return in that case looks good too.

The tests still have a much more complex diff than they should. Currently you are inadvertently undoing a lot of the changes @bdice made in #733 to use `pytest` rather than `unittest`. Once you fix those, I think this PR will be ready for a final round of reviews.

Charlottez112:
> The tests still have a much more complex diff than they should. Currently you are inadvertently undoing a lot of the changes @bdice made in #733 to use `pytest` rather than `unittest`. Once you fix those, I think this PR will be ready for a final round of reviews.

I'm not sure why this happened. Sorry about that. I just fixed those, and I guess it should be good now.

cpp/box/Box.h
```diff
@@ -336,12 +336,17 @@ class Box
     /*! \param vecs Vectors to wrap, updated to the minimum image obeying the periodic settings
      *  \param Nvecs Number of vectors
      */
-    void wrap(vec3<float>* vecs, unsigned int Nvecs) const
+    void wrap(vec3<float>* vecs, unsigned int Nvecs, vec3<float> *out = nullptr) const
```
bdice: Match the convention of the other pointers (`*` on the type rather than the name).
```suggestion
    void wrap(vec3<float>* vecs, unsigned int Nvecs, vec3<float>* out = nullptr) const
```

freud/box.pyx
```diff
@@ -292,6 +292,10 @@ cdef class Box:
         Args:
             vecs (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`):
                 Unwrapped vector(s).
+            out (:class:`numpy.ndarray` or :code:`None`):
```
bdice: Add sizes to match the other parameter docstrings.
```suggestion
            out (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray` or :code:`None`):
```

freud/box.pyx
```diff
@@ -292,6 +292,10 @@ cdef class Box:
...
+                The array in which to place the wrapped vectors. It must be of
+                dtype `np.float32` or larger. If ``None``, this function will
```
bdice: "Or larger"? That doesn't seem quite right. If it's not `float32` then a new array is allocated, right? I think this should be `float32` only.

freud/box.pyx
```diff
@@ -300,13 +304,18 @@ cdef class Box:
...
         cdef unsigned int Np = l_points.shape[0]
-        self.thisptr.wrap(<vec3[float]*> &l_points[0, 0], Np)
+        cdef const float[:, ::1] l_out = out
 
-        return np.squeeze(vecs) if flatten else vecs
+        self.thisptr.wrap(<vec3[float]*> &l_points[0, 0],
+                          Np, <vec3[float]*> &l_out[0, 0])
+        if flatten:
+            out = np.squeeze(out)
+        return out
```
bdice: Write this as one line so that this method matches the style of similar methods.
```suggestion
        return np.squeeze(out) if flatten else out
```

freud/_box.pxd
```diff
@@ -39,7 +39,8 @@ cdef extern from "Box.h" namespace "freud::box":
         void makeAbsolute(vec3[float]*, unsigned int) const
         void makeFractional(vec3[float]*, unsigned int) const
         void getImages(vec3[float]*, unsigned int, vec3[int]*) const
-        void wrap(vec3[float]* vs, unsigned int Nv) const
+        void wrap(vec3[float]* vs, unsigned int Nv,
+                  vec3[float]* out) const
```
bdice: Delete the variable names here, none of the other methods include them.

freud/box.pyx
```diff
@@ -300,13 +304,18 @@ cdef class Box:
         vecs = np.asarray(vecs)
         flatten = vecs.ndim == 1
         vecs = np.atleast_2d(vecs)
-        vecs = freud.util._convert_array(vecs, shape=(None, 3)).copy()
+        vecs = freud.util._convert_array(vecs, shape=(None, 3))
+        out = freud.util._convert_array(out, shape=vecs.shape)
 
         cdef const float[:, ::1] l_points = vecs
```
bdice, line 310: Let's rename `l_points` to `l_vecs` to match the variable name used elsewhere.

freud/util.pyx
```diff
@@ -212,8 +212,12 @@ def _convert_array(array, shape=None, dtype=np.float32):
     Returns:
         :class:`numpy.ndarray`: Array.
     """
+    if array is None:
+        return np.empty(shape, dtype=dtype)
```
bdice, line 225: We should specify in the docstring that if `array is None`, the calling code is responsible for setting the initial data. Otherwise it is possible that the caller might not realize that and expect an array of zeros.

freud/util.pyx
```diff
@@ -212,8 +212,12 @@ def _convert_array(array, shape=None, dtype=np.float32):
...
+
     array = np.asarray(array)
     return_arr = np.require(array, dtype=dtype, requirements=['C'])
```
bdice: I would try to combine these steps so that the case where `array is None` doesn't have to _also_ convert to an array and require C-ordering / contiguity. Basically just do something like:
```python
if array is None:
    # Construct a valid return_array that meets the array requirements.
else:
    # Perform the "asarray" and enforce array requirements.
```

tests/test_box_Box.py
```diff
@@ -112,6 +112,38 @@ def test_wrap_multiple_images(self):
         points = np.array(points)
         npt.assert_allclose(box.wrap(points)[0, 0], -2, rtol=1e-6)
 
+    def test_out_default(self):
+        box = freud.box.Box(2, 2, 2, 1, 0, 0)
+        points = [[10, -5, -5], [0, 0.5, 0]]
+        npt.assert_allclose(box.wrap(points)[0, 0], -2, rtol=1e-6)
```
bdice: Could we check the entire array is correct, instead of just one value? Applies to all tests below.

tests/test_box_Box.py
```diff
@@ -112,6 +112,38 @@ def test_wrap_multiple_images(self):
...
+
+    def test_out_provided_with_input_array(self):
+        box = freud.box.Box(2, 2, 2, 1, 0, 0)
+        points = [[10, -5, -5], [0, 0.5, 0]]
+        npt.assert_allclose(box.wrap(points, out=points)[0, 0], -2, rtol=1e-6)
```
bdice: One way to test the `out` behavior more carefully would be to make the input array read-only. You can do that by setting flags on the array with NumPy. I think that should fail if you let `out` be the default (overwrite in place).

tests/test_box_Box.py
```diff
@@ -112,6 +112,38 @@ def test_wrap_multiple_images(self):
...
+
+    def test_out_provided_with_new_array(self):
+        box = freud.box.Box(2, 2, 2, 1, 0, 0)
+        points = [[10, -5, -5], [0, 0.5, 0]]
+        points = np.array(points, dtype=np.float32)
+        new_array = np.zeros(points.shape, dtype=np.float32)
+        npt.assert_allclose(box.wrap(points, out=new_array)[0, 0], -2, rtol=1e-6)
```
bdice: Also assert that the old `points` array did _not_ change.

tests/test_box_Box.py
```diff
@@ -112,6 +112,38 @@ def test_wrap_multiple_images(self):
...
+        new_array = np.zeros((np.asarray(points.shape) - 1), dtype=np.float32)
+        with pytest.raises(ValueError):
+            npt.assert_allclose(box.wrap(points, out=new_array)[0, 0], -2, rtol=1e-6)
+
+    def test_out_provided_with_array_with_wrong_dtype(self):
+        box = freud.box.Box(2, 2, 2, 1, 0, 0)
+        points = [[10, -5, -5], [0, 0.5, 0]]
+        points = np.array(points, dtype=np.int)
+        new_array = np.zeros(points.shape, dtype=np.int64)
+        npt.assert_allclose(box.wrap(points, out=new_array)[0, 0], -2, rtol=1e-6)
```
bdice: Should this fail?

bdice (changes requested):

freud/box.pyx
Charlottez112: Just to clarify: do you mean the ```l_points``` in ```box.wrap``` or all the  ```l_points``` in ```box.pyx```?

freud/util.pyx
Charlottez112: I'm not sure I understand why cases where ```array is None``` also has to convert I thought what we're doing here is that we generate an empty array that has the given shape and type if ```array is None```, and we just skip the converting part?

freud/box.pyx
bdice: Nevermind, I see there are lots of exceptions. I would have liked the C++ and Python parameter names to match, e.g. `const vec3<float>* vecs` in C++ and `l_vecs` / `vecs` in Cython. It doesn't really matter, this comment can be ignored. We could make an issue to refactor all of that to improve readability but the benefit is small.

freud/util.pyx
bdice: That's my mistake, I read the code incorrectly. I saw `return_arr = np.empty(shape, dtype=dtype)` instead of `return np.empty(shape, dtype=dtype)`. This change is fine as-is.

tests/test_box_Box.py
Charlottez112: Yeah I think you're right. I guess what happens here is that if the ```new_array``` passed to```out```  is not ```float32```, the returned array of ```box.wrap``` should give the right thing, but the ```new_array``` itself wouldn't be overwritten? (Because ```_convert_array``` doesn't make changes inplace?) In this case, should we add a check in ```box.wrap``` and throw an error if the array passed to ```out``` doesn't have the right type?

cpp/box/Box.h
```diff
@@ -336,12 +336,17 @@ class Box
...
     {
+        if (out == nullptr)
```
bdice: An alternative approach would be for us to change the signature of this method, and _require_ `out` (no default value of `nullptr`). This would allow us to make the first argument `const vec3<float>*` (constant). This method is only used by Cython; I could not find any instances of `wrap` applied to multiple vectors in the C++ code so this wouldn't be breaking. @vyasr @Charlottez112 What do you think?

edit: It would slightly break the consistency between the C++ and Python APIs to have a required `out` argument (Python won't require it).

tests/test_box_Box.py
bdice: I'm not immediately sure what is supposed to happen, we can come back to this after resolving the other comments.

Charlottez112: I added a type check in ```box.pyx``` just to make the test pass first.

freud/box.pyx
```diff
@@ -300,13 +304,21 @@ cdef class Box:
         vecs = np.asarray(vecs)
         flatten = vecs.ndim == 1
         vecs = np.atleast_2d(vecs)
-        vecs = freud.util._convert_array(vecs, shape=(None, 3)).copy()
+        vecs = freud.util._convert_array(vecs, shape=(None, 3))
+        if out is not None:
+            if out.dtype != np.float32:
+                raise ValueError('expected out array type: np.float32')
```
bdice: See: https://stackoverflow.com/questions/26921836/correct-way-to-test-for-numpy-dtype
```suggestion
        if out is not None and isinstance(out, np.ndarray) and out.dtype != np.float32:
            raise ValueError('expected out array type: np.float32')
```

bdice: This has an example of what an invalid out parameter looks like: https://riptutorial.com/numpy/example/10938/the-out-parameter

Here's the NumPy C API that performs error checking: https://github.com/numpy/numpy/blob/cb557b79fa0ce467c881830f8e8e042c484ccfaa/numpy/core/src/multiarray/common.c#L401-L409

I think we're covering our bases reasonably well but just wanted to share that info for reference.

bdice:
@Charlottez112 Apologies if this PR review process has felt slow -- we have dealt with a lot of edge cases in freud regarding how input data arrays are received and converted. Introducing a new API for "out" parameters requires a high level of scrutiny to make sure that our bases are covered and we're not introducing undefined behavior or bugs. There is a lot to consider, with the various behaviors of NumPy, Cython, and C++.

cpp/box/Box.h
vyasr: Yeah I considered that possibility as well, and chose to stick with this implementation precisely because I wanted to maintain `out` as an optional parameter. However, on second though I've changed my mind and would like to require the `out` parameter in C++. Here's my reasoning: we already have different semantics in Python and C++, because in Python we create a new array for the user if one is not provided. I think it would be a less obvious, and therefore more insidious, inconsistency if the default behavior when `out` is not provided is to overwrite in place in C++ while it is to return a copy in Python. C++ users are much more likely to care about the performance impact of minimizing copies, so it's much more consistent with best practices in the language to require them to handle allocating new arrays if they need to; otherwise, they can just pass a pointer to the same array to trigger in-place overwrites.

Note that in order for this to be a good idea we'd want to make sure that all the C++ APIs for box functions behave the same way. I think that shouldn't be an issue though.

freud/box.pyx
vyasr: I guess if we really want to be thorough we should also check `out.flags['C_CONTIGUOUS']` 🤷

Charlottez112: I added the check for ```out.flags['C_CONTIGUOUS']```

tests/test_box_Box.py
Charlottez112: I also added a test for this.

cpp/box/Box.h
Charlottez112: I agree with @vyasr. I'm not sure I understand the benefits of requiring ```out```, and I don't know enough about the C++ user cases to make comments on that. But one thing I've been thinking is that if we're doing all the checks (in ```box.wrap```) for ```out``` when it's given, why don't we just make a new array for it in ```box.wrap``` instead of making one in ```_convert_array```, when it's the default ```None```.

vyasr: @Charlottez112 I'm not sure I follow, which part of what I said are you agreeing with? It sounds like not the part about requiring `out`, so I'm a bit confused.

Regarding where we make the new array, I assume that you mean in `box.wrap` in Python, not in the C++ function? If so, the main reason I suggested doing it in `_convert_array` is because then the exact same logic can be used for all the other box functions. Anywhere that a function requires an array to have certain properties but where a user might not pass that array, those functions could rely on `_convert_array` to create an appropriate array. Does that make sense?

vyasr: @bdice let me know whether we're on the same page here before we ask @Charlottez112 to make any changes.

vyasr (commented):
@Charlottez112 it looks like there's still a few open conversations. The one about whether or not to require the `out` parameter in C++ still needs a response from @bdice, but can you resolve all of the others? Then once the `out` parameter's behavior is decided re-request my review.

Thanks for being a trooper and going through all these iterations! Like Bradley said, this PR involves getting a lot of small decisions just right, we really appreciate you sticking with it.

cpp/box/Box.h
bdice: @vyasr I'm on board with the proposal to require `out` in C++, and particularly this line of reasoning that you mentioned above:

> I think it would be a less obvious, and therefore more insidious, inconsistency if the default behavior when out is not provided is to overwrite in place in C++ while it is to return a copy in Python.

I don't entirely understand the subsequent discussion, it sounds like there is some confusion / other topics. Please let me know if you're asking for input on some piece of that.

Charlottez112: > @Charlottez112 I'm not sure I follow, which part of what I said are you agreeing with? It sounds like not the part about requiring `out`, so I'm a bit confused.

Sorry for the confusion. I meant to ask @bdice if he could explain a bit more about what he thinks.

> Regarding where we make the new array, I assume that you mean in `box.wrap` in Python, not in the C++ function? If so, the main reason I suggested doing it in `_convert_array` is because then the exact same logic can be used for all the other box functions. Anywhere that a function requires an array to have certain properties but where a user might not pass that array, those functions could rely on `_convert_array` to create an appropriate array. Does that make sense?

Yeah that makes sense. I guess I just don't know if there's a better way to combine checking the ```out``` array with generating one if it's none.

Charlottez112: I went ahead and changed this part in ```Box.h```. Not sure if there's anything else we need to change

freud/box.pyx
```diff
@@ -300,13 +304,23 @@ cdef class Box:
...
         vecs = np.atleast_2d(vecs)
-        vecs = freud.util._convert_array(vecs, shape=(None, 3)).copy()
+        vecs = freud.util._convert_array(vecs, shape=(None, 3))
+        if out is not None and isinstance(out, np.ndarray):
+            if out.dtype != np.float32 or out.flags['C_CONTIGUOUS'] == 0:
+                raise ValueError('output array is not acceptable '
+                                 '(must have the right type, nr dimensions, '
+                                 'and be a C-Array)')
+
+        out = freud.util._convert_array(out, shape=vecs.shape)
```
vyasr:
```suggestion
        out = freud.util._convert_array(out, shape=vecs.shape, allow_copy=False)
```

freud/util.pyx
```diff
@@ -201,7 +201,8 @@ def _convert_array(array, shape=None, dtype=np.float32):
     and converts to a supplied dtype.
 
     Args:
-        array (:class:`numpy.ndarray`): Array to check and convert.
+        array (:class:`numpy.ndarray` or :code:`None`): Array to check and convert.
+            If :code:`None`, an empty array of given shape and type will be initialized.
```
vyasr:
```suggestion
            If :code:`None`, an empty array of given shape and type will be initialized
            (Default value: :code:`None`).
```

vyasr (approved):
LGTM! Thanks for all the hard work @Charlottez112! I had a couple of minor change requests left, but they were small so I went ahead and made them. Feel free to look at the last couple of commits if you want to see what I changed.

freud/util.pyx
```diff
@@ -196,24 +196,40 @@ cdef class _Compute(object):
...
     Returns:
         :class:`numpy.ndarray`: Array.
     """
-    array = np.asarray(array)
-    return_arr = np.require(array, dtype=dtype, requirements=['C'])
+    if array is None:
+        return np.empty(shape, dtype=dtype)
+
+    return_arr = np.require(array, dtype=dtype, requirements=requirements)
+
```
vyasr:
```suggestion

    return_arr = np.require(array, dtype=dtype, requirements=requirements)

```

vyasr:
@Charlottez112 @bdice only one question left to address on my end do we want to merge this straight to master, or accumulate all in-place operations into one branch and merge them en masse at the end? @bdice if you prefer the latter, please create a new target branch and repoint this PR before you merge it.

freud/util.pyx
```diff
@@ -209,15 +210,25 @@ def _convert_array(array, shape=None, dtype=np.float32):
...
 
     Returns:
         :class:`numpy.ndarray`: Array.
     """
     if array is None:
         return np.empty(shape, dtype=dtype)
 
-    array = np.asarray(array)
-    return_arr = np.require(array, dtype=dtype, requirements=['C'])
+    return_arr = np.require(array, dtype=dtype, requirements=requirements)
```
Charlottez112, line 227: Could you explain why we don't want ```array = np.asarray(array)``` here? Is this for reducing the number of copies? I guess this is what's causing the errors

vyasr: It's redundant. `np.require` internally has to convert `array` to a numpy array anyway. The problems is that I didn't change enough of the code below, which is still using `array` instead of `return_arr`.

bdice:
I'm going to hold off and merge this after we release v2.5.1 on Tuesday. This can go into v2.6.0.

bdice:
@Charlottez112 Can you please review the checklist and update the changelog and credits before we merge? Thanks.

bdice:
> do we want to merge this straight to master, or accumulate all in-place operations into one branch and merge them en masse at the end?

@glotzerlab/freud-developers I welcome others' thoughts on this. I suggest that we make a list of the operations that we wish to support in-place, and then decide.

tommy-waltmann:
Let's make a list, look at the size of the list, and then decide how we want to proceed.

bdice (approved):

vyasr:
In the interest of not dragging this out any further, I'm going to go ahead and merge this to master once tests pass. Adding this feature for other functions shouldn't be a problem before 2.6.

PR #676, opened 2020-10-27
Enable scikit-build on Windows (without static members)

bdice:
## Description
Windows compatibility for scikit-build. Follows up on #668 and supersedes #669.
- Handles MSVC symbol exports. MSVC requires explicit declarations of symbols that need to be exported, with `__declspec(dllexport)`. When we compile `libfreud`, we can tell CMake to automatically export all symbols. This lifts most of the burden for us.
- Fixes some problems where CMake wasn't working as expected. In particular, `target_link_libraries` didn't recognize `TBB` and had to be given `${TBB_LIBRARY}` instead.
- Some symbols, specifically static class members like `NeighborQuery::DEFAULT_NUM_NEIGHBORS` or `NeighborQueryIterator::ITERATOR_TERMINATOR` are not exported by CMake's setting and we must do that manually. **This PR improves on the solution used in #669 by using namespace constexpr variables, instead of static member variables.**
- Installs CMake 3.14 on Windows only, since it's the oldest version that supports Visual Studio 2019 (which is installed on this CI image and is used for builds).

## Motivation and Context
Windows builds are useful for integration with OVITO. Also it's our second most popular build after Linux:
image
(data collected for freud 2.3.0 only, on Oct. 13, 2020)

## How Has This Been Tested?
Builds locally and on CI.

## Types of changes

## Checklist:


bdice:
@vyasr I think this PR should be squashed when it is merged into #668.

cpp/locality/CMakeLists.txt
```diff
@@ -18,9 +18,7 @@ add_library(
     NeighborComputeFunctional.h
     NeighborList.cc
     NeighborList.h
-    NeighborPerPointIterator.cc
     NeighborPerPointIterator.h
-    NeighborQuery.cc
```
bdice, line 23: Now header only! 😄

cpp/locality/NeighborQuery.h
```diff
@@ -22,6 +22,23 @@
 
 namespace freud { namespace locality {
 
+//! Enumeration for types of queries.
+enum QueryType
```
bdice, line 26: I moved this out of the scope of `QueryArgs`. I would prefer to use an `enum class` instead of an `enum` (since it is strongly typed) but support for that was only added to Cython somewhat recently, and I suspect it would cause us headaches in the interim. https://github.com/cython/cython/pull/3640

.circleci/config.yml
```diff
@@ -96,6 +98,13 @@ references:
         # Max 4 threads to not overtax CI.
         python${PYVER} setup.py build_ext --inplace -- -DCOVERAGE=ON -- -j 4 -v
 
+  build_windows: &build_windows
+    run:
+      name: Build with MSVC
+      command: |
+        python${PYVER} --version
+        python${PYVER} setup.py build_ext --inplace -- -G "Visual Studio 16 2019" -DCOVERAGE=ON
```
bdice, line 106: I think we could simplify this later, since `build_windows` and `build` are now the same except for some platform-specific flags. I would prefer to cut the scope where it is, and do that in #668 or afterward.

tommy-waltmann (approved):

cpp/locality/NeighborQuery.h
vyasr: Yeah I'm pretty sure I tried that (nested inside `QueryArgs`) when I first made this PR and found that limitation, didn't know they got around to adding the feature. We can change it later if we like, but I agree that it's not worth doing until Cython has proper support. There are other workarounds (using an additional namespace, nesting inside another class, etc) to enforce better scoping, but the benefits are pretty minimal.

cpp/locality/NeighborQuery.h
```diff
@@ -22,6 +22,23 @@
...
+};
+
+constexpr auto DEFAULT_MODE = QueryType::none;  //!< Default mode.
+constexpr unsigned int DEFAULT_NUM_NEIGHBORS(0xffffffff); //!< Default number of neighbors.
+constexpr float DEFAULT_R_MAX(-1.0);  //!< Default maximum query distance.
+constexpr float DEFAULT_R_MIN(0);  //!< Default minimum query distance.
+constexpr float DEFAULT_R_GUESS(-1.0);  //!< Default guess query distance.
+constexpr float DEFAULT_SCALE(-1.0);  //!< Default scaling parameter for AABB nearest neighbor queries.
+constexpr bool DEFAULT_EXCLUDE_II(false);  //!< Default for whether or not to include self-neighbors.
+constexpr auto ITERATOR_TERMINATOR = NeighborBond(-1, -1, 0);  //!< The object returned when iteration is complete.
```
vyasr, line 40: Nice side effect of making these non-static that we can make this header-only.

vyasr (approved):
Thanks! Glad we talked through this, I like this solution much more than the old one. Anything we can do to avoid Windows-specific code will dramatically increase the chances that Windows builds don't eventually break. Agreed on the squash, I'll go ahead with that.

.circleci/config.yml
vyasr: I'd say that's a post-skbuild refactor, I've noticed in a couple recent iterations that our CI scripts in general could use some cleanup. We can do it when we also do things like enabling pre-commit and clang-format.

PR #677, opened 2020-10-28
Feature/gaussiandensity values

### LINKED ISSUE #666: Gaussian convolution of the parameter
Hi,

I think that might be useful to extend GuassianDensity class to allow the convolution of the parameter. FX, if some parameter p is assigned to every atom in the box, its density function could be written as convolution of parameter for atom *l* with Dirac's delta function:

image

For some tasks (e.g. studying phase transitions) the blurring of the parameter density function might be needed ([see for example](https://elifesciences.org/articles/13150)). Adding such option to GuassianDensity class is straightforward and would have solved the issue. The parameter list might be passed as an argument to *compute* method and C++ code at the very last step, the gaussian distribution should be multiplied by the parameter for the current point.

pbuslaev:
This a pull request for issue #666. I have implemented the convolution of the parameter array `p` with Gaussians. The changes concern 

- `freud/density.pyx`: Added the parameter values=None to compute method. The type of the values is float

-  `freud/_density.pxd`: Added an argument for the array of values

- `cpp/density/GaussianDensity.h`: Added argument matching the type of that added to _density.pxd

- `cpp/density/GaussianDensity.cc`: Added the argument and used the input values to calculate the convolution

### DIFF
--- a/cpp/density/GaussianDensity.cc
+++ b/cpp/density/GaussianDensity.cc
@@ -84,7 +84,7 @@ void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const fl
         for (size_t idx = begin; idx < end; ++idx)
         {
             const vec3<float> point = (*nq)[idx];
-            const float value = (*values)[idx]
+            const float value = values[idx];
 
             // Find which bin the particle is in
             int bin_x = int((point.x + Lx / 2.0f) / grid_size_x);
--- a/cpp/density/GaussianDensity.cc
+++ b/cpp/density/GaussianDensity.cc
@@ -32,7 +32,7 @@ vec3<unsigned int> GaussianDensity::getWidth()
 }
 
 //! Compute the density array.
-void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const float* values)
+void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, float* values)
 {
     // set the number of dimensions for the calculation the first time it is done
     if (!m_has_computed || nq->getBox().is2D() == m_box.is2D())
@@ -83,8 +83,8 @@ void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const fl
         // for each reference point
         for (size_t idx = begin; idx < end; ++idx)
         {
-            const vec3<float> point = (*nq)[idx];
-            const float value = values[idx];
+            const vec3<float> point = (*nq)[idx]; 
+            float value = (values != NULL) ? values[idx] : 1.0;
 
             // Find which bin the particle is in
             int bin_x = int((point.x + Lx / 2.0f) / grid_size_x);
--- a/cpp/density/GaussianDensity.cc
+++ b/cpp/density/GaussianDensity.cc
@@ -32,7 +32,7 @@ vec3<unsigned int> GaussianDensity::getWidth()
 }
 
 //! Compute the density array.
-void GaussianDensity::compute(const freud::locality::NeighborQuery* nq)
+void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const T* values)
 {
     // set the number of dimensions for the calculation the first time it is done
     if (!m_has_computed || nq->getBox().is2D() == m_box.is2D())
--- a/cpp/density/GaussianDensity.cc
+++ b/cpp/density/GaussianDensity.cc
@@ -32,7 +32,7 @@ vec3<unsigned int> GaussianDensity::getWidth()
 }
 
 //! Compute the density array.
-void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const T* values)
+void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const float* values)
 {
     // set the number of dimensions for the calculation the first time it is done
     if (!m_has_computed || nq->getBox().is2D() == m_box.is2D())
@@ -84,6 +84,8 @@ void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const T*
         for (size_t idx = begin; idx < end; ++idx)
         {
             const vec3<float> point = (*nq)[idx];
+            const float value = (*values)[idx]
+
             // Find which bin the particle is in
             int bin_x = int((point.x + Lx / 2.0f) / grid_size_x);
             int bin_y = int((point.y + Ly / 2.0f) / grid_size_y);
@@ -130,7 +132,7 @@ void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const T*
                         if (r_sq < r_max_sq)
                         {
                             // Evaluate the gaussian
-                            const float gaussian = normalization * std::exp(-r_sq / (float(2.0) * sigmasq));
+                            const float gaussian = value * normalization * std::exp(-r_sq / (float(2.0) * sigmasq));
 
                             // Assure that out of range indices are corrected for storage
                             // in the array i.e. bin -1 is actually bin 29 for nbins = 30
--- a/cpp/density/GaussianDensity.cc
+++ b/cpp/density/GaussianDensity.cc
@@ -32,7 +32,7 @@ vec3<unsigned int> GaussianDensity::getWidth()
 }
 
 //! Compute the density array.
-void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, float* values)
+void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const float* values)
 {
     // set the number of dimensions for the calculation the first time it is done
     if (!m_has_computed || nq->getBox().is2D() == m_box.is2D())
--- a/cpp/density/GaussianDensity.cc
+++ b/cpp/density/GaussianDensity.cc
@@ -84,7 +84,7 @@ void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const fl
         for (size_t idx = begin; idx < end; ++idx)
         {
             const vec3<float> point = (*nq)[idx]; 
-            float value = (values != NULL) ? values[idx] : 1.0;
+            const float value = (values != NULL) ? values[idx] : 1.0f;
 
             // Find which bin the particle is in
             int bin_x = int((point.x + Lx / 2.0f) / grid_size_x);
--- a/cpp/density/GaussianDensity.h
+++ b/cpp/density/GaussianDensity.h
@@ -49,7 +49,7 @@ public:
     }
 
     //! Compute the density.
-    void compute(const freud::locality::NeighborQuery* nq);
+    void compute(const freud::locality::NeighborQuery* nq, const T* values);
 
     //! Get a reference to the last computed density.
     const util::ManagedArray<float>& getDensity() const;
--- a/cpp/density/GaussianDensity.h
+++ b/cpp/density/GaussianDensity.h
@@ -49,7 +49,7 @@ public:
     }
 
     //! Compute the density.
-    void compute(const freud::locality::NeighborQuery* nq, float* values = NULL);
+    void compute(const freud::locality::NeighborQuery* nq, const float* values = NULL);
 
     //! Get a reference to the last computed density.
     const util::ManagedArray<float>& getDensity() const;
--- a/cpp/density/GaussianDensity.h
+++ b/cpp/density/GaussianDensity.h
@@ -49,7 +49,7 @@ public:
     }
 
     //! Compute the density.
-    void compute(const freud::locality::NeighborQuery* nq, const float* values);
+    void compute(const freud::locality::NeighborQuery* nq, float* values = NULL);
 
     //! Get a reference to the last computed density.
     const util::ManagedArray<float>& getDensity() const;
--- a/cpp/density/GaussianDensity.h
+++ b/cpp/density/GaussianDensity.h
@@ -49,7 +49,7 @@ public:
     }
 
     //! Compute the density.
-    void compute(const freud::locality::NeighborQuery* nq, const T* values);
+    void compute(const freud::locality::NeighborQuery* nq, const float* values);
 
     //! Get a reference to the last computed density.
     const util::ManagedArray<float>& getDensity() const;
--- a/freud/_density.pxd
+++ b/freud/_density.pxd
@@ -27,7 +27,7 @@ cdef extern from "GaussianDensity.h" namespace "freud::density":
         const freud._box.Box & getBox() const
         void reset()
         void compute(const freud._locality.NeighborQuery*,
-                     float*) except +
+                     const float*) except +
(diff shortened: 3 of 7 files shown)


vyasr:
@pbuslaev thanks for the contribution! I'll look at your code as soon as I can.

Currently your code is failing our style checks, which we do using `flake8`. If you `pip install flake8` you'll be able to run `flake8 freud` from the root of the repo and it will tell you how to address those issues. For future commits, you can set up a pre-commit hook, which [we've documented in our development guide](https://freud.readthedocs.io/en/stable/reference/development/howtoadd.html). We do plan to switch to using the [`pre-commit`](https://pre-commit.com/) framework for this at some point, but for now the old style flake8 hook will get the job done. Thanks!

pbuslaev:
@vyasr Thanks for your comments! I have updated the style.

freud/density.pyx
```diff
@@ -228,17 +228,33 @@ cdef class GaussianDensity(_Compute):
...
         R"""Calculates the Gaussian blur for the specified points.
 
         Args:
             system:
                 Any object that is a valid argument to
                 :class:`freud.locality.NeighborQuery.from_system`.
+            values ((:math:`N_{points}`) :class:`numpy.ndarray`):
+                Values associated with the system points used to calculate the
+                convolution. Calculates Gaussian blur if :code:`None`.
+                (Default value = :code:`None`).
```
bdice: Mention that the "Gaussian blur" is equivalent to every point having a value of 1.
```suggestion
                convolution. Calculates Gaussian blur (equivalent to providing
                a value of 1 for every point) if :code:`None`. (Default value
                = :code:`None`).
```

freud/density.pyx
```diff
@@ -228,17 +228,33 @@ cdef class GaussianDensity(_Compute):
...
         """
         cdef freud.locality.NeighborQuery nq = \
             freud.locality.NeighborQuery.from_system(system)
-        self.thisptr.compute(nq.get_ptr())
+
+        if values is None:
+            values = freud.util._convert_array(
+                np.ones(nq.points.shape[0]), shape=(nq.points.shape[0], ),
+                dtype=np.float32)
```
bdice: 1. We can avoid an unnecessary double-float conversion here. We should specify the dtype for `np.ones` since it defaults to `float64`, but we don't need to specify the dtype for `_convert_array` since it defaults to `float32`. 
2. We don't need to check the shape in `_convert_array` since we already know its shape is correct (by construction). We only have to check the shape of a user-provided array of values.

```suggestion
                np.ones(nq.points.shape[0], dtype=np.float32))
```

freud/density.pyx
```diff
@@ -228,17 +228,33 @@ cdef class GaussianDensity(_Compute):
...
+        else:
+            values = freud.util._convert_array(
+                values, shape=(nq.points.shape[0], ), dtype=np.float32)
+
+        cdef float[::1] l_values = values
```
bdice: ~This can be marked as `const`.~ _Edit:_ I don't think this can be marked as `const` if you use the code above to support `NULL`. (Cython requires variable declaration to be outside of a conditional, but you have to assign the variable conditionally so you can't make it `const`.)
```suggestion
        cdef const float[::1] l_values = values
```

bdice (changes requested):
Overall this PR is very good! Nice work, @pbuslaev. I made a few suggestions that you can apply directly.

Please add tests for using a set of distinct values. This is the file that contains tests: https://github.com/glotzerlab/freud/blob/master/tests/test_density_GaussianDensity.py

One minor improvement I would like to see is allowing `NULL` for the values array in C++. (We use `NULL` instead of `nullptr` for compatibility with Cython, though `nullptr` is typically preferred.) We are considering making our C++ API "public" in the future so that other libraries could use freud internally, so having the "None" behavior match in Python and C++ would be nice. You can see an example of how to do that here:
https://github.com/glotzerlab/freud/blob/deef8b5678e10c1a8157e4b8e617f80aea2f1706/freud/box.pyx#L344-L385
https://github.com/glotzerlab/freud/blob/deef8b5678e10c1a8157e4b8e617f80aea2f1706/cpp/box/Box.h#L364-L390

freud/density.pyx
pbuslaev: @bdice thanks a lot for your comments. I have added tests and `nullptr` to the latest commit

pbuslaev: @bdice Thanks for your valuable comments. I have committed your suggestions about Gaussian blur and double-float conversion, added tests and nullptr

cpp/density/GaussianDensity.cc
```diff
@@ -32,7 +32,7 @@ vec3<unsigned int> GaussianDensity::getWidth()
 }
 
 //! Compute the density array.
-void GaussianDensity::compute(const freud::locality::NeighborQuery* nq)
+void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, float* values)
```
vyasr:
```suggestion
void GaussianDensity::compute(const freud::locality::NeighborQuery* nq, const float* values)
```

cpp/density/GaussianDensity.h
```diff
@@ -49,7 +49,7 @@ class GaussianDensity
     }
 
     //! Compute the density.
-    void compute(const freud::locality::NeighborQuery* nq);
+    void compute(const freud::locality::NeighborQuery* nq, float* values = NULL);
```
vyasr:
```suggestion
    void compute(const freud::locality::NeighborQuery* nq, const float* values = NULL);
```

freud/_density.pxd
```diff
@@ -26,7 +26,8 @@ cdef extern from "GaussianDensity.h" namespace "freud::density":
         GaussianDensity(vec3[unsigned int], float, float) except +
         const freud._box.Box & getBox() const
         void reset()
-        void compute(const freud._locality.NeighborQuery*) except +
+        void compute(const freud._locality.NeighborQuery*,
+                     float*) except +
```
vyasr:
```suggestion
                     const float*) except +
```

tests/test_density_GaussianDensity.py
```diff
@@ -103,6 +103,38 @@ def test_sum_3d(self):
...
+        r_max = 49
+        sigma = 10
+        num_points = 1
+        box_size = width
+        box, points = freud.data.make_random_system(
+            box_size, num_points, is2D=True)
+        values = np.random.rand(num_points)
+
+        gd = freud.density.GaussianDensity(width, r_max, sigma)
+        gd.compute(system=(box, points), values)
```
vyasr:
```suggestion
        gd.compute(system, values)
```

This line is the reason your tests are currently failing, you're running into test errors because you're providing system as a kwarg but not values.

tests/test_density_GaussianDensity.py
```diff
@@ -103,6 +103,38 @@ def test_sum_3d(self):
...
+        r_max = 49
+        sigma = 10
+        num_points = 1
+        box_size = width
+        box, points = freud.data.make_random_system(
+            box_size, num_points, is2D=False)
+        values = np.random.rand(num_points)
+
+        gd = freud.density.GaussianDensity(width, r_max, sigma)
+        gd.compute(system=(box, points), values)
```
vyasr:
```suggestion
        gd.compute(system, values)
```

Same as above.

tests/test_density_GaussianDensity.py
```diff
@@ -103,6 +103,38 @@ def test_sum_3d(self):
...
         assert np.isclose(np.sum(gd.density), 1, atol=1e-4)
 
+    def test_sum_val_2d(self):
+        # Ensure that the Gaussian convolution sums to average of the values
+        width = 100
+        r_max = 49
+        sigma = 10
+        num_points = 1
+        box_size = width
+        box, points = freud.data.make_random_system(
```
vyasr:
```suggestion
        system = freud.data.make_random_system(
```

Since you're just handing the system around without using the box or points independently, you can just pass the tuple around.

tests/test_density_GaussianDensity.py
```diff
@@ -103,6 +103,38 @@ def test_sum_3d(self):
...
+        assert np.isclose(np.sum(gd.density), np.mean(values), atol=1e-4)
+
+    def test_sum_val_3d(self):
+        # Ensure that the Gaussian convolution sums to average of the values
+        width = 100
+        r_max = 49
+        sigma = 10
+        num_points = 1
+        box_size = width
+        box, points = freud.data.make_random_system(
```
vyasr:
```suggestion
        system = freud.data.make_random_system(
```

vyasr (changes requested):
This is looking really good! I've left some comments that you should be able to directly apply. One additional request: can you update the docstring for `GaussianDensity` to include the formula (just the simple product-sum from #666)? It's not a complicated formula, but it's always helpful for new users to be able to associate the code with the underlying math without having to think too hard.

vyasr:
@pbuslaev I wasn't sure if you had enabled our modification of your branch. You did, so I went ahead and applied my proposed edits and merged in master for you. You can still see all my comments (even though they're marked as resolved) if you're curious what I changed. Only outstanding request I have is to update the documentation. Specifically, above the `Args` section of the `GaussianDensity` docstring if you could add an integral (sum) showing what the GaussianDensity actually does, and how the default convolution kernel is the indicator function but can be modified using the `values` argument, that would be fantastic!

cpp/density/GaussianDensity.cc
```diff
@@ -83,7 +83,9 @@ void GaussianDensity::compute(const freud::locality::NeighborQuery* nq)
         // for each reference point
         for (size_t idx = begin; idx < end; ++idx)
         {
-            const vec3<float> point = (*nq)[idx];
+            const vec3<float> point = (*nq)[idx]; 
+            float value = (values != NULL) ? values[idx] : 1.0;
```
bdice:
```suggestion
            const float value = (values != NULL) ? values[idx] : 1.0f;
```

freud/density.pyx
```diff
@@ -228,17 +228,31 @@ cdef class GaussianDensity(_Compute):
...
         """
         cdef freud.locality.NeighborQuery nq = \
             freud.locality.NeighborQuery.from_system(system)
-        self.thisptr.compute(nq.get_ptr())
+
+        cdef float* l_values_ptr = NULL
+        cdef float[::1] l_values
+        if values is not None:
+            l_values = freud.util._convert_array(
+                values, shape=(nq.points.shape[0], ), dtype=np.float32)
```
bdice: We can leave off the `dtype=np.float32` since it's currently the default.
```suggestion
                values, shape=(nq.points.shape[0], ))
```

tests/test_density_GaussianDensity.py
```diff
@@ -103,6 +103,38 @@ def test_sum_3d(self):
...
+        num_points = 1
+        box_size = width
+        system = freud.data.make_random_system(
+            box_size, num_points, is2D=True)
+        values = np.random.rand(num_points)
+
+        gd = freud.density.GaussianDensity(width, r_max, sigma)
+        gd.compute(system, values)
+        # This has discretization error as well as single-precision error
+        assert np.isclose(np.sum(gd.density), np.mean(values), atol=1e-4)
```
bdice: I think this only works because `num_points = 1`. I think the total should be `num_points * np.mean(values)`, otherwise written as `np.sum(values)`. We should update the tests to `assert np.isclose(np.sum(gd.density), num_points, atol=1e-4)` above as well.
```suggestion
        assert np.isclose(np.sum(gd.density), np.sum(values), atol=1e-4)
```

tests/test_density_GaussianDensity.py
```diff
@@ -103,6 +103,38 @@ def test_sum_3d(self):
...
+        num_points = 1
+        box_size = width
+        system = freud.data.make_random_system(
+            box_size, num_points, is2D=False)
+        values = np.random.rand(num_points)
+
+        gd = freud.density.GaussianDensity(width, r_max, sigma)
+        gd.compute(system, values)
+        # This has discretization error as well as single-precision error
+        assert np.isclose(np.sum(gd.density), np.mean(values), atol=1e-4)
```
bdice:
```suggestion
        assert np.isclose(np.sum(gd.density), np.sum(values), atol=1e-4)
```

freud/density.pyx
```diff
@@ -187,8 +187,17 @@ cdef class GaussianDensity(_Compute):
...
-    the grid cell from the center of the Gaussian. The resulting data is a
-    regular grid of particle densities that can be used in standard algorithms
+    the grid cell from the center of the Gaussian. The weights for the
+    Gaussians could be additionally specified in the compute method. The
+    convolution of the weights with the Gaussians is calculated in this case:
+
+    .. math::
+
+        p(r) = \sum_i \frac{1}{2\pi \sigma^2}
+        \exp \left(-\frac{(r-r_i)^2}{2\sigma^2}\right) p_i
```
bdice: Added vector notation -- I'm going to go ahead and commit this.
```suggestion
        p(\vec{r}) = \sum_i \frac{1}{2\pi \sigma^2}
        \exp \left(-\frac{(\vec{r}-\vec{r}_i)^2}{2\sigma^2}\right) p_i
```

bdice (approved):
Really great work @pbuslaev, thanks for the PR! I made some final updates to make the tests a little more rigorous, and I approve.

bdice:
@pbuslaev One more request: please complete the PR checklist below (you might have deleted this checklist when creating the PR message). The primary things are indicating that you agree to the CONTRIBUTING document and updating the changelog and credits to include your contributions.

## Checklist:

bdice:
@pbuslaev Don't worry about the failing CI checks from ReadTheDocs, I will rebase this PR on `master` and fix that before merging.

pbuslaev:
> @pbuslaev One more request: please complete the PR checklist below (you might have deleted this checklist when creating the PR message). The primary things are indicating that you agree to the CONTRIBUTING document and updating the changelog and credits to include your contributions.

For some reason, I can't select anything in the checklist. However, I did everything according to it

## Checklist: 

 - [x]  I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
 - [x]  My code follows the code style of this project.
 - [x]  I have updated the documentation (if relevant).
 - [x]  I have added tests that cover my changes (if relevant).
 - [x]  All new and existing tests passed.
 - [x]  I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst). 
 - [x]  I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

bdice:
@pbuslaev Fantastic, thanks! I'll get this updated and merged (we changed our build system to use scikit-build while you were working on this PR). I appreciate the great work!

bdice:
This can be merged when tests pass.

PR #685, opened 2020-11-01
Feature/pre commit

vyasr:
## Description

Add support for pre-commit, including basic style checks as well as flake8, cmake-format, and clang-format.

I don't see any documentation in pre-commit for how to handle "optional" hooks conditional upon (for example) a tool being installed. In particular, I would like to enable things like cppcheck and clang-tidy, but I don't necessarily want to deal with the overhead of making sure these are installed and working on our CI rig. One option is to run pre-commit specifying which hooks to run (`pre-commit run flake8 clang-format cmake-format --all-files`) in CI, and then ask users to either use this command or install additional tools to their machine if they want to run the full pre-commit suite. This feels a bit clunky, though, and I would prefer to avoid this, but it maybe preferable to not having such checks in place at all. Open to discussion from reviewers.

I will update documentation etc once we come to a consensus.

**Update: All C++ checks have been moved to #687. This PR now only includes checks of the Python code and CMake linting.**

## Motivation and Context

`pre-commit` is the modern solution for hooks, `flake8`'s built-in hook support is deprecated. Moving to `pre-commit` also allows us to easily integrate checks for other languages, like C++.

## How Has This Been Tested?

Runs locally and CI has been updated so will be tested there.

## Types of changes

## Checklist:


bdice:
@vyasr “Optional” checks may not be a good idea if it can cause failures that are unrelated to the changes in a particular commit (caused by previous changes).

vyasr:
My question is basically how much burden we want to put on developers, or at least how we choose to mitigate that burden. Our typical pre-commit checks like `flake8` are easy to require because anyone using freud should have no trouble with `pip install flake8`. What happens now that I've introduced pre-commit checks for `clang-format`, or if I add further checks for `clang-tidy` or `cppcheck`? If developers run `pre-commit install`, they will no longer even be able to `git commit` without having these installed. They could choose to manually specifying which hooks to run to `pre-commit run HOOK1 [HOOK2] ... --all-files`, or alternatively, they could choose to just push and rely on our CI. Are these solutions acceptable to us? I’m fine just documenting one of these solutions but i want to think through this and make sure we’re not making development unnecessarily cumbersome.

bdice:
The solution we've had thus far is to only _require_ style checks that are easily installed (e.g. from PyPI/conda-forge), and to run C++ code cleanup manually when we release. Our CI won't fail if someone commits poorly-formatted C++ code, and I think that's ideal because it reduces the burden on developers. If we enable all checks in pre-commit, a small change to docs or the Cython code would also require all C++ linters to be installed. I would be happy to have a script that runs all the C++ checks that can be used during the release process, but I'm hesitant to require all of that in pre-commit.

vyasr:
I am also hesitant to require users to _install_ extra things for precisely the types of cases you describe. However, I would also prefer that CI fails if users commit poorly formatted C++ code. A compromise would be to include a separate pre-commit config file (e.g. `.pre-commit-config-dev.yaml`) that can be used by CI (or devs locally). The `pre-commit run` command accepts an alternative config file, which could include the full suite of hooks.

b-butler:
I think only using software that is easy to install by default is the right way to go. It would be nice to have more extensive linting or reformatting as an option perhaps in a separate config file, but adding more difficulty in contributing, I don't believe is a good idea.

vyasr:
I'm splitting this PR into two. This PR will only be concerned with Python-related hooks, and these will be part of the default pre-commit infrastructure that all contributors are expected to use. I'll make a follow-up PR shortly with the C++-related changes.

vyasr:
Windows builds are failing due to more certificate issues, specifically when trying to build MDAnalysis (which I think is trying to download the most recently uploaded version of numpy, 1.19.4, as of a few hours ago). Preinstalling numpy does not help, I'll try to figure out the certificate issue later.

vyasr:
Just to be 100% sure, I reran the last Windows build that succeeded on master, and [it does indeed fail now](https://app.circleci.com/pipelines/github/glotzerlab/freud/2278/workflows/d1dfd2ed-186f-4cc2-bc0d-3985eb3e3a4e/jobs/7415). So whatever has changed here isn't really related. I'd be OK with merging this PR (I can revert the last commit) and trying to resolve the Windows issues separately, if reviewers approve.

.circleci/config.yml
```diff
@@ -44,7 +44,7 @@ references:
     run:
       name: Install dependencies
       command: |
-        # Since we're testing against a very old skbuild, wheels may not exist
+        # Since we're testing against a very old cmake, wheels may not exist
```
bdice, line 47: FYI: cmake 3.6.3.post1 has builds tagged for different Python versions (`py27`, `py34`, ..., `py37`) but [cmake >=3.14 uses universal wheels](https://pypi.org/project/cmake/3.14.3/#files), so the builds are tagged `py3-none`. The problem of "wheels not existing for newer minor version of Python" should disappear once we require cmake >=3.14. Since this comment is _already_ quite verbose, it might be useful to add that information.

.circleci/config.yml
```diff
@@ -69,28 +69,19 @@ references:
...
+
+  pre-test-checks: &pre-test-checks
     run:
-      name: flake8
+      name: Run all static checks
       command: |
-        python${PYVER} -m flake8 --show-source .
+        # Modifying path in previous steps doesn't seem to hold, so specifying
+        # the full path explicitly here seems cleaner.
+        "`python${PYVER} -m site --user-base`/bin/pre-commit" run --all-files
```
bdice: Changes to environment variables are not persistent between "run" steps in CircleCI. This might also be fixed by adding `conda activate base` to this step, which should alter the path.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+    -   id: debug-statements
+-   repo: local
+    hooks:
+    -   id: flake8
+        name: flake8
+        entry: flake8
+        language: python
+        types: [python]
+        args:
+        - --show-source
```
bdice, line 18: Can this option be configured with `setup.cfg` or `pyproject.toml` instead?

doc/source/reference/development/howtoadd.rst
```diff
@@ -11,12 +11,19 @@ Python
 Python (and Cython) code in **freud** should follow `PEP 8 <https://www.python.org/dev/peps/pep-0008/>`_.
 
 During continuous integration (CI), all Python and Cython code in **freud** is tested with `flake8 <http://flake8.pycqa.org/>`_ to ensure PEP 8 compliance.
-It is strongly recommended to `set up a pre-commit hook <http://flake8.pycqa.org/en/latest/user/using-hooks.html>`_ to ensure code is compliant before pushing to the repository:
+Additionally, all CMake code is tested using `cmakelang <https://cmake-format.readthedocs.io/en/latest/index.html>`__.
```
bdice: This change specifies both the project and tool name, for easier googling.
```suggestion
Additionally, all CMake code is tested using `cmakelang's cmake-format <https://cmake-format.readthedocs.io/en/latest/index.html>`__.
```

doc/source/reference/development/howtoadd.rst
```diff
@@ -11,12 +11,19 @@ Python
...
+It is strongly recommended to `set up a pre-commit hook <https://pre-commit.com/>`_ to ensure code is compliant before pushing to the repository:
 
 .. code-block:: bash
 
-    flake8 --install-hook git
-    git config --bool flake8.strict true
+    pip install -r requirements-precommit.txt
+    pre-commit install
+
+To manually run `Pre-commit`_ for all the files present in the repository, run the following command:
```
bdice: Does this link somewhere?

requirements-precommit.txt
```diff
@@ -0,0 +1,3 @@
+flake8==3.8.3
+pre-commit==2.7.1
+cmakelang==0.6.13
```
bdice: Use the latest versions or else we will get update PRs from dependabot next month. (We still might, if they update the packages.)
```suggestion
flake8==3.8.4
pre-commit==2.8.2
cmakelang==0.6.13
```

requirements-testing.txt
```diff
@@ -4,8 +4,8 @@ garnett==0.7.1
 GitPython==3.1.11
 gsd==2.3.0
 matplotlib>=3.0.0
-MDAnalysis==1.0.0
 numpy==1.19.3
+MDAnalysis==1.0.0
```
bdice: Was this reordering related to Windows tests? Revert (not alphabetical).

setup.cfg
```diff
@@ -7,8 +7,8 @@ message = Bump up to version {new_version}.
 [flake8]
 filename = *.py,*.pyx,*.pxi,*.pxd
 exclude = .eggs,*.egg,build,extern,doc/source/gettingstarted/examples
-ignore = E225,E226,E227,E741,E999,W504,N
-per-file-ignores = 
+ignore = E225,E226,E227,E741,E999,W504,N,D,RST
+per-file-ignores =
```
bdice, line 11: This removal of trailing whitespace will be reverted on the next release, because `bumpversion` always outputs trailing whitespace here... (no action needed, just FYI)

doc/source/reference/development/howtoadd.rst
bdice: Nope: 
image

setup.cfg
vyasr: Actually there is an action item here, thanks for the reminder. I meant to add setup.cfg to the exclude for the trailing-whitespace hook so that we don't go through this every time.

requirements-testing.txt
vyasr: Yep, see [my last comment above](https://github.com/glotzerlab/freud/pull/685#issuecomment-721193525), reverting it was indeed my intention.

doc/source/reference/development/howtoadd.rst
vyasr: Good catch, I copied this particular note from signac-docs and forgot to replace the link.

.pre-commit-config.yaml
vyasr: It can, but that would affect all runs of `flake8` in the repo, including manual user runs. Is that really what we want? We don't do that her or for any other repo because it gets pretty verbose and makes it harder to work with when you need to manually go through and fix many things.

.circleci/config.yml
vyasr: We only use conda for the Windows build :P but yes, I forgot about that until after I tried this. I can make a note to that effect here.

.pre-commit-config.yaml
bdice: That's fine, good point -- marking as resolved.

.circleci/config.yml
bdice: Oh, right, ignore what I said about conda. However, it might possible to use `python -m pre-commit` instead of the path hack?

bdice (approved):
Nice, @vyasr! Approved.

bdice:
@vyasr I'll let you handle merging when you're ready, since #687 depends on this.

.circleci/config.yml
vyasr: I'm dumb. I had tried exactly that and it failed and I was too lazy to figure out why. Of course, it's because python package names don't have hyphens; `python -m pre_commit` works.

PR #687, opened 2020-11-03
Feature/pre commit cpp

### LINKED ISSUE #629: Introduce pipeline for undefined behavior in C++

### LINKED ISSUE #673: Re-enable clang-format in CI
Currently `clang-format` is disabled in the freud CI due to some past compatibility issues. @vyasr suggested we should re-enable it. I might consider enabling pre-commit for the whole repo to run `clang-format` and other checks.

vyasr:
## Description

*This should be merged after #685. Once that PR is merged, this one can be updated and additional documentation added to describe how the C++ hooks can be run.*

Adds additional validation of the C++ code (clang-tidy, cppcheck, clang-format) and updates all code so that these validation steps succeed.

## Motivation and Context

Resolves: #629, Resolves: #673

I think this is as much as we can do with purely static checks, and as part of CI. As we move forward if we run into more in-depth issues we can consider adding config files for valgrind or similar tools to run more dynamic checks, but those will be more extensive.

## How Has This Been Tested?

CI is updated to run these checks.

## Types of changes

## Checklist:


CMake/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@
 # List commands don't propagate up, so we have to manually set the parent scope
 # variable after.
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/CMake")
```
tommy-waltmann:
```suggestion
list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/CMake" PARENT_SCOPE)
```
I think this can all be done in one line by applying the suggestion and then remove the set() statement after this.

cpp/box/Box.h
```diff
@@ -5,10 +5,10 @@
 #define BOX_H
 
 #include "utils.h"
+#include <algorithm>
 #include <complex>
 #include <sstream>
 #include <stdexcept>
-#include <algorithm>
```
tommy-waltmann: What is the need for this?

cpp/density/LocalDensity.cc
```diff
@@ -14,7 +14,7 @@ LocalDensity::LocalDensity(float r_max, float diameter)
     : m_box(box::Box()), m_r_max(r_max), m_diameter(diameter)
 {}
 
-LocalDensity::~LocalDensity() {}
+LocalDensity::~LocalDensity() = default;
```
tommy-waltmann: Since this PR seems to be mostly about cleaning source code, we should move this to the header

cpp/environment/Registration.h
```diff
@@ -307,7 +313,7 @@ class RegisterBruteForce
         return makeVec3Matrix(T);
     }
 
-    double getRMSD()
+    float getRMSD() const
```
tommy-waltmann, line 316: There are many other places in this file where floating-point values are changed from using double to using float. What is the logic behind this?

cpp/locality/AABBQuery.h
```diff
@@ -115,31 +118,31 @@ class AABBIterator : public NeighborQueryPerPointIterator
...
+    ~AABBIterator() override = default;
 
     //! Computes the image vectors to query for
     void updateImageVectors(float r_max, bool _check_r_max = true);
 
 protected:
     const AABBQuery* m_aabb_query;         //!< Link to the AABBQuery object
     std::vector<vec3<float>> m_image_list; //!< List of translation vectors
-    unsigned int m_n_images;               //!< The number of image vectors to check
+    unsigned int m_n_images {0};           //!< The number of image vectors to check
```
tommy-waltmann, line 137: Does this actually have any different behavior than what was there before? Initializing instance variables like this doesn't seem like good code convention to me.

cpp/locality/BondHistogramCompute.h
```diff
@@ -20,13 +20,11 @@ class BondHistogramCompute
...
-    BondHistogramCompute()
-        : m_box(box::Box()), m_frame_counter(0), m_n_points(0), m_n_query_points(0), m_reduce(true),
-          m_histogram(), m_local_histograms()
-    {}
+    BondHistogramCompute() : m_box(box::Box()), m_histogram(), m_local_histograms() {}
 
     //! Destructor
-    virtual ~BondHistogramCompute() {};
+    virtual ~BondHistogramCompute() = default;
+    ;
```
tommy-waltmann: What's with the random semicolon here?

cpp/locality/BondHistogramCompute.h
```diff
@@ -115,17 +113,17 @@ class BondHistogramCompute
...
 protected:
     box::Box m_box;
-    unsigned int m_frame_counter;  //!< Number of frames calculated.
-    unsigned int m_n_points;       //!< The number of points.
-    unsigned int m_n_query_points; //!< The number of query points.
-    bool m_reduce;                 //!< Whether or not the histogram needs to be reduced.
+    unsigned int m_frame_counter {0};  //!< Number of frames calculated.
+    unsigned int m_n_points {0};       //!< The number of points.
+    unsigned int m_n_query_points {0}; //!< The number of query points.
+    bool m_reduce {true};              //!< Whether or not the histogram needs to be reduced.
```
tommy-waltmann, line 118: I mentioned this before, but what is the logic behind initializing instance variables this way?

cpp/order/HexaticTranslational.h
```diff
@@ -26,10 +26,10 @@ template<typename T> class HexaticTranslational
 {
 public:
     //! Constructor
-    HexaticTranslational(T k, bool weighted = false) : m_k(k), m_weighted(weighted) {}
+    explicit HexaticTranslational(T k, bool weighted = false) : m_k(k), m_weighted(weighted) {}
```
tommy-waltmann, line 29: This is more of a cpp question, but what does explicit do to a contructor?

tommy-waltmann (commented):
Wow .. that was a lot of files. I mostly had some suggestions and questions, but there was one small thing that must be addressed.

cpp/order/HexaticTranslational.h
bdice: `explicit` is a good thing to add to one-argument constructors because it prevents unintended typecasting. (The second argument has a default value so it's like having a one argument constructor.)

https://en.cppreference.com/w/cpp/language/explicit
https://www.reddit.com/r/Cplusplus/comments/ejuf7k/must_i_use_explicit_keyword_in_c/

CMake/CMakeLists.txt
vyasr: Unfortunately no, you can't do this. `list` does not have a mechanism to propagate changes up in scope, see the [last sentence of the Introduction](https://cmake.org/cmake/help/latest/command/list.html#introduction).

cpp/box/Box.h
vyasr: `clang-format` has a preferred sort order for includes (alphabetical). Up to now reformatting has only been something we do on releases, and this include was probably added since the last release by #590.

vyasr:
> Wow .. that was a lot of files. I mostly had some suggestions and questions, but there was one small thing that must be addressed.

I did promise a hefty PR on Slack ;)

cpp/density/LocalDensity.cc
vyasr: That's a good call, I'll make that change. Please leave this conversation unresolved until then, thanks.

.clang-tidy
```diff
@@ -0,0 +1,23 @@
+# Specific checks ignored:
+# readability-avoid-const-params-in-decls: While it may be semantically confusing, it's easier to read when declaration and definition consts match.
+# readability-redundant-member-init: If the constructor changes or if there are multiple constructor signatures, prefer to be explicit.
+# modernize-use-trailing-return-type: This syntax is still generally unfamiliar to many developers and will make the code harder to maintain.
+# misc-no-recursion: Rewriting things like the NeighborList to avoid recursion is out of scope for now.
+# readability-magic-numbers: Wigner3j.
```
bdice:
```suggestion
# readability-magic-numbers: Wigner3j contains (necessary) magic numbers.
```

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,40 @@
+# This file contains pre-commit hooks for checking the C++ code. Since these
+# hooks require tools that may not trivial to install (unlike the Python hooks
+# like flake8), we keep these in a separate file to be applied manually.
```
bdice:
```suggestion
# This file contains pre-commit hooks for checking the C++ code. Since these
# hooks require tools that may not be trivial to install (unlike the Python
# hooks like flake8), we keep these in a separate file to be applied manually.
```

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,40 @@
...
+    rev: python
+    hooks:
+      - id: clang-format
+        # Unclear if this is necessary, it seems like clang-format uses
+        # .clang-format if found by default, at least for newer versions. I'll
+        # specify this here just in case.
+        args:
+        - --style=file
+        - -i
+        exclude: '.*extern/.*'
```
bdice: Is the beginning `.*` necessary?
```suggestion
        exclude: 'extern/.*'
```

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - id: clang-tidy
+        # Ignore external files and those taken verbatim from HOOMD.
```
bdice, line 16: I think that all of these files have nontrivial changes relative to their HOOMD counterparts at this point, so it would be fine with me to include them.

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,40 @@
...
+        args:
+        - cpp/
+        - --enable=warning,style,performance,portability # Ignore unused (could be used by Cython).
+        - --std=c++14
+        - --language=c++ # Force C++ for .h header files.
+        - --inline-suppr # Enable inline suppresions of warnings.
+        - -icpp/locality/AABBTree.cc
+        - --suppress=*:cpp/locality/AABB.h
+        - --suppress=*:cpp/locality/AABBTree.h
+        - --suppress=*:*extern/*
```
bdice:
```suggestion
        - --suppress=*:extern/*
```

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,40 @@
...
+# Set up the cppcheck hook manually, the one provided by the above repo is configured too rigidly.
+-   repo: local
+    hooks:
+      - id: cppcheck
+        # Ignore external files and those taken verbatim from HOOMD.
+        name: cppcheck
+        entry: cppcheck
+        language: system
+        files: '.*\.(h|cc)'
+        exclude: '.*AABB\.h|.*extern.*|.*AABBTree\.(h|cc)|VectorMath.h'
```
bdice:
```suggestion
        exclude: '.*AABB\.h|extern.*|.*AABBTree\.(h|cc)|VectorMath.h'
```

cpp/environment/Registration.h
vyasr: A lot of objects were being implicitly casted, e.g. a function returning `double` would be assigned to a `float`. In many places this was being done multiple steps removed from the initial place (e.g. in a member variable being assigned a map containing the output of a function returning double), but none of these cases that I could see had any reason to prefer doubles for numerical precision reasons (which Voronoi does and therefore explicitly chooses to use doubles for). For instance, if you trace the rmsd all the way back, it ultimately goes back to just adding a large number of `float`s together. In theory I suppose there is some precision loss, but I don't think there's any reason to expect that it's anything more than noise for the use case here. I changed everything to be a float to be consistent with the rest of freud, with the anticipation that implementing #464 would allow us to regain true double precision support if really necessary.

cpp/box/Box.h
```diff
@@ -367,7 +368,7 @@ class Box
      *  \param masses Optional array of masses, of length Nvecs
      *  \return Center of mass as a vec3<float>
      */
-    vec3<float> centerOfMass(vec3<float>* vecs, size_t Nvecs, float* masses = NULL) const
+    vec3<float> centerOfMass(vec3<float>* vecs, size_t Nvecs, const float* masses = nullptr) const
```
bdice, line 371: I'm not actually sure if this is a correct change. I believe there's some magic with Cython requiring NULL instead of nullptr. I briefly searched for this but couldn't come up with a reference.

cpp/environment/LocalBondProjection.cc
```diff
@@ -11,9 +11,9 @@
 
 namespace freud { namespace environment {
 
-LocalBondProjection::LocalBondProjection() {}
+LocalBondProjection::LocalBondProjection() = default;
 
-LocalBondProjection::~LocalBondProjection() {}
+LocalBondProjection::~LocalBondProjection() = default;
```
bdice: This should go in the header.

cpp/locality/AABBQuery.h
vyasr: Yeah actually of all the changes that clang-tidy and cppcheck recommended, this syntax was the only thing that was completely new to me. It's documented [here](https://en.cppreference.com/w/cpp/language/data_members) (in the member initialization section).

I was at first hesitant to make this change, but I was convinced because it does have significant benefits over just using initializer lists in the constructor. This initializer becomes the default for all constructors, but can be overridden by constructor initializer lists. The big benefit here is that it prevents any issues where a constructor _forgets_ to initialize parameters, something that was the case in a few places in freud. The ancillary benefit is that it's nice and readable here, the default value is right next to the declaration (and hopefully, the docstring).

cpp/locality/BondHistogramCompute.h
vyasr: Typo :) I'll fix.

cpp/locality/BondHistogramCompute.h
vyasr: See the other comment.

cpp/order/HexaticTranslational.h
vyasr: Yeah you don't want to be able to do something like `Hexatic hex = 6`, which would otherwise be possible.

cpp/environment/MatchEnv.h
```diff
@@ -27,23 +27,22 @@ namespace freud { namespace environment {
...
     }
 
-    unsigned int env_ind;          //!< The index of the environment
+    unsigned int env_ind {0};      //!< The index of the environment
     std::vector<vec3<float>> vecs; //!< The vectors that define the environment
     //! Is this environment a ghost? Do we ignore it when we compute actual
     //  physical quantities associated with all environments?
     bool ghost;
-    unsigned int num_vecs; //!< The number of vectors defining the environment currently
+    unsigned int num_vecs {0}; //!< The number of vectors defining the environment currently
```
bdice:
```suggestion
    unsigned int num_vecs {0}; //!< The number of vectors currently defining the environment
```

vyasr:
Since @bdice asked and it's probably useful information to maintain, I'll say that in terms of where the changes came from these changes were probably about 50-30-20 clang-tidy-me-cppcheck, noting that I ran clang-tidy first so I might have fixed things that cppcheck would have found. Both tools apply quite different heuristics and it's definitely worth using both. clang-tidy was much better at finding things like variables that could be const, places to use range-based loops, and in general identifying places where more modern C++ syntax could be exploited. cppcheck was more useful for identifying unintialized variables, places we could use std::algorithm, and in general performing more _correctness_ checks. Neither of them consistently finds everything since they must use plenty of complex heuristics internally, so a large number of the changes were me taking their suggestion in one place and applying it to many others. Both tools are definitely handy and worth using.

There are also a handful of changes (extensive use of static_cast in particular) that I made independently since they helped improve code quality.

cpp/locality/AABBQuery.cc
```diff
@@ -183,7 +183,7 @@ NeighborBond AABBQueryBallIterator::next()
                         // Check ii exclusion before including the pair.
                         if (r_sq < r_max_sq && r_sq >= r_min_sq)
                         {
-                            return NeighborBond(m_query_point_idx, j, std::sqrt(r_sq));
+                            return {m_query_point_idx, j, std::sqrt(r_sq)};
```
bdice: We lost the name of the class here...

cpp/locality/AABBQuery.cc
```diff
@@ -275,15 +279,16 @@ NeighborBond AABBQueryIterator::next()
...
-                     || ((m_all_distances.size() - m_query_points_below_r_min.size()) >= m_num_neighbors))
+
+            if ((m_r_cur >= m_r_max) || (m_r_cur >= max_plane_distance)
+                || ((m_all_distances.size() - m_query_points_below_r_min.size()) >= m_num_neighbors))
             {
                 // Once this condition is reached, either we found enough
                 // neighbors beyond the normal min_plane_distance
                 // condition or we conclude that there are not enough
                 // neighbors left in the system.
                 for (std::map<unsigned int, float>::const_iterator it(m_all_distances.begin());
```
bdice: You can probably use `auto it(m_all_distances.cbegin())`? Just get a const iterator with `cbegin` / `cend`.
```suggestion
                for (auto it(m_all_distances.cbegin());
```

cpp/locality/LinkCell.cc
```diff
@@ -298,16 +311,17 @@ void LinkCell::computeCellList(const vec3<float>* points, unsigned int n_points)
...
     {
         m_cell_list[n_points + cell] = LINK_CELL_TERMINATOR;
     }
 
-    // generate the cell list
-    for (int i = n_points - 1; i >= 0; i--)
+    // Generate the cell list. The condition in the for loop exploits the fact
+    // that i-- returns the pre-decrement value, so when i first reaches 0 the
+    // comparison will be 1 > 0 and ensure that the 0 case runs.
+    for (unsigned int i = n_points; i-- > 0;)
```
bdice: I'm not really a fan of this. The original `for` was easier to understand and didn't require a comment to explain the order of decrement/return.

cpp/locality/LinkCell.cc
```diff
@@ -322,10 +336,10 @@ vec3<unsigned int> LinkCell::indexToCoord(unsigned int x) const
     // For backwards compatibility with the Index1D layout, the indices and
     // the dimensions are passed in reverse to the indexer. Changing this would
     // also require updating the logic in IteratorCellShell.
-    return vec3<unsigned int>(coord[2], coord[1], coord[0]);
+    return {coord[2], coord[1], coord[0]};
```
bdice: Another case where we're losing the class name. I think that's obscuring important information and I find it less readable.

cpp/locality/LinkCell.cc
```diff
@@ -500,7 +518,7 @@ NeighborBond LinkCellQueryBallIterator::next()
 
             if (r_sq < r_max_sq && r_sq >= r_min_sq)
             {
-                return NeighborBond(m_query_point_idx, j, std::sqrt(r_sq));
+                return {m_query_point_idx, j, std::sqrt(r_sq)};
```
bdice: Lost class name.

cpp/locality/LinkCell.h
```diff
@@ -179,7 +179,7 @@ class LinkCell : public NeighborQuery
 {
 public:
     //! iterator to iterate over particles in the cell
-    typedef IteratorLinkCell iteratorcell;
+    using iteratorcell = IteratorLinkCell;
```
bdice: Is this definition actually useful?

cpp/locality/NeighborList.cc
```diff
@@ -77,12 +82,11 @@ void NeighborList::updateSegmentCounts() const
     {
         m_counts.prepare(m_num_query_points);
         m_segments.prepare(m_num_query_points);
-        int index(-1);
         int last_index(-1);
         unsigned int counter(0);
         for (unsigned int i = 0; i < getNumBonds(); i++)
         {
-            index = m_neighbors(i, 0);
+            int index = m_neighbors(i, 0);
```
bdice: This is `const` within the body of the loop.
```suggestion
            const int index = m_neighbors(i, 0);
```

cpp/locality/NeighborList.cc
```diff
@@ -106,44 +110,55 @@ void NeighborList::updateSegmentCounts() const
...
 {
-    // Can't use vector<bool> because that is specialized for compact storage
-    std::unique_ptr<bool[]> dist_filter(new bool[getNumBonds()]);
+    std::vector<bool> dist_filter(getNumBonds());
     for (unsigned int i(0); i < getNumBonds(); ++i)
     {
         dist_filter[i] = (m_distances[i] >= r_min && m_distances[i] < r_max);
     }
-    return filter(dist_filter.get());
+    return filter(dist_filter.begin());
```
bdice: const iterator?
```suggestion
    return filter(dist_filter.cbegin());
```

cpp/locality/NeighborList.cc
```diff
@@ -182,22 +197,29 @@ void NeighborList::copy(const NeighborList& other)
...
+    {
         throw std::runtime_error("NeighborList found inconsistent array sizes.");
+    }
     if (num_points != m_num_points)
+    {
         throw std::runtime_error("NeighborList found inconsistent array sizes.");
+    }
 }
 
 unsigned int NeighborList::bisection_search(unsigned int val, unsigned int left, unsigned int right) const
```
bdice, line 210: It might be possible to write this to use `std::lower_bound`. https://en.cppreference.com/w/cpp/algorithm/lower_bound

cpp/locality/NeighborList.cc
```diff
@@ -212,14 +234,14 @@ bool compareNeighborDistance(const NeighborBond& left, const NeighborBond& right
...
+    if (!left.empty() && !right.empty())
     {
         return compareNeighborBond(left[0], right[0]);
     }
-    else
-    {
-        return left.size() < right.size();
-    }
+    // cppcheck erroneously flags this check as redundant because it follows
+    // both branches independently.
```
bdice: Any further insight here? I'm not sure if I understand what you mean by "it follows both branches independently."

freud/_pmft.pxd
```diff
@@ -42,8 +42,8 @@ cdef extern from "PMFTXY.h" namespace "freud::pmft":
         PMFTXY(float, float, unsigned int, unsigned int) except +
 
         void accumulate(const freud._locality.NeighborQuery*,
-                        float*,
-                        vec3[float]*,
+                        const float*,
+                        vec3[const float]*,
```
bdice: This doesn't match the others, which say `const vec3[float]*`.
```suggestion
                        const vec3[float]*,
```

cpp/util/diagonalize.cc
```diff
@@ -3,7 +3,7 @@
 namespace freud { namespace util {
 
 void diagonalize33SymmetricMatrix(const util::ManagedArray<float>& mat, util::ManagedArray<float>& eigen_vals,
-                                  util::ManagedArray<float>& eigen_vecs)
+                                  const util::ManagedArray<float>& eigen_vecs)
```
bdice: I don't think this is `const`... The eigenvectors are calculated by Eigen and written to this ManagedArray.

cpp/util/ManagedArray.h
```diff
@@ -300,12 +325,15 @@ template<typename T> class ManagedArray
...
         size_t cur_prod = 1;
         size_t idx = 0;
-        for (int i = indices.size() - 1; i >= 0; --i)
+        // In getting the linear bin, we must iterate over bins in reverse
+        // order to build up the value of cur_prod because each subsequent axis
+        // contributes less according to row-major ordering.  The condition in
+        // the for loop exploits the fact that i-- returns the pre-decrement
+        // value, so when i first reaches 0 the comparison will be 1 > 0 and
+        // ensure that the 0 case runs.
+        for (unsigned int i = indices.size(); i-- > 0;)
```
bdice: This isn't as readable as the previous code, especially for new developers. See other comment above.

cpp/util/ManagedArray.h
```diff
@@ -261,12 +283,15 @@ template<typename T> class ManagedArray
...
```
bdice: This isn't as readable as the previous code, especially for new developers. See other comment above.

cpp/util/ManagedArray.h
```diff
@@ -235,11 +261,7 @@ template<typename T> class ManagedArray
      */
     static inline std::vector<size_t> getMultiIndex(std::vector<size_t> shape, size_t index)
     {
-        size_t cur_prod = 1;
-        for (auto it = shape.begin(); it != shape.end(); ++it)
-        {
-            cur_prod *= *it;
-        }
+        size_t cur_prod = std::accumulate(shape.begin(), shape.end(), 1, std::multiplies<>());
```
bdice: Use const iterators. Also does it make sense to name it "_current_" product now that it's computed in one line?

```suggestion
        size_t cur_prod = std::accumulate(shape.cbegin(), shape.cend(), 1, std::multiplies<>());
```

cpp/util/ManagedArray.h
```diff
@@ -218,7 +238,13 @@ template<typename T> class ManagedArray
...
         size_t cur_prod = 1;
         size_t idx = 0;
-        for (int i = indices.size() - 1; i >= 0; --i)
+        // In getting the linear bin, we must iterate over bins in reverse
+        // order to build up the value of cur_prod because each subsequent axis
+        // contributes less according to row-major ordering.  The condition in
+        // the for loop exploits the fact that i-- returns the pre-decrement
+        // value, so when i first reaches 0 the comparison will be 1 > 0 and
+        // ensure that the 0 case runs.
+        for (unsigned int i = indices.size(); i-- > 0;)
```
bdice: This isn't as readable as the previous code, especially for new developers. See other comment above.

cpp/util/ManagedArray.h
```diff
@@ -205,7 +219,13 @@ template<typename T> class ManagedArray
...
         size_t cur_prod = 1;
         size_t idx = 0;
-        for (int i = indices.size() - 1; i >= 0; --i)
+        // In getting the linear bin, we must iterate over bins in reverse
+        // order to build up the value of cur_prod because each subsequent axis
+        // contributes less according to row-major ordering. The condition in
+        // the for loop exploits the fact that i-- returns the pre-decrement
+        // value, so when i first reaches 0 the comparison will be 1 > 0 and
+        // ensure that the 0 case runs.
+        for (unsigned int i = indices.size(); i-- > 0;)
```
bdice: This isn't as readable as the previous code, especially for new developers. See other comment above.

cpp/util/ManagedArray.h
```diff
@@ -96,8 +98,14 @@ template<typename T> class ManagedArray
                 (*m_size) *= (*m_shape)[i];
             }
 
-            m_data = std::shared_ptr<std::shared_ptr<T>>(
-                new std::shared_ptr<T>(new T[size()], std::default_delete<T[]>()));
+            // We make use of C-style arrays here rather than any alternative
+            // because we need the underlying data representation to be
+            // compatible with numpy on the Python side. We _could_ do this
+            // with a different data structure like std::vector, but it would
```
bdice, line 104: Or `std::span`, coming soon to a C++20 near you. https://en.cppreference.com/w/cpp/container/span

Multi-dimensional spans have been proposed as well.

cpp/util/Histogram.h
```diff
@@ -22,20 +23,22 @@ namespace freud { namespace util {
  */
 template<typename T> struct Weight
 {
-    Weight() : value(1), is_default(true) {}
-    Weight(T value) : value(value), is_default(false) {}
+    Weight() : value(1) {}
```
bdice: Is it possible to set the default weight below where the member is declared, like `T value {1};` or is that not possible because `T` is a template type parameter?

bdice (changes requested):
Some initial comments are attached. I covered about 70% of the diff (I lost track somewhere along the way, I'll use the "Viewed" checkbox on the next pass). I'll need to take another pass through this after we discuss/resolve some of these comments. Nice work, these are excellent improvements to the core of freud!

CMake/CMakeLists.txt
tommy-waltmann: What you can do is `set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR} ${CMAKE_MODULE_PATH} PARENT_SCOPE)` if you want it to be one line. My original suggestion was wrong.

cpp/locality/AABBQuery.h
tommy-waltmann: It also seems to help limit the size of initializer lists, which can help with readability. Let's stick with it.

cpp/order/HexaticTranslational.h
tommy-waltmann: This seems like something that should just always happen implicitly for safety reasons. I wouldn't want to have 2, 3, 4, ... argument constructors doing what Vyas showed above either.

CMake/CMakeLists.txt
vyasr: Yes, that you can do. I don't have a strong preference between these, so I'm just going to leave it as is to avoid changing, but thanks for the suggestion.

cpp/order/HexaticTranslational.h
vyasr: While it is true that the same principle applies to (n>1)-argument constructors, in practice you're not very likely to accidentally do something like `Hexatic hex = {6, true}`, so I don't think linters check for that. We could go through and mark all constructors this way, but it seems like overkill.

cpp/box/Box.h
vyasr: Cython doesn't have any concept of nullptr, so you do have to pass NULL instead everywhere in Cython. The reasons to use nullptr have more to do with properly disambiguating different function signatures (and in general avoiding ambiguities in things like casting, for instance `int x = NULL` is perfectly valid but undesirable). In practice, assuming no ambiguity is present, `NULL` is either 0 or `(void *) 0` (I believe that's implementation-defined), and the result of calling a function `f(float *x)` in Cython as `f(NULL)` is equivalent to an implicit cast to a null `float *` pointer.

The upshot of all of this is that while we do have to use NULL in Cython, it's safe to use nullptr in C++ in all the corresponding places.

cpp/locality/AABBQuery.cc
vyasr: This change is the result of [modernize-return-braced-init-list](https://clang.llvm.org/extra/clang-tidy/checks/modernize-return-braced-init-list.html). I am fine with ignoring that warning as well, I didn't have a problem with making this change in the first place but the redundancy is probably outweighed by clarity in this case.

cpp/locality/LinkCell.cc
vyasr: Yeah I think I just made this change because I found a cool way to do it :P. However, I also don't like that the original used an integer for something that should always be unsigned, I had always wanted to find a better idiom and clang-tidy complaining about narrowing was just motivation. Would you prefer
`for (unsigned int i = n_points; i != static_cast<unsigned int>(-1); --i)`?

cpp/locality/LinkCell.h
vyasr: A fair point, it doesn't shorten anything really. I can just switch this to use the fully qualified name everywhere.

cpp/locality/NeighborList.cc
vyasr: I think that's out of scope for this PR since it requires nonzero thought, but I'm not opposed to that in general.

cpp/locality/NeighborList.cc
vyasr: If we reached this point, that means that the previous `if` statement was false, which means that one of the two lists was empty. `cppcheck` looks at the two possible cases independently: if `left` is empty then `left.size() < right.size()` is trivially true, and if `right` is empty then the statement is trivially false (for slightly different reasons, hence the two different warnings I'm suppressing). We could rewrite this line using more if statements to avoid the warnings, but I find the current idiom to be readable and easy to understand (not to mention simpler than any alternative).

cpp/util/ManagedArray.h
vyasr: And by coming soon, you mean coming to your compiler of choice by... 2026. Maybe 2023, they are getting faster with updating to standards haha.

cpp/util/Histogram.h
vyasr: Nope that should work, I just didn't notice it since it didn't get flagged. Of course, that would fail if you did something weird like `Weight<std::map<T, U>> x`... but I think we can safely assume nobody will ever use a non-numeric type `T`.

cpp/locality/AABBQuery.cc
vyasr: At that point this can probably just use a range-based loop.

cpp/util/diagonalize.cc
vyasr: Great catch. clang-tidy thought this was OK because `ManagedArray.get` is marked as `const`. I _think_ this was necessary to enable all the Cython-related gymnastics that `get` was added to support, but then Eigen represents one particular case where we break this paradigm. I don't remember all the details of that, so I'll investigate whether we can convert `get` into a non-const method without causing compiler failures elsewhere.

.pre-commit-config-cpp.yaml
vyasr: So the problem is that AABB.h uses lots of non-portable intrinsics and therefore throws a bunch of warnings. I'd be fine including AABBTree.* and VectorMath and excluding AABB.h, does that sound reasonable?

cpp/util/diagonalize.cc
vyasr: Yes, so this problem is because we require a non-const pointer in Cython for constructing any views. We also need to use the `get` function internally in a few places. For some reason I guess I decided a good compromise would be to define the method as `const` despite returning a non-const pointer... not sure what I was thinking then. I'm just going to overload the method, and it'll be up to calling code to not break the underlying array with invalid access patterns.

cpp/order/HexaticTranslational.h
bdice: > This seems like something that should just always happen implicitly for safety reasons.

Did you mean _explicitly_? I think adding `explicit` is the right choice for single-argument constructors like this. As Vyas mentions, it's not really necessary for constructors with more than one argument.

cpp/box/Box.h
bdice: Sounds good to me. Resolving.

cpp/util/diagonalize.cc
vyasr: Ah now I'm really remembering all the headaches that come up here. The `ManagedArray` objects owned by `ManagedArrayContainer` are marked `const`, because they shouldn't be modified by any method of the `ManagedArrayContainer` (or really by anything at this point). However, ultimately you have to get the underlying pointer in Cython to pass to numpy, and that pointer cannot be const. All the attempts at const-ness were designed around this one point where we can't control the need for const-ness. Since the output array from `ManagedArrayContainer.__array__` is set to be read-only, I think the best thing we can do is just remove all the `const` qualifiers in the Cython util code and just rely on the fact that the only way the underlying array gets exposed is via the read-only numpy array from `__array__`. That will allow proper overloads and simplify almost everything else.

cpp/locality/LinkCell.cc
bdice: I see what you're going for. I'm not sure. We're just trading off between clarity of intent, unsigned wrapping, and narrowing conversions, and I don't see a clear winner. The last suggestion with `static_cast<unsigned int>(-1)` might be the best. Another alternative could be what you wrote with `i != std::numeric_limits<unsigned int>::max()`. Resolve this however you want.

cpp/locality/NeighborList.cc
bdice: @vyasr This is marked as resolved but I don't see the change. Did I make a mistake here?

cpp/locality/AABBQuery.cc
tommy-waltmann: I vote for the NeighborBond way here too.

cpp/util/Histogram.h
bdice: Let's put in a template check or static assert for [is_arithmetic](https://en.cppreference.com/w/cpp/types/is_arithmetic).

cpp/locality/NeighborList.cc
vyasr: Just not pushed yet. I'll push shortly.

cpp/locality/NeighborList.cc
vyasr: Issue for this was created #689

bdice:
@vyasr Can you fix the failing tests?
```
test_ld (test_environment_LocalDescriptors.TestLocalDescriptors)
Verify the behavior of LocalDescriptors by explicitly calculating ... terminate called after throwing an instance of 'std::invalid_argument'
  what():  Attempted to access index 1984 in an array of size 200
```

.circleci/config.yml
```diff
@@ -85,6 +91,10 @@ references:
         # run steps in CircleCI, so we specify the full path to pre-commit
         # explicitly here.
         python${PYVER} -m pre_commit run --all-files
+        # We need to generate the compile_commands.json database, which
+        # requires the build.
+        python${PYVER} setup.py build_ext --inplace -- -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -- -j 4
+        "`python${PYVER} -m site --user-base`/bin/pre-commit" run -c .pre-commit-config-cpp.yaml --all-files
```
bdice: Use `python -m pre_commit` here as in #685.
```suggestion
        python -m pre_commit run -c .pre-commit-config-cpp.yaml --all-files
```

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,39 @@
...
+        exclude: '.*AABB\.h|extern.*'
+        args:
+        - cpp/
+        - --enable=warning,style,performance,portability # Ignore unused (could be used by Cython).
+        - --std=c++14
+        - --language=c++ # Force C++ for .h header files.
+        - --inline-suppr # Enable inline suppresions of warnings.
+        - -icpp/locality/AABBTree.cc
+        - --suppress=*:cpp/locality/AABB.h
+        - --suppress=*:cpp/locality/AABBTree.h
```
bdice: Did you mean to make the exclusions here for `AABBTree` as well?

cpp/util/diagonalize.cc
```diff
@@ -2,7 +2,7 @@
 
 namespace freud { namespace util {
 
-void diagonalize33SymmetricMatrix(const util::ManagedArray<float>& mat, util::ManagedArray<float>& eigen_vals,
+void diagonalize33SymmetricMatrix(util::ManagedArray<float>& mat, util::ManagedArray<float>& eigen_vals,
```
bdice: If `get` has a const overload, then can this be const again? (I don't see why not)

cpp/util/VectorMath.h
```diff
@@ -45,7 +45,9 @@ template<class Real> struct vec3
     //! Swap with another vector
     void swap(vec3<Real>& v)
     {
-        Real tx, ty, tz;
+        Real tx;
+        Real ty;
+        Real tz;
         tx = v.x;
         ty = v.y;
         tz = v.z;
```
bdice: Alternatively, just replace with `std::swap`. I'm just noting refactors as I read this -- feel free to leave it unresolved and it can be handled in a subsequent PR.
```suggestion
        Real tx(v.x);
        Real ty(v.y);
        Real tz(v.z);
```

cpp/util/ManagedArray.h
```diff
@@ -300,12 +324,12 @@ template<typename T> class ManagedArray
...
 
+        size_t cur_prod = 1;
+        size_t idx = 0;
         // In getting the linear bin, we must iterate over bins in reverse
         // order to build up the value of cur_prod because each subsequent axis
         // contributes less according to row-major ordering.
-        size_t cur_prod = 1;
-        size_t idx = 0;
-        for (int i = indices.size() - 1; i >= 0; --i)
+        for (unsigned int i = indices.size(); i != static_cast<unsigned>(-1); --i)
```
bdice: This should just call the static method instead of duplicating it.

cpp/util/ManagedArray.h
```diff
@@ -261,12 +285,12 @@ template<typename T> class ManagedArray
      */
     static inline size_t getIndex(std::vector<size_t> shape, std::vector<size_t> indices)
```
bdice:
```suggestion
    static inline size_t getIndex(const std::vector<size_t>& shape, const std::vector<size_t>& indices)
```

bdice: Also change to `const&` for vector arguments in the non-static method below, and several other places in this file.

cpp/util/ManagedArray.h
```diff
@@ -218,7 +243,10 @@ template<typename T> class ManagedArray
     {
         size_t cur_prod = 1;
         size_t idx = 0;
-        for (int i = indices.size() - 1; i >= 0; --i)
+        // In getting the linear bin, we must iterate over bins in reverse
+        // order to build up the value of cur_prod because each subsequent axis
+        // contributes less according to row-major ordering.
+        for (unsigned int i = indices.size(); i != static_cast<unsigned>(-1); --i)
```
bdice: Can we use `unsigned int` instead of `unsigned`? I know they're equivalent but I don't feel comfortable not specifying `int` since we use the full name elsewhere. This applies in several places.

vyasr:
@bdice sorry didn't see that until most of the way through group meeting. Fixed.

cpp/order/SolidLiquid.h
```diff
@@ -96,12 +97,11 @@ class SolidLiquid
...
-        std::vector<unsigned int> sizes;
         auto keys = m_cluster.getClusterKeys();
-        for (auto cluster = keys.begin(); cluster != keys.end(); cluster++)
-        {
-            sizes.push_back(cluster->size());
-        }
+        std::vector<unsigned int> sizes;
+        sizes.reserve(keys.size());
+        std::transform(keys.begin(), keys.end(), std::back_inserter(sizes),
+                       [](auto& key) { return key.size(); });
```
bdice, line 104: Just initialize with the correct size (instead of reserving after construction), then use `.begin()` as the output iterator.
```suggestion
        std::vector<unsigned int> sizes(keys.size());
        std::transform(keys.cbegin(), keys.cend(), sizes.begin(),
                       [](const auto& key) { return key.size(); });
```

cpp/order/RotationalAutocorrelation.cc
```diff
@@ -70,7 +69,7 @@ void RotationalAutocorrelation::compute(const quat<float>* ref_orientations, con
         for (unsigned int b = 0; b <= m_l; b++)
         {
             unit_harmonics.push_back(std::conj(hypersphere_harmonic(xi, zeta, a, b)));
-            prefactors[a][b] = m_factorials[a] * m_factorials[m_l - a] * m_factorials[b]
+            prefactors[a][b] = static_cast<float>(m_factorials[a]) * m_factorials[m_l - a] * m_factorials[b]
                 * m_factorials[m_l - b] / (float(m_l) + 1);
```
bdice: This seems undesirable to cast only one factorial value. Make that consistent (either cast every value or cast their product). Also use `static_cast<float>(m_l)`.
```suggestion
            prefactors[a][b] = static_cast<float>(m_factorials[a]) * m_factorials[m_l - a] * m_factorials[b]
                * m_factorials[m_l - b] / (static_cast<float>(m_l) + 1);
```

cpp/order/Nematic.h
```diff
@@ -23,10 +23,11 @@ class Nematic
 {
 public:
     //! Constructor
-    Nematic(vec3<float> u);
+    explicit Nematic(const vec3<float>& u);
 
     //! Destructor
-    virtual ~Nematic() {};
+    virtual ~Nematic() = default;
+    ;
```
bdice:
```suggestion
```

cpp/order/Nematic.cc
```diff
@@ -13,7 +13,8 @@
 namespace freud { namespace order {
 
 // m_u is the molecular axis, normalized to a unit vector
-Nematic::Nematic(vec3<float> u) : m_n(0), m_u(u / std::sqrt(dot(u, u))), m_nematic_tensor_local({3, 3}) {}
+Nematic::Nematic(const vec3<float>& u) : m_n(0), m_u(u / std::sqrt(dot(u, u))), m_nematic_tensor_local({3, 3})
```
bdice: Can `m_n` and `m_nematic_tensor_local` use default member initializers in the header instead?

cpp/order/Cubatic.cc
```diff
@@ -155,14 +164,14 @@ Cubatic::Cubatic(float t_initial, float t_final, float scale, unsigned int n_rep
 tensor4 Cubatic::calcCubaticTensor(quat<float>& orientation)
 {
     tensor4 calculated_tensor = tensor4();
-    for (unsigned int i = 0; i < 3; i++)
+    for (auto& m_system_vector : m_system_vectors)
```
bdice, line 164: Could be written with `std::accumulate`.

cpp/order/Cubatic.cc
```diff
@@ -81,7 +76,7 @@ tensor4 tensor4::operator*(const float& b) const
 
 void tensor4::copyToManagedArray(util::ManagedArray<float>& ma)
 {
-    memcpy(ma.get(), (void*) &data, sizeof(float) * 81);
+    std::copy(data.begin(), data.end(), ma.get());
```
bdice, line 76:
```suggestion
    std::copy(data.cbegin(), data.cend(), ma.get());
```

tommy-waltmann (approved):

cpp/order/Cubatic.cc
```diff
@@ -16,27 +16,22 @@
...
-
-tensor4::tensor4(vec3<float> vector)
+tensor4::tensor4(const vec3<float>& vector)
 {
     unsigned int cnt = 0;
-    float v[3];
+    std::array<float, 3> v;
     v[0] = vector.x;
     v[1] = vector.y;
     v[2] = vector.z;
```
bdice:
```suggestion
    std::array<float, 3> v = {vector.x, vector.y, vector.z};
```

cpp/locality/Voronoi.cc
```diff
@@ -23,31 +24,22 @@ void Voronoi::compute(const freud::locality::NeighborQuery* nq)
...
-    boxLatticeVectors[1] = box.getLatticeVector(1);
-    if (box.is2D())
-    {
-        boxLatticeVectors[2] = vec3<float>(0, 0, 1);
-    }
-    else
-    {
-        boxLatticeVectors[2] = box.getLatticeVector(2);
-    }
+    vec3<float> v1 = box.getLatticeVector(0);
```
bdice: There are quite a few values in here that could be made `const`, I think.

cpp/locality/RawPoints.h
```diff
@@ -44,21 +45,21 @@ class RawPoints : public NeighborQuery
...
      */
-    virtual std::shared_ptr<NeighborQueryIterator>
-    query(const vec3<float>* query_points, unsigned int n_query_points, QueryArgs query_args) const
+    std::shared_ptr<NeighborQueryIterator> query(const vec3<float>* query_points, unsigned int n_query_points,
+                                                 QueryArgs query_args) const override
     {
         if (!aq)
         {
-            aq = std::unique_ptr<AABBQuery>(new AABBQuery(m_box, m_points, m_n_points));
+            aq = std::make_unique<AABBQuery>(m_box, m_points, m_n_points);
```
bdice, line 53: Thanks, C++14!

cpp/locality/NeighborQuery.h
```diff
@@ -254,27 +257,27 @@ class NeighborQueryPerPointIterator : public NeighborPerPointIterator
...
           m_exclude_ii(exclude_ii)
     {}
 
     //! Empty Destructor
-    virtual ~NeighborQueryPerPointIterator() {}
+    ~NeighborQueryPerPointIterator() override = default;
 
     //! Indicate when done.
-    virtual bool end()
+    bool end() override
```
bdice:
```suggestion
    bool end() const override
```

cpp/locality/NeighborQuery.h
```diff
@@ -128,7 +123,9 @@ class NeighborQuery
         // pair calculations using non-periodic boxes should fail
         vec3<bool> periodic = m_box.getPeriodic();
         if (!(periodic.x && periodic.y && periodic.z))
-            std::domain_error("Cannot execute pair queries in a non-periodic box");
+        {
+            throw std::domain_error("Cannot execute pair queries in a non-periodic box");
```
bdice: _Cannot_ or "this is not implemented"? I think it would just take a bit of work on AABBQuery and LinkCell to ignore periodic images.

cpp/locality/NeighborList.cc
```diff
@@ -212,14 +234,14 @@ bool compareNeighborDistance(const NeighborBond& left, const NeighborBond& right
...
+    // cppcheck-suppress [knownConditionTrueFalse,unsignedLessThanZero]
+    return left.size() < right.size();
```
bdice: Let's replace this function body, since it's just as simple as having a weird linter exception:
```c++
if (right.empty())
{
    return false;
}
if (left.empty())
{
    return true;
}
return compareNeighborBond(left[0], right[0]);
```

cpp/environment/Registration.h
```diff
@@ -396,17 +405,19 @@ class RegisterBruteForce
...
+        {
             return false;
+        }
 
         bool bRetVal = false;
 
         for (int i = k - 1; i >= 0; i--)
         {
-            if (comb[i] + 1 < size_t(N + i - k + 1))
+            if (comb[i] + 1 < (N + i - k + 1))
```
bdice, line 420: It's odd that both of these have a "+1". Is this due to some fear of unsigned comparison with -1?

cpp/environment/MatchEnv.h
```diff
@@ -27,23 +27,22 @@ namespace freud { namespace environment {
 struct Environment
 {
     //! Constructor.
-    Environment(bool ghost = false) : env_ind(0), vecs(0), ghost(ghost), num_vecs(0), vec_ind(0), proper_rot()
-    {}
+    Environment(bool ghost = false) : vecs(0), ghost(ghost), vec_ind(0), proper_rot() {}
```
bdice: Should some of these be default member initializers?

cpp/environment/MatchEnv.cc
```diff
@@ -655,9 +653,9 @@ unsigned int EnvironmentCluster::populateEnv(EnvDisjointSet dj)
 
     // Now update the vector of environments from the map.
     m_cluster_environments.resize(cluster_env.size());
-    for (auto it = cluster_env.begin(); it != cluster_env.end(); ++it)
+    for (auto& it : cluster_env)
```
```suggestion
    for (const auto& it : cluster_env)
```

bdice (commented):
See comments. @vyasr If there are changes I suggested that you think are out of scope, or you want me to do them, just say so.

cpp/util/diagonalize.cc
vyasr: Yeah this is possible, the template parameter for Eigen just needs to be modified as well which I didn't realize the first time I tried to get the const version to compile.

cpp/order/SolidLiquid.h
vyasr: I chose this method since it's marginally more efficient, but that difference is almost certainly negligible in this case. I'm going to leave it since it's already written this way, but signing off on this change for future uses.

cpp/order/RotationalAutocorrelation.cc
vyasr: The intent was to cast the product so that the multiplication is integer multiplication, looks like I misplaced the closing parenthesis.

cpp/order/Cubatic.cc
vyasr: There is a _lot_ of the code base that could be rewritten using `std::algorithm`. I'll make this particular change as requested, but let's avoid making any more related requests on this PR; we can do that more gradually if and when we have other reasons to modify the relevant parts of the code (as we make other changes cppcheck will probably  automatically start identifying more opportunities as well).

vyasr: Actually changed my mind, I'm skipping this for now. Currently tensor4 only supports in place addition. We could add the extra operator, but the result of making this change would end up being a net increase in lines of code with limited extra clarity anyway since the accumulation involves a nontrivial operation. We can revisit in the future.

cpp/locality/NeighborQuery.h
vyasr: At the time this was written it was closer to "cannot" in the sense that _other_ parts of freud would need modification first (namely, Box.h). Now that #209 has been resolved this is really only not implemented at the NeighborQuery level. I can update the error.

cpp/environment/Registration.h
vyasr: I noticed this as well, and I'm not sure. It's possible, and I didn't feel comfortable making any changes without testing.

bdice (approved):
Looks good to me!

PR #690, opened 2020-11-09
Enable Python 3.9 in CI.

bdice:
Enables Python 3.9 in continuous integration. I am also doing some pre-release work in this branch so that I can experiment with CI changes for the upcoming 2.4 release.

### DIFF
--- a/.pre-commit-config-cpp.yaml
+++ b/.pre-commit-config-cpp.yaml
@@ -2,19 +2,21 @@
 # hooks require tools that may not be trivial to install (unlike the Python
 # hooks like flake8), we keep these in a separate file to be applied manually.
 repos:
--   repo: https://github.com/pocc/pre-commit-hooks
-    rev: python
+-   repo: local
     hooks:
       - id: clang-format
+        name: clang-format
+        entry: clang-format
+        language: system
+        types: [c++]
+        # Ignore external files. Also include AABB.h because it uses intrinsics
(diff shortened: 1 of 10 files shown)


.pre-commit-config-cpp.yaml
```diff
@@ -2,23 +2,31 @@
...
-    rev: python
+-   repo: local
     hooks:
       - id: clang-format
+        name: clang-format
+        entry: clang-format
+        language: system
+        types: [c++]
+        # Ignore external files. Also include AABB.h because it uses intrinsics
+        # heavily and these are not considered portable.
```
bdice:
```suggestion
        # Ignore external files.
```

.pre-commit-config-cpp.yaml
```diff
@@ -2,23 +2,31 @@
...
         - --style=file
-        exclude: 'extern/.*'
       - id: clang-tidy
-        # Ignore external files and those taken verbatim from HOOMD.
+        name: clang-tidy
+        entry: clang-tidy
+        language: system
+        types: [c++]
+        # Ignore external files. Also include AABB.h because it uses intrinsics
+        # heavily and these are not considered portable.
```
bdice:
```suggestion
        # Exclude external files. Also exclude AABB.h because it uses intrinsics
        # heavily and these are not considered portable.
```

PR #692, opened 2020-11-10
Skip deprecation warning hack on Windows.

bdice:
## Description
There's a specific exclusion for Python 3.8 in our CMake, which adds `-Wno-deprecated-declarations` to the compile options. MSVC doesn't recognize this flag.

## Motivation and Context
Fixes Windows builds for Python 3.8.
This change was needed to build on conda-forge (I added a patch). We didn't catch this locally because we don't test Windows with Python 3.8 (only 3.7).

## How Has This Been Tested?
Tested via conda-forge: https://github.com/conda-forge/freud-feedstock/pull/32

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,8 +34,8 @@ references:
         # universal wheels instead, so once we upgrade to cmake >=3.14 we
         # won't be dependent on them updating wheels for specific new Python
         # versions.
-        ${PYTHON} -m pip install --user scikit-build==0.10.0
-        ${PYTHON} -m pip install --user -r requirements-testing.txt
+        ${PYTHON} -m pip install --user --progress-bar off scikit-build==0.10.0
+        ${PYTHON} -m pip install --user --progress-bar off -U -r requirements-testing.txt
 
   get_conda_requirements: &get_conda_requirements
     run:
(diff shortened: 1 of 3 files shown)


tommy-waltmann:
Would it be worth it to run Windows tests on py38 as well as py37 since they seem to behave differently in some ways?

vyasr:
This particular issue is a very special case because it's a check that I put in _specifically_ for Python3.8. Having said that, I don't have any issue with adding more CI builds, the only downside is that it potentially increases the latency of our PR iteration. My understanding is that orbs are completely independent of Docker, so this wouldn't increase our pull rate (which is important to consider due to the upcoming change in the Docker pricing model). If we want to do this, one thing I would consider is making one of the test steps (say linux python3.8) a prerequisite for all of the other testing steps. Alternatively, we could just start running tests as part of what is currently the pre-test checks job, since all the extra C++ linting forced me to build freud in that step anyway. Running all tests on one particular version prior to launching the full suite would be a good way to limit our Docker usage going forward.

The latter set of changes is of course out of scope for this PR, but would be something to consider if we decided to add more Windows testing.

bdice:
I'm going to try to improve CI as suggested above.

bdice:
We now have tests for Windows Python 3.6, 3.7, 3.8, and 3.9. There are some errors in pip installation of requirements that are resolved by later installation steps. It doesn't affect the tests so I'm not going to worry about it for now.

tommy-waltmann (approved):

PR #694, opened 2020-11-12
Force --build-type=ReleaseWithDocs to avoid conflicts with UseCython.cmake.

bdice:
## Description
The flags Cython relies on for documentation were changed with the merge of #668 when scikit-build was introduced. This means that the 2.4.0 release does not have functioning documentation. I opened an issue https://github.com/scikit-build/scikit-build/issues/518 to detail what is going wrong. In this PR, I introduce a workaround that forces `--build-type=ReleaseWithDocs`, an arbitrary build type that is not detected by scikit-build's `UseCython.cmake` configuration.

## Motivation and Context
Fixes broken documentation. This will be released as freud 2.4.1.

## How Has This Been Tested?
Works locally for me. I will verify the output of ReadTheDocs. I would like other developers to try it as well.

## Types of changes

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,13 +29,15 @@ def setup(*args, **kwargs):
         if arg == '--':
             # Insert at the end of the options that go to scikit-build
             break
-        elif arg.startswith('--build-type') and arg != BUILD_TYPE:
-            raise RuntimeError(f'Specifying --build-type is not allowed. '
-                               f'freud requires {BUILD_TYPE}')
+        elif arg.startswith('--build-type'):
+            # Don't override user-specified build type
+            index = False
+            break
     else:
         # Insert at the end of the provided arguments
         index = len(sys.argv)
-    sys.argv.insert(index, BUILD_TYPE)
+    if index:
+        sys.argv.insert(index, BUILD_TYPE)
     skbuild_setup(*args, **kwargs)
 
 
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
+import sys
 import os
-from skbuild import setup
+from skbuild import setup as skbuild_setup
 
 version = '2.4.0'
 
@@ -12,6 +13,31 @@ try:
         readme = f.read()
 except ImportError:
     readme = desc
+
+
+def setup(*args, **kwargs):
+    """This wrapper exists to force the option --build-type=ReleaseWithDocs.
+
+    Neither Release nor RelWithDebInfo will work, due to hard-coded options in
+    scikit-build's UseCython.cmake that disable docstrings. The choice of
+    ReleaseWithDocs is arbitrary, as a string that won't overlap with any build
+    type handled in UseCython.cmake. See this issue for details:
+    https://github.com/scikit-build/scikit-build/issues/518
+    """
+    BUILD_TYPE = '--build-type=ReleaseWithDocs'
+    for index, arg in enumerate(sys.argv):
+        if arg == '--':
+            # Insert at the end of the options that go to scikit-build
+            break
+        elif arg.startswith('--build-type') and arg != BUILD_TYPE:
+            raise RuntimeError(f'Specifying --build-type is not allowed. '
+                               f'freud requires {BUILD_TYPE}')
+    else:
+        # Insert at the end of the provided arguments
+        index = len(sys.argv)
+    sys.argv.insert(index, BUILD_TYPE)
+    skbuild_setup(*args, **kwargs)
+
 
 setup(
     name='freud-analysis',
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ references:
       name: Build with MSVC
       command: |
         ${PYTHON} --version
-        ${PYTHON} setup.py build_ext --inplace -- -G "Visual Studio 16 2019" -DCOVERAGE=ON
+        ${PYTHON} setup.py build_ext --inplace -- -G "Visual Studio 16 2019" -DCOVERAGE=ON -- /m
 
   test: &test
     run:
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -13,6 +14,31 @@
...
+def setup(*args, **kwargs):
+    """This wrapper exists to force the option --build-type=ReleaseWithDocs.
+
+    Neither Release nor RelWithDebInfo will work, due to hard-coded options in
+    scikit-build's UseCython.cmake that disable docstrings. The choice of
+    ReleaseWithDocs is arbitrary, as a string that won't overlap with any build
+    type handled in UseCython.cmake. See this issue for details:
+    https://github.com/scikit-build/scikit-build/issues/518
+    """
+    BUILD_TYPE = '--build-type=ReleaseWithDocs'
```
vyasr, line 27: I would prefer to use something that is clearly a fake value `--build-type=DummyType` or similar. Otherwise someone will inevitably see this at some point and think it's meaningful, no matter how well you document it. Giving it an apparently meaningful name is misleading.

setup.py
```diff
@@ -13,6 +14,31 @@
...
+    for index, arg in enumerate(sys.argv):
+        if arg == '--':
+            # Insert at the end of the options that go to scikit-build
+            break
+        elif arg.startswith('--build-type') and arg != BUILD_TYPE:
+            raise RuntimeError(f'Specifying --build-type is not allowed. '
+                               f'freud requires {BUILD_TYPE}')
```
vyasr: I don't see any reason to do this. Users should be free to override build types if they wish, that's "here be dragons" territory. Enabling debug flags in particular could be of use in certain cases.

vyasr (changes requested):
You found the solution that I avoided suggesting 😂 but it might be the best that we can do for now.

setup.py
bdice: I disagree. The important part is that the name is unique and findable in the source. The name is arbitrary but that doesn't mean that it's a meaningless choice. I would rather convey the intent of the value in its name and avoid the possibility of raising false alarms when users see the build output and ask "Why is this saying I'm building a `FakeValue`? Seems wrong." (I would totally see `--build-type=FakeValue` and worry.)

setup.py
bdice: Agreed. I will remove this failure and instead just rely on a default value.

setup.py
vyasr: We have opposite reactions. I would see ReleaseWithDocs and spend a lot of time Googling it or searching CMake code to find what that means. As someone who is relatively new to CMake but has built enough projects to know that such names are common, wouldn't you interpret that name as something meaningful unless you wrote it?

I think we are maybe coming at it from different perspectives though. You are coming at this from the perspective of a user installing this, but actually inspecting the build output. In that case I see your point. I'm coming at this from the perspective of a developer who doesn't know CMake all that well. In that case looking at the CMake output or the CMakeCache.txt file would make it seem like there's some special feature being used. So maybe getting an opinion from @b-butler @tommy-waltmann and @klywang would be helpful.

bdice:
I merged #695 into this PR for testing, since both are necessary for fixing ReadTheDocs. As of commit 08a08ef, the build on ReadTheDocs appears to be successful.

bdice:
It appears that defining a custom configuration type does not "just work" on Windows. I looked into [this Stack Overflow suggestion](https://stackoverflow.com/a/60092265/2505544) and [this blog post](https://www.trentreed.net/blog/advanced-cmake-usage/) and [this CMake wiki/guide](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#how-can-i-extend-the-build-modes-with-a-custom-made-one-) and I got something that worked locally in 69ca42e.

setup.py
tommy-waltmann: I think documenting the build type option is what's most important here. I like that the name of the build type option means something, but I think we do need to clarify that this option is our custom build type. Some comments in the CMake where the build option is defined would be helpful here, as well as in setup.py, and in the package documentation would make the intent of this build type clear to both developers and users who are building from source.

klywang: I second @tommy-waltmann's comment that documentation is most important.

With that being said, I think a descriptive name like `ReleaseWithDocs` is helpful. If I understand correctly, running `python setup.py install` without the `--build-type=ReleaseWithDocs` flag or with `--build-type=Release` would result in a build without documentation. So, even if the name is arbitrary,  a label like `ReleaseWithDocs` makes the purpose of this additional option very clear, while an option like `DummyType` would raise questions of its usefulness and purpose, and I would probably end up Googling the flag (at least, from my perspective).

bdice: @tommy-waltmann @klywang Thanks for the feedback! I will make sure this is documented in the CMake help strings and package docs as well as `setup.py`. Also just to clarify, the point of this wrapper in `setup.py` is so that `python setup.py install` "just works" without needing to specify a `--build-type` option (we default to building with docs).

vyasr: OK seems like everyone is in agreement that the most important thing is lots of docs. 

I don't know CMake well enough to know if the extra variables you're setting are comprehensively overriding the behavior. However, since this is just a temporary patch until your scikit-build PR gets merged (and released) I'm OK with moving forward with this.

bdice: > I don't know CMake well enough to know if the extra variables you're setting are comprehensively overriding the behavior.

I am decently confident that I'm defining a "comprehensive" set of variables here. I did a lot of reading on this and found several examples of custom configurations that match the behavior in this PR. I would say this is robust / not-hacky.

vyasr (approved):
I think we're set.

PR #700, opened 2020-11-16
Release/2.4.1

bdice:
Release v2.4.1. Fixes several issues with docs and installation introduced with the new scikit-build system, and one bug in Box.


bdice:
This is currently broken because CMake relied on `TBBROOT`, which changed to `TBB_ROOT` in 1e61c186f1de820eedb8dcccfb9a46fe614e43ee. That breaks the script that sets environment variables that comes with TBB. https://github.com/oneapi-src/oneTBB/blob/eca91f16d7490a8abfdee652dadf457ec820cc37/build/generate_tbbvars.sh#L24

@vyasr Please offer insight on your preferred solution, since you made that change. Feel free to commit directly to this branch.

PR #713, opened 2021-01-02
Try testing oldest dependency versions in a separate configuration.

bdice:
## Description
In some Slack discussions, we decided it might be good to test newest and oldest dependency versions in separate configurations. This provides a configuration for the "oldest" tests.

Along the way, I found quite a few places where our stated package requirements were inaccurate. 😬

This PR separates our tests into "newest" (most of the configurations) and "oldest," which is the lowest version numbers that we know we support. I recommend the following policies for testing in the future:

The oldest dependencies should only be changed in a limited number of cases:
- Dropping Python versions (only the earliest supported version needs to be tested with "oldest" dependencies)
- Adding new features that rely on newer dependency versions

For the newest dependencies:
- New test configurations should be added when new Python versions are released and added to the list of supported versions
- Pinned testing dependencies are automatically updated by dependabot
- Configurations can be dropped from CI if "newest" dependencies drop support. I am hoping that our dependencies drop support no faster than NEP 29 would. I am also hoping we only need to support releasing for configurations that are covered by NEP 29 at release time.

## How Has This Been Tested?
Added new CI configuration.

## Types of changes

## Checklist:


.circleci/config.yml
```diff
@@ -27,15 +27,39 @@ references:
...
-        # won't be dependent on them updating wheels for specific new Python
-        # versions.
-        ${PYTHON} -m pip install --user --progress-bar off scikit-build==0.10.0
-        ${PYTHON} -m pip install --user --progress-bar off -U -r requirements-testing.txt
+        # Create a virtual environment to ignore the CI image's pre-installed packages
+        ${PYTHON} -m pip install --progress-bar off --upgrade virtualenv --user
+        mkdir -p ./venv
+        ${PYTHON} -m virtualenv ./venv --clear
+        . venv/bin/activate
+        ${PYTHON} -m pip install --progress-bar off --upgrade pip virtualenv
```
b-butler: Why are we upgrading `virtualenv` again?

b-butler (commented):
I only have one comment otherwise the PR LGTM.

.circleci/config.yml
bdice: That might have been an oversight on my part. I'll change it to only upgrade pip.
```suggestion
        ${PYTHON} -m pip install --progress-bar off --upgrade pip
```

bdice:
I fixed the one comment from @b-butler and I will merge this when tests pass.

Issue #635, opened 2020-07-30
Unclear default value for `normalize` kwarg in `freud.density.RDF`


jennyfothergill:
**Describe the bug**
The help for `freud.density.RDF` says this about the normalize kwarg:
 |      normalize (bool, optional):
 |          Scale the RDF values by
 |          :math:`\frac{N_{query\_points}}{N_{query\_points}-1}`. This
 |          argument primarily exists to deal with standard RDF calculations
 |          where no special ``query_points`` or ``neighbors`` are provided,
 |          but where the number of ``query_points`` is small enough that the
 |          long-ranged limit of :math:`g(r)` deviates significantly from
 |          :math:`1`. It should not be used if :code:`query_points` is
 |          provided as a different set of points, or if unusual query
 |          arguments are provided to :meth:`~.compute`, specifically if
 |          :code`exclude_ii` is set to :code:`False`. This normalization is
 |          not meaningful in such cases and will simply convolute the data.

I think the default value is `True`, but this could be more clear. Please forgive me if this is obvious or I have just missed it somewhere!

**To Reproduce**
`help(freud.density.RDF)`

**Error output**
n/a

**System configuration (please complete the following information):**
 - OS: OSX 10.14.6 (18G103)
 - Version of Python 3.7.6
 - Version of freud '2.2.0'

**Additional context**
n/a

vyasr:
Actually, the default is False, and the docstring should be updated to include the usual indicator `Default value: False`. However, I can also see that the naming is a bit confusing in light of what we're discussing on #396. I'm wondering if we instead need to add a string argument `normalization_mode`. The mode would default to `'infer'`, in which case it would handle the two obvious cases discussed in #396 (`query_points` either being identical to or completely disjoint from `points`). Setting normalize to `'none'` would completely remove normalization by particle number, in which case it would be up to the user to compute the normalization factor. Setting normalize to `'finite_size'` would apply the current normalization (N-1/N), which is really only relevant for small systems. @jennyfothergill does that make sense? Then we would update the docstrings accordingly (including with defaults) and deprecate the current `normalize` argument, marking it for removal in freud 3.0 (to be replaced by the `normalization_mode`).

bdice:
@vyasr Does this need to be API-breaking? I'd prefer to just handle a third option for `normalize` instead of creating a new argument named `normalization_mode`. Then `True` and `False` would be removed in 3.0.
```
True or 'finite_size'  # Handled the same way
[...]

PR #738, opened 2021-03-28
Store and re-use computed neighbor vectors.

bdice:
## Description
I am dealing with an unusual case with Voronoi neighbors in a sparse system (only a few particles), where particles can be their own neighbors or share multiple bonds with the same neighbor. For this case, it's insufficient to identify neighbors by their distance. Instead, I need the actual vectors computed by the Voronoi tessellation.

- Make vec3, vec2, quat constexpr-qualified (also constexpr implies inline).
- Add vectors to NeighborBond, NeighborList.
- Store neighbor vectors in all neighbor bonds, use this data instead of re-computing bond vectors from the query points and points.
- Ignore E402 because flake8 doesn't like some changes made by isort; I prefer isort so I'm ignoring the flake8 warning.

There is an API break worth mentioning here: `NeighborList.from_arrays` can no longer accept `distances`, and instead requires `vectors`. (The distances are computed internally to match the vectors.) I could not think of a way to avoid this API break.

## Motivation and Context
In this PR, I re-work the NeighborBond and related classes to store the actual *vector* for each bond. This vector can be directly re-used in many analysis methods, avoiding the need to perform computing `box.wrap(points[point_index] - query_points[query_point_index])`. I have benchmarks below. This should help improve correctness in a few edge cases like the Voronoi case I mentioned above, and should make it easier if we ever choose to let freud find neighbors beyond the nearest periodic image. I also caught a bug in the tests because of this change.

## How Has This Been Tested?
Existing tests pass with minor tweaks.

Performance:
- RDF is basically unchanged
- BondOrder is ~10% faster
- PMFT is ~10% slower

Overall I'm not concerned about the performance impact of this PR.

## Types of changes

## Checklist:


cpp/locality/NeighborList.h
```diff
@@ -93,6 +98,11 @@ class NeighborList
     {
         return m_weights;
     }
+    //! Access the vectors array for reading and writing
```
b-butler:
```suggestion
    //! Access the vectors array for reading
```

tests/test_environment_LocalDescriptors.py
```diff
@@ -404,15 +404,15 @@ def test_query_point_ne_points(self):
...
+            query_points, dict(exclude_ii=False, num_neighbors=num_neighbors)
         ).toNeighborList()
 
         ld = freud.environment.LocalDescriptors(l_max, mode="global")
         ld.compute((box, points), query_points, neighbors=nl)
 
         # Loop over the sphs and compute them explicitly.
-        for idx, (i, j) in enumerate(nl):
-            bond = box.wrap(points[j] - query_points[i])
+        for idx, bond in enumerate(nl.vectors):
```
b-butler, line 410: What is the point of enumerating here if we are going to overwrite `bond` in the next line?

tests/test_locality_NeighborList.py
```diff
@@ -129,76 +129,77 @@ def test_segments(self):
...
-                4, 4, query_point_indices, point_indices[:-1], distances
+                4, 4, query_point_indices, point_indices[:-1], vectors
             )
         with pytest.raises(ValueError):
             freud.locality.NeighborList.from_arrays(
-                4, 4, query_point_indices, point_indices, distances[:-1]
+                4, 4, query_point_indices, point_indices, vectors[:-1]
             )
         with pytest.raises(ValueError):
             weights = np.ones((len(query_point_indices) - 1,))
```
b-butler: Why do we create a new set of weights, but just use indexing in the other cases to test for error raising?

b-butler (commented):
Overall looks good to me. I left a few comments.

vyasr:
@bdice I don't mind the API break, I think that it's probably reasonable for us at this stage. However, I think that since the 1.0 release the project has been much more consistent in adhering to best practices overall, which includes versioning. I don't have time to review this right now, I will get to it soon, but if we can't find a way to avoid the API break we should stick to semantic versioning and require a 3.0 version. I don't think that's a huge issue; we shouldn't be scared of a major version bump if that's the best way to convey the breakage.

bdice:
> 
> 
> @bdice I don't mind the API break, I think that it's probably reasonable for us at this stage. However, I think that since the 1.0 release the project has been much more consistent in adhering to best practices overall, which includes versioning. I don't have time to review this right now, I will get to it soon, but if we can't find a way to avoid the API break we should stick to semantic versioning and require a 3.0 version. I don't think that's a huge issue; we shouldn't be scared of a major version bump if that's the best way to convey the breakage.

I agree. I would like to get input from Josh (who is on vacation, so not tagged) or @b-butler about their takes on semver strictness and what they would do in the case of HOOMD. I would be fine with bumping to 3.0. It's also fine to just defer on this PR (or merge into a `next` branch) in case we want to do a review of other small changes that would be good to make for 3.0, and punt on that for a month or two before making the change.

tests/test_locality_NeighborList.py
bdice: Good catch! I fixed it.

tests/test_environment_LocalDescriptors.py
bdice: This is a mistake on my part that occurred during some manual testing. Fixed.

b-butler:
@bdice and @vyasr I would support a major version bump. For the hoomd beta's we are allowing breakages in-between beta versions, but the are not considered official releases of 3.0. I think the more people who follow semantic versioning the better as it leads to less headaches for downstream developers/scripters. I also don't think that a major version bump need be a big ordeal. Many packages have very large major versions, and I think that is fine as long as the changes are in the project's interest. The only caveat is when considering maintenance burden when supporting multiple major versions depending on the level of changes.

vyasr:
Cool. I think a major bump would be the right way to go. I was also going to propose holding off on merging this into master until we have an idea of what other breaking changes might be useful. Offhand I recall a few other awkward locality APIs and maybe the open RDF normalization work introducing something (I'll need to look back at that eventually).

b-butler:
@tommy-waltmann what do you think of this PR? Should we just close it or is this still in the future interest of the project in your opinion?

bdice:
@b-butler @tommy-waltmann If you accept the API breakage, I’d like to see this feature go in when we are ready for 3.0 development. This is needed for correctness in some Voronoi cases and simplifies some analysis method code at the expense of the neighbor list requiring more information. If you would like, I can resolve conflicts and/or pull out some of the nonbreaking improvements from this PR (into separate PRs to 2.x) so that the diff is smaller. Let me know what you think.

tommy-waltmann:
@b-butler If you have a use case (please share) motivating the use of this addition, then I think its worth continuing to pursue.

One option as far as implementation goes that is worth discussion is storing the distance as well as the vector in each `NeighborBond` to prevent performance regressions in any modules. The tradeoff there is memory usage and data consistency between the distances and vectors vs performance.

If we decide to continue to pursue this change, I think we will have enough momentum to start concretely planning a v3 release, making a project board for it, etc.

@bdice yes, I would agree that the base branch for this PR should be changed. I would like to see all the changes for this not be released until v3, just for better organization.

tests/test_locality_Voronoi.py
```diff
@@ -328,6 +334,26 @@ def test_voronoi_weights_fcc(self):
         )
         npt.assert_allclose(wrapped_distances, vor.nlist.distances)
 
+    def test_diamond(self):
```
bdice, line 294: @b-butler @tommy-waltmann This is an example of a test that fails without this PR. Before this PR, the Voronoi neighbor bonds are defined by a minimum distance and do not distinguish the actual vectors between particles. This fails for many systems that are small unit cells. In this case, the two particles are bonded to each other with several unique vectors that could have different lengths. The bonds between the particle and itself should have a nonzero vector (pointing to a periodic image outside the cell). When computing the distance based on the wrapped i-j vector, we lose information that can be retained if we track the vector when the neighbor is found.

Without this PR, it is likely that small unit cells will give incorrect results for most quantities related to bonding, like PMFTs and Bond Order Diagrams, when the bonds are defined using Voronoi neighbors.

Also, I updated this branch to resolve conflicts with `master`.

vyasr:
I definitely think we want to move forward with this PR, and that it's just stalled waiting for a sufficient push to make v3 happen. I think the other major blocker for v3 is getting someone to address #635, which is nontrivial. If devs are willing to make a push I think that task could be completed and we'd be largely on track for a v3 release.

tommy-waltmann:
I won't personally have much time to work on this, or other related issues for v3 until mid-August. I'm currently getting ready for multiple conferences and working on writing multiple papers. I have, however, made a `next` branch to accumulate changes that will be on the v3 release. Please rebase this PR so the target branch is `next`

tests/test_locality_Voronoi.py
tommy-waltmann: Okay, this example makes the motivation for this PR more concrete, thanks for the explanation.

cpp/environment/MatchEnv.cc
```diff
@@ -502,7 +501,7 @@ Environment MatchEnv::buildEnv(const freud::locality::NeighborQuery* nq,
         const size_t j(nlist->getNeighbors()(bond, 1));
         if (i != j)
         {
-            vec3<float> delta(bondVector(locality::NeighborBond(i, j), nq, nq->getPoints()));
```
tommy-waltmann, line 505: Can remove the NeighborBond header from this file, since this was the only usage of NeighborBond here

tommy-waltmann:
Regarding the only failing test, the difference between the values is ~`2e-5`, when the tolerance is set at `1e-5`. Is the reason this is failing now just due to numerical precision, or there something deeper going on?

tommy-waltmann:
> Regarding the only failing test, the difference between the values is ~`2e-5`, when the tolerance is set at `1e-5`. Is the reason this is failing now just due to numerical precision, or there something deeper going on?

For whatever reason this test is failing now, it appears to stem from how the neighbor distances are computed in the Voronoi module.

tommy-waltmann:
Just made some small changes to address my own comments on this PR and push this along a bit faster.

vyasr:
I think the next step for moving this forward is to make a plan for v3.

tommy-waltmann:
The plan is to fix the issues on the v3 milestone, targeting the `next` branch and make a release when they are completed. Is there anything else you would like to see in a plan for v3?

vyasr:
No, I just mean that I think this PR has been close to ready for a long time and the reason it stalled is the fact that it was going into v3. I don't think anything has been merged into `next` yet, and as soon as we start we'll have to deal with keeping it in sync with `master` etc so it's probably best to have a plan for all of those changes before starting the merges.

tommy-waltmann:
What would you like to see in "a plan for all of those changes" other than what we already have? As far as merges go. I would expect that after every 2.x release, we will merge `master` into `next` until it is time to release v3.0.0

vyasr:
I'm mostly responding to this statement:
> push this along a bit faster.

Faster than what? It's not clear exactly what we're waiting on here. Is this PR ready to merge into `next`? If so, do devs have the bandwidth to address the remaining issues for v3 sometime soon? Maintaining a long-running `next` branch could also get painful when it comes to resolving conflicts depending on the scope of proposed v3 changes.

tommy-waltmann:
The scope of the v3 changes aren't going to be massive, its not going to be like hoomd v3. I would expect that we find one or two more things that aren't on the v3 milestone currently as we address what's there, but I don't expect the number of things that need to be addressed to snowball.

Timewise, I am going to have more time in the fall than I had last summer to work towards a v3 release. The current rate of freud development on the master branch is not fast, so I wouldn't expect a lot of v2.x releases between now and when we release v3. We shouldn't have too much issue with maintaining a `next` branch before the v3 release.

As far as this PR goes, its just waiting for approving reviews from those who have been requested, and there is one test that is failing, likely just because of the need for a tolerance change on a floating point comparison.

vyasr:
OK that sounds fine. In that case I'll try to review this again sometime soon.

cpp/locality/NeighborBond.h
```diff
@@ -14,15 +19,16 @@ struct NeighborBond
     // For now, id = query_point_idx and ref_id = point_idx (into the NeighborQuery).
     constexpr NeighborBond() = default;
 
-    constexpr NeighborBond(unsigned int query_point_idx, unsigned int point_idx, float d = 0, float w = 1)
-        : query_point_idx(query_point_idx), point_idx(point_idx), distance(d), weight(w)
+    constexpr NeighborBond(unsigned int query_point_idx, unsigned int point_idx, float d, float w,
```
b-butler, line 30: Shouldn't distance be either be computed on the fly or in the constructor. I see no reason to pass it any more. This way we have to be careful to ensure that the vector is equivalent to the distance rather than being sure by the object's methods or construction.

cpp/locality/NeighborList.cc
```diff
@@ -50,8 +51,9 @@ NeighborList::NeighborList(unsigned int num_bonds, const unsigned int* query_poi
         }
         m_neighbors(i, 0) = index;
         m_neighbors(i, 1) = point_index[i];
+        m_vectors[i] = vectors[i];
+        m_distances[i] = std::sqrt(dot(vectors[i], vectors[i]));
```
b-butler: Should we automatically compute this or only when called through `NeighborBond`.

cpp/locality/Voronoi.cc
```diff
@@ -124,8 +125,10 @@ void Voronoi::compute(const freud::locality::NeighborQuery* nq)
 
             // Compute cell neighbors
             size_t neighbor_counter(0);
+            size_t face_vertices_index(0);
             for (auto neighbor_iterator = neighbors.begin(); neighbor_iterator != neighbors.end();
-                 neighbor_iterator++, neighbor_counter++)
+                 neighbor_iterator++, neighbor_counter++,
```
b-butler:
```suggestion
                 ++neighbor_iterator, ++neighbor_counter,
```

tests/test_environment_BondOrder.py
```diff
@@ -113,7 +113,7 @@ def test_points_ne_query_points(self):
         r_max = 1.6
 
         num_neighbors = 12
-        n_bins_theta = 30
+        n_bins_theta = 35
```
b-butler, line 116: This seems unrelated.

tests/test_environment_LocalDescriptors.py
```diff
@@ -239,11 +239,12 @@ def test_ql(self, unit_cell):
             # in cases where there is no symmetry. Since simple cubic
             # should have a 0 ql value in many cases, we need to set high
             # tolerances for those specific cases.
-            atol = 1e-3 if unit_cell == freud.data.UnitCell.sc else 1e-6
+            atol = 1e-3 if unit_cell == freud.data.UnitCell.sc else 1e-5
```
b-butler, line 242: This seems unrelated.

tests/test_environment_LocalDescriptors.py
```diff
@@ -239,11 +239,12 @@ def test_ql(self, unit_cell):
...
             npt.assert_allclose(
                 steinhardt.particle_order,
                 ql[:, L],
                 atol=atol,
+                rtol=1e-5,
```
b-butler, line 247: This seems unrelated.

tests/test_locality_Voronoi.py
```diff
@@ -25,14 +25,15 @@ def test_random_2d(self):
...
+        # Verify the neighbor vectors
+        wrapped_points = box.wrap(
+            points[vor.nlist.query_point_indices] + vor.nlist.vectors
         )
-        npt.assert_allclose(wrapped_distances, vor.nlist.distances)
+        npt.assert_allclose(wrapped_points, points[vor.nlist.point_indices], atol=1e-5)
+
+        # Verify the neighbor distances
+        vector_lengths = np.linalg.norm(vor.nlist.vectors, axis=-1)
+        npt.assert_allclose(vector_lengths, vor.nlist.distances)
```
b-butler: This check is done repeatedly. It would make a good check function.

b-butler (changes requested):
Overall the code looks good. I have a few comments/questions on some of the implementation.

cpp/locality/NeighborBond.h
tommy-waltmann: Good point! There is now an invariant to maintain in this struct. We should actually go a step further and convert it to a class with private members and getters/setters to maintain data consistency. I think I noticed this at one point too and forgot to go back and change this part.

tests/test_environment_BondOrder.py
tommy-waltmann: It is related, surprisingly. The changes in this PR have, for whatever reason, created a situation where on Windows only, choosing exactly 30 bins created a situation where a small fraction of the bin counts leaked over into neighboring bins and made the test fail.

cpp/locality/NeighborList.cc
tommy-waltmann: This will get fixed when I address the first comment.

tommy-waltmann: I would allow this to be done automatically. The neighborlist has to expose its distances to python as a linear array, so it makes sense to store distances and vectors as separate arrays to avoid recomputing the distances too many times.

I may however, make some changes to the neighborlist class similar to the proposed in NeighborBond to ensure consistency between vectors and distances.

tests/test_environment_LocalDescriptors.py
tommy-waltmann: The changes on this PR created some loss of precision in the `LocalDescriptors` code. I'm not sure what the origin is.

cpp/locality/NeighborList.cc
b-butler: I agree that distances in Python need to be a linear array. I was just wondering if for our purposes it would be sufficiently fast to create a NumPy array on the fly when distances are requested. I could even see an argument that distances in Python shouldn't exist as it is a simple NumPy one-liner. Either decision is fine with me though as it isn't generally a costly operation.

tests/test_environment_BondOrder.py
b-butler: Ah, well that is fine then. Numerical precision and all that.

tests/test_environment_LocalDescriptors.py
b-butler: Hmm well if it is only sc I guess that is fine. I didn't notice anything in the code myself.

cpp/locality/NeighborList.cc
tommy-waltmann: Creating the arrays on-the-fly seems like more work than is necessary, and honestly may even be a performance hit if a user isn't aware the grabbing the distances is an O(N) operation.

There are non-python places in the code that use the distances, so I'd like to only compute them once and cache them for performance considerations.

Memory usage in the neighborlist isn't a concern either; if it was, this PR would likely not exist.

bdice: Correct, the design here is precomputing and caching distances because of their common use and to avoid the O(Nbonds) property fetch. I did some very loose calculations/benchmarking and decided this was the right tradeoff (and that the additional memory consumption was irrelevant). Feel free to check me on that, but my recollection is that a C++ cached array of distances is best to precompute because if you get to the point where it is needed, you’re usually in a TBB parallel operation where it would be difficult to compute the whole array AND if you compute the distance at that point, it causes issues with centralizing the floating point correctness. We used to have sqrt’s littered through the various methods’ implementations (when we computed both neighbor vector and distance on the fly) and the conventions and casting were wacky and subtlely different across the code base (not all even used the dot function for the dot product iirc?). That was not good. Having a single source for precomputed distances was better than the previous mess.

b-butler: @bdice @tommy-waltmann I still don't see how this is one source of truth then? `NeighborBonds` also has distances so it is another source of truth. We could for performance reasons say that the vector and distance passed must be compatible and not check, but then we have an untested assumption in `NeighborBond`.

Also, if `NeighborBond` is responsible for computing distances then it would be centralized and all distances would be computed the same. Ultimately, I am not a maintain or even common developer for freud. I trust that you both have better notions for what makes sense in freud. I am just offering my thoughts which if not helpful feel free to ignore (and mark this as resolved).

tommy-waltmann: There are two places right now where distances are computed: in the `NeighborList` constructor where they are computed from the corresponding vectors and taken directly as an argument for the `NeighborBond` constructor.

I have no problems with computing the distances in the `NeighborList` constructor. It is necessary, in fact, to do this because we need to support custom `NeighborList`s where users give the indices of the neighbors and their vectors. There is no potential for data inconsistency between vectors and distances in the neighborlists because of some recent changes I've made to the neighborlist class on this PR. This design is totally fine with me.

The second place is in the `NeighborBond` class, where both the distance and vector are given directly in the constructor. The design is this way because `NeighborQuery` objects have to compute the distance between a pair of particles to decide if they fit the query arguments before creating the `NeighborBond`. This results in a situation where the distance must be computed before even creating the `NeighborBond`, so to prevent redundant calculations, both the neighbor vector and the distance are given to the `NeighborBond` constructor. In this use case, we trust that the `NeighborQuery` constructs the `NeighborBond` with the correct corresponding distance/vector pair. If other parts of the code were to use this constructor, the opportunity for data inconsistency exists. My plan for this class is to create a constructor for the `NeighborBond` where only the vector is given and the distance is computed automatically, and document that the constructor which takes both the distance and vector should only be used by `NeighborQuery` objects. That seems like the right tradeoff to me in terms of performance vs reducing the potential for errors.

Ultimately, the distances must be able to be computed in multiple places because the `NeighborQuery` and `NeighborList` objects play different roles in iterating over and constructing sets of neighbors at the C++ level. Does that clear things up @b-butler ?

tommy-waltmann: I should also add, after having a closer look, that there are also places where `NeighborBond`s are constructed from `NeighborList` entries. This is another case where the correct distance is known beforehand, so re-computing the `NeighborBond` distance would be redundant. I'll heavily document which `NeighborBond` constructor should be used in which case.

freud/locality.pyx
```diff
@@ -198,7 +198,8 @@ cdef class NeighborQueryResult:
 
         npoint = dereference(iterator).next()
         while npoint != ITERATOR_TERMINATOR:
-            yield (npoint.query_point_idx, npoint.point_idx, npoint.distance)
+            yield (npoint.getQueryPointIdx(), npoint.getPointIdx(),
+                   npoint.getDistance())
```
tommy-waltmann, line 202: One more point for discussion: When iterating over a `NeighborQueryResult` in python: In addition to returning the two indices, should we now return the vector, the vector/distance, the distance, or return the vector/weight/distance? 

My personal vote is for at least returning the vector in some form, although I'm not very opinionated on the other data we return with it.

cpp/locality/NeighborList.cc
b-butler: Yes this makes sense. If we wanted to (not saying it is necessary, but we could use friend functions and a private constructor to ensure we one construct `NeighborBond` objects in methods in select methods.

freud/locality.pyx
b-butler: I could see us doing the full tuple where users can pattern match to grab what they want.

bdice: In case you're curious: when we made the `NeighborList` iterable, we only returned the point indices because we thought those would be the most commonly requested. It's already possible to do something like:
```python
for bond_idx, (query_idx, point_idx) in enumerate(nlist):
    weight = nlist.weights[bond_idx]
    print(query_idx, point_idx, weight)
```
(Untested snippet, based on my recollection of the API).

tommy-waltmann: @b-butler I'm not super familiar with the developer side of pattern matching, what would that look like implementation-wise?

b-butler: just a `yield query_idx, point_idx, vector, distance, weight`

tommy-waltmann: For sake of completeness, we can just have it return all 5 values in case a user wants to use them. What @bdice says about iterating over the `NeighborList` is true, but I don't think the same holds for iterating over a `NeighborQueryResult`.

bdice: I think it would be best if iterating over either object type gave the same kinds of information. `NeighborQueryResult` is kind of like a lazy/generated `NeighborList`.

tommy-waltmann (approved):

freud/locality.pyx
tommy-waltmann: Since the iterating is done lazily in `NeighborQueryResult`, we cannot index into the `NeighborList` to get all 5 pieces of information like we can with the `NeighborList`. To make iterating over the `NeighborQueryResult` the most useful, we would need to return all 5 pieces of information. If we decide we want to keep it symmetric between `NeighborList` and `NeighborQueryResult`, that would mean that iterating over a `NeighborList` should return all 5 pieces of information also.

In general, I don't think users find the `NeighborQueryResult` to be a very useful object in python (I've had people in lab ask me why it even exists in the first place), so I don't have a huge opinion on whether they should match or not. It also takes very little time to generate the full `NeighborList` from the `NeighborQueryResult`, so after thinking it over again, I'm okay leaving everything as is for now.

tommy-waltmann:
Does anyone else have any more comments or concerns on this PR?

cpp/locality/NeighborBond.h
```diff
@@ -9,20 +14,38 @@ namespace freud { namespace locality {
...
+     * */
+    constexpr NeighborBond(unsigned int query_point_idx, unsigned int point_idx, float d, float w,
+                           const vec3<float>& v)
+        : query_point_idx(query_point_idx), point_idx(point_idx), distance(d), weight(w), vector(v)
+    {}
+
+    /**
+     * This constructor should be used for the majority of cases when instantiating
+     * a NeighborBond object.
+     * */
```
bdice:
```suggestion
     */
```

cpp/locality/NeighborBond.h
```diff
@@ -87,10 +110,58 @@ struct NeighborBond
         return weight < n.weight;
     }
 
+    unsigned int getQueryPointIdx() const
+    {
+        return query_point_idx;
+    }
+
+    void setQueryPointIdx(unsigned int new_qpidx)
```
bdice: How about a simpler name?
```suggestion
    void setQueryPointIdx(unsigned int idx)
```

cpp/locality/NeighborBond.h
```diff
@@ -87,10 +110,58 @@ struct NeighborBond
...
+    {
+        point_idx = new_pidx;
+    }
+
+    float getWeight() const
+    {
+        return weight;
+    }
+
+    void setWeight(float new_weight)
```
bdice:
```suggestion
    void setWeight(float w)
```

cpp/locality/NeighborBond.h
```diff
@@ -87,10 +110,58 @@ struct NeighborBond
...
+    {
+        weight = new_weight;
+    }
+
+    const vec3<float>& getVector() const
+    {
+        return vector;
+    }
+
+    void setVector(const vec3<float>& new_vector)
```
bdice:
```suggestion
    void setVector(const vec3<float>& v)
```

cpp/locality/NeighborBond.h
```diff
@@ -87,10 +110,58 @@ struct NeighborBond
...
+    {
+        return distance;
+    }
+
+private:
     unsigned int query_point_idx {0}; //! The query point index.
     unsigned int point_idx {0};       //! The reference point index.
     float distance {0};               //! The distance between the points.
     float weight {0};                 //! The weight of this bond.
+    vec3<float> vector;               //! The directed vector from query point to reference point
```
bdice:
```suggestion
    vec3<float> vector;               //! The directed vector from query point to reference point.
```

cpp/locality/NeighborList.h
```diff
@@ -106,6 +87,18 @@ class NeighborList
         return m_segments;
     }
 
+    /**
+     * Set the values for the neighbor index to be that of the given neighborbond
+     * */
```
bdice:
```suggestion
     */
```

freud/locality.pyx
```diff
@@ -518,10 +517,10 @@ cdef class NeighborList:
             point_indices (:class:`np.ndarray`):
                 Array of integers corresponding to indices in the set of
                 points.
-            distances (:class:`np.ndarray`):
-                Array of distances between corresponding query points and
+            vectors (:math:`\left(N_{bonds}, 3\right)` :class:`numpy.ndarray`):
+                Array of bond vectors from query points to corresponding
                 points.
-            weights (:class:`np.ndarray`, optional):
+            weights (:math:`\left(N_{bonds} \right)` :class:`np.ndarray`):
```
bdice: Mention “optional” if there’s a default.
```suggestion
            weights (:math:`\left(N_{bonds} \right)` :class:`np.ndarray`, optional):
```

freud/locality.pyx
```diff
@@ -632,6 +632,14 @@ cdef class NeighborList:
             &self.thisptr.getDistances(),
             freud.util.arr_type_t.FLOAT)
 
+    @property
+    def vectors(self):
+        R"""(:math:`N_{bonds}`, 3) :class:`np.ndarray`: The vectors for each
```
```suggestion
        r"""(:math:`N_{bonds}`, 3) :class:`np.ndarray`: The vectors for each
```

bdice (commented):
@tommy-waltmann A couple minor comments. Otherwise LGTM! (I can’t approve since I opened the PR.)

b-butler (approved):
LGTM

freud/locality.pyx
vyasr: I originally added the `NeighborQueryResult` in #241 as a way for users to prototype new neighbor-based analysis methods in Python. It's a more Pythonic approach than building and iterating over a `NeighborList` since it functions more naturally as an iterator. In practice I don't think anyone has really been doing new methods development so the use case probably hasn't really come up. For that reason, I agree that most users wouldn't find it useful, any more than most users find iterating directly over a `NeighborList` useful.

As far as making iteration over a `NeighborList` and a `NeighborQueryResult` identical, I agree that would be optimal but there are meaningful differences here. A `NeighborList` can encapsulate the other data into members/properties, whereas the result object cannot, so there's good reason for it to behave differently. The alternative would be to implement `NeighborList.__iter__` and explicitly have it return all 5 properties instead of having it fall back to `__getitem__`. That raises questions about whether it's odd to have `next(iter(nlist))` return something different than `nlist[0]`, though, so there's no perfect solution, but that may be the best compromise.

bdice: I would like to provide a little further argumentation for the Python layer treating (query) point indices as “the iterable thing” and not including weight/distance/vector in the iterable. The indices are the most fundamental part of the neighbor list concept in my eyes, and all else is auxiliary. I think it’s sufficient for users to zip-iterate over multiple arrays of indices/weights/etc as needed. The easy path `for i, j in nlist` should keep it minimal and only yield indices.

vyasr: I'm inclined to agree that we shouldn't change how NeighborList iteration behaves. However, that doesn't address the case where no NeighborList is constructed. The implication is that the only way to iterate over the extra arrays (weights/distances/vectors) is to construct a `NeighborList`, i.e. there's no way to do this without materialization. I don't really like that, but that's the tradeoff for keeping the iteration consistent with `NeighborList`.

One other option (not necessarily advocating for it, just throwing out ideas) is taking a page from numpy's book and adding `return_*` arguments (or maybe a single `return_opts={key: val}` parameter) like `numpy.unique`.

tommy-waltmann: Based on this discussion, I think the current behavior wrt iterating over `NeighborList`/`NeighborQueryResult` objects is at least acceptable. If someone has strong feelings for changing the current behavior, we can further discuss in a new issue because this PR is about ready for merge.

PR #746, opened 2021-04-09
Build wheels with cibuildwheel.

bdice:
## Description
Enabling cibuildwheel.

Game plan:
- Use TBB tarball release. It includes a script to set all environment variables. https://github.com/oneapi-src/oneTBB/releases/tag/v2021.2.0
- @bdice will iterate on the CI config until it passes, and ask for review/help when needed.
- Ideally we could replace most of our existing CI with cibuildwheel, because if it can build a wheel and run tests... that's pretty much everything (matrix of OS and Python versions). I think the only special configuration we'd need to test outside of cibuildwheel is the one with oldest dependencies.

## Motivation and Context
Simplifying wheel building.

## How Has This Been Tested?
Need to test wheels manually.

## Types of changes

## Checklist:


.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,111 @@
...
+        with:
+          python-version: '3.9'
+
+      - name: Install build requirements
+        run: python3 -m pip --disable-pip-version-check install scikit-build cmake -r requirements.txt --upgrade --progress-bar=off
+
+      - name: Build sdist
+        run: python setup.py sdist
+
+#     - uses: actions/upload-artifact@v2.2.2
```
joaander: I found a few issues when testing this part of the GSD `build_wheels.yml`. The latest version also adds TestPyPI uploads on non-tagged builds.

https://github.com/glotzerlab/gsd/blob/fdd624f65ce71a88057a4c8a5707c93c7fa604c8/.github/workflows/build_wheels.yml#L72-L115

bdice:
Just an update on this PR -- the Ubuntu and MacOS builds are pretty close to working. We just need to fix up the tests, perhaps by skipping some non-essential integration tests that rely on MDAnalysis. I updated MDAnalysis to 1.0.1 so that Mac builds work (there are no wheels, only a tarball) but this appears to cause trouble with Linux (looks like Python 3.6 incompatibility but still unclear to me why that is happening).

For Windows builds, we need to debug the TBB installation script `cibuildwheel-before-build.ps1`, there is some issue with building TBB and a second (separate) issue with detecting the system compilers while building freud.

joaander:
> For Windows builds, we need to debug the TBB installation script `cibuildwheel-before-build.ps1`, there is some issue with building TBB and a second (separate) issue with detecting the system compilers while building freud.

As far as I know, Windows builds on PyPI don't have the same stringent manylinux requirements. You may be able to use the provided TBB binaries on windows.

bdice:
It appears there is an issue with how the Mac wheels use delocate. It only appears for the Python 3.7 wheel, so it's possible that there's something that only works the first time (for 3.6) or there's some incompatibility resulting from the Python version.

```python
+ delocate-listdeps /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibuildwheelkfabgbr9/built_wheel/freud_analysis-2.5.1-cp37-cp37m-macosx_10_9_x86_64.whl && delocate-wheel --require-archs x86_64 -w /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibuildwheelkfabgbr9/repaired_wheel /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibuildwheelkfabgbr9/built_wheel/freud_analysis-2.5.1-cp37-cp37m-macosx_10_9_x86_64.whl
  /Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/delocate/libsana.py:111: UserWarning: Couldn't find @rpath/libfreud.dylib on paths:
  @rpath/libfreud.dylib
  	/Users/runner/work/freud/freud/@loader_path
  @rpath/libtbb.12.dylib
    '\n\t'.join(realpath(path) for path in rpaths),
  /Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/delocate/libsana.py:111: UserWarning: Couldn't find @rpath/libtbb.12.dylib on paths:
  	/Users/runner/work/freud/freud/@loader_path
    '\n\t'.join(realpath(path) for path in rpaths),
```

bdice:
The errors with `delocate` appear on Python 3.8, when it's the first version tested, so it seems to be related to the Python version and not that the first build works and later builds fail. But I'm not sure why Python 3.6 passes but 3.7 and 3.8 fail. The PyPI homepage for `delocate` gives some additional information that may be relevant: https://pypi.org/project/delocate/ My guess is that the fix will involve some level of rpath manipulation, but I can't figure out whether that's related to the Python version being used or if it's something like Python 3.6 being the default version in the CI image...?

joaander:
`delocate` suggests that it ignores `@loader_path` and `@executable_path`. Try setting `INSTALL_RPATH_USE_LINK_PATH True` when building wheels. Once it resolves it from an absolute path, `delocate` will replace it with a relative path.

bdice:
@joaander Thanks for the hint! It looks like delocate still shows a warning, but the built wheels appear to pass tests now. However, I'm unsure if the wheels are actually "delocated" because of that warning. I enabled the `INSTALL_RPATH_USE_LINK_PATH` CMake option for all `APPLE` builds, I don't think I can enable it _only_ for building wheels. We could enable it for only the cibuildwheel environment, based on environment variables -- when do you think it should be enabled?

Previously @vyasr had used this command, I tried to read some docs but wasn't sure whether it's needed. If so, it's possible that a similar trick would be required for libfreud, since it's also not being found by delocate...
```bash
# Force the TBB path to use an absolute path to itself for others to find.
install_name_tool -id "${PWD}/libtbb.dylib" libtbb.dylib
```

joaander:
> @joaander Thanks for the hint! It looks like delocate still shows a warning, but the built wheels appear to pass tests now. However, I'm unsure if the wheels are actually "delocated" because of that warning. 

You could dowload the artifact from the CI, unpack it, and check it with `otool` to see.

> I enabled the `INSTALL_RPATH_USE_LINK_PATH` CMake option for all `APPLE` builds, I don't think I can enable it _only_ for building wheels. We could enable it for only the cibuildwheel environment, based on environment variables -- when do you think it should be enabled?

Yes, you could check the env var for only the cibuildwheel environment.

> Previously @vyasr had used this command, I tried to read some docs but wasn't sure whether it's needed. If so, it's possible that a similar trick would be required for libfreud, since it's also not being found by delocate...
> 
> ```shell
> # Force the TBB path to use an absolute path to itself for others to find.
> install_name_tool -id "${PWD}/libtbb.dylib" libtbb.dylib
> ```

Install names on mac os are a confusing topic and can interfere with relative linking. If TBB has a relative install name, it is most likely the wrong relative install name and it might fix issues if you change the install name to an absolute one: https://stackoverflow.com/questions/35220111/install-name-tool-difference-between-change-and-id

.github/workflows/build_wheels.yml
bdice: Thanks for the tip. I updated this section.

bdice:
@glotzerlab/freud-developers This PR is ready for an initial round of review. Linux and Mac builds appear to work but need to be tested. I would like reviewers to help by downloading the build artifacts and trying to install, particularly on Mac.

.circleci/ci-oldest-reqs.txt
```diff
@@ -1,7 +1,7 @@
 cmake==3.6.3.post1
 cython==0.29.14
 garnett==0.7.1
-gsd==2.0.0
+gsd==2.4.2
```
tommy-waltmann, line 4: Why the large bump in the gsd dependency?

.github/dependabot.yml
```diff
@@ -1,7 +1,13 @@
...
-  schedule:
-    interval: monthly
-  open-pull-requests-limit: 10
+  - package-ecosystem: pip
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    target-branch: maint
```
tommy-waltmann: We don't have a maint branch, what is this for?

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,128 @@
...
+  build_wheels:
+    name: Build wheel for ${{ matrix.os }}, Python ${{ matrix.pyver }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15] #, windows-2019]
+        pyver: [3.6, 3.7, 3.8, 3.9]
+
+    steps:
+      - uses: actions/checkout@v2.3.4
```
tommy-waltmann, line 32: This is more of a question, but what do these uses statements do exactly?

setup.py
```diff
@@ -93,7 +93,6 @@ def setup(*args, **kwargs):
         "gsd>=2.0",
         "garnett>=0.7.1",
         "matplotlib>=3.0",
-        "MDAnalysis>=0.20.1",
```
tommy-waltmann, line 96: Is MDAnalysis a soft dependency now?

joaander:
I will review this next week.

.circleci/ci-oldest-reqs.txt
joaander: This is the first GSD version that provides mac wheels and fixes support PEP 517 enabled builds.

.github/dependabot.yml
joaander: It was copied from `gsd`. It should be removed in `freud`.

.github/dependabot.yml
```diff
@@ -1,7 +1,13 @@
...
 updates:
-- package-ecosystem: pip
-  directory: "/"
-  schedule:
-    interval: monthly
-  open-pull-requests-limit: 10
+  - package-ecosystem: pip
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
joaander, line 6:
```suggestion
      interval: "monthly"
  open-pull-requets-limit: 10
```
In this PR, you are removing the `open-pull-requests-limit: 10`. Is that intentional?

.github/dependabot.yml
```diff
@@ -1,7 +1,13 @@
...
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    target-branch: maint
+    schedule:
+      interval: "monthly"
+    pull-request-branch-name:
+      separator: "-"
```
joaander:
```suggestion
```
You use the default separator above for `pip`.

.github/workflows/build_wheels.yml
joaander: They use scripts written by others:
https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions

.github/workflows/cibuildwheel-before-build.sh
```diff
@@ -0,0 +1,36 @@
...
+PACKAGE_DIR=$1
+PLATFORM=$2
+TBB_VERSION="2021.2.0"
+#if [[ "${TBB_PLATFORM}" == "win" ]]; then
+#    TBB_ZIP="oneapi-tbb-${TBB_VERSION}-${TBB_PLATFORM}.zip"
+#else
+#    TBB_ZIP="oneapi-tbb-${TBB_VERSION}-${TBB_PLATFORM}.tgz"
+#fi
+TBB_ZIP="v${TBB_VERSION}.zip"
+#curl -L -O "https://github.com/oneapi-src/oneTBB/releases/download/v${TBB_VERSION}/${TBB_ZIP}"
```
joaander:
```suggestion
```

.github/workflows/cibuildwheel-before-build.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# Move to a hard-coded path (defined by CIBW_ENVIRONMENT)
+mv "oneTBB-${TBB_VERSION}" "${PACKAGE_DIR}/tbb"
+cd "${PACKAGE_DIR}/tbb"
+mkdir -p build
+cd build
+cmake ../ -DTBB_TEST=OFF
+cmake --build . -j
+cmake -DCOMPONENT=runtime -P cmake_install.cmake
+cmake -DCOMPONENT=devel -P cmake_install.cmake
```
joaander:
```suggestion
cmake --install .
```
Is this a documented method to install TBB from source?

I use `cmake --install` in `glotzerlab-software` and it works, provided I set an installation path with `-DCMAKE_INSTALL_PREFIX=` in the first `cmake` invocation. This installs both headers and libs in standard `<PREFIX>/include` and `<PREFIX>/lib` directories.

You may need to adjust your `CMAKE_PREFIX_PATH` and `LD_LIBRARY_PATH` if you make this change.

CMakeLists.txt
```diff
@@ -111,6 +111,6 @@ set(ignoreMe "${SKBUILD}")
 add_subdirectory(cpp)
 add_subdirectory(freud)
 
-if(_using_conda)
+if(_using_conda OR DEFINED ENV{CIBUILDWHEEL})
   set_target_properties(libfreud PROPERTIES INSTALL_RPATH_USE_LINK_PATH True)
```
joaander, line 115: With this change, I don't think the above `LD_LIBRARY_PATH` setting is necessary.

joaander (changes requested):
Thanks, this looks like a good maintainable solution going forward. I have a few minor inline suggestions to improve below.

We do need to remove the now unused deploy scripts in `.circleci` and all related jobs in `.circleci/config.yml`

.github/workflows/cibuildwheel-before-build.sh
bdice: I think I found this option and chose to use it because it was cross-platform and didn't require me to know which make tool to use. I don't really know what to say about `CMAKE_PREFIX_PATH` or `LD_LIBRARY_PATH` but I would welcome a suggestion if you think there is a simpler way (perhaps one that avoids the workaround in the CMake file to determine `INSTALL_RPATH_USE_LINK_PATH` based on the environment variable `CIBUILDWHEEL`).

setup.py
bdice: MDAnalysis is not required for freud, but we have historically used it when running freud's tests because we provide interoperability (freud can extract periodic boxes and positions from MDAnalysis timesteps). Those intergration tests are skipped if MDAnalysis is unavailable.

I removed MDAnalysis as a test requirement because it has been really difficult for us to find versions that work on all the platforms and Python/OS versions that freud supports. Because of the challenges we've faced both in this PR and historically (there are workarounds in our existing deploy scripts), we have to drop it as a testing dependency. The MDAnalysis package won't support Python 3.9 until version 2.0, which isn't released yet, and it also has had a lot of issues on Mac (no wheels are available, and building from source fails). You can see a number of related issues on this search page: https://github.com/MDAnalysis/mdanalysis/issues?q=3.9

bdice:
@joaander @tommy-waltmann I think I have addressed all your review comments. This is ready for a second look.

.github/workflows/cibuildwheel-before-build.sh
joaander: `cmake --install .` is platform independent (like `cmake --build` it forwards the install command to whatever tool it generated build files for). However, it will not remove the need for `INSTALL_RPATH_USE_LINK_PATH`. It will just put the files under a standard `include/ lib/` hierarchy under $PREFIX.

joaander (approved):
Thanks!

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: Upload to PyPI
+        # upload to PyPI on every tag starting with 'v'
+        if: startsWith(github.ref, 'refs/tags/v')
+        uses: pypa/gh-action-pypi-publish@v1.4.2
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN }}
+
+      - name: Upload to TestPyPI
```
tommy-waltmann, line 113: We seem to be uploading to testPyPI after we upload to PyPI. Should it be the reverse of that?

.github/workflows/cibuildwheel-before-build.sh
bdice: Oh, sorry. I was confused. I misread the original comment and thought your code suggestion _was_ the current code, so I didn't understand what you were trying to indicate. I'll try your suggestion now.

tommy-waltmann (commented):
Just one last quick question

.github/workflows/build_wheels.yml
bdice: The order shouldn't matter -- the conditions are different for each type of upload. Josh has tested this for GSD so I'm not concerned.

tommy-waltmann (approved):

PR #752, opened 2021-04-21
Added reset to DiffractionPattern.

### LINKED ISSUE #648: Add a reset option to DiffractionPattern.compute()
I've been playing around with the `diffractometer` a bit and everything works well.  However, when I try to use it in a similar way I would the `rdf` and `msd` and call `.compute()` over multiple frames (and in this case view orientations) it looks like only the last iteration is stored and returned.

To be able to iterate over multiple snapshots and/or view orientations and accumulate the results of `.compute()` when generating a diffraction pattern.

Something along the lines of:

```
for frame in traj:
  for view in careful_list_of_normla_distribution_views:
     dp.compute(system=frame, view_orientation=view, reset=False)
```

klywang:
## Motivation and Context

Resolves: #648 

## How Has This Been Tested?

Added test to make sure `reset=True` works and that `reset=False` runs.

## Types of changes

## Checklist:

### DIFF
--- a/doc/source/reference/development/design.rst
+++ b/doc/source/reference/development/design.rst
@@ -18,7 +18,7 @@ a Python interpreter.
 
 In order to achieve this goal, **freud** takes the following viewpoints:
 
-* **freud** works directly with `NumPy <http://www.numpy.org/>_`
+* **freud** works directly with `NumPy <https://numpy.org/>`__
   arrays to retain maximum flexibility. Integrations with other tools
   should be performed via the common data representations of NumPy arrays.
 * For ease of maintenance, **freud** uses Git for version control;
@@ -27,7 +27,7 @@ In order to achieve this goal, **freud** takes the following viewpoints:
   to read.
 * To ensure correctness, **freud** employs unit testing using the
   Python :mod:`pytest` framework. In addition, **freud** utilizes
-  `CircleCI <https://circleci.com>`_ for continuous integration to
+  `CircleCI <https://circleci.com>`__ for continuous integration to
   ensure that all of its code works correctly and that any changes or
   new features do not break existing functionality.
 
--- a/freud/diffraction.pyx
+++ b/freud/diffraction.pyx
@@ -63,7 +63,8 @@ cdef class DiffractionPattern(_Compute):
     cdef double[:, :, :] _k_vectors_orig
     cdef double[:] _k_values
     cdef double[:, :, :] _k_vectors
-    cdef list _diffraction
+    cdef double[:, :] _diffraction
+    cdef list _diffraction_fft
     cdef double _box_matrix_scale_factor
     cdef double[:] _view_orientation
     cdef cbool _k_values_cached
@@ -82,8 +83,8 @@ cdef class DiffractionPattern(_Compute):
         # Store these computed arrays which are exposed as properties.
         self._k_values = np.empty_like(self._k_values_orig)
         self._k_vectors = np.empty_like(self._k_vectors_orig)
-        # self._diffraction = np.empty((self.output_size, self.output_size))
-        self._diffraction = []
+        self._diffraction = np.empty((self.output_size, self.output_size))
+        self._diffraction_fft = []
 
     def _calc_proj(self, view_orientation, box):
         """Calculate the inverse shear matrix from finding the projected box
@@ -207,7 +208,7 @@ cdef class DiffractionPattern(_Compute):
                 value: True).
         """
         if reset:
-            self._diffraction = []
+            self._diffraction_fft = []
 
         system = freud.locality.NeighborQuery.from_system(system)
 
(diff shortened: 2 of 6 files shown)


klywang:
How do we want to handle plots when `reset=False`? The first thing that comes to mind is to allow a list of axes, where the number of axes objects are equal to the number of frames. By default, plot could generate a separate figure for each plot.

Please let me know what you think.

bdice:
> 
> 
> How do we want to handle plots when `reset=False`? The first thing that comes to mind is to allow a list of axes, where the number of axes objects are equal to the number of frames. By default, plot could generate a separate figure for each plot.
> 
> Please let me know what you think.

Other modules use `reset=False` to indicate an accumulation. For example, computing a diffraction pattern over several frames with `reset=False` should compute a time-averaged quantity (possibly after normalizing by the number of frames). Thus there should only be one plot.

klywang:
I tried averaging with two different methods (RMS averaging and vector averaging) for a system with 3 frames.
The diffraction for each frame (`reset=True`) looked like this:
frame_diffraction


With RMS averaging, the diffraction looked like this:
rms_diffraction_averaged
Where the left is the diffraction of the system with no noise (from `freud.data.UnitCell.fcc().generate_system(4)`), and the right is the data averaged over 3 frames with noise applied to the position.
FFT average was computed from the square root of the mean of the the FFT data for each frame: `np.sqrt(np.mean(np.absolute(fft_data)**2, axis=0))`

With vector averaging, the diffraction looked like this:
vector_diffraction_averaged
Where the left is the diffraction of the system with no noise (from `freud.data.UnitCell.fcc().generate_system(4)`), and the right is the data averaged over 3 frames with noise applied to the position.
FFT average was computed by averaging the imaginary and real components separately.

Should I add a mode option as well, or would one method be preferred over another? Or was there a different way we wanted to do this?

Using the following code:
```python
import freud
import numpy as np

dp = freud.diffraction.DiffractionPattern()
dp_ref = freud.diffraction.DiffractionPattern()
box, positions = freud.data.UnitCell.fcc().generate_system(4)

n_frames = 3
np.random.seed(42)
noises = np.random.normal(scale=1e-4, size=(n_frames, len(positions), 3))
fig, (ax_ref, ax_avg) = plt.subplots(1, 2)
fig, axes = plt.subplots(1, 3)

for n, noise in enumerate(noises):
    noisy_positions = positions + noise
    dp_ref.compute((box, noisy_positions))
    dp_ref.plot(ax=axes[n])
    dp.compute((box, noisy_positions), reset=False)

dp_ref.compute((box, positions))
dp_ref.plot(ax=ax_ref)
dp.plot(ax=ax_avg)
plt.show()
```


Edit: Fixed plots/code

doc/source/reference/development/design.rst
```diff
@@ -18,7 +18,7 @@ a Python interpreter.
 
 In order to achieve this goal, **freud** takes the following viewpoints:
 
-* **freud** works directly with `NumPy <http://www.numpy.org/>_`
+* **freud** works directly with `NumPy <https://numpy.org/>`__
```
bdice, line 21: @klywang This is a tiny unrelated fix. I noticed this link wasn't rendering correctly while you were showing the docs during our call today.

freud/diffraction.pyx
```diff
@@ -200,7 +202,15 @@ cdef class DiffractionPattern(_Compute):
             peak_width (float):
                 Width of Gaussian convolved with points, in system length units
                 (Default value = 1).
+            reset (bool):
+                Whether to erase the previously computed valuse before adding
+                the new computations; if False, will accumulate data (Default
```
bdice:
```suggestion
                Whether to erase the previously computed values before adding
                the new computations; if False, will accumulate data (Default
```

freud/diffraction.pyx
```diff
@@ -227,18 +237,24 @@ cdef class DiffractionPattern(_Compute):
         # Compute FFT and convolve with Gaussian
         cdef double complex[:, :] diffraction_fft
         diffraction_fft = np.fft.fft2(im)
+
```
bdice:
```suggestion
```

bdice (approved):
@klywang I pushed a few changes to this branch,, it looks good to me. I changed how the accumulation works, so that it's a pure time-average (sum all the diffraction patterns, divide by the number of frames). I think this is the standard way to compute this quantity, but if you know the other way is correct then we can revisit that. I'd want to reference the literature if we use any other formula for the accumulation.

Here is an updated notebook that I used for testing. It's mostly similar to the notebook above, but with some reformatting and a different noise tolerance that makes it easier to see the effect of averaging.
[Diffraction_reset.zip](https://github.com/glotzerlab/freud/files/6362476/Diffraction_reset.zip)


Can you take a quick look at this PR and then merge it once you're happy?

vyasr:
+1 for straightforward averaging or justifying any alternative with literature.

klywang:
@bdice @vyasr I'm fine with the averaging. I double checked and it looks like it is standard.

PR #754, opened 2021-04-26
Shorten tests

vyasr:
## Description

This PR reduces the runtime of most of the most expensive tests in freud. On my Mac, this reduces the total time for tests from ~5:15 to ~1:30. The remaining expensive tests are I think justified in their length: environment matching with global search, the RDF ideal gas test, and Minkowski structure metrics account for roughly half the time, but are important and not easy to simplify further.

## Motivation and Context

Resolves: #751 

## How Has This Been Tested?

Existing tests pass.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation improvement (updates to user guides, docstrings, or developer docs)

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [ ] I have updated the documentation (if relevant).
- [ ] I have added tests that cover my changes (if relevant).
- [ ] All new and existing tests passed.
- [ ] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst).
- [ ] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

I don't think this needs a changelog entry or an update to the credits.

PR #762, opened 2021-04-29
Double tolerance on failing PCF test.

bdice:
## Description
A test comparing the positional correlation function (PCF) is failing unexpectedly. I think it's just a matter of having too low of a tolerance. This started occurring in PR #754 but seems unrelated to the actual content of that PR. The failures have only been observed on GitHub Actions / cibuildwheel, not on CircleCI.

I will merge this when tests pass because it is affecting all of CI. @vyasr I believe you wrote this test. If you have any comments on this change, please feel free to reply and tag me.

## Motivation and Context
Fixes a failing test.

## How Has This Been Tested?
CI should pass with these larger values.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_pmft.py
+++ b/tests/test_pmft.py
@@ -186,7 +186,7 @@ class PMFTTestBase:
         radii = build_radii(get_bin_centers(pmft))
         rdf = pmft_to_rdf(pmft, radii)
 
-        assert np.isclose(np.nanmean(rdf), 1, rtol=1e-2, atol=1e-2)
+        assert np.isclose(np.nanmean(rdf), 1, rtol=2e-2, atol=2e-2)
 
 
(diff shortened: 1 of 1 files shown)


vyasr:
Yes I wrote this. Your changes seems fine to me. None of my PMFT/RDF ideal gas tests are particularly rigorous. What we really should be doing is analyzing whether the deviation from unity is statistically meaningful using something like a t test, but I never went down that route because it's _far_ more expensive computationally. If we ever go through the effort of separating validation tests from unit tests (and writing enough validation tests for this to be worthwhile), and if we switch to running those on dedicated hardware or using a CI platform with more runners, then we can revisit writing true validation tests for more of our methods. For now I think we have to settle for tests like this.

PR #767, opened 2021-05-05
Use np.asarray to convert subclasses. Resolves #766.

### LINKED ISSUE #766: OVITO support broken by changes in _convert_array
**Describe the bug**
I cannot convert an OVITO DataCollection to a freud system-like object. This worked with freud v2.5.1 and I believe it was broken in 8a811d7cc4d8465f81567f01b81a614611250949 or some other piece of #674. @vyasr @Charlottez112

My best guess is that the native dtype of the OVITO Property is `float64` and freud is requiring `float32` but this requirement cannot be met by the read-only array view provided by OVITO's `PropertyView`. While `PropertyView` is seemingly "array-like", it can't be used by `np.require` in the same way as the result of `np.asarray` (view? copy?) that was previously passed in.

**To Reproduce**
1. Install the latest freud (from source) into an OVITO Python environment
2. Open a trajectory in OVITO
3. Use the following viewport modifier:
 
```python
import freud

def render(args):
    pipeline = args.scene.selected_pipeline
    if not pipeline:
        return
    data = pipeline.compute(args.frame)
    system = freud.AABBQuery.from_system(data)
    print(system)
```

**Error output**
```python
The Python script has exited with an error.
  File "<string>", line 12, in render
  File "locality.pyx", line 371, in freud.locality.NeighborQuery.from_system
  File "locality.pyx", line 836, in freud.locality.AABBQuery.__cinit__
  File "util.pyx", line 227, in freud.util._convert_array
  File "/home/bdice/miniconda3/envs/dice/lib/python3.8/site-packages/numpy/core/_asarray.py", line 400, in require
    arr = array(a, dtype=dtype, order=order, copy=False, subok=subok)
  File "/home/bdice/miniconda3/envs/dice/lib/python3.8/site-packages/ovito/data/stdobj/property.py", line 77, in _Property__array__
    return PropertyView(self.__array__impl__(dtype), self)
RuntimeError: Property: Cannot create NumPy array view with dtype other than the native data type of the property.
```

 - OS: Windows / Linux
 - Version of Python: 3.8
 - Version of freud: c68dbce048d77bf5497fbf1066e68088f87497d9

bdice:
## Description
Resolves #766. See issue for details. I'm not adding a changelog / credits entry for this because the breaking change in #674 had not yet been released.

## Motivation and Context
Resolves: #766

## How Has This Been Tested?
Tested with local installation of OVITO, and it fixed the problem. I don't know how to write an effective test for this behavior without extensive mocking of `np.ndarray` subclasses.

## Types of changes

## Checklist:

### DIFF
--- a/freud/util.pyx
+++ b/freud/util.pyx
@@ -224,6 +224,7 @@ def _convert_array(array, shape=None, dtype=np.float32, requirements=("C", ),
     if array is None:
         return np.empty(shape, dtype=dtype)
 
+    array = np.asarray(array)
     return_arr = np.require(array, dtype=dtype, requirements=requirements)
 
     if not allow_copy and return_arr is not array:
--- a/tests/integration/test_reader_integrations.py
+++ b/tests/integration/test_reader_integrations.py
@@ -7,13 +7,6 @@ import pytest
 
 import freud
 
-try:
-    import MDAnalysis
-
-    MDANALYSIS = True
-except ImportError:
-    MDANALYSIS = False
-
 
 def _relative_path(*path):
     return os.path.join(os.path.dirname(__file__), *path)
@@ -37,13 +30,13 @@ class TestReaderIntegrations:
     @pytest.mark.skipif(
         sys.platform.startswith("win"), reason="Not supported on Windows."
     )
-    @pytest.mark.skipif(not MDANALYSIS, reason="MDAnalysis is not installed.")
     def test_mdanalysis_gsd(self):
+        MDAnalysis = pytest.importorskip("MDAnalysis")
         reader = MDAnalysis.coordinates.GSD.GSDReader(LJ_GSD)
         self.run_analyses(reader)
 
-    @pytest.mark.skipif(not MDANALYSIS, reason="MDAnalysis is not installed.")
     def test_mdanalysis_dcd(self):
+        MDAnalysis = pytest.importorskip("MDAnalysis")
         reader = MDAnalysis.coordinates.DCD.DCDReader(LJ_DCD)
         self.run_analyses(reader)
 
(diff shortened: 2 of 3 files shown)


vyasr:
Can you add an integration test for OVITO? The Python module is `pip install`able, and this issue is revealing that we don't have tests for that, whereas we already test `garnett`, `gsd`, and `mdanalysis` (although I know that you've removed the latter from CI due to installation headaches). This isn't the first time that we've run into OVITO-specific issues, so I think we need to try to add that test.

vyasr (changes requested):
See previous comment re: test.

klywang:
It looks like Ovito pro has a [python code generator](https://www.ovito.org/docs/current/advanced_topics/code_generation.html?highlight=scripts) function. Maybe this would be helpful in setting up tests as well? It seems like it automatically generates a Python script from your active visualization workflow, that can be executed in Python either from `ovitos` or you `pip install` ovito

vyasr:
> It looks like Ovito pro has a [python code generator](https://www.ovito.org/docs/current/advanced_topics/code_generation.html?highlight=scripts) function. Maybe this would be helpful in setting up tests as well? It seems like it automatically generates a Python script from your active visualization workflow, that can be executed in Python either from `ovitos` or you `pip install` ovito

[We requested this feature :)](https://gitlab.com/stuko/ovito/-/issues/194)

bdice:
@vyasr @klywang I added a test that runs if `ovito` is present on the system. I don't think it's a good idea to make it a test dependency -- it requires a properly-linked Qt installation for `import ovito` to work. I think it would be a huge headache to maintain that testing dependency, given the challenges I've had with installing OVITO's Python module locally. This test will function correctly if the import succeeds, and I think that's sufficient for an integration test.

bdice:
Also @klywang's idea to use the code generator was very helpful! 😄

vyasr (approved):
I'm fine with not running the test on CI as long as it exists. That's a tacit acknowledgment that we're OK with breaking OVITO support occasionally and fixing later, which I think is true. Otherwise this PR LGTM.

PR #769, opened 2021-05-06
Implement remaining in place operations

### LINKED ISSUE #578: Reduce copying of input data

vyasr:
## Description

Implement in-place `unwrap`, `make_fractional`, and `make_absolute`. This is a continuation of #674.

## Motivation and Context

Resolves: #578 

## How Has This Been Tested?

New tests added.

## Types of changes

## Checklist:


cpp/box/Box.h
```diff
@@ -228,33 +228,33 @@ class Box
...
     }
 
-    //! Compute the position of the particle in box relative coordinates
-    /*! \param p point
-     *  \returns alpha
+    //! Convert a point's coordinate from absolute to relative box coordinates.
+    /*! \param v The vector of the point in absolute coordinates.
+     *  \returns The vector of the point in relative coordinates.
      *
      *  alpha.x is 0 when \a x is on the far left side of the box and
```
bdice: This docstring needs to be rewritten. There is no `alpha`.

cpp/box/Box.h
```diff
@@ -263,13 +263,18 @@ class Box
         return delta;
     }
 
-    void makeFractional(vec3<float>* vecs, unsigned int Nvecs) const
+    //! Convert point coordinates from absolute to relative box coordinates.
```
bdice: I don't think "relative" and "fractional" mean the same thing. "Relative" suggests a translation (in both physics and CSS contexts), whereas fractional conveys the desired meaning to me.
```suggestion
    //! Convert point coordinates from absolute coordinates to fractional box coordinates.
```

cpp/box/Box.h
```diff
@@ -228,33 +228,33 @@ class Box
...
-                vecs[i] = makeAbsolute(vecs[i]);
+                out[i] = makeAbsolute(vecs[i]);
             }
         });
     }
 
-    //! Compute the position of the particle in box relative coordinates
-    /*! \param p point
-     *  \returns alpha
+    //! Convert a point's coordinate from absolute to relative box coordinates.
```
bdice: I don't think "relative" and "fractional" mean the same thing. "Relative" suggests a translation (in both physics and CSS contexts), whereas fractional conveys the desired meaning to me.
```suggestion
    //! Convert a point's coordinate from absolute coordinates to fractional box coordinates.
```

cpp/box/Box.h
```diff
@@ -263,13 +263,18 @@ class Box
...
+     *  \param out The array in which to place the wrapped vectors.
+     */
+    void makeFractional(const vec3<float>* vecs, unsigned int Nvecs, vec3<float>* out) const
     {
         util::forLoopWrapper(0, Nvecs, [=](size_t begin, size_t end) {
-            for (size_t i = begin; i < end; ++i)
-            {
-                vecs[i] = makeFractional(vecs[i]);
-            }
+                for (size_t i = begin; i < end; ++i)
```
bdice: This seems to be indented too far...?

freud/box.pyx
```diff
@@ -180,50 +180,72 @@ cdef class Box:
...
 
-    def make_fractional(self, absolute_coordinates):
+    def make_fractional(self, absolute_coordinates, out=None):
         R"""Convert absolute coordinates into fractional coordinates.
 
         Args:
             absolute_coordinates (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`):
                 Absolute coordinate vector(s).
+            out (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray` or :code:`None`):
+                The array in which to place the absolute positions. It must be
```
bdice:
```suggestion
                The array in which to place the fractional coordinates. It must be
```

freud/box.pyx
```diff
@@ -180,50 +180,72 @@ cdef class Box:
...
-    def make_absolute(self, fractional_coordinates):
+    def make_absolute(self, fractional_coordinates, out=None):
         R"""Convert fractional coordinates into absolute coordinates.
 
         Args:
             fractional_coordinates (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`):
                 Fractional coordinate vector(s), between 0 and 1 within
                 parallelepipedal box.
+            out (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray` or :code:`None`):
+                The array in which to place the absolute positions. It must be
```
bdice:
```suggestion
                The array in which to place the absolute coordinates. It must be
```

freud/box.pyx
```diff
@@ -180,50 +180,72 @@ cdef class Box:
...
+                of dtype `np.float32`. If ``None``, this function will return a
+                newly allocated array (Default value = None).
 
         Returns:
             :math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`:
-                Absolute coordinate vector(s).
+                Absolute coordinate vector(s). If `out` is provided, a
```
bdice:
```suggestion
                Absolute coordinate vector(s). If ``out`` is provided, a
```

freud/box.pyx
```diff
@@ -180,50 +180,72 @@ cdef class Box:
...
                 Absolute coordinate vector(s).
+            out (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray` or :code:`None`):
+                The array in which to place the absolute positions. It must be
+                of dtype `np.float32`. If ``None``, this function will return a
+                newly allocated array (Default value = None).
 
         Returns:
             :math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`:
-                Fractional coordinate vector(s).
+                Fractional coordinate vector(s). If `out` is provided, a
```
bdice:
```suggestion
                Fractional coordinate vector(s). If ``out`` is provided, a
```

freud/box.pyx
```diff
@@ -299,13 +321,15 @@ cdef class Box:
 
         Returns:
             :math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`:
-                Vector(s) wrapped into the box.
+                Vector(s) wrapped into the box. If `out` is provided, a
```
bdice:
```suggestion
                Vector(s) wrapped into the box. If ``out`` is provided, a
```

freud/box.pyx
```diff
@@ -326,10 +350,15 @@ cdef class Box:
...
                 Image indices for vector(s).
+            out (:math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray` or :code:`None`):
+                The array in which to place the unwrapped vectors. It must be of
+                dtype `np.float32`. If ``None``, this function will
+                return a newly allocated array (Default value = None).
 
         Returns:
             :math:`\left(3, \right)` or :math:`\left(N, 3\right)` :class:`numpy.ndarray`:
-                Unwrapped vector(s).
+                Unwrapped vector(s). If `out` is provided, a reference to it is
```
bdice:
```suggestion
                Unwrapped vector(s). If ``out`` is provided, a reference to it is
```

cpp/box/Box.h
```diff
@@ -228,33 +228,33 @@ class Box
...
             }
         });
     }
 
-    //! Compute the position of the particle in box relative coordinates
-    /*! \param p point
-     *  \returns alpha
+    //! Convert a point's coordinate from absolute to relative box coordinates.
+    /*! \param v The vector of the point in absolute coordinates.
+     *  \returns The vector of the point in relative coordinates.
```
```suggestion
     *  \returns The vector of the point in fractional box coordinates.
```

bdice (changes requested):
Comments attached.

cpp/box/Box.h
vyasr: I agree.

freud/box.pyx
```diff
@@ -180,50 +180,72 @@ cdef class Box:
...
         flatten = fractions.ndim == 1
         fractions = np.atleast_2d(fractions)
         fractions = freud.util._convert_array(fractions, shape=(None, 3))
+        out = freud.util._convert_array(
+                out, shape=fractions.shape, allow_copy=False)
 
         cdef const float[:, ::1] l_points = fractions
         cdef unsigned int Np = l_points.shape[0]
-        self.thisptr.makeAbsolute(<vec3[float]*> &l_points[0, 0], Np)
+        cdef const float[:, ::1] l_out = out
```
bdice: Probably should not be const. I’m guessing this only passes tests because the const-ness is not preserved when passing to C++?
```suggestion
        cdef float[:, ::1] l_out = out
```

freud/box.pyx
```diff
@@ -180,50 +180,72 @@ cdef class Box:
...
         """  # noqa: E501
         fractions = np.asarray(fractional_coordinates).copy()
         flatten = fractions.ndim == 1
         fractions = np.atleast_2d(fractions)
         fractions = freud.util._convert_array(fractions, shape=(None, 3))
+        out = freud.util._convert_array(
+                out, shape=fractions.shape, allow_copy=False)
 
         cdef const float[:, ::1] l_points = fractions
         cdef unsigned int Np = l_points.shape[0]
```
bdice, line 208: Not sure what our typical convention is here but const might be appropriate.
```suggestion
        cdef const unsigned int Np = l_points.shape[0]
```

bdice (approved):
A couple minor questions about const-ness. These apply to all the methods but I only wrote the comments once. One thing I’m not sure about — since the output memory view is currently (incorrectly) declared as const, my guess is that constness doesn’t apply when calling into C++. But does that constness translate into a non-writable NumPy array flag? That is, with the current const, is the returned output modifiable?

LGTM after constness fixes.

freud/box.pyx
vyasr: The way that Cython translates to C you cannot declare stack variables like this `const` because their declaration will happen well before they are initialized.

vyasr:
> A couple minor questions about const-ness. These apply to all the methods but I only wrote the comments once. One thing I’m not sure about — since the output memory view is currently (incorrectly) declared as const, my guess is that constness doesn’t apply when calling into C++. But does that constness translate into a non-writable NumPy array flag? That is, with the current const, is the returned output modifiable?
> 
> LGTM after constness fixes.

The `const` only applies to the memoryview itself, so in this case it affects operations on `l_out`. The value we pass to C++ is just the pointer acquired by taking the address of the initial value on the array, so you are correct that the `const`ness doesn't translate. It also does not translate to the output because the memoryview is just that, a view on memory owned by another object. That other object is a numpy array created via `_convert_array`, which does not set the `WRITEABLE` flag to `False`, and that is the object that gets returned, so it is not read-only.

PR #770, opened 2021-05-06
Some code cleanup / improvements separated from PR #660.

bdice:
## Description
Pulling out some unrelated changes from #660 to make that PR easier to review and merge.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/locality/BondHistogramCompute.h
+++ b/cpp/locality/BondHistogramCompute.h
@@ -1,5 +1,5 @@
-#ifndef HISTOGRAM_COMPUTE_H
-#define HISTOGRAM_COMPUTE_H
+#ifndef BOND_HISTOGRAM_COMPUTE_H
+#define BOND_HISTOGRAM_COMPUTE_H
 
 #include "Box.h"
 #include "Histogram.h"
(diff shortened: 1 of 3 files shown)


vyasr:
Why make all the getters `const`? Most (although I can't say definitively _all_) are returning copies of data that are constructed when the method is called, so it's not intrinsically unsafe in those cases. I guess it's fine, most of them aren't things that need to get changed anyway, the non-`const` version would make it easy (if we ever expose a C++ API) to construct one histogram by modifying the outputs acquired from the getters of another.

bdice:
> 
> 
> Why make all the getters `const`? Most (although I can't say definitively _all_) are returning copies of data that are constructed when the method is called, so it's not intrinsically unsafe in those cases. I guess it's fine, most of them aren't things that need to get changed anyway, the non-`const` version would make it easy (if we ever expose a C++ API) to construct one histogram by modifying the outputs acquired from the getters of another.

We can revert this if you don't think it's a good idea. My intention was to convey meaning / semantics about the Histogram, as well as making the API a little more restrictive against users doing things with the bin locations that just seem silly. I have no strong preference.

vyasr:
Not a huge deal either way, now that you've I'd say leave it as is for now and we can reevaluate if and when we actually expose the C++ API.

vyasr (approved):

PR #771, opened 2021-05-07
Skip TestPyPi upload when secrets are not available

joaander:
## Description

Use GitHub Actions conditions to skip the TestPyPi upload when secrets are not available.

## Motivation and Context

This resolves failing `build_wheels.yml` workflows on pull requests from forks and dependabot.

## How Has This Been Tested?

I tested this change on GSD with a test pull request from a fork: https://github.com/glotzerlab/gsd/pull/97

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -111,8 +111,8 @@ jobs:
           password: ${{ secrets.PYPI_API_TOKEN }}
 
       - name: Upload to TestPyPI
-        # otherwise, upload to TestPyPi
-        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        # otherwise, upload to TestPyPi when secrets are available
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') && (github.event_name != 'pull_request' || (github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]')) }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -111,8 +111,8 @@ jobs:
           password: ${{ secrets.PYPI_API_TOKEN }}
 
       - name: Upload to TestPyPI
-        # otherwise, upload to TestPyPi
-        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        # otherwise, upload to TestPyPi when secrets are available
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') && (github.event_name != 'pull_request' || (github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]')) }}
```
bdice, line 115: Should we also exclude `pre-commit-ci`? https://github.com/glotzerlab/freud/pull/765

joaander: Those PRs are passing, so it seems we don't need to now. We could add the exception if we think `pre-commit` pull requests might run with reduced privileges in the future. However, the pre-commit auto-fix commits are applied directly to user-generated pull requests, so the app already has indirect access to our secrets.

bdice (approved):

PR #773, opened 2021-05-13
Add Support for Multiple l in Steinhardt

### LINKED ISSUE #747: Steinhardt OP Performance Improvements and Potential API Change
## Description

Using multiple `Steinhardt` objects to compute different `l` values is significantly slower than using `LocalDescriptors` with a custom Numba function to compute Steinhardt OP from the spherical harmonics.

I profiled both code paths with pre-computed neighbor lists for computing all Minkowski structure metrics (i.e. weighted Steinhardt) up to `l=12` and found three reason for the difference in speed.
1. `PointSphEvaluator::compute` computes for all `l` values up to some maximum (in the `fsph` library).
2. The construction of `PointSphEvaluator` for each `Steinhardt` object is costly as the constructor is not a trivial one.
3. The repeated iteration over the neighbor list.

The difference in performance is mostly due to the first two reason.

## Potential Solutions

* Allow for `fsph` to only compute the desired spherical harmonics.
* Move the computational cost of constructing `PointSphEvaluator` objects outside the constructor (likely more involved than the point above).
* Allow `Steinhardt` to compute the Steinhardt OP for multiple `l` (negates the need for the above bullet). We can do this without changing API by
  + Returning the same shaped arrays when a single `l` is chosen
  + Allowing either a sequence of ints or an int in the constructor of `Steinhardt`
  + Returning a list of ints or an int conditionally for the `l` property based on the number of `l` chosen
 
With respect to `fsph` we could allow for two types of selections either ranges (e.g. [4, 6], [0, 4]) or arbitrary `l` (e.g. [1, 4, 3, 9]). The first is likely easier to implement, but this could also change how we would go about extending the `Steinhardt` class if we chose to do so.

## Alternatives

I can keep using my custom Numba solution for now.

@klarh @bdice let me know of any thoughts you have regarding this.

Edit:
I am linking the notebook I used for the original benchmarking, and the Python scripted used for profiling with
[freud-benchmark.zip](https://github.com/glotzerlab/freud/files/6288469/freud-benchmark.zip)
 `py-spy`.

b-butler:
## Description

This adds support for multiple `l` values in the `Steinhardt` class. I also introduce significant speed ups through three changes
- Only construct `fsph::PointSPHEvaluator` once per thread per `baseCompute` (the construction is somewhat costly)
- Reduce allocations of `std::vector` this surprisingly comprised a decent bit of the time computing.
- Reduce calls to `ManagedArray::getIndex` or the use of the `operator()`. This reduces `std::vector` allocations further and reduces computation by accessing the data directly through `operator[]`.

The changes are implemented in a backwards compatible way. Currently if the user only provides a single `l` value, everything behaves as before, but if they provide multiple, we return a list of parameters for all the order parameters provided by the compute.

Edit: The new approach uses a multidimensional `ManagedArray` and NumPy arrays.

## Motivation and Context

Resolves: #747

This is desired since using multiple `Steinhardt` objects is significantly slower than using `LocalDescriptors` and a custom Numba solution to compute the Steinhardt order parameters (even in serial).

Here is a brief benchmark comparing the new code to the old and the custom code mentioned in issue #747. For more details see the comment below.

| Name | Time| Commit |
| ------- | ------- | -------- |
| New Code (l = [0, 12]) | ~169ms | 345a0dd |
| New Code with `ManagedArray`s | ~330ms | 8371e9a |
| Custom Code (l = [0, 12]) | ~334ms | NA |
| New Code (l=6) | ~57.7ms | 345a0dd |
| Old Code (l=6) | ~190ms | f9599480 |

\* The commit hash for "Old Code" may be wrong. I just used the HEAD of the master branch for comparison.

Edit: Due to rebasing the new code commit hash does not exist on the branch any more.

## How Has This Been Tested?

I added a test like the one that checks the validity of the spherical harmonic computation using multiple `l` values on a single compute.

## Types of changes

## Checklist:


cpp/order/Steinhardt.cc
```diff
@@ -90,15 +99,37 @@ void Steinhardt::compute(const freud::locality::NeighborList* nlist,
...
     freud::locality::loopOverNeighborsIterator(
         points, points->getPoints(), m_Np, qargs, nlist,
         [=](size_t i, const std::shared_ptr<freud::locality::NeighborPerPointIterator>& ppiter) {
             float total_weight(0);
             const vec3<float> ref((*points)[i]);
+            // Construct PointSPHEvaluator outside loop since the construction is costly.
+            auto max_l = std::max_element(m_ls.begin(), m_ls.end());
+            if (max_l == m_ls.end())
+            {
+                throw std::invalid_argument("Selected l's do not have a maximum.");
```
bdice: Is this code path reachable? I don't think this is a case we need to catch. We would be able to simplify the code if this was removed (make `max_l` `const auto` and dereference the pointer immediately).

cpp/order/Steinhardt.cc
```diff
@@ -184,80 +237,113 @@ void Steinhardt::computeAve(const freud::locality::NeighborList* nlist,
...
-                        // vector function signature for indexing into the
-                        // arrays for speed.
-                        m_qlmiAve({static_cast<unsigned int>(i), k}) += m_qlmi({nb2.point_idx, k});
+                        auto& qlmiAve = m_qlmiAve[l_index];
+                        auto& qlmi = m_qlmi[l_index];
+                        for (unsigned int k = 0; k < m_num_ms[l_index]; ++k)
+                        {
+                            // Adding all the qlm of the neighbors. We use the
+                            // vector function signature for indexing into the
+                            // arrays for speed.
```
bdice, line 249: Would this be better to cache the base index and re-use `base_index + k` for all `k`? That pattern is used in a few other places that were changed for this PR.
_edit: I made this change in my branch with other suggestions_

freud/order.pyx
```diff
@@ -543,11 +544,12 @@ cdef class Steinhardt(_PairCompute):
         to observe a value of 0 for the per-particle order parameter even with
         a finite number of neighbors. If you would like to ignore this
         distinction, you can mask the output order parameter values using
-        NumPy: :code:`numpy.nan_to_num(particle_order)`.
+        NumPy: :code:`numpy.nan_to_num(particle_order[0])`.
```
bdice: I think the `[0]` is not helpful here -- it's assuming you have an array named `particle_order` and this `[0]` only applies if users have multiple l values.
```suggestion
        NumPy: :code:`numpy.nan_to_num(particle_order)`.
```

freud/order.pyx
```diff
@@ -543,11 +544,12 @@ cdef class Steinhardt(_PairCompute):
...
 
     Args:
-        l (unsigned int):
-            Spherical harmonic quantum number l.
+        l (unsigned int or list of unsigned int):
```
```suggestion
        l (unsigned int or sequence of unsigned int):
```

freud/order.pyx
```diff
@@ -543,11 +544,12 @@ cdef class Steinhardt(_PairCompute):
...
+            One or more spherical harmonic quantum number l's to compute the
+            Steinhardt order parameter for.
```
bdice: Avoid ending with a preposition.
```suggestion
            One or more spherical harmonic quantum number l's used to compute
            the Steinhardt order parameter.
```

freud/order.pyx
```diff
@@ -568,6 +570,8 @@ cdef class Steinhardt(_PairCompute):
 
     def __cinit__(self, l, average=False, wl=False, weighted=False,
                   wl_normalize=False):
+        if not hasattr(l, "__len__"):
```
bdice: Use `isinstance(l, collections.abc.Sequence)`.

freud/order.pyx
```diff
@@ -598,41 +602,64 @@ cdef class Steinhardt(_PairCompute):
     @property
     def l(self):  # noqa: E743
         """unsigned int: Spherical harmonic quantum number l."""
-        return self.thisptr.getL()
+        ls = self.thisptr.getL()
+        if ls.size() == 1:
+            return ls[0]
+        return ls
```
```suggestion
        return ls[0] if ls.size() == 1 else ls
```

freud/order.pyx
```diff
@@ -598,41 +602,64 @@ cdef class Steinhardt(_PairCompute):
...
-        """float: The system wide normalization of the :math:`q_l` or
-        :math:`w_l` order parameter."""
-        return self.thisptr.getOrder()
+        """float or list of float: The system wide normalization of the
+        :math:`q_l` or :math:`w_l` order parameter for each ``l`` selected. If
+        only 1 ``l`` was selected returns a list of normalizations."""
+        order = self.thisptr.getOrder()
+        if order.size() == 1:
+            return order[0]
+        return order
```
```suggestion
        return order[0] if order.size() == 1 else order
```

freud/order.pyx
```diff
@@ -598,41 +602,64 @@ cdef class Steinhardt(_PairCompute):
...
+        of `numpy.ndarray`: Variant of the Steinhardt order parameter for each
+        particle (filled with :code:`nan` for particles with no neighbors).
+        Is a list of NumPy arrays if more than one ``l`` was specified."""
+        order_arrays = self.thisptr.getParticleOrder()
+        if order_arrays.size() == 1:
+            return freud.util.make_managed_numpy_array(
+                &order_arrays[0], freud.util.arr_type_t.FLOAT)
+        return [freud.util.make_managed_numpy_array(&order_arrays[i],
+                                                    freud.util.arr_type_t.FLOAT)
+                for i in range(order_arrays.size())]
```
bdice: Simplify this using a similar construction to the suggestions above. Also applies below. Avoid repeating the complicated syntax of `freud.util.make_managed_numpy_array`.

freud/order.pyx
```diff
@@ -598,41 +602,64 @@ cdef class Steinhardt(_PairCompute):
...
-        """:math:`\\left(N_{particles}\\right)` :class:`numpy.ndarray`: Variant
-        of the Steinhardt order parameter for each particle (filled with
-        :code:`nan` for particles with no neighbors)."""
-        return freud.util.make_managed_numpy_array(
-            &self.thisptr.getParticleOrder(),
-            freud.util.arr_type_t.FLOAT)
+        """:math:`\\left(N_{particles}\\right)` :class:`numpy.ndarray` or list
+        of `numpy.ndarray`: Variant of the Steinhardt order parameter for each
+        particle (filled with :code:`nan` for particles with no neighbors).
+        Is a list of NumPy arrays if more than one ``l`` was specified."""
```
bdice: **Design question:** As a user, I would prefer to get higher-dimensional NumPy arrays instead of lists-of-arrays (or vectors of ManagedArrays in C++) in any place where the dimensions match (i.e. not with a size that varies with `l`, of course).

bdice (changes requested):
Some suggested changes here: https://github.com/glotzerlab/freud/compare/feature/multi-sph-l...feature/multi-sph-l-bdice

Please share whatever numbers/benchmarks you have handy. No need to do a _ton_ of work for that, but it would be nice to know how much of a performance difference there is for a system with 4000 particles and `l = [4, 5, 6, 7, 8, 9, 10, 11, 12]`.

Please re-request my review once you've had a chance to merge in my suggested changes and address comments. Thanks!

freud/_order.pxd
```diff
@@ -70,20 +70,20 @@ cdef extern from "HexaticTranslational.h" namespace "freud::order":
...
         unsigned int getNP() const
         void compute(const freud._locality.NeighborList*,
                      const freud._locality.NeighborQuery*,
                      freud._locality.QueryArgs) except +
-        const freud.util.ManagedArray[float] &getQl() const
-        const freud.util.ManagedArray[float] &getQlm() const
-        const freud.util.ManagedArray[float] &getParticleOrder() const
-        float getOrder() const
+        const vector[freud.util.ManagedArray[float]] &getQl() const
+        const vector[freud.util.ManagedArray[float complex]] &getQlm() const
```
bdice: This is making our CI sad. Is there a compatibility issue with that version of Cython, or something weird with this syntax? Might try a typedef of `float complex` that is a single symbol, for Cython to recognize this syntax...
```
Error compiling Cython file:
------------------------------------------------------------
...
        unsigned int getNP() const
        void compute(const freud._locality.NeighborList*,
                     const freud._locality.NeighborQuery*,
                     freud._locality.QueryArgs) except +
        const vector[freud.util.ManagedArray[float]] &getQl() const
        const vector[freud.util.ManagedArray[float complex]] &getQlm() const
                                                  ^
------------------------------------------------------------

/root/ci/freud/freud/_order.pxd:79:51: Expected ']', found 'complex'
```

freud/order.pyx
b-butler: For some reason at the time I was trying to avoid the import, but will do.

freud/order.pyx
b-butler: I thought about this, but this was the easiest thing to do first pass. I think I can modify it to have that behavior without many changes though.

freud/_order.pxd
b-butler: I am not sure, it passes on some tests, and locally, but not for some :man_shrugging: ? But I can typedef and see if that helps.

Edit: use emoji rather than text that formats weirdly.

cpp/order/Steinhardt.cc
b-butler: Only if they pass an empty list, which we can check for elsewhere.

cpp/order/Steinhardt.cc
b-butler: Yes it would be better, I focused on `baseCompute`, but I should extend that focus.

cpp/order/Steinhardt.cc
bdice: That should raise an error in the constructor. Can't believe I didn't think of the empty case, I stared at that for a while. 🙃

b-butler:
Some rough bench-marking on a FCC system of 16,000 particles with some noise. The benchmarks were run on my laptop which has 4 cores with hyperthreading (an i5-8250U for those interested). I did not bother to close all my programs to get the cleanest reading. I ran the benchmarks in a Jupyter notebook with the `%%timeit` magic command a few times and made sure they were sufficiently close. All neighborlist were precomputed to prevent this from having an effect. This is for the standard computation (not averaged or `Wl`). 

| Name | Time| Commit |
| ------- | ------- | -------- |
| New Code (l = [0, 12]) | ~169ms | 345a0dd |
| New Code with `ManagedArray`s | ~330ms | 8371e9a |
| Custom Code (l = [0, 12]) | ~334ms | NA |
| New Code (l=6) | ~57.7ms | 345a0dd |
| Old Code (l=6) | ~190ms | f9599480 |

I am going to put this at the top of the PR but wanted to record when I added it as well.

Edit: Added with`ManagedArray` and commit hashes. The one for the old code may not be correct, I used the current HEAD on master for comparison.

freud/order.pyx
```diff
@@ -598,41 +605,59 @@ cdef class Steinhardt(_PairCompute):
     @property
     def l(self):  # noqa: E743
         """unsigned int: Spherical harmonic quantum number l."""
-        return self.thisptr.getL()
+        ls = self.thisptr.getL()
+        # list conversion is necessarily explicit as otherwise CI Cython
+        # complains about compiling the expression with two different types.
+        return ls[0] if ls.size() == 1 else list(ls)
```
bdice: Or just dodge the Cython confusion entirely by pre-emptively creating a list. Also applies below.
```suggestion
        ls = list(self.thisptr.getL())
        return ls[0] if len(ls) == 1 else ls
```

b-butler:
I changed to using `ManagedArray` objects over vectors of `ManagedArray` objects. I did notice it was slower about the same speed for the custom code benchmarked above and mentioned in #747.

bdice:
> I changed to using `ManagedArray` objects over vectors of `ManagedArray` objects. I did notice it was slower about the same speed for the custom code benchmarked above and mentioned in #747.

We don’t have to make this change if there is a measurable performance difference. I am a little surprised, however. Can you share your benchmarks with commit ids for each version tested? I can review the implementation and see what might be wrong with the performance if you agree that this design choice (one large array) is preferable.

b-butler:
@bdice I added commit hashes. I may not get back to this today to fix the tests, but it works locally for me, and feel free to test our the commits locally to see if you find what might be causing the slow down. I think the singular NumPy array is nice if we can do it without sacrificing significant performance.

vyasr:
I opted for simplicity, generality, and safety in the implementation of `ManagedArray` over performance wherever there was an obvious choice to be made. In code where we have higher dimensional arrays (like PMFTs) or in general in any code where we index into ManagedArray objects in nested loops, those choices have noticeable performance penalties. I've worked around these before, but I can work with @b-butler to resolve this more fundamentally if we think it's worthwhile. There will by necessity be some tradeoffs to be made if we choose that, at the very least in the form of more code to maintain if we try to maintain both slow/safe and fast/unsafe code paths, but if the bottlenecks are hindering analysis then maybe it's worth it.

b-butler:
@vyasr  I think this was the right choice especially since generally we don't need this high performance multidimensional access. I am not sure if this case would warrant creating unsafe paths for performance reasons, but if you have thoughts on improvements on my usage please suggest them.

b-butler:
@vyasr and @bdice what are your thoughts on what we should do from here with both working versions.

bdice:
I see the same performance between commit 345a0dd7, which used a vector of ManagedArrays, and commit 13e856a, which uses a 2D Managed Array. Both commits are about 2x faster than the `master` branch for large systems and one value of `l`.

345a0dd7:
image

13e856a:
image

cpp/order/Steinhardt.cc
```diff
@@ -184,80 +234,117 @@ void Steinhardt::computeAve(const freud::locality::NeighborList* nlist,
...
                     {
-                        // Adding all the qlm of the neighbors. We use the
-                        // vector function signature for indexing into the
-                        // arrays for speed.
-                        m_qlmiAve({static_cast<unsigned int>(i), k}) += m_qlmi({nb2.point_idx, k});
+                        auto& qlmiAve = m_qlmiAve[l_index];
+                        auto& qlmi = m_qlmi[l_index];
+                        const auto ave_index = qlmiAve.getIndex({static_cast<unsigned int>(i), 0});
+                        const auto nb2_index = qlmi.getIndex({nb2.point_idx, 0});
+                        for (unsigned int k = 0; k < m_num_ms[l_index]; ++k)
```
bdice: Double-check that we use `size_t` for indexing.
```suggestion
                        for (size_t k = 0; k < m_num_ms[l_index]; ++k)
```

cpp/order/Steinhardt.cc
```diff
@@ -184,80 +234,117 @@ void Steinhardt::computeAve(const freud::locality::NeighborList* nlist,
...
-                m_qlm_local.local()[k] += m_qlmiAve[index] / float(m_Np);
-                // Add the norm, which is the complex squared magnitude
-                m_qliAve[i] += norm(m_qlmiAve[index]);
+                auto& qlmiAve = m_qlmiAve[l_index];
+                auto& qlmi = m_qlmi[l_index];
+                auto& qlm_local = m_qlm_local[l_index];
+                const size_t index = qlmiAve.getIndex({static_cast<unsigned int>(i), 0});
+                const size_t qliAve_index = qliAve_i_start + l_index;
+
+                for (unsigned int k = 0; k < m_num_ms[l_index]; ++k)
```
bdice:
```suggestion
                for (size_t k = 0; k < m_num_ms[l_index]; ++k)
```

bdice (approved):
A couple minor suggestions, otherwise LGTM. I read through this kind of quickly but I think it's fine, the diffs are just messy with the additional level of indentation. We have pretty good tests for this class now, so I feel relatively comfortable with the changes.

cpp/order/Steinhardt.cc
b-butler: We do, I think I got confused with something converting to `unsigned int` rather than `size_t`.

cpp/order/Steinhardt.h
```diff
@@ -58,15 +61,19 @@ class Steinhardt
...
-     *  \param l Spherical harmonic number l.
-     *           Must be a positive number.
+     *  \param l Spherical harmonic numbers l. Must be non-negative integers.
      */
-    explicit Steinhardt(unsigned int l, bool average = false, bool wl = false, bool weighted = false,
-                        bool wl_normalize = false)
-        : m_l(l), m_num_ms(2 * l + 1), m_average(average), m_wl(wl), m_weighted(weighted),
-          m_wl_normalize(wl_normalize), m_qlm_local(2 * l + 1)
+    explicit Steinhardt(const std::vector<unsigned int>& ls, bool average = false, bool wl = false,
+                        bool weighted = false, bool wl_normalize = false)
```
vyasr, line 67: Define a delegating constructor with the old signature that wraps `l` in an initializer list and forwards everything else along. That will make things easier for users intending to continue using this code the old way, especially using the C++ API.

cpp/order/SolidLiquid.cc
```diff
@@ -10,7 +10,7 @@ namespace freud { namespace order {
 
 SolidLiquid::SolidLiquid(unsigned int l, float q_threshold, unsigned int solid_threshold, bool normalize_q)
     : m_l(l), m_num_ms(2 * l + 1), m_q_threshold(q_threshold), m_solid_threshold(solid_threshold),
-      m_normalize_q(normalize_q), m_steinhardt(l), m_cluster()
+      m_normalize_q(normalize_q), m_steinhardt({l}), m_cluster()
```
vyasr: If you define a delegating constructor as I proposed this change can be removed.

cpp/order/Steinhardt.h
```diff
@@ -48,6 +49,8 @@ namespace freud { namespace order {
  * If the flag wl_normalize is set, the third-order invariant wl order parameter
  * will be normalized.
  *
+ * We compute the Steinhardt order parameter for one or more l specified in m_ls.
+ *
```
vyasr: I don't think this is needed, the parameter is documented (probably should rewrite that to clearly emphasize that it's one or more values).

cpp/order/Steinhardt.h
```diff
@@ -58,15 +61,19 @@ class Steinhardt
 public:
     //! Steinhardt Class Constructor
     /*! Constructor for Steinhardt analysis class.
-     *  \param l Spherical harmonic number l.
-     *           Must be a positive number.
+     *  \param l Spherical harmonic numbers l. Must be non-negative integers.
```
vyasr:
```suggestion
     *  \param l One or more values of the spherical harmonic parameter `l`. Must be non-negative integers.
```

cpp/order/Steinhardt.h
```diff
@@ -137,16 +144,18 @@ class Steinhardt
...
 
 private:
     //! \internal
     //! Spherical harmonics calculation for Ylm filling a
-    //  std::vector<std::complex<float> > with values for m = 0, 1, ..., l, -1, ..., -l
-    void computeYlm(const float theta, const float phi, std::vector<std::complex<float>>& Ylm) const;
+    //  std::vector<std::vector<std::complex<float> > > with values for m = 0, 1, ..., l, -1, ..., -l
+    //  for all l in m_ls
+    void computeYlm(fsph::PointSPHEvaluator<float>& sph_eval, const float theta, const float phi,
+                    std::vector<std::vector<std::complex<float>>>& Ylms) const;
```
vyasr: Use the `YlmsType` once it's moved.

cpp/order/Steinhardt.cc
```diff
@@ -4,55 +4,63 @@
...
 #include "utils.h"
+#include <vector>
 
 /*! \file Steinhardt.cc
     \brief Computes variants of Steinhardt order parameters.
 */
 
 namespace freud { namespace order {
 
+    using YlmsType = std::vector<std::vector<std::complex<float>>>;
```
vyasr: Move this to the header so you can also use it there.

cpp/order/Steinhardt.cc
```diff
@@ -90,15 +101,35 @@ void Steinhardt::compute(const freud::locality::NeighborList* nlist,
...
+    }
+
     freud::locality::loopOverNeighborsIterator(
         points, points->getPoints(), m_Np, qargs, nlist,
         [=](size_t i, const std::shared_ptr<freud::locality::NeighborPerPointIterator>& ppiter) {
             float total_weight(0);
             const vec3<float> ref((*points)[i]);
+            // Construct PointSPHEvaluator outside loop since the construction is costly.
+            auto max_l = *std::max_element(m_ls.begin(), m_ls.end());
+            fsph::PointSPHEvaluator<float> sph_eval(max_l);
```
vyasr, line 121: It's worth double-checking this in case I'm wrong (haven't looked at fsph much in a while), but IIRC the evaluator is stateless and constructing the iterators is entirely stateless, so this logic can be removed from the lambda entirely and the evaluator captured by reference in the lambda. Probably won't have any meaningful performance impact relative to doing this in parallel, but it's cleaner if you can do this once and is a slightly reduction in the code's memory footprint.

cpp/order/Steinhardt.cc
```diff
@@ -122,32 +153,49 @@ void Steinhardt::baseCompute(const freud::locality::NeighborList* nlist,
...
-                    m_qlmi({static_cast<unsigned int>(i), k}) += weight * Ylm[k];
+                    auto& qlmi = m_qlmi[l_index];
+                    auto& Ylm = Ylms[l_index];
+                    // Get the initial index and iterate using ++ for faster iteration
+                    // Profiling showed using operator() to slow the code significantly.
+                    const size_t index = qlmi.getIndex({i, 0});
+                    for (size_t k = 0; k < m_num_ms[l_index]; ++k)
+                    {
+                        qlmi[index + k] += weight * Ylm[k];
+                    }
```
vyasr, line 166:
```suggestion
                    const size_t index = qlmi.getIndex({i, 0});
                    const size_t max_index = index + m_num_ms[l_index];
                    for (size_t k = index; k < max_index; ++k)
                    {
                        qlmi[k] += weight * Ylm[k];
                    }
```

cpp/order/Steinhardt.cc
```diff
@@ -122,32 +153,49 @@ void Steinhardt::baseCompute(const freud::locality::NeighborList* nlist,
...
+                auto& qlmi = m_qlmi[l_index];
+                auto& qlm_local = m_qlm_local[l_index];
+                const size_t index = qlmi.getIndex({i, 0});
+                const size_t qli_index = qli_i_start + l_index;
+
+                for (size_t k = 0; k < m_num_ms[l_index]; ++k)
                 {
-                    m_qlm_local.local()[k] += m_qlmi[index] / float(m_Np);
+                    // Cache the index for efficiency.
+                    qlmi[index + k] /= total_weight;
```
vyasr: You can make the same change that I made above (starting the loop index `k` at `index`) to avoid having to shift the index inside this loop. That'll help you more here since it's happening multiple times.

cpp/order/Steinhardt.cc
```diff
@@ -184,80 +236,117 @@ void Steinhardt::computeAve(const freud::locality::NeighborList* nlist,
...
                     {
-                        // Adding all the qlm of the neighbors. We use the
-                        // vector function signature for indexing into the
-                        // arrays for speed.
-                        m_qlmiAve({static_cast<unsigned int>(i), k}) += m_qlmi({nb2.point_idx, k});
+                        auto& qlmiAve = m_qlmiAve[l_index];
+                        auto& qlmi = m_qlmi[l_index];
+                        const auto ave_index = qlmiAve.getIndex({i, 0});
+                        const auto nb2_index = qlmi.getIndex({nb2.point_idx, 0});
+                        for (size_t k = 0; k < m_num_ms[l_index]; ++k)
```
vyasr, line 245: Loop over `[ave_index, ave_index + m_num_ms[l_index)`.

cpp/order/Steinhardt.cc
```diff
@@ -184,80 +236,117 @@ void Steinhardt::computeAve(const freud::locality::NeighborList* nlist,
...
-                m_qlm_local.local()[k] += m_qlmiAve[index] / float(m_Np);
-                // Add the norm, which is the complex squared magnitude
-                m_qliAve[i] += norm(m_qlmiAve[index]);
+                auto& qlmiAve = m_qlmiAve[l_index];
+                auto& qlmi = m_qlmi[l_index];
+                auto& qlm_local = m_qlm_local[l_index];
+                const size_t index = qlmiAve.getIndex({i, 0});
+                const size_t qliAve_index = qliAve_i_start + l_index;
+
+                for (size_t k = 0; k < m_num_ms[l_index]; ++k)
```
vyasr, line 268: Shift loop indices.

cpp/order/Steinhardt.cc
```diff
@@ -184,80 +236,117 @@ void Steinhardt::computeAve(const freud::locality::NeighborList* nlist,
...
+            const auto target_particle_index = target.getIndex({i, 0});
+            const auto norm_particle_index = normalization_source.getIndex({i, 0});
+            for (size_t l_index = 0; l_index < m_ls.size(); ++l_index)
             {
-                const float normalization = std::sqrt(normalizationfactor) / normalization_source[i];
-                target[i] *= normalization * normalization * normalization;
+                const auto l = m_ls[l_index];
+                const auto& source_l = source[l_index];
+
+                const auto normalizationfactor = static_cast<float>(4.0 * M_PI / m_num_ms[l_index]);
```
vyasr, line 337: Considering how often we compute this normalization factor, perhaps it would be worth defining a compile-time evaluated template function `getNormalizationFactor<num_ms>()` encoding the values for the supported range of `l` values? Currently we're limited anyway by the values of the Wigner3j coefficients we hard code (I plan to replace that eventually), so you could match that with a compile-time generation of the normalization. I'm happy to take that on myself at a later point if you want to punt, though.

freud/order.pyx
```diff
@@ -598,41 +605,59 @@ cdef class Steinhardt(_PairCompute):
     @property
     def l(self):  # noqa: E743
         """unsigned int: Spherical harmonic quantum number l."""
-        return self.thisptr.getL()
+        # list conversion is necessary as otherwise CI Cython complains about
+        # compiling the below expression with two different types.
```
vyasr, line 609: :(

vyasr (changes requested):
This looks good overall. I'll defer to you on performance changes, I trust that you and Bradley have tested sufficiently. Regarding the ManagedArray discussion, the thing to keep in mind is that conceptually `ManagedArray` is really a low-level RAII wrapper around a pointer to a contiguous memory buffer whose ownership can be transferred. `std::vector` should be used more like a list, whereas `ManagedArray` should be like a numpy array. Since `l` values are arbitrary and not necessarily sequential, a `std::vector<ManagedArray>` makes sense. The expectation is that Python users acquiring `ManagedArray` variables expect that those variables should be zero-copy but remain unchanged if a recompute is triggered. Since it looks like you're replacing all the `ManagedArray` data in the vector on recompute, I think you'll be fine. I'll double-check this on my final pass though since I was a bit cursory with checking that this time around.

cpp/order/Steinhardt.cc
b-butler: Since `qlm_local` is indexed by the current `k` I would have to have two index variables still, but I can do this.

b-butler:
@vyasr most of your comments on loops I disagree with shifting the loop bounds would simplify the code, since the different arrays are indexed differently. In all cases at least two four loop variables would need to be defined and incremented. If you prefer multiple indices to be created in a for loop, then I will make the change.

cpp/order/Steinhardt.cc
b-butler: I could be wrong, but since the `PointSPHEvaluator` precomputes quantities based on theta and phi we wouldn't want to capture by reference since this would lead to race conditions when `compute` is called with the current thread's theta and phi.

Edit: add permalink https://github.com/glotzerlab/fsph/blob/11d4e640f14d25e74212d9bd1ca784e39206ba7c/src/spherical_harmonics.hpp#L181-L195

cpp/order/SolidLiquid.cc
```diff
@@ -29,7 +29,9 @@ void SolidLiquid::compute(const freud::locality::NeighborList* nlist,
 
     // Compute Steinhardt using neighbor list (also gets ql for normalization)
     m_steinhardt.compute(&m_nlist, points, qargs);
-    const auto& qlm = m_steinhardt.getQlm();
+    const auto& qlm = m_steinhardt.getQlm()[0];
+    // Even though this is a 2-D array since Steinhardt can compute multiple l the second dimension is always
+    // 1 since we only allow one l for SolidLiquid.
```
bdice: Put the comment above the line to which it applies. Also this comment can be much simpler.
```suggestion
    // SolidLiquid only has one l value so we index the 2D array from Steinhardt.
    const auto& qlm = m_steinhardt.getQlm()[0];
```

cpp/order/Steinhardt.cc
```diff
@@ -4,55 +4,62 @@
...
 #include "utils.h"
+#include <vector>
 
 /*! \file Steinhardt.cc
     \brief Computes variants of Steinhardt order parameters.
 */
 
 namespace freud { namespace order {
 
+
```
bdice:
```suggestion
```

cpp/order/Steinhardt.cc
bdice: I think that might not work here -- we also need to index by `nb2_index + k`.

bdice (commented):
A couple more comments.

cpp/order/Steinhardt.cc
vyasr: You're right, I'm resolving this suggestion.

cpp/order/Steinhardt.cc
vyasr: You're right, ignore this suggestion.

cpp/order/Steinhardt.cc
vyasr: Punting

vyasr:
> @vyasr most of your comments on loops I disagree with shifting the loop bounds would simplify the code, since the different arrays are indexed differently. In all cases at least two four loop variables would need to be defined and incremented. If you prefer multiple indices to be created in a for loop, then I will make the change.

Caught up with @b-butler briefly offline. I agree that in many cases you still need to maintain multiple indices, so just shifting the bounds may not be so helpful. I am a little hesitant to rely on the compiler to optimize out the multiple computations of the loop index, so I have a slight preference for computing it once and storing it to make the compiler's job easier, but aside from that I don't think any changes are necessary on the indexing front. I'll leave it to @b-butler to make the call there, then please re-request my review and we can wrap this up.

bdice:
@b-butler Let me know if you want help on anything here, it'd be helpful for my research code to get this merged. 👍

vyasr (approved):
Looks good, we can make further changes in later iterations if we need to.

PR #775, opened 2021-05-25
Added tests and documentation for cases zero neighbor cases.

### LINKED ISSUE #678: Check how order parameters treat particles with zero neighbors
In #672 and #624, we discussed how the Steinhardt order parameter treats particles with zero neighbors. Steinhardt assigns a value of `NaN`. We need to check to see how other methods (Hexatic, Translational, SolidLiquid, LocalDescriptors) handle cases with zero neighbors.

> The follow-up PR would basically be to review freud for other methods where this problem could arise and copy over the test to make sure the same behavior is consistent, as well as updating docs with the note and snippet added here.

_Originally posted by @vyasr in https://github.com/glotzerlab/freud/pull/672#issuecomment-718754948_

klywang:
We checked the behavior of Hexatic, Translational, SolidLiquid, and LocalDescriptors handle cases with zero neighbors.

For particles without neighbors:
- Steinhardt: Returns `nan` for `wl=True` and `wl=False`
- Hexatic: Returns `nan`
- Translational: Returns 0
- SolidLiquid: `particle_harmonics` returns `nan`
- LocalDescriptors: `num_sphs` returns 0, `sph=[]`

## Description
Added tests for edge cases and updated docs where `nan` is expected.

TODO: Had some issue building the docs. Most of the notes did not show up.

## Motivation and Context

Resolves: #678 

## How Has This Been Tested?

Added unit tests.

## Types of changes

## Checklist:


cpp/order/HexaticTranslational.cc
```diff
@@ -37,7 +37,14 @@ void HexaticTranslational<T>::computeGeneral(Func func, const freud::locality::N
             }
             if (normalize_by_k)
             {
-                m_psi_array[i] /= std::complex<float>(m_k);
+                if (total_weight == 0.)
+                {
+                    m_psi_array[i] /= std::complex<float>(total_weight);
```
klywang: Force this to be `nan` by dividing by zero. Is there a better way to do this? I saw `std::nanf` but I wasn't sure what to pick as the argument.

vyasr: I haven't reviewed this PR thoroughly yet, but I'm not sure this is generating the output you want. `std::complex<float>(1)/std::complex<float>(0)` will return (inf, nan), not just nan ([here's an example](https://godbolt.org/z/dr9K3nYa8))

vyasr:
@klywang any updates on this?

cpp/order/HexaticTranslational.cc
klywang: So tried to set `m_psy_array` equal to just nan like this:

```python
m_psi_array[i] = std::complex<float>(std::nanf)
```

and then like this

```python
m_psi_array[i] = std::complex<float>(std::nanf)
```

All tests pass on my laptop, but fail on CI, so I reverted back to a divide by zero. I double checked the values from the no neighbor test for the Translational order parameter, and in python, it registers as nan and not inf.

Issue #750, opened 2021-04-20
Diffraction pattern depends on shape of simulation box


tcmoore3:
**Describe the bug**
When calculating a diffraction pattern, I expect the symmetry of the diffraction to reflect that of the particle positions. However, if the box has edges of different lengths, then this is not the case. E.g., if I make a square lattice with different numbers of unit cells in each direction, then I get a diffraction pattern that does not have a 4-fold symmetric rotation axis. See [this notebook](https://github.com/glotzerlab/freud/files/6344229/example.zip) for an example.

**System configuration (please complete the following information):**
 - I get this behavior with versions 2.4 and 2.5

bdice:
Thanks for filing this issue — @tcmoore3 I know you’ve been working on this, have you been able to make any progress on fixing the problem?

tcmoore3:
@bdice I've figured out the origin of this bug in the code, which comes from how the k-vectors are generated, relevant lines [here](https://github.com/glotzerlab/freud/blob/ff9d561c2b8d719eea8e7167f083da3fbcc296a6/freud/diffraction.pyx#L257) and [here](https://github.com/glotzerlab/freud/blob/ff9d561c2b8d719eea8e7167f083da3fbcc296a6/freud/diffraction.pyx#L300). I need help figuring out how to proceed from here, since fixing this requires decisions about what kinds of assumptions/approximations you have to make to do this calculation.

joaander:
When I last looked at it with @tcmoore3 I was looking at the coordinate transforms used to map the system from the box to a unit cube on which the FFT is performed. The k-vectors should be spaced at 1/width (or 2pi/width) in this space. There should be an inverse transformation to map the k-vectors back to simulation units that is a function of the shear and scale matrices.

PR #777, opened 2021-05-26
Fix DiffractionPattern k values/vectors

### LINKED ISSUE #723: Document/fix units for k vectors in diffraction
These units are not currently documented, but they should be. Note, the units are not too important for visualization, as a large fraction of users use the diffraction to detect the presence of order within systems. However, if you want to do quantitative analyses with the diffraction pattern, the magnitudes of the k-vectors is quite important.

bdice:
## Description
This PR rescales k values and k vectors so that they correspond to the usual conventions in condensed matter physics.

This corrects the calculation of k values/vectors such that the peaks in the image should now correspond to G vectors that satisfy the following relationship for any lattice vector R ([from Wikipedia](https://en.wikipedia.org/wiki/Reciprocal_lattice)):

image

However, this has only been tested in a limited case (simple cubic system in a cubic box) and needs further testing. I am concerned that the current calculation using `np.max(system.box.to_matrix())` is incorrect for boxes like `{"Lx": 10, "Ly": 10, "Lz": 20}` when viewing down the z-axis. It should probably be based on the plane used to project the points (the box face most aligned with the view axis).

I also changed the behavior of `zoom`, please see screenshots below for a description.

The notebooks used to test this behavior (and develop the correct scaling factors for NumPy's FFT frequencies) are attached here.
[fft_tests.zip](https://github.com/glotzerlab/freud/files/6549580/fft_tests.zip)

To-do:

- [ ] Add tests for different grid size, output size, zoom values with a cubic system.
- [ ] Add documentation about the conventions of k vectors (formulas).
- [ ] Investigate behavior for boxes that are short or long in z dimension.
- [ ] Investigate behavior for boxes that aren't orthorhombic.
- [ ] Investigate behavior for boxes that are 2D.

Once this is resolved, then we can investigate the problems in issue #750.

## Motivation and Context
Resolves: #723

## How Has This Been Tested?
Tests are needed. See to-dos above.

## Screenshots

Previously, the constructor argument `grid_size` was divided by the compute argument `zoom`, and the resulting value was used as the "true" grid resolution of the histogram image and thus the FFT of that histogram. I experimented with changing this so that the `zoom` parameter does not affect the histogram resolution. Here is a screenshot of how that affects the behavior:

### Truncated resolution, old behavior: `grid_size/zoom`

zoom = 4, grid_size = 512, output_size = 512
old_small

zoom = 2, grid_size = 512, output_size = 512
old_large

### Full resolution, new behavior: `grid_size`

zoom = 4, grid_size = 512, output_size = 512
new_small

zoom = 2, grid_size = 512, output_size = 512
new_large

The difference is that the truncated resolution diffraction patterns appear to decay more quickly. I'm unsure about this, so I changed the code's behavior to use the full `grid_size` resolution.

## Types of changes

## Checklist:


alacour (commented):
The changes to code look straightforward, although (visually) the intensities of the k-vectors only appear to be decaying slightly slower. Are you planning to examine noncubic boxes within this pull request?

bdice:
> 
> 
> The changes to code look straightforward, although (visually) the intensities of the k-vectors only appear to be decaying slightly slower. Are you planning to examine noncubic boxes within this pull request?

I haven't decided on a final scope for this pull request. I will discuss with @AKerr9509 and we will make a plan. Separating the to-do list at the top of this PR description into a series of smaller changes may be best. I propose the following:
- In this first PR: Add very basic tests for the cubic system and then merge.
- Second PR: Add tests for different grid size, output size, zoom values with a cubic system.
- Third PR: Add documentation about the conventions of k vectors (formulas).
- Fourth PR:
  - Investigate behavior for boxes that are short or long in z dimension.
  - Investigate behavior for boxes that aren't orthorhombic.
  - Investigate behavior for boxes that are 2D.

bdice:
In a future PR, we're also going to change the normalization criteria such that `S(k=0) = N`. Currently it is `S(k=0) = 1`. The typical formulation in physics gives a result of `N` but we changed it to `1` so it's easier to plot. However, we can just adapt the plotting code to re-scale as needed, and keep the normal physics convention.

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +160,79 @@ def test_k_values_and_k_vectors(self):
...
+                    view_orientation, [first_k_bin, first_k_bin, 0]
+                )
+                last_k_vector = rowan.rotate(
+                    view_orientation, [last_k_bin, last_k_bin, 0]
+                )
+
+                npt.assert_allclose(dp.k_vectors[0, 0], first_k_vector)
+                npt.assert_allclose(dp.k_vectors[-1, -1], last_k_vector)
+
+            if output_size % 2 == 1:
```
bdice: Try to unify these code paths. Most of the code is the same for both, so try to only have an "if/else" for the portion that is different. Avoid using two "ifs" because the cases should be exclusive and if/else makes that clear.

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +160,79 @@ def test_k_values_and_k_vectors(self):
...
+
+                first_k_vector = rowan.rotate(
+                    view_orientation, [first_k_bin, first_k_bin, 0]
+                )
+                last_k_vector = rowan.rotate(
+                    view_orientation, [last_k_bin, last_k_bin, 0]
+                )
+
+                npt.assert_allclose(dp.k_vectors[0, 0], first_k_vector)
+                npt.assert_allclose(dp.k_vectors[-1, -1], last_k_vector)
```
bdice: @AKerr9509 You had a sentence in a commit message asking about whether to check all four corners. That would be a good idea! You could even check the left-center, top-center, right-center, and bottom-center values for a total of 9 values (including the center). I think those should be aligned with vectors like `[first_k_value, 0, 0]`, `[last_k_value, 0, 0]`, `[0, first_k_value, 0]`, and `[0, last_k_value, 0]`.

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +160,79 @@ def test_k_values_and_k_vectors(self):
...
+            # np.fft.n
+            # fftfreq and the _k_scale_factor to write an expression for
+            # the first and last bins' k-values and k-vectors.
+
+            scale_factor = 2 * np.pi * output_size / (np.max(box.to_matrix()) * zoom)
+
+            if output_size % 2 == 0:
+
+                first_k_bin = -0.5 * scale_factor
+                last_k_bin = (0.5 - (1 / output_size)) * scale_factor
```
bdice: Maybe these variable names would be better?
```suggestion
                first_k_value = -0.5 * scale_factor
                last_k_value = (0.5 - (1 / output_size)) * scale_factor
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +160,79 @@ def test_k_values_and_k_vectors(self):
...
+
+            output_size = dp.output_size
+            npt.assert_allclose(dp.k_values[output_size // 2], 0)
+            center_index = (output_size // 2, output_size // 2)
+            npt.assert_allclose(dp.k_vectors[center_index], [0, 0, 0])
+
+            # Tests for the left and right k_value bins. Uses the math for
+            # np.fft.n
+            # fftfreq and the _k_scale_factor to write an expression for
+            # the first and last bins' k-values and k-vectors.
```
bdice: Clean up comment.
```suggestion
            # Tests for the left and right k_value bins. Uses the math for
            # np.fft.fftfreq and the _k_scale_factor to write an expression for
            # the first and last bins' k-values and k-vectors.
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +161,147 @@ def test_k_values_and_k_vectors(self):
...
+
+            if output_size % 2 == 0:
+
+                first_k_value = -0.5 * scale_factor
+                last_k_value = (0.5 - (1 / output_size)) * scale_factor
+
+                npt.assert_allclose(dp.k_values[0], first_k_value)
+                npt.assert_allclose(dp.k_values[-1], last_k_value)
+
+                first_k_vector = rowan.rotate(
```
bdice: For consistency, please rename:
```suggestion
                top_left_k_vector = rowan.rotate(
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +161,147 @@ def test_k_values_and_k_vectors(self):
...
+                    view_orientation, [first_k_value, first_k_value, 0]
+                )
+                last_k_vector = rowan.rotate(
```
bdice: For consistency, please rename:
```suggestion
                bottom_right_k_vector = rowan.rotate(
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +161,147 @@ def test_k_values_and_k_vectors(self):
...
+                bottom_left_k_vector = rowan.rotate(
+                    view_orientation, [last_k_value, first_k_value, 0]
+                )
+
+                npt.assert_allclose(dp.k_vectors[0, 0], first_k_vector)
+                npt.assert_allclose(dp.k_vectors[-1, -1], last_k_vector)
+                npt.assert_allclose(dp.k_vectors[0, -1], top_right_k_vector)
+                npt.assert_allclose(dp.k_vectors[-1, 0], bottom_left_k_vector)
+
+                centre = output_size // 2
```
bdice: We prefer American English spellings. 😄 This suggestion also applies several other places.
```suggestion
                center = output_size // 2
```

freud/diffraction.pyx
```diff
@@ -46,6 +46,15 @@ cdef class DiffractionPattern(_Compute):
...
     can be accessed as a square array of shape ``(output_size, output_size)``.
 
+    The :math:`\vec{k}=0` peak is always located at index
+    ``(output_size // 2, output_size // 2)`` and is normalized to have a value
+    of :math:`S(\vec{k}=0) = 1` (not :math:`N`, a common convention). The
+    remaining :math:`\vec{k}`-vectors are computed such that each peak in the
+    diffraction pattern satisfies the relationship :math:`\vec{k} \cdot
+    \vec{R} = 2\piN` for some integer :math: `N` and lattice vector of
+    the system :math: `R` (see
+    `this article <https://en.wikipedia.org/wiki/Reciprocal_lattice>`).
```
bdice: Needs some syntax fixes. You can see the live docs rendered here: https://freud--777.org.readthedocs.build/en/777/modules/diffraction.html#freud.diffraction.DiffractionPattern
```suggestion
    \vec{R} = 2 \pi N` for some integer :math:`N` and lattice vector of
    the system :math:`R`. See the
    `reciprocal lattice Wikipedia page <https://en.wikipedia.org/wiki/Reciprocal_lattice>`__
    for more information.
```
You may also need to add `# noqa: E501` to the end of this docstring since that line with the link is fairly long. Search "E501" to see examples in the rest of the code.

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +161,176 @@ def test_k_values_and_k_vectors(self):
...
+                    view_orientation, [first_k_value, 0, 0]
+                )
+                right_center_k_vector = rowan.rotate(
+                    view_orientation, [last_k_value, 0, 0]
+                )
+
+                npt.assert_allclose(dp.k_vectors[center, 0], top_center_k_vector)
+                npt.assert_allclose(dp.k_vectors[center, -1], bottom_center_k_vector)
+                npt.assert_allclose(dp.k_vectors[0, center], left_center_k_vector)
+                npt.assert_allclose(dp.k_vectors[-1, center], right_center_k_vector)
```
bdice: This whole block doesn't need to be copied twice in the if/else. Try this structure:
```python
if output_size % 2 == 0:
    first_k_value = -0.5 * scale_factor
    last_k_value = (0.5 - (1 / output_size)) * scale_factor
else:
    first_k_value = (-0.5 + 1 / (2 * output_size)) * scale_factor
    last_k_value = (0.5 - 1 / (2 * output_size)) * scale_factor

npt.assert_allclose(dp.k_values[0], first_k_value)
npt.assert_allclose(dp.k_values[-1], last_k_value)

# ... continue with all the rest of the common code
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +161,176 @@ def test_k_values_and_k_vectors(self):
...
+
+                    # Locate brightest areas of diffraction pattern (intensity > threshold),
+                    # and check that the ideal diffraction peak locations, given by
+                    # k * R = 2*pi*N for some lattice vector R and integer N, are contained
+                    # within these regions. This test only checks N in range [-2, 2].
+                    threshold = 0.2
+                    xs, ys = np.nonzero(dp.diffraction > threshold)
+                    xy = np.dstack((xs, ys))[0]
+
+                    ideal_peaks = {-2: "f", -1: "f", 0: "f", 1: "f", 2: "f"}
```
bdice:
```suggestion
                    ideal_peaks = {i: False for i in [-2, -1, 0, 1, 2]}
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +161,176 @@ def test_k_values_and_k_vectors(self):
...
+                    all_peaks = True
+
+                    for peak in ideal_peaks:
+                        for x, y in xy:
+                            k_vector = dp.k_vectors[x, y]
+                            lattice_vector = [1, 1, 1]
+                            dot_prod = np.dot(k_vector, lattice_vector)
+
+                            if dot_prod == (peak * 2 * np.pi):
+                                ideal_peaks[peak] = "t"
```
bdice:
```suggestion
                                ideal_peaks[peak] = True
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -160,11 +161,176 @@ def test_k_values_and_k_vectors(self):
...
+
+                    for peak in ideal_peaks:
+                        if ideal_peaks[peak] == "f":
+                            all_peaks = False
+
+                    assert all_peaks == True
```
bdice: You can just make sure that all the peaks are found, and remove the `all_peaks` variable.
```suggestion
                    assert all(ideal_peaks.values())
```

bdice (commented):
A few comments for improvement. @AKerr9509 it seems like you're on the right track. Some tests are failing, take a look at the output from CI and you can see where your code isn't working (some variables aren't defined because you changed some of the names but not all). Ping me on Slack if get stuck or want help with anything.

bdice:
@AKerr9509 Good work on this! I'm going to merge since this fixes bugs and is at a good stopping point. We'll start a new pull request for follow-up work.

PR #798, opened 2021-06-17
Fix Steinhardt averaging to use one neighbor shell instead of two.

### LINKED ISSUE #795: Steinhardt average value are incorrectly computed.
Analysis by @arvdee identified that freud's Steinhardt order parameters are not being computed correctly for `average=True`.

This issue is primarily meant to document the existence of the problem. This thread on the user forum includes extensive details: https://groups.google.com/g/freud-users/c/zpR9UEkDhow/m/KGoixKrrBAAJ

I have created a fix and plan to apply it once #794 and #773 are merged.

The cause is that the `computeAve` function is looping over two neighbor shells instead of one.

I have a repository here that I plan to adapt for freud's validation tests: https://github.com/bdice/freud_boops_accuracy

@arvdee Can you and your colleague who provided the "GC" data provide us with permission to include the `Test_Configuration` files and the corresponding Steinhardt/MSM reference data in freud's source code for use as validation tests?

This bug affects all versions of freud 2.x up to now.

bdice:
## Description
This PR fixes an error in the computation of the Steinhardt order parameter when `average=True`. This PR replaces #796.

I added validation tests using data from @Arvdee and collaborators, with permission. I also improved and modernized the existing validation tests to use pytest.

## Motivation and Context
Resolves #795.

## How Has This Been Tested?
I created a repository of validation tests at https://github.com/bdice/freud_boops_accuracy and stripped down the contents for the validation tests in this PR.

## Types of changes

## Checklist:


tests/validation/test_Minkowski_Structure_Metrics.py
```diff
@@ -19,63 +20,53 @@ def _get_structure_data(structure, qtype):
...
-                # Test Average ql
-                comp = freud.order.Steinhardt(sph_l, average=True, weighted=True)
-                comp.compute((box, positions), neighbors=voro.nlist)
-                npt.assert_allclose(
-                    comp.order, expected_avql[:, sph_l], rtol=5e-5, atol=1e-3
-                )
+            # Test average q'l
+            comp = freud.order.Steinhardt(sph_l, average=True, weighted=True)
+            comp.compute(snap, neighbors=voro.nlist)
+            npt.assert_allclose(comp.order, expected_avql[:, sph_l], atol=1e-5)
```
bdice, line 56: The tolerances on these computations were far too wide before (an absolute tolerance of 1e-3 is really big). I wonder if the high tolerances were hiding the fact that the averaged quantities were incorrectly calculated. 🤔 These tests act on "perfect" structures so it would be easy to hide the issue that was averaging over two neighbor shells instead of one, and mask it with a higher numerical tolerance. Regardless, the measured tolerances are much lower now. I keep a minimum tolerance of 1e-5 even though it could be a little smaller, because some of these tests are known to be flaky due to numerical differences on exotic systems/compilers/etc.

tests/validation/test_Steinhardt_Average.py
```diff
@@ -0,0 +1,177 @@
...
+
+def _compute_msms(system, lmax, average=False, wl=False):
+    """Returns Minkowski Structure Metrics up to a maximum l value."""
+    voro = freud.locality.Voronoi().compute(system=system)
+    return _compute_steinhardts(
+        system,
+        [
+            dict(average=average, weighted=True, l=i, wl=wl, wl_normalize=wl)
+            for i in range(lmax + 1)
+        ],
```
b-butler: Since the parameters besides `l` are not changing here couldn't all the structure metrics be computed at once using `list(range(lmax + 1))`?

tests/validation/test_Steinhardt_Average.py
```diff
@@ -0,0 +1,177 @@
...
+            # If most of the particles agree, this test is probably correct.
+            # Close investigation showed that numerical issues (truncated
+            # particle positions in the dat file?) and bonds with facet weights
+            # of the order of 1e-9 caused some deviations in the neighbor
+            # counts, which in turn affected the averaged ql values. This
+            # affected 37 out of 3288 particles in the sample data.
+            close_values = np.isclose(
+                msm_data[:, sph_l], freud_data[:, sph_l], atol=1e-5
+            )
+            assert sum(close_values) > (0.985 * len(freud_data)), f"{name} failed"
```
b-butler:
```suggestion
            assert np.sum(close_values) > (0.985 * len(freud_data)), f"{name} failed"
```
should be a little faster.

tests/validation/test_Steinhardt_Average.py
```diff
@@ -0,0 +1,177 @@
...
+        names = ["ave. q'4", "ave. q'6", "ave. w'4", "ave. w'6"]
+        for i, name in enumerate(names):
+            # If most of the particles agree, this test is probably correct.
+            # Close investigation showed that numerical issues (truncated
+            # particle positions?) and bonds with facet weights of the order of
+            # 1e-9 caused some deviations in the neighbor counts, which in turn
+            # affected the averaged ql values. This affected 37 out of 3288
+            # particles in the sample data for one of the cases checked.
+            close_values = np.isclose(rvd_data[:, i], freud_data[:, i], atol=5e-5)
+            assert sum(close_values) > (0.985 * len(freud_data)), f"{name} failed"
```
b-butler:
```suggestion
            assert np.sum(close_values) > (0.985 * len(freud_data)), f"{name} failed"
```

b-butler (changes requested):
Probably not worth the effort, but it would cut down a lot on the file number to have an HDF5 file store all the numerical data (and it would be more efficient storage wise) and could keep descriptive information about the data still.

I don't feel strongly enough about any of my comments to require them, so just resolve them and re-request my review if you don't want to implement the changes.

tests/validation/test_Steinhardt_Average.py
bdice: As of #773, yes. But that feature hadn't been merged when I started work on this. I updated it in 9dadc6de05e8e3a7dea0be0c386d3ef6785d7241.

bdice:
> Probably not worth the effort, but it would cut down a lot on the file number to have an HDF5 file store all the numerical data (and it would be more efficient storage wise) and could keep descriptive information about the data still.

I considered something like this and decided against it. If we were to include a binary format, I would want to use NumPy's compressed `.npz` to avoid the dependency on HDF5. However, I thought it would be less transparent to only commit a binary file here. Sticking with the text-based output format of the reference code makes it easier to cross-check in the future, if needed, and it matches the existing formats we used for other validation data. If we only included binary data, we would want to include the scripts used to generate the reference binary data from the text-based output, and that seemed like needless effort. I did shrink the reference GSD file by dropping some unnecessary log/state fields, and the total file size of the new validation data is 1.4M which seemed small enough. The freud repository is already enormous due to the past history of committing Cython's C++ outputs, so we're not changing the repository size very significantly.

I don't feel very strongly about this and am open to others' thoughts if someone thinks this should be changed.

b-butler (approved):
These are good tests to have. Thanks for writing them.

PR #802, opened 2021-06-24
Fix doctests to work with pytest.

bdice:
## Description
Examples in docstrings were not being tested since the switch to pytest.

This is because the `load_tests` function is not recognized by pytest, only unittest.

I attempted to use `pytest --doctest-modules` but that isn't quite what we want, and it appears to only work with pure Python (not Cython) modules.

Instead, I adopted a variation on the strategy proposed in this StackOverflow post:
https://stackoverflow.com/questions/16982514/python-test-discovery-with-doctests-coverage-and-parallelism

## Motivation and Context
Ensures that docstrings are runnable. I noticed the issue while reviewing #793.

## How Has This Been Tested?
Tests should pass and show output like `test_docstrings[freud.box.Box.center]`.

## Types of changes

## Checklist:


vyasr:
The current error looks to be related to the upload of coverage artifacts I think. FYI, the Windows test is a false positive, the tests actually failed on import but it's reporting a success on this dashboard anyway.

For some reason all doctests are doubled up, with a copy in some `*.__test__.*` namespaces. I suspect it's related to https://github.com/pytest-dev/pytest/issues/1558, but I've never used this feature of pytest to skip tests before so I'm not sure. It looks like all doctests are somehow being added twice.

bdice:
@vyasr I couldn’t figure out how the `__test__` module attribute is created. The previous tests were also doubled for the same reason. I think it might be a Cython thing. The docstring tests don’t point to line numbers but the `__test__` tests do have pyx line info, so it is probably being generated at Cython compilation time.

I’ll look at the failing tests later, I have a lot going on right now. Anyone can feel free to take a look or push to this branch.

bdice:
I got the doctests to run successfully, but I am seeing an unrelated error in CI from either `coverage` or `pytest-cov`. I commented on this related issue to document the problem: https://github.com/nedbat/coveragepy/issues/1160

@glotzerlab/freud-maintainers Since I'm not sure if our code coverage is being reported correctly anyway, I think the next step is to disable coverage/codecov before merging this PR. Thoughts on this? See also: codecov deprecation: https://github.com/glotzerlab/signac/issues/630

bdice:
The issue upstream in `coverage` has been fixed:
- https://github.com/nedbat/coveragepy/issues/1160#issuecomment-946304541
- https://github.com/nedbat/coveragepy/commit/afe6cf34d022e8dbbaa47826c487a98ca6832721

This PR can be merged once a new version of coverage is released and we upgrade the pinning. I don't really want to disable coverage in our CI (even though it's not currently working), but that would be another option.

bdice:
@tommy-waltmann I went ahead and merged this since it affects tests and had been stalled. We want any new PRs to have their documentation examples tested. This was blocked until coverage 6.1 was released, which fixed the failing tests.

PR #815, opened 2021-07-29
Fix box docs

bdice:
## Description
@ramanishsingh caught an error in the box class documentation: masses are an array of shape `(N,)`, not `(N, 3)`.

I also added return types to a few methods that were missing them.

This PR also fixes a minor nitpick that has always confused me. Raw strings can be written with `R"` or `r"`, but I am used to seeing the lowercase form everywhere except in freud. For clarity (and to use the more common format), I flipped them to use the lowercase form.

## Motivation and Context
Fixes docs. No changelog line needed.

## How Has This Been Tested?
ReadTheDocs should render correctly.

## Types of changes

## Checklist:

### DIFF
--- a/doc/source/topics/optimizing.rst
+++ b/doc/source/topics/optimizing.rst
@@ -65,4 +65,4 @@ Minor speedups may also be gained from passing properly structured data to **fre
 The package was originally designed for analyzing particle simulation trajectories, which are typically stored in single-precision binary formats.
 As a result, the **freud** library also operates in single precision and therefore converts all inputs to single-precision.
 However, NumPy will typically work in double precision by default, so depending on how data is streamed to **freud**, the package may be performing numerous data copies in order to ensure that all its data is in single-precision.
-To avoid this problem, make sure to specify the appropriate data types (`numpy.float32 <https://docs.scipy.org/doc/numpy/user/basics.types.html>`_) when constructing your NumPy arrays.
+To avoid this problem, make sure to specify the appropriate data types (:attr:`numpy.float32`) when constructing your NumPy arrays.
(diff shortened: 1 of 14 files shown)


j-proc (commented):
Looks fine to me, the (N,3) shape is obviously nonsensical and a missed clean up from copy/paste for the documentation structure.

@bdice as for the r/R issue. It's my preference as well, just because the way it reads is more clearly a formatting syntax and not the start of a line you should actually read. However, I think if you're going to make this change, we need to do something to make its consistency enforced. I'm not sure if that means making it explicit in some documentation or if it could be added to pre-commit. You're more of a pre-commit expert to know if it's possible.

bdice:
> @bdice as for the r/R issue. It's my preference as well, just because the way it reads is more clearly a formatting syntax and not the start of a line you should actually read. However, I think if you're going to make this change, we need to do something to make its consistency enforced. I'm not sure if that means making it explicit in some documentation or if it could be added to pre-commit. You're more of a pre-commit expert to know if it's possible.

I'm not familiar with a way to enforce this, unfortunately. Most of the code in freud is written by copy-paste, so I suspect this won't require too much ongoing maintenance. There is one test with a random failure -- we can ignore it for now, this happens on rare occasions. We could increase the tolerance for that test at some point but it doesn't fit in this PR's scope. @j-proc Please approve once you're ready. 👍

itsagoodbrain:
> I'm not familiar with a way to enforce this, unfortunately. Most of the code in freud is written by copy-paste, so I suspect this won't require too much ongoing maintenance.

That's a good point. This is somewhat unrelated to the main fix which is needed. Approving

j-proc (approved):
That's a good point. It's unrelated to the main issue anyways. Approving.

PR #820, opened 2021-08-25
Static structure factor "direct" algorithm

bdice:
## Description
This PR adds static structure factor calculations using the same methods implemented in the dynasor library (https://dynasor.materialsmodeling.org/), re-implemented in TBB-parallelized C++ for efficiency and to match the rest of freud's code base.

We will give credit to dynasor and cite its paper.

I'm copying the current to-do list discussed with @DomFijan into this PR.
- [x] Convert freud-like class into an actual freud _Compute class
- [x] Docs
- [x] Tests

As for dynasor / dependency management:
- [x] First pass: Leave hard dependency on dynasor
- [x] Second pass: Copy C code into a C++ module in freud (eliminate calc_rho_k)
- [x] Third pass: Copy [reciprocal_isotropic](https://gitlab.com/materials-modeling/dynasor/-/blob/master/dsf/reciprocal.py#L131)  into freud (eliminate dynasor dependency)
  - Re-implement k-space sampler to sample chunks of k-space, to permit larger k-values without allocating huge grids of k-points in memory
- [x] Fourth pass: Move as much logic as possible into C++

## Motivation and Context
Implements "gold standard" method (not based on Debye formula like #660).

## How Has This Been Tested?
Tests are in progress.

## Types of changes

## Checklist:


bdice:
@DomFijan @andkerr @tcmoore3 I pushed some additional work to this branch over the weekend. Just wanted to update you all on the progress. I think this PR is in a nearly-complete state.

I have implemented the k-point sampler in C++ and optimized a number of the core methods. The code is quite fast. I haven't done head-to-head comparisons with dynasor but I know that the algorithms in this PR are substantially improved from the original. Every step of the process is parallelized, including the k-point sampler (which is expensive). The k-point sampler can prune k-points that are outside the bounds of (k_min, k_max) and does not pre-allocate a full array of k-points, meaning that it can sample from a grid of k-points that is larger than system memory so long as the small fraction of points being sampled fits into memory. This permits calculations at very large k values, which may not be very useful in practice but is essential for testing the limiting behavior of S(k) as k goes to infinity (see tests for several examples of limiting behavior).

Work that I know still needs to be done:
- [x] Add support for 2D boxes (or just error out if we don't want to support them yet)
- [ ] Revise freud-examples notebooks, merge, and update git submodule in the freud repo. https://github.com/glotzerlab/freud-examples/pull/24

bdice:
@DomFijan I noticed there are some issues with installing the dynasor package. The package that was uploaded to PyPI may be invalid. We should report that upstream: https://gitlab.com/materials-modeling/dynasor

If you need to install a package directly from a git repo, please pin it to a tag as described in [this StackOverflow answer](https://stackoverflow.com/a/13754517/). We pin all our testing dependencies to isolate issues caused by upstream packages used for testing from actual issues in new code.

freud/diffraction.pyx
```diff
@@ -617,3 +884,31 @@ cdef class DiffractionPattern(_Compute):
             return freud.plot._ax_to_bytes(self.plot())
         except (AttributeError, ImportError):
             return None
+
+def reciprocal_isotropic(box, k_max, k_min=0, max_k_points=10000):
```
bdice: I created this method just so that I could test and optimize this function directly from Python. @DomFijan Do you think there is any need to expose this in the public API? If not, I'd like to remove it from the Python API and just leave it in C++.

DomFijan: I agree. No need to expose it in python API. I would advocate to move some of the documentation information to the StaticStructurefactorDirect class such as the similarity with dynasor implementation and at what max_k_points they start to give same results.

cpp/diffraction/StaticStructureFactorDirect.cc
```diff
@@ -0,0 +1,298 @@
...
+{
+    // Compute k vectors by sampling reciprocal space
+    const auto& box = neighbor_query->getBox();
+    if (box.is2D()) {
+        throw std::invalid_argument("2D boxes are not currently supported.");
+    }
+    const auto k_bin_edges = m_structure_factor.getBinEdges()[0];
+    const auto k_min = k_bin_edges.front();
+    const auto k_max = k_bin_edges.back();
+    if (!reuse_box || !m_k_grid_assigned) {
```
bdice: @DomFijan I'm not sure if the `reuse_box` flag is consistent with the rest of freud. An alternative would be to resample if (`reset=True` or if the current box object is not equal to the previous box object). That is, we would store `m_box` as a member variable for the last used box object (many other classes do this), and re-sample if `reset || (m_box != box)`.

DomFijan: Saving and comparing the box might be a good solution. It is also probably more elegant then current solution.

bdice: @DomFijan If you have time to implement that, go ahead. Otherwise I'll probably be able to look at it next weekend.

requirements/requirements-test.txt
```diff
@@ -6,6 +6,7 @@ cython==0.29.24
 garnett==0.7.1
 GitPython==3.1.18
 gsd==2.4.2
+https://gitlab.com/materials-modeling/dynasor/-/archive/1.1/dynasor-1.1.tar.gz
```
bdice: I have requested a new release of dynasor with a fixed MANIFEST. Once this PR is merged and the new release is made, we can replace this with a PyPI package name / version. https://gitlab.com/materials-modeling/dynasor/-/merge_requests/16

freud/diffraction.pyx
```diff
@@ -263,6 +265,277 @@ cdef class StaticStructureFactorDebye(_Compute):
...
+            raise ValueError(
+                "If query_points are provided, N_total must also be provided "
+                "in order to correctly compute the normalization of the "
+                "partial structure factor."
+            )
+        if reuse_box and reset:
+            raise ValueError(
+                "The reuse_box option has to be used with reset = False"
+            )
+        # Convert box to orthorhombic and wrap points inside this new box.
```
bdice: @DomFijan This orthorhombic conversion needs to be done in C++, or else we will end up having the Python and C++ APIs returning different results.

Can you provide further explanation on why this step is needed? It doesn't seem like this would be necessary if we're correctly computing the reciprocal space vectors that are being sampled isotropically. From my brief search in the dynasor docs, it does not appear that dynasor requires orthorhombic boxes. If I missed something, let me know!

freud/diffraction.pyx
```diff
@@ -263,6 +265,277 @@ cdef class StaticStructureFactorDebye(_Compute):
...
+    For particle types :math:`\alpha` and :math:`\beta`, we compute the total
+    scattering function as a sum of the partial scattering functions as:
+
+    .. math::
+
+        S(k) - 1 = \sum_{\alpha}\sum_{\beta} \frac{N_{\alpha}
+        N_{\beta}}{N_{total}^2} \left(S_{\alpha \beta}(k) - 1\right)
+
+    This class is based on the MIT licensed `Dynasor library
+    <https://gitlab.com/materials-modeling/dynasor/>`__.
```
bdice: We need to add a citation here. Citation info:
- https://dynasor.materialsmodeling.org/credits.html
- https://onlinelibrary.wiley.com/doi/10.1002/adts.202000240

freud/diffraction.pyx
bdice: Done. There was one change that I wanted to mention:

> and at what max_k_points they start to give same results.

The information that was in the docstring was incorrect. It said "This implementation approaches Dynasor implementation for number of max_k_points around 50000, but is also faster." I don't know where that came from. [By default, Dynasor uses 20,000 points.](https://gitlab.com/materials-modeling/dynasor/-/blob/1.1/dynasor#L146-147) The scaling should be exactly the same since we're using the same probability function for each k vector to ensure that reciprocal space is sampled isotropically. However, this also depends on the value of `k_max`, because for smaller values of `k_max` we can sample the entire grid in reciprocal space up to the target magnitude of k. That's met by [this condition for skipping the pruning step](https://github.com/glotzerlab/freud/blob/aba73322c379a232aed8ea9b69fb4ebb4b4d376d/cpp/diffraction/StaticStructureFactorDirect.cc#L186-L190).

cpp/diffraction/StaticStructureFactorDirect.h
```diff
@@ -0,0 +1,150 @@
...
+    using StructureFactorHistogram = util::Histogram<float>;
+    using KBinHistogram = util::Histogram<unsigned int>;
+
+public:
+    //! Constructor
+    StaticStructureFactorDirect(unsigned int bins, float k_max, float k_min = 0,
+                                unsigned int max_k_points = 10000);
+
+    //! Destructor
+    virtual ~StaticStructureFactorDirect() = default;
```
tommy-waltmann: I had the same comment on the last structure factor PR, this doesn't need to be virtual unless we plan on having classes that inherit from this one.
```suggestion

```

tests/test_diffraction_StaticStructureFactorDirect.py
```diff
@@ -0,0 +1,248 @@
+import matplotlib
```
tommy-waltmann, line 1: Most of the tests in this file are basically copies of tests which were added in #660 . To reduce code duplication, let's write a single version of the tests that take either a `StaticStructureFactorDebye` or `StaticStructureFactorDirect` as an argument and then perform the test.

freud/diffraction.pyx
```diff
@@ -263,6 +265,277 @@ cdef class StaticStructureFactorDebye(_Compute):
             return None
 
 
+cdef class StaticStructureFactorDirect(_Compute):
```
tommy-waltmann: Since this class exposes the same interface as `StaticStructureFactorDebye`, it may be worth having a base class (at least on the python side) so we only have to write the code for common properties once.

tests/test_diffraction_StaticStructureFactorDirect.py
```diff
@@ -0,0 +1,248 @@
...
+import numpy as np
+import numpy.testing as npt
```
tommy-waltmann, line 3: #660 was merged before I could say this, but our tests for that PR didn't cover some properties (like `bin_centers`, `bin_edges`, etc.) and I would like to see those tests added in this PR.

tommy-waltmann (changes requested):
Great work on this so far! I have some comments mostly related to testing and improving code reuse.

cpp/diffraction/StaticStructureFactorDirect.cc
DomFijan: I'll give this a go this week. I briefly check the == operator on the box class and it seems it only checks the lengths and tilts, but not PBCs. Do you think PBCs should also be checked in this case? @bdice

bdice: @DomFijan Yes. PBCs are not well supported in freud.

freud/diffraction.pyx
DomFijan: @bdice I will attempt to move it sometime this week. 

So I tried computing S(k) using this method for a system that had a non-orthorombic box and then converted and wrapped it to an orthorombic box and got slightly different results. This is most likely (although I'm not 100% sure) due to the fact that our k-space grid is "orthorombic" (or rather rectangular?), so the corners of a tilted box probably don't get treated properly. I haven't checked this but if this is correct then the boxes with higher tilt might have a bigger error stemming from this...
In any case the system configuration and structural properties should be exactly the same no matter what box shape is chosen? By converting to an orthorombic box we sort of stick to something that we know works correctly. At least that was my thought process behind this.

freud/diffraction.pyx
DomFijan: I was under the impression that the two implementations are different, and I went to test this at some point and was getting different results for lower numbers of max_k_points (or at least I thought I did this). I did a couple of tests on current build and it seems that this is no longer the case. Everything seems to be in line with what you have written...

freud/diffraction.pyx
bdice: > our k-space grid is "orthorombic" (or rather rectangular?)

I don't think that's true. The k-space grid is based on multiples of the reciprocal lattice vectors (b1, b2, b3) which should correctly handle triclinic boxes. The k vectors aren't necessarily aligned with the (x, y, z) axes if there are nonzero tilt factors. See [implementation here](https://github.com/glotzerlab/freud/blob/4dfec608d641323190f7bf7f32d69c5d57cb1837/cpp/diffraction/StaticStructureFactorDirect.cc#L207-L226).

As we ran into with the diffraction pattern calculation, wrapping the points into an orthorhombic box will change the distances between some pairs of points.

As for "slightly different results" -- are you confident that the orthorhombic wrapped results were "right" and the previous implementation was "wrong"? I don't think the statement "structural properties should be exactly the same no matter what box shape is chosen" is true unless you have proof otherwise. (edit: I mean in the case that you simulate in one box and wrap into another box shape -- I would agree that a large equilibrated system in a cubic shape and in a triclinic shape should give the same results up to finite size effects and any strain induced by the periodic box itself).

If we want a definitive test of this, we can re-run the dynasor sample code for liquid aluminum and use a box with tilt factors, then compare results. I think that dynasor supports arbitrary box matrices (which is even less restrictive than freud) and my previous implementation (without wrapping) should match dynasor's results.

tests/test_diffraction_StaticStructureFactorDirect.py
```diff
@@ -0,0 +1,248 @@
...
+        box, points = freud.data.make_random_system(L, N)
+        system = freud.AABBQuery.from_system((box, points))
+        A_points = system.points[: N // 3]
+        B_points = system.points[N // 3 :]
+        sf = freud.diffraction.StaticStructureFactorDirect(bins=100, k_max=10)
+        sf.compute((system.box, B_points), query_points=A_points, N_total=N)
+        S_AB = sf.S_k
+        sf.compute((system.box, A_points), query_points=B_points, N_total=N)
+        S_BA = sf.S_k
+        npt.assert_allclose(S_AB, S_BA, rtol=1e-6, atol=1e-6)
```
bdice: Tests are failing [here](https://app.circleci.com/pipelines/github/glotzerlab/freud/3165/workflows/5359a0d2-ac86-4dd3-80cf-b45dd7ee63b0/jobs/12387) with:
> Max absolute difference: 1.4603138e-06
> Max relative difference: 0.00017976

We should increase this tolerance.
```suggestion
        npt.assert_allclose(S_AB, S_BA, rtol=1e-5, atol=1e-5)
```

freud/diffraction.pyx
DomFijan: I totally forgot that k-space grid generator works in that way. You are right! Perhaps there shouldn't be an issue there.
Just to be sure I'll remove the conversion to orthorombic and try to test this again and come back with some concrete plots and numbers and test against dynasor as well. Perhaps I made a mistake there as well.

I cannot see though how choice of the box can influence the configuration (structure) of a snapshot of a system if one has periodic boundaries. Consider a snapshot of an infinite system of particles periodic in all directions with a given periodicity. Now attempt to tesselate the space with a shape that obeys the periodicity. Any shape that obeys the periodicity will produce the same system configuration. See the image for an example. The key here is that Lx and Ly parameters obey the periodicity in both box cases.
boxes
 
When it comes to simulations, trajectories in different box types could produce different results. But this is a consequence of the fact that system might be relaxing more slowly (or might not be able to relax at all) to equilibrium due to strain as you suggested.

bdice: @DomFijan Distances are not preserved when shearing, even if Lx, Ly are the same. In the example below, I use a system with two points. I use a square box and a sheared box. The points are the same after wrapping (they are within both boxes). However, the distances between the points under periodic boundary conventions are not the same. In the square system, the distance is `0.707...` while in the sheared case the distance is `1.118...`. I expect this difference in the minimum vectors to affect S(q) but could be convinced otherwise by a test as you proposed.

```python
import freud
import numpy as np
import matplotlib
matplotlib.use("Agg")

box_square = freud.Box.square(3)
box_shear = freud.Box(3, 3, xy=0.5, is2D=True)

points = np.array([[-1.5, -1.5, 0], [-1, 1, 0]])
square_points = box_square.wrap(points)
shear_points = box_shear.wrap(points)

print(box_square)
print(box_shear)

print(points)
print(square_points)
print(shear_points)

print(box_square.compute_all_distances(points, points))
print(box_shear.compute_all_distances(points, points))

system_square = freud.AABBQuery.from_system((box_square, square_points))
system_shear = freud.AABBQuery.from_system((box_shear, shear_points))

fig, ax = plt.subplots()
system_square.plot(ax=ax)
system_shear.plot(ax=ax)
fig.savefig("result.png")
```

result

DomFijan: You are right. Actually my argument doesn't even hold for the figure I posted. It took me good 25 mins of staring at it to realize that. Symmetry is satisfied only in x direction, while it doesn't work for y  😆 - because of the shear. 
I will be rolling back the changes sometime today or tomorrow.

freud/diffraction.pyx
DomFijan: Is there a good example of this already in freud? Would this class be visible to users and on docs? What kind of documentation would we have for that class? @bdice  @tommy-waltmann

cpp/diffraction/StaticStructureFactorDirect.cc
```diff
@@ -0,0 +1,306 @@
...
+    const auto& box = neighbor_query->getBox();
+    if (box.is2D())
+    {
+        throw std::invalid_argument("2D boxes are not currently supported.");
+    }
+    const auto k_bin_edges = m_structure_factor.getBinEdges()[0];
+    const auto k_min = k_bin_edges.front();
+    const auto k_max = k_bin_edges.back();
+    if ((!box_assigned)
+        || (box != previous_box && box.getPeriodicX() != previous_box.getPeriodicX()
```
bdice: Refactor the equality operator instead: https://github.com/glotzerlab/freud/blob/9b4db2bf4a073c0c3483de6cccafcecadc0c8efc/cpp/box/Box.h#L82

freud/diffraction.pyx
tommy-waltmann: Have a look at the PMFT module, each PMF class inherits from `_PMFT`. In the case of the PMFT module, there is also a base class on the cpp side, and after looking at the static structure factor cpp classes, I think we should have a base class for those also.

tests/test_diffraction_StaticStructureFactorDirect.py
DomFijan: I added tests for bin_edges, bin_centers, bounds and min_valid_kmin for both debye and direct.

tests/test_diffraction_StaticStructureFactorDirect.py
DomFijan: Could this be resolved directly by constructing the base class (see below) and then testing the base class?

freud/diffraction.pyx
DomFijan: I'll check. This might take a while, especially for cpp side.

tests/test_diffraction_StaticStructureFactorDirect.py
tommy-waltmann: I would not advise this. The base class should stay abstract since it doesn't really represent a method of calculating the structure factor, it should just exist to prevent defining the same properties and documenting them multiple times.

bdice:
I pushed a fix for our Python 3.6 "oldest CI," which uses NumPy 1.14.0. NumPy 1.15.0 is required for `np.histogram_bin_edges`.

DomFijan:
@Charlottez112 discovered yesterday that in some cases S_k generated by this code is not consistent for normal (default) values of `max_k_points`. Below are 3 images to showcase the problem. In first image the default value of `max_k_points` is used (20000) on a small system and everything is in order:
bcc_size_4
When the size of the box is increased by putting more unit cells in it, two subsequent calculations of S_k using direct method for default number of `max_k_points` the two calculations stop matching:
bcc_size_10
Finally if `max_k_points` is increased to 60000 for the new larger system size the problem goes away:
bcc_size_10_max_k_points_60000
Note that the fact that peak heights do not match between different system sizes is not an error as far as I was able to gather from literature. It seems that this is a known deficiency in powder diffraction and simulated spectra where the peak heights are system size dependent for crystalline systems.

The origin of this issue lies in the grid pruning method which is non deterministic. If sufficiently large `max_k_points` is given the problem is alleviated. 

This now raises the problem that `max_k_points` is actually not a good way to define this value because the quality of calculation is then system size dependent. I think we should consider introducing a replacement for this variable in from of some k point density - which would guarantee that the quality of calculation is system size independent. This would in turn make the run time more dependent on system size then it currently is, but it feels the right trade-off to make here. I can make necessary adjustments to the code.   @bdice  any thoughts on this?

bdice:
@Charlottez112 @DomFijan Are the results deterministically wrong when too few k-points are sampled? Or are some random samples of k-points different than others? Just trying to verify that it is in fact a sampling problem. 

How much noise is in the system? What is a “small” system? Python code to reproduce would be helpful here.

DomFijan:
> @Charlottez112 @DomFijan Are the results deterministically wrong when too few k-points are sampled? Or are some random samples of k-points different than others? Just trying to verify that it is in fact a sampling problem.
> 
> How much noise is in the system? What is a “small” system? Python code to reproduce would be helpful here.

The results seem to be changing from run to run. I.e. different peaks don't match when the code is rerun. 

No noise. Small is `system_size = 4` and large is `system_size=10` . For large system `max_k_points=60000` works fine.

        bins = 100
        k_max = 30
        max_k_points = 20000
        system_size = 10

        box, points = freud.data.UnitCell.bcc().generate_system(system_size)
        print(len(points))
        print(box.volume/max_k_points, max_k_points/box.volume)
        # Compute structure factor from freud once
        sf1 = freud.diffraction.StaticStructureFactorDirect(
            bins=bins, k_max=k_max, max_k_points=max_k_points
        )
        sf1.compute(system)

        # Compute structure factor from freud twice
        sf2 = freud.diffraction.StaticStructureFactorDirect(
            bins=bins, k_max=k_max, max_k_points=max_k_points
        )
        sf2.compute(system)

        plt.plot(sf1.bin_centers[1:],sf1.S_k[1:])
        plt.plot(sf2.bin_centers[1:],sf2.S_k[1:])
        plt.show()

bdice:
If there is no noise in the system, then I think the peak heights will depend on whether the exact k-points with nonzero values are included in the randomly sampled k-points. I don't think this is a bug in the code but we could improve the API if users wish to ensure that all k-points are sampled.

The reason you observe this is that the default value for `max_k_points` is probably high enough that it is already sampling all k-points for the smaller system, but does not sample all k-points for the larger system. You can investigate this by looking at the section of the code dealing with pruning and Cardano's formula. That part of the code determines the number of k-points in a sphere of radius `k_max` and determines the sampling criteria based on that.

For cases like this, we could add a special value, like if `max_k_points == 0` then all valid k-points will be considered (we ensure the sample set contains all k-points up to `k_max`, and do not randomly drop any). This could be implemented through a small alteration here to return a prune distance of infinity if `max_k_points == 0`:
https://github.com/glotzerlab/freud/blob/49263cbc8fcd7b6119bb8cbc421844342fe8ebde/cpp/diffraction/StaticStructureFactorDirect.cc#L187

DomFijan:
@bdice  I agree that this is not a bug. However we should make sure that users are aware of this and to try and refactor the API and pick default values such that this is avoided when using default values in as many situations as possible.

Correct me if I'm wrong but wouldn't setting `max_k_points = bins * bins * bins`  be equivalent to not doing any pruning? If user sets `max_k_points` like that, same effect would be achieved as compared to what you are suggesting? Would perhaps make more sense to expand docs to state something like that? 

What would you suggest for the default value for `max_k_points`? I'm not particularly happy with having a random number  there (20 000). Would it make sense to have it be `max_k_points = bins * bins * bins`  by default? 
Or would it make more sense to have some sort of `max_k_point` density measure instead of `max_k_points` along with `max_bins`? This way a default k point density could be provided that is independent of system size? 
Or alternatively we could just document this very carefully with huge exclamation marks?

@Charlottez112  what do you think?

bdice:
> However we should make sure that users are aware of this and to try and refactor the API and pick default values such that this is avoided when using default values in as many situations as possible.

That's a good idea. I would make the default `max_k_points=0` and implement the pruning logic I mentioned previously.

> Correct me if I'm wrong but wouldn't setting `max_k_points = bins * bins * bins` be equivalent to not doing any pruning?

No, that's not quite right. The number of bins is not related to the density of sampling. To compute the exact number of k-points that will be sampled, you'd need to know how many k-points fit on a grid within a sphere of radius `k_max`... but we shouldn't do that because it's complicated and prone to errors at the boundary. The easiest solution is to alter `get_prune_distance` to return infinity if `max_k_points==0` and set 0 as the default value. It will skip the pruning and the code will be very efficient.

> Or alternatively we could just document this very carefully with huge exclamation marks?

Typically "correctness" is at the top of our priorities in freud, closely followed by a balance of performance and usability (which sometimes conflict but not often). I think that defaulting to using all points up to `k_max` when the default value of `max_k_points=0` is provided is aligned with these goals. If possible, users should expect correct and well-formed results without consulting the documentation for edge cases / asterisks on usage.

The behavior to document would be something like "If `max_k_points` is greater than zero, the k-points used will be sampled from the full grid of k-points with uniform radial density, resulting in a sample of `max_k_points` on average."

It's possible that `max_k_points` isn't the right name because it isn't a maximum -- it's a desired number of k-points that then determines the probability of pruning each point such that the sampling produces roughly `max_k_points` k-points.

tommy-waltmann:
As far as naming the variable goes, we could just call it `num_k_points` or `num_sampled_k_points`. I agree that making the default behavior be an amount of sampling that is guaranteed to be sufficient is the right decision here.

requirements/requirements-test.txt
DomFijan: As of today (10/11/2021) PyPI hasn't been updated with the working version of dynasor even though changes to fix the issue were merged on gitlab side.

ChangeLog.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to
 
 ### Added
 * `freud.diffraction.StaticStructureFactorDebye` class (unstable) can be used to compute the static structure factor S(k) using the Debye formula.
+* `freud.diffraction.StaticStructureFactorDirect` class (unstable) can be used to compute the static structure factor S(k) by sampling reciprocal space vectors.
```
bdice: 2.7.0 has been released, this will need to go into 2.8.0 / Added.

cpp/box/Box.h
```diff
@@ -83,7 +83,10 @@ class Box
...
         return ((this->getL() == b.getL()) && (this->getTiltFactorXY() == b.getTiltFactorXY())
                 && (this->getTiltFactorXZ() == b.getTiltFactorXZ())
-                && (this->getTiltFactorYZ() == b.getTiltFactorYZ()));
+                && (this->getTiltFactorYZ() == b.getTiltFactorYZ())
+                && (this->getPeriodicX() == b.getPeriodicX())
+                && (this->getPeriodicY() == b.getPeriodicY())
+                && (this->getPeriodicZ() == b.getPeriodicZ()));
     }
 
     inline bool operator!=(const Box& b) const
```
bdice, line 91: We need to refactor `operator!=` to be implemented in the same way as the equality operator. The best way to do this is to replace the current implementation with one that just checks "!(lhs == rhs)" like this answer: https://stackoverflow.com/a/24938533

Here's the code I would recommend (please verify that it works correctly).
```c++

inline bool operator!=(const Box& b) const
{
    return !(b == *this);
}
```

We should also add Python tests to make sure that boxes with different periodic flags do not compare equal.

cpp/diffraction/StaticStructureFactor.cc
```diff
@@ -0,0 +1,14 @@
...
+    \brief Base class for structure factor classes.
+*/
+
+#include "StaticStructureFactor.h"
+
+namespace freud { namespace diffraction {
+
+    StaticStructureFactor::StaticStructureFactor()
+
+}; }; // namespace freud::diffraction
```
bdice, line 40: I think this file can be deleted. It's not listed in the `CMakeLists.txt` so it isn't being built to my knowledge.

cpp/diffraction/StaticStructureFactorDebye.h
```diff
@@ -18,73 +19,32 @@
 
 namespace freud { namespace diffraction {
 
-class StaticStructureFactorDebye
+class StaticStructureFactorDebye : public StaticStructureFactor
 {
-    using S_kHistogram = util::Histogram<float>;
+    using StructureFactorHistogram = util::Histogram<float>;
```
bdice: I think this definition is being duplicated from the parent class and can be deleted here.

cpp/diffraction/StaticStructureFactor.cc
DomFijan: This changed. We added a constructor with 3 arguments here. Now it's part of the build.

cpp/box/Box.h
DomFijan: So I think the box comparisons work differently on cpp and python level (I might be wrong?). 

Currently, on python level the equal operator gets defined in box.pyx line 690:

    def _eq(self, other):
        return self.to_dict() == other.to_dict()

    def __richcmp__(self, other, int op):
        r"""Implement all comparisons for Cython extension classes"""
        if op == Py_EQ:
            return self._eq(other)
        if op == Py_NE:
            return not self._eq(other)
        else:
            raise NotImplementedError("This comparison is not implemented")

I'm not sure if it works correctly though. If I try this code:

        box = freud.box.Box(2, 2, 2, 1, 0.5, 0.1)
        box2 = freud.box.Box(2, 2, 2, 1, 0.5, 0.1)
        box2.periodic = False
        box3 = freud.box.Box(2, 2, 2, 1, 0.5, 0.1)
        box3.periodic_x = False
        box4 = freud.box.Box(2, 2, 2, 1, 0.5, 0.1)
        box4.periodic_z = False
        print(not(box==box2))
        print(box.periodic,  box.periodic_x,  box.periodic_z)
        print(box2.periodic, box3.periodic_x, box4.periodic_z)
        print(box==box2,box==box3,box==box4

I get following output:
```
False
[ True  True  True] True True
[False False False] False False
True True True
```
So python knows that periodic flags are different but compares the boxes incorrectly? 

As for cpp code, I can add the suggested != operator, it does look much neater. Do we have cpp tests? Would it make sense then that python references cpp equality operators? How would one do that in cython?

bdice: @DomFijan You're right, this is a bug. We should make Python's `__richcmp__` call through to the C++ comparison operator. I'll work on this and push a fix.

cpp/diffraction/StaticStructureFactorDirect.cc
```diff
@@ -0,0 +1,281 @@
...
+
+namespace freud { namespace diffraction {
+
+StaticStructureFactorDirect::StaticStructureFactorDirect(unsigned int bins, float k_max, float k_min,
+                                                         unsigned int num_sampled_k_points)
+    : StaticStructureFactor(bins, k_max, k_min), m_num_sampled_k_points(num_sampled_k_points)
+{
+    // We must construct two separate histograms, one for the counts and one
+    // for the actual S(Q). The counts are used to normalize the S(Q) function.
+    // move semantics in Histogram constructor might cause issues? @bdice
```
bdice: I think move semantics are not an issue because the constructor accepts axes by value, which should copy the vector. Then `m_axes(std::move(axes))` avoids a second copy.

bdice: I'm not 100% sure though.

cpp/box/Box.h
bdice: Done.

freud/diffraction.pyx
```diff
@@ -30,7 +32,94 @@ from freud.util cimport _Compute, vec3
...
+        Args:
+            ax (:class:`matplotlib.axes.Axes`, optional): Axis to plot on. If
+                :code:`None`, make a new figure and axis.
+                (Default value = :code:`None`)
+
+        Returns:
+            (:class:`matplotlib.axes.Axes`): Axis with the plot.
+        """
+        import freud.plot
+        return freud.plot.line_plot(self.bin_centers[self.bin_centers>self.min_valid_k],
```
bdice, line 105: I noticed that the `min_valid_k` property is defined on the child classes, despite having the same implementation for both classes. (The docstrings might be different.) I'm not sure if the property should be used in the parent class's plot method since it's defined in the child class. I would try to define that in the parent class, maybe use a private `def _min_valid_k` (as a "computed property"). Then call that in the parent's plotting method and in the child classes' implementations of `min_valid_k` with their unique docstrings? Not sure exactly what to do here.

freud/diffraction.pyx
```diff
@@ -171,96 +259,209 @@ cdef class StaticStructureFactorDebye(_Compute):
...
+                the new computation; if False, will accumulate data (Default
+                value = True).
+        """  # noqa E501
+        if (query_points is None) != (N_total is None):
+            raise ValueError(
+                "If query_points are provided, N_total must also be provided "
+                "in order to correctly compute the normalization of the "
+                "partial structure factor."
+            )
+        # Convert points to float32 to avoid errors when float64 is passed
```
bdice, line 400: What is the error that happens here? `NeighborQuery.from_system` should accept float32 or float64 with no problems.

```python
>>> import numpy as np
>>> import freud
>>> points = np.array([[0., 0., 0.], [1., 1., 1.]])
>>> points.dtype
dtype('float64')
>>> box = freud.Box.cube(3)
>>> freud.locality.NeighborQuery.from_system((box, points))
<freud.locality._RawPoints object at 0x7fbe34083830>
```

freud/diffraction.pyx
```diff
@@ -171,96 +259,209 @@ cdef class StaticStructureFactorDebye(_Compute):
...
+            const float[:, ::1] l_points = nq.points
+            unsigned int num_points = l_points.shape[0]
+            const vec3[float]* l_query_points_ptr = NULL
+            const float[:, ::1] l_query_points
+            unsigned int num_query_points
+
+
+        if query_points is not None:
+            # Convert points to float32 to avoid errors when float64 is passed
+            l_query_points = np.asarray(query_points).astype(np.float32)
```
bdice: This should use freud's existing functions for converting arrays to float32. Look for examples of `freud.util._convert_array`.

freud/diffraction.pyx
```diff
@@ -171,96 +259,209 @@ cdef class StaticStructureFactorDebye(_Compute):
...
-            This function plots :math:`S(k)` for values above
-            :py:attr:`min_valid_k`.
-
-        Args:
-            ax (:class:`matplotlib.axes.Axes`, optional): Axis to plot on. If
-                :code:`None`, make a new figure and axis.
-                (Default value = :code:`None`)
+    @property
+    def num_sampled_k_points(self):
+        r"""int: The maximum number of :math:`\vec{k}` points to use when constructing :math:`k` space
```
bdice: This isn't a maximum, it's more like a desired number / expected average.

freud/diffraction.pyx
DomFijan: @tommy-waltmann  and I discussed this when we were refactoring the cython classes and introduced the base class. We opted for the current state because of the docs. However, we missed connection to the plotting function...

freud/diffraction.pyx
DomFijan: I'm not sure if its for the points or query points, but the error that last example in freud-examples notebook gives without conversion to float32 is :

```
ValueError: Buffer dtype mismatch, expected 'const float' but got 'double'
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_14363/638289685.py in <module>
     36     Ge_positions = frame[ptypes == "Ge"]
     37     S_positions = frame[ptypes == "S"]
---> 38     sfGe_Ge.compute(
     39         (box, Ge_positions),
     40         query_points=Ge_positions,

util.pyx in freud.util._Compute.__getattribute__.compute_wrapper()

diffraction.pyx in freud.diffraction.StaticStructureFactorDirect.compute()

ValueError: Buffer dtype mismatch, expected 'const float' but got 'double'
```

freud/diffraction.pyx
DomFijan: I've added the def to the base class. Docs seem to be held together correctly from what I can tell. Let me know if this is ok.

cpp/diffraction/StaticStructureFactorDirect.cc
tommy-waltmann: We could always write a quick test script to determine if this is an issue, but your explanation makes sense.

bdice: I think the easiest test would be to see if you can construct multiple new histograms from the original histogram's axes. If it works, that's an indication that it is probably safe.

In the constructor:
```cpp
m_k_histogram = KBinHistogram(m_structure_factor.getAxes());
another_histogram = KBinHistogram(m_structure_factor.getAxes());
std::cout << "Original size: " << m_structure_factor.getAxes()[0].size() << std::endl;
std::cout << "KBins size: " << m_k_histogram.getAxes()[0].size() << std::endl;
std::cout << "Another size: " << another_histogram.getAxes()[0].size() << std::endl;
```

tommy-waltmann (changes requested):
After looking closer at the inheritance heirarchies in freud, I think our base `StaticStructureFactor` class should be a `BondHistogramCompute` on the cpp side, and our `_StaticStructureFactor` class should be a `_SpatialHistogram1D`. I think this class hierarchy makes a lot more sense, and will further consolidate exposing the properties and storing the data needed by the `StaticStructureFactor` classes.

bdice:
@tommy-waltmann I found there were some catches with that design, but you can give it a try if it makes sense now. I can't recall exactly what went wrong but I think it had to do with the fact that the parallelism shouldn't be distributed in the same way as a per-thread bond histogram since the bins represent `k` and choosing the bin isn't a function of the bond.

You might find some benefit to defining a base class `HistogramCompute` that is more general than `BondHistogramCompute` and then sharing the common logic at that level?

(It may also be better for the project momentum to pick a stopping point for this PR and push further refactoring to a follow-up PR.)

tommy-waltmann:
> @tommy-waltmann I found there were some catches with that design, but you can give it a try if it makes sense now. I can't recall exactly what went wrong but I think it had to do with the fact that the parallelism shouldn't be distributed in the same way as a per-thread bond histogram since the bins represent `k` and choosing the bin isn't a function of the bond.
> 
> You might find some benefit to defining a base class `HistogramCompute` that is more general than `BondHistogramCompute` and then sharing the common logic at that level?
> 
> (It may also be better for the project momentum to pick a stopping point for this PR and push further refactoring to a follow-up PR.)

The only thing about the `BondHistogramCompute` class that suggests bonds must be involved is the `accumulateGeneral` function; all the other methods, data members, etc. apply equally well to compute classes that use histograms with no regard to how they compute the histograms. It would be simple to put one member function into a new class, call the new class `BondHistogramCompute`, and rename the old class `HistogramCompute`.

The cython equivalent class, `_SpatialHistogram` and `_SpatialHistogram1D`, would not need to have any changes either.

I am not opposed to the idea of putting this in a separate PR, but if the small set of changes could be done in this PR, that would be nice.

tommy-waltmann:
Actually, I take some of what I said about cython changes back, there appears to be more that would need to be done there. This might require a bit more thought before moving forward with a decision. I'll make an issue and we can move that extra level of refactoring to a new PR, if we decide to pursue it at all.

cpp/diffraction/StaticStructureFactorDirect.h
```diff
@@ -0,0 +1,109 @@
...
+class StaticStructureFactorDirect : public StaticStructureFactor
+{
+    using KBinHistogram = util::Histogram<unsigned int>;
+
+public:
+    //! Constructor
+    StaticStructureFactorDirect(unsigned int bins, float k_max, float k_min = 0,
+                                unsigned int num_sampled_k_points = 0);
+
+    //! Destructor
```
tommy-waltmann:
```suggestion
```

doc/source/reference/credits.rst
```diff
@@ -336,6 +337,7 @@ Charlotte Zhao
 Domagoj Fijan
 
 * Contributed code, design, documentation, and testing for ``StaticStructureFactorDebye`` class.
+* Contributed code, design, documentation, and testing for ``StaticStructureFactorDirect`` class.
```
tommy-waltmann, line 342: Put under my name "contributed code for StaticStructureFactor classes"

freud/_diffraction.pxd
```diff
@@ -1,15 +1,23 @@
...
 cimport freud.util
 from freud.util cimport vec3
 
 
+cdef extern from "StaticStructureFactor.h" namespace "freud::diffraction":
+    cdef cppclass StaticStructureFactor:
+        void reset()
+        const freud.util.ManagedArray[float] &getStructureFactor()
+        const vector[float] getBinEdges() const
+        const vector[float] getBinCenters() const
```
tommy-waltmann: Does each class need to export these separately, or can we get rid of their declarations in the derived classes?

freud/_diffraction.pxd
```diff
@@ -19,3 +27,16 @@ cdef extern from "StaticStructureFactorDebye.h" namespace "freud::diffraction":
         const vector[float] getBinEdges() const
         const vector[float] getBinCenters() const
         float getMinValidK() const
```
tommy-waltmann: This is now a part of the base class, so we can move the declaration to the base class too.

freud/_diffraction.pxd
```diff
@@ -19,3 +27,16 @@ cdef extern from "StaticStructureFactorDebye.h" namespace "freud::diffraction":
...
+
+cdef extern from "StaticStructureFactorDirect.h" namespace "freud::diffraction":
+    cdef cppclass StaticStructureFactorDirect(StaticStructureFactor):
+        StaticStructureFactorDirect(unsigned int, float, float, unsigned int) except +
+        void accumulate(const freud._locality.NeighborQuery*,
+                        const vec3[float]*, unsigned int, unsigned int) except +
+        void reset()
```
tommy-waltmann, line 31: This derived class declares reset separately from the base class also.

freud/box.pyx
```diff
@@ -687,17 +688,19 @@ cdef class Box:
...
+        cdef Box c_other
+        try:
+            c_other = <Box?> other
+            if op == Py_EQ:
+                return dereference(self.thisptr) == dereference(c_other.thisptr)
+            if op == Py_NE:
+                return dereference(self.thisptr) != dereference(c_other.thisptr)
+        except TypeError:
+            # Cython cast to Box failed
+            pass
```
tommy-waltmann, line 702: I don't know exactly how this gets used, but in this case does `__richcmp__` just return none? It seems like in all cases a comparison operator should return either True or False.

freud/diffraction.pyx
```diff
@@ -30,7 +32,99 @@ from freud.util cimport _Compute, vec3
...
+    def k_min(self):
+        """float: Minimum value of k at which to calculate the structure
+        factor."""
+        return self.bounds[0]
+
+
+    @_Compute._computed_property
+    def min_valid_k(self):
+        return self.thisptr.getMinValidK()
+
```
tommy-waltmann:
```suggestion
```

freud/diffraction.pyx
```diff
@@ -30,7 +32,99 @@ from freud.util cimport _Compute, vec3
...
+        factor."""
+        return self.bounds[1]
+
+    @property
+    def k_min(self):
+        """float: Minimum value of k at which to calculate the structure
+        factor."""
+        return self.bounds[0]
+
+
```
tommy-waltmann:
```suggestion
```

freud/diffraction.pyx
```diff
@@ -171,96 +264,203 @@ cdef class StaticStructureFactorDebye(_Compute):
...
+    method. The normalization criterion is based on the Faber-Ziman formalism.
+    For particle types :math:`\alpha` and :math:`\beta`, we compute the total
+    scattering function as a sum of the partial scattering functions as:
+
+    .. math::
+
+        S(k) - 1 = \sum_{\alpha}\sum_{\beta} \frac{N_{\alpha}
+        N_{\beta}}{N_{total}^2} \left(S_{\alpha \beta}(k) - 1\right)
+
+    This class is based on the MIT licensed `Dynasor library
```
tommy-waltmann: We mention earlier in this docstring that this is based on the `dynasor` library, we should consider removing it here.

freud/diffraction.pyx
```diff
@@ -171,96 +264,203 @@ cdef class StaticStructureFactorDebye(_Compute):
...
+            The desired number of :math:`\vec{k}` vectors to sample from the
+            reciprocal lattice grid. If set to 0, all :math:`\vec{k}` vectors
+            are used. If greater than 0, the :math:`\vec{k}` vectors are sampled
+            from the full grid with uniform radial density, resulting in a
+            sample of ``num_sampled_k_points`` vectors on average (Default
+            value = 0).
+    """
+
+    cdef:
+        freud._diffraction.StaticStructureFactorDirect * thisptr
```
tommy-waltmann:
```suggestion
    cdef freud._diffraction.StaticStructureFactorDirect * thisptr
```
I don't see a need to use two lines here.

tests/test_diffraction_StaticStructureFactorDirect.py
```diff
@@ -0,0 +1,206 @@
...
+        # and no other k-points. Thus the smallest bin is measuring S(0) = N.
+        sf = freud.diffraction.StaticStructureFactorDirect(bins=100, k_max=10)
+        L = 10
+        N = 1000
+        box, points = freud.data.make_random_system(L, N)
+        system = freud.AABBQuery.from_system((box, points))
+        sf.compute(system)
+        assert np.isclose(sf.S_k[0], N)
+
+    def test_accumulation(self):
```
tommy-waltmann, line 76: Why not have an accumulation test in `StaticStructureFactorDebye`?

tommy-waltmann (changes requested):
With the above comments about the decision to punt on the more extensive internal refactoring, I gave this PR a full review. I really appreciate all the work you have all done so far on this, it's looking pretty good and I have comments on some more minor things.

tests/test_diffraction_StaticStructureFactorDirect.py
DomFijan: This fails. I've also made test_S_0_is_N for Debye and that one fails as well - this is the origin of the problem. The reason S(0)!=N is that histogram that is created when k_min=0 has bin centre at somewhere > 0. This is not a problem in Direct method because of how the reciprocal_isotropic deals with the k_space grid where all vectors are constructed in loop that starts at zero so that zero vector is allowed and exists. Both methods however show that bin_center of first bin is at the same position >0. We have to think about how to settle this. Should we perhaps create a separate bin that is truly centered at zero and only exists for zero when k_min is forced to be zero? @bdice

DomFijan: As per discussion with @bdice on slack, I've filed a bug report.

requirements/requirements-test.txt
bdice: It _has_ been updated on PyPI, it's just that the version number is not newer than 1.1. 😄 I filed this issue and a PR to fix it. For now, I'll use PyPI with a tag for 1.1b0 and we can let dependabot fix it for us at a later date when my PR is merged and uploaded. I think using a tarball is worse than using a pinned release from PyPI (even though it's versioned incorrectly). books. https://gitlab.com/materials-modeling/dynasor/-/issues/37

cpp/diffraction/StaticStructureFactorDirect.cc
bdice: I tested this a bit, seems fine as it is. Resolving.

freud/diffraction.pyx
bdice: The base class needed to use `self.ssfptr` instead of `self.thisptr`. Fixed in 6c9baefa4bcbc694e3794f5efefc519ce61cc15a. Resolving.

freud/_diffraction.pxd
bdice: This is a good question. It seems to work whether these are defined on the base class or derived classes in `_diffraction.pxd`. Cython doesn't have much for dealing with virtual functions and we don't have many great examples in freud about how to handle this (all the examples I located felt like special cases...). I tweaked it a little bit, so that `getMinValidK` is on the base class (it's not overridden) and `accumulate` is only on the derived classes (since it's overridden).

edit: I also applied Tommy's other suggestion for `reset`. Dom resolved the comments but it didn't look like the changes were applied so I did it.

freud/box.pyx
bdice: This Cython casting error should fall through and return `NotImplemented`. If two objects cannot be compared, comparison methods are supposed to return `NotImplemented` according to the rules of the Python data model.

tl;dr `a == b` will try `a.__eq__(b)`, then `b.__eq__(a)`, then `a is b`. Each fallback happens if the `__eq__` doesn't exist or returns `NotImplemented`.

Info on these rules:
- [`https://docs.python.org/3/reference/datamodel.html#object.__eq__`](https://docs.python.org/3/reference/datamodel.html#object.__eq__)
- This page, under `NotImplemented`: https://docs.python.org/3/reference/datamodel.html#the-standard-type-hierarchy
- https://eev.ee/blog/2012/03/24/python-faq-equality/

tests/test_diffraction_StaticStructureFactorDirect.py
bdice: See #850. Resolving.

bdice (commented):
@DomFijan @tommy-waltmann I have pushed some changes. Everything looks good now except for tests. I filed #853 for follow-up work on that. I approve and recommend merging this PR, but I'll leave it to @DomFijan @tommy-waltmann to decide to press the button (I can't officially approve my own PR).

tommy-waltmann (approved):
LGTM!

tommy-waltmann:
There are codecov tests that are failing, but those are because of how we wrote the tests, and should stop failing as soon as #853 is completed.

PR #843, opened 2021-09-29
Fix compiler optimization flags

vyasr:
## Description

#694 introduced an arbitrary build type ReleaseWithDocs in order to bypass scikit-build's behavior of removing docstrings in Cython code when code is compiled for a Release build. This build type also bypasses CMake's normal logic associated with Release builds, which includes building with optimization turned on. This PR manually adds back optimization flags.

I was wary of problems like this at the time #694 was created. I had avoided suggesting this solution precisely because I was concerned about undesirable side effects, but as that was within a month of my PhD defense and I didn't have time to investigate further into better alternatives I didn't want to hold up the package having documentation (and clearly I didn't check the effect of that PR carefully enough). In lieu of scikit-build/scikit-build#518 ever being resolved, we may want to consider shipping our own slightly modified versions of their CMake modules (with attribution of course) to allow us to take advantage of standard build types again. I did a cursory check and the compilation of ReleaseWithDocs looks sane to me now, but it's a brittle solution to the underlying problem.

## Motivation and Context

Turning off optimization is a significant performance regression that has been present in freud since 2.4.1, when the hack was introduced. As a specific use case, our [example comparison to MDAnalysis](https://github.com/glotzerlab/freud-examples/blob/master/examples/Benchmarking%20RDF%20against%20MDAnalysis.ipynb) shows a 3-6x performance difference between the optimized cases.

## How Has This Been Tested?

Local runs show the expected performance, and inspection of the generated build commands look as expected.

## Screenshots

Performance of MDAnalysis comparison benchmark before the PR:
image

Performance of MDAnalysis comparison benchmark after the PR:
image

## Types of changes

## Checklist:


CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ set(CMAKE_CONFIGURATION_TYPES
 
 # CMake looks for these variables to be defined for the custom configuration.
 set(CMAKE_CXX_FLAGS_RELEASEWITHDOCS
-    ${CMAKE_CXX_FLAGS_RELEASE}
+    "${CMAKE_CXX_FLAGS_RELEASE} -O3"
```
joaander:
```suggestion
    "-O3 -DNDEBUG"
```
`-DNDEBUG` makes `assert` no-ops and may deactivate other debug code paths depending on the library.

I recommend making these explicit here. The reason using `${CMAKE_CXX_FLAGS_RELEASE}` didn't work to set this is likely because that variable is not populated with its defaults until later during the configure step.

`-O3 -DNDEBUG` is the default for gcc/clang type compilers. This may not be valid for MSVC. You may want to add a conditional based on the compiler to set optimizations in the windows builds as well.

CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ set(CMAKE_CONFIGURATION_TYPES
...
     CACHE STRING "")
 set(CMAKE_C_FLAGS_RELEASEWITHDOCS
     ${CMAKE_C_FLAGS_RELEASE}
```
joaander:
```suggestion
    "-O3 -DNDEBUG"
```
For completeness to ensure that any `.c` files are also compiled with optimizations.

joaander (changes requested):

CMakeLists.txt
vyasr: Good idea. I observed DNDEBUG in my build but making it explicit here is probably safer.

CMakeLists.txt
vyasr: Sure no harm in doing that.

CMakeLists.txt
vyasr: Made the changes locally just to be sure I wouldn't observe any changes. It looks fine to me.

joaander (approved):

CMakeLists.txt
```diff
@@ -30,10 +30,10 @@ set(CMAKE_CONFIGURATION_TYPES
 
 # CMake looks for these variables to be defined for the custom configuration.
 set(CMAKE_CXX_FLAGS_RELEASEWITHDOCS
-    ${CMAKE_CXX_FLAGS_RELEASE}
+    "${CMAKE_CXX_FLAGS_RELEASE} -DNDEBUG -O3"
     CACHE STRING "")
 set(CMAKE_C_FLAGS_RELEASEWITHDOCS
-    ${CMAKE_C_FLAGS_RELEASE}
+    "${CMAKE_C_FLAGS_RELEASE}  -DNDEBUG -O3"
```
bdice: Remove extra space.
```suggestion
    "${CMAKE_C_FLAGS_RELEASE} -DNDEBUG -O3"
```

bdice (approved):
Removed one extra space, otherwise LGTM! Thanks for the fix. I’ll follow up on the notebook PR afterward.

PR #867, opened 2021-11-05
update cmake to only use TBB target

### LINKED ISSUE #866: Build Failures with TBB on Ubuntu 20.04
**Describe the bug**
Builds fail, saying they cannot find TBB's include directories.

**To Reproduce**
Install tbb using the apt package manager with the system configuration defined below, and then build freud with
```
python setup.py install --user
```

**Error output**
CMake cannot find TBB's include directories. The CMake output complains about the variable `TBB_INCLUDE_DIR` being set to `NOTFOUND`.

 - OS: Ubuntu 20.0.4 on WSL
 - Version of Python: 3.8
 - Version of freud: 2.7.0

**Additional context**
I found this error while trying to help a new student @marxd1 build freud with the above system configuration. The good news is that we have already pinpointed the problem: line 64 of the top-level `CMakeLists.txt` should read
```
if(NOT TBB_INCLUDE_DIR OR NOT TBB_LIBRARY)
```
instead of
```
if(NOT DEFINED TBB_INCLUDE_DIR OR NOT DEFINED TBB_LIBRARY)
```

Freud's FindTBB.cmake necessarily defines the variables `TBB_INCLUDE_DIR` and `TBB_LIBRARY`, but if it doesn't actually find them on the system, it sets their value to `NOTFOUND`. In this case, the check with `DEFINED` evaluates to true instead of what it should evaluate to, false. 

After some more discussion with @joaander, he noted that freud doesn't use the modern CMake approach of just linking the build target `TBB::tbb` to `libfreud` via `target_link_libraries` after a call to `find_package`. This is another thing worth fixing in a PR to fix the above build issue.

tommy-waltmann:
## Description
Freud's CMake used the `TBB_INCLUDE_DIR` and `TBB_LIBRARY` variables, instead of just linking to the TBB build target, which caused build issues on #866 . This PR makes changes to fix the build issues on certain systems and use better modern CMake style.

## Motivation and Context
Resolves: #866 

## How Has This Been Tested?
The CI will test the new CMake code on many different systems, and I will verify that the system configuration referenced in #866 builds freud correctly.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -218,42 +218,6 @@ references:
       - *comparison
       - *store_benchmarks
 
-  build_linux_test_wheels: &build_linux_test_wheels
-    steps:
-      - *load_code
-      - *update_submodules
-      - run:
-          name: Compile Linux wheels
-          command: |
-            bash .circleci/deploy-linux.sh testpypi
-
-  build_linux_wheels: &build_linux_wheels
-    steps:
-      - *load_code
-      - *update_submodules
-      - run:
-          name: Compile Linux wheels
-          command: |
-            bash .circleci/deploy-linux.sh pypi
-
-  build_macos_test_wheels: &build_macos_test_wheels
-    steps:
-      - *load_code
-      - *update_submodules
-      - run:
-          name: Compile Mac wheels
-          command: |
-            bash .circleci/deploy-macos.sh testpypi
-
-  build_macos_wheels: &build_macos_wheels
-    steps:
-      - *load_code
-      - *update_submodules
-      - run:
-          name: Compile Mac wheels
-          command: |
-            bash .circleci/deploy-macos.sh pypi
-
 jobs:
   check-style:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -337,7 +337,7 @@ workflows:
           requires:
             - check-style
             - linux-python-39
-      - linux-python-36-oldest:
+      - linux-python-37-oldest:
           requires:
             - check-style
             - linux-python-39
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -287,12 +287,12 @@ jobs:
       PYTHON: "python3.7"
     <<: *build_and_test_linux_with_cov
 
-  linux-python-37-oldest:
+  linux-python-36-oldest:
     docker:
-      - image: glotzerlab/ci:2021.11-gcc7_py37
+      - image: glotzerlab/ci:2021.11-gcc7_py36
     environment:
-      PYVER: "3.7"
-      PYTHON: "python3.7"
+      PYVER: "3.6"
+      PYTHON: "python3.6"
       DEPENDENCIES: "OLDEST"
     <<: *build_and_test_linux
 
@@ -337,7 +337,7 @@ workflows:
           requires:
             - check-style
             - linux-python-39
-      - linux-python-37-oldest:
+      - linux-python-36-oldest:
           requires:
             - check-style
             - linux-python-39
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -287,12 +287,12 @@ jobs:
       PYTHON: "python3.7"
     <<: *build_and_test_linux_with_cov
 
-  linux-python-36-oldest:
+  linux-python-37-oldest:
     docker:
-      - image: glotzerlab/ci:2020.10-py36
+      - image: glotzerlab/ci:2021.11-gcc7_py37
     environment:
-      PYVER: "3.6"
-      PYTHON: "python3.6"
+      PYVER: "3.7"
+      PYTHON: "python3.7"
       DEPENDENCIES: "OLDEST"
     <<: *build_and_test_linux
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -289,7 +289,7 @@ jobs:
 
   linux-python-36-oldest:
     docker:
-      - image: glotzerlab/ci:2021.11-gcc7_py36
+      - image: glotzerlab/ci:2020.10-py36
     environment:
       PYVER: "3.6"
       PYTHON: "python3.6"
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,7 +13,7 @@ and this project adheres to
 * Added error checking for `r_min`, `r_max` arguments in `freud.density.RDF` and `freud.locality.NeighborList`.
 * Doctests are now run with pytest.
 * Cleaned up tests for the static structure factor classes.
-* CMake build system only uses references to TBB target
+* CMake build system only uses references to TBB target.
 
 ## v2.7.0 -- 2021-10-01
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,7 @@ and this project adheres to
 * Added error checking for `r_min`, `r_max` arguments in `freud.density.RDF` and `freud.locality.NeighborList`.
 * Doctests are now run with pytest.
 * Cleaned up tests for the static structure factor classes.
+* CMake build system only uses references to TBB target
 
 ## v2.7.0 -- 2021-10-01
 
--- a/doc/source/gettingstarted/installation.rst
+++ b/doc/source/gettingstarted/installation.rst
@@ -43,7 +43,7 @@ The following are **required** for installing **freud**:
 .. note::
 
     Depending on the generator you are using, you may require a newer version of CMake.
-    In particular, Visual Studio 2019 requires CMake >= 3.14.
+    In particular, Visual Studio 2019 requires CMake >= 3.14.0.
     For more information on specific generators, see the `CMake generator documentation <https://cmake.org/cmake/help/git-stage/manual/cmake-generators.7.html>`__.
 
 The **freud** library uses scikit-build and CMake to handle the build process itself, while the other requirements are required for compiling code in **freud**.
--- a/doc/source/gettingstarted/installation.rst
+++ b/doc/source/gettingstarted/installation.rst
@@ -38,13 +38,13 @@ The following are **required** for installing **freud**:
 - `Intel Threading Building Blocks <https://www.threadingbuildingblocks.org/>`__ (>=2017.03.R2)
 - `Cython <https://cython.org/>`__ (>=0.29.14)
 - `scikit-build <https://scikit-build.readthedocs.io/>`__ (>=0.10.0)
-- `CMake <https://cmake.org/>`__ (>=3.12.4)
+- `CMake <https://cmake.org/>`__ (>=3.12.0)
 
 .. note::
 
     Depending on the generator you are using, you may require a newer version of CMake.
-    In particular, Visual Studio 2019 requires CMake 3.14. For more information on specific
-    generators, see the `CMake generator documentation <https://cmake.org/cmake/help/git-stage/manual/cmake-generators.7.html>`__.
+    In particular, Visual Studio 2019 requires CMake >= 3.14.
+    For more information on specific generators, see the `CMake generator documentation <https://cmake.org/cmake/help/git-stage/manual/cmake-generators.7.html>`__.
 
 The **freud** library uses scikit-build and CMake to handle the build process itself, while the other requirements are required for compiling code in **freud**.
 These requirements can be met by installing the following packages from the `conda-forge channel <https://conda-forge.org/>`__:
--- a/doc/source/gettingstarted/installation.rst
+++ b/doc/source/gettingstarted/installation.rst
@@ -35,7 +35,7 @@ The following are **required** for installing **freud**:
 - A C++14-compliant compiler
 - `Python <https://www.python.org/>`__ (>=3.6)
 - `NumPy <https://www.numpy.org/>`__ (>=1.14)
-- `Intel Threading Building Blocks <https://www.threadingbuildingblocks.org/>`__
+- `Intel Threading Building Blocks <https://www.threadingbuildingblocks.org/>`__ (>=2017.03.R2)
 - `Cython <https://cython.org/>`__ (>=0.29.14)
 - `scikit-build <https://scikit-build.readthedocs.io/>`__ (>=0.10.0)
 - `CMake <https://cmake.org/>`__ (>=3.12.4)
--- a/doc/source/gettingstarted/installation.rst
+++ b/doc/source/gettingstarted/installation.rst
@@ -38,13 +38,13 @@ The following are **required** for installing **freud**:
 - `Intel Threading Building Blocks <https://www.threadingbuildingblocks.org/>`__
 - `Cython <https://cython.org/>`__ (>=0.29.14)
 - `scikit-build <https://scikit-build.readthedocs.io/>`__ (>=0.10.0)
-- `CMake <https://cmake.org/>`__ (>=3.6.3)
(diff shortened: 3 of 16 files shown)


bdice:
One thing the CI will not test is older versions of TBB. It would be nice to test an older version on Ubuntu (perhaps 4.3, before the switch to calendrical versions - conda also has this but I suspect testing on Ubuntu is a better case for CMake than testing with conda) and update our docs with a required minimum version number. A version is currently not listed. https://freud.readthedocs.io/en/latest/gettingstarted/installation.html

tommy-waltmann:
Do you think installing TBB 4.3 from source would be best? The ubuntu package manager doesn't have any versions of `libtbb-dev<=4.3`.

tommy-waltmann:
Also, the only failing test is failing because of something that is more of a warning, and its an incorrect warning.

bdice:
> Do you think installing TBB 4.3 from source would be best? The ubuntu package manager doesn't have any versions of `libtbb-dev<=4.3`.

I would test the oldest version you can install with Ubuntu 18.04, hopefully something at least 3 years old? Might be easier to try in a Docker container.

tommy-waltmann:
> > Do you think installing TBB 4.3 from source would be best? The ubuntu package manager doesn't have any versions of `libtbb-dev<=4.3`.
> 
> I would test the oldest version you can install with Ubuntu 18.04, hopefully something at least 3 years old? Might be easier to try in a Docker container.

My laptop happens to have Ubuntu 18.04, which I have tested and confirms works with TBB version 2017~U7-8

tommy-waltmann:
I have confirmed that this set of changes fixes the build issues on the machine we originally found the issue on. All that should be left here is removing some comments, defining a minimum required TBB, and decided whether to merge with the "failing" CI build.

bdice:
> and decide whether to merge with the "failing" CI build.

We need to make sure our "oldest" tests pass here. This may require us to bump our minimum requirement for cmake, which is currently very conservative (supports a very old version of cmake).

```
-- Found TBB: /usr/lib/x86_64-linux-gnu/libtbb.so (found version "2017.0") 
CMake Error at cpp/cluster/CMakeLists.txt:4 (target_link_libraries):
  Object library target "_cluster" may not link to anything.


-- Configuring incomplete, errors occurred!
See also "/root/ci/freud/_skbuild/linux-x86_64-3.6/cmake-build/CMakeFiles/CMakeOutput.log".
  File "/root/ci/freud/venv/lib/python3.6/site-packages/skbuild/setuptools_wrap.py", line 574, in setup
    languages=cmake_languages
  File "/root/ci/freud/venv/lib/python3.6/site-packages/skbuild/cmaker.py", line 232, in configure
    os.path.abspath(CMAKE_BUILD_DIR())))

An error occurred while configuring with CMake.
  Command:
    "cmake" "/root/ci/freud" "-G" "Ninja" "-DCMAKE_INSTALL_PREFIX:PATH=/root/ci/freud/_skbuild/linux-x86_64-3.6/cmake-install" "-DPYTHON_EXECUTABLE:FILEPATH=/root/ci/freud/venv/bin/python3.6" "-DPYTHON_VERSION_STRING:STRING=3.6.9" "-DPYTHON_INCLUDE_DIR:PATH=/usr/include/python3.6m" "-DPYTHON_LIBRARY:FILEPATH=/usr/lib/x86_64-linux-gnu/libpython3.6m.so" "-DSKBUILD:BOOL=TRUE" "-DCMAKE_MODULE_PATH:PATH=/root/ci/freud/venv/lib/python3.6/site-packages/skbuild/resources/cmake" "-DCOVERAGE=ON" "-DCMAKE_BUILD_TYPE:STRING=ReleaseWithDocs"
  Source directory:
    /root/ci/freud
  Working directory:
    /root/ci/freud/_skbuild/linux-x86_64-3.6/cmake-build
Please see CMake's output for more information.
```

tommy-waltmann:
With regard to oldest versions, since we install tbb from conda-forge and cmake from pypi to do ci, we may need to rework our ci-oldest-reqs file, or make another defining an oldest set of conda-forge dependencies.

ChangeLog.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to
 * Added error checking for `r_min`, `r_max` arguments in `freud.density.RDF` and `freud.locality.NeighborList`.
 * Doctests are now run with pytest.
 * Cleaned up tests for the static structure factor classes.
+* CMake build system only uses references to TBB target
```
```suggestion
* CMake build system only uses references to TBB target.
```

bdice:
> 
> 
> With regard to oldest versions, since we install tbb from conda-forge and cmake from pypi to do ci, we may need to rework our ci-oldest-reqs file, or make another defining an oldest set of conda-forge dependencies.

Just wanted to clarify a few things (some may already be aware of some of this info).

Only our Windows CircleCI gets TBB from a conda package:
https://github.com/glotzerlab/freud/blob/3dad11885dcfebfa79fe2575f4e8b556ab3caba2/.circleci/config.yml#L61

The Linux CircleCI uses whatever version of TBB is installed in the `glotzerlab/ci:2020.10-clang11_py39 ` CI container:
https://github.com/glotzerlab/freud/blob/3dad11885dcfebfa79fe2575f4e8b556ab3caba2/.circleci/config.yml#L260

The "oldest" version pinnings are meant to ensure that we support users building from source in HPC environments that receive updates less frequently. Our conda-forge packages depend on compatible TBB libraries, but it matches the rest of conda-forge's pinnings and can be mass-migrated as was done here: https://github.com/conda-forge/freud-feedstock/pull/37. Our PyPI packages are independent of the user's system-installed (or conda-installed) TBB library because we vendor a copy of TBB in the freud wheel.

One way we could run CI tests for our "oldest" TBB version would be to create another cibuildwheel job that doesn't upload a wheel, but only runs the build/test with a different pinning here: https://github.com/glotzerlab/freud/blob/3dad11885dcfebfa79fe2575f4e8b556ab3caba2/.github/workflows/cibuildwheel-before-build.sh#L17-L19 (I don't think we necessary _need_ to run more CI tests but that's one way you could do it.)

I think I would recommend the following actions:
- Figure out what minimal version of CMake is needed to support the CMake changes you made (since the older CMake version pinned in `.circleci/ci-oldest-reqs.txt` is failing now) and bump that version. https://github.com/glotzerlab/freud/blob/3dad11885dcfebfa79fe2575f4e8b556ab3caba2/.circleci/ci-oldest-reqs.txt#L2
- Identify some version of TBB to claim is our "minimum" (even if older versions might work, we can pick some claim and then not provide official support for older versions). It sounds like "2017 Update 7" might be a good minimum from your tests. Put this version in the docs.

joaander:
Use the docker image `glotzerlab/ci:2021.11-gcc7_py37` to test. This is based on Ubuntu 18.04 LTS which ships TBB 2017.03.R2. Users are not likely to install against dependencies older than the oldest currently supported Ubuntu LTS release.

tommy-waltmann:
> Use the docker image `glotzerlab/ci:2021.11-gcc7_py37` to test. This is based on Ubuntu 18.04 LTS which ships TBB 2017.03.R2. Users are not likely to install against dependencies older than the oldest currently supported Ubuntu LTS release.

Does this mean we should also not claim to support python 3.6 anymore?

bdice:
> 
> 
> > Use the docker image `glotzerlab/ci:2021.11-gcc7_py37` to test. This is based on Ubuntu 18.04 LTS which ships TBB 2017.03.R2. Users are not likely to install against dependencies older than the oldest currently supported Ubuntu LTS release.
> 
> Does this mean we should also not claim to support python 3.6 anymore?

The Python 3.6 [container `glotzerlab/ci:2021.11-gcc7_py36`](https://hub.docker.com/layers/glotzerlab/ci/2021.11-gcc7_py36/images/sha256-a1aa5fa2bfd108cf54f556246c755e3454bd4caea37b026067b72d3a06f417e1?context=explore) would work the same way. We can choose to drop Python 3.6 support in a separate PR if needed, but I don't think it is needed for this PR.

The CI container's packages are defined here: https://github.com/glotzerlab/ci-containers/blob/c8796ccca7509281fb037032aae49b1e664a863b/ubuntu18.04/Dockerfile#L19

This uses `libtbb-dev` on Ubuntu 18.04 (bionic) which is 2017~U7-8. https://packages.ubuntu.com/bionic/libtbb-dev

> Users are not likely to install against dependencies older than the oldest currently supported Ubuntu LTS release.

I agree this is a good idea @joaander! That's a reasonable minimum version for freud.

tommy-waltmann:
Can someone help me with the build error that the linux-python-36-oldest test is giving? The error is related to changing the CI image that is used, but I don't know what the best way to go about fixing it is.

bdice:
@tommy-waltmann The newer CI image labeled `glotzerlab/ci:2021.11-gcc7_py36` appears to be already inside of a virtual environment, which is different from the current CI images labeled `glotzerlab/ci:2020.10-py37`. This is also true for `glotzerlab/ci:2021.11-gcc10_py39`. It's just a change in the underlying CI container.

I debugged this by running the following:
```
$ docker run -it glotzerlab/ci:2021.11-gcc10_py39
root@0ada996ced55:/# which python
/opt/venv/bin/python
root@0ada996ced55:/# pip install -U virtualenv --user
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

However the older containers do not use a virtual environment:
```
$ docker run -it glotzerlab/ci:2020.10-py36
root@e2903989b9f3:/# which python
/usr/bin/python
root@e2903989b9f3:/# pip install -U virtualenv --user
Collecting virtualenv
  Downloading virtualenv-20.10.0-py2.py3-none-any.whl (5.6 MB)
  |################################| 5.6 MB 1.7 MB/s
  ...
```

Luckily, this doesn't matter because the old container has the same TBB version, so we don't need to change the Python 3.6 container at all:

```
$ docker run -it glotzerlab/ci:2020.10-py36
root@d8408b0ad4a1:/# dpkg --list | grep tbb
ii  libtbb-dev:amd64           2017~U7-8                           amd64        parallelism library for C++ - development files
ii  libtbb2:amd64              2017~U7-8                           amd64        parallelism library for C++ - runtime files
```

I will take the following steps:
- Revert the changes to the Python 3.6 CI image. The old image has the desired version of TBB but switching to the new image would require changes to our CI scripts because of a difference in the container's use of a Python virtual environment.
- Using a local copy of the container, try running/re-running with progressively newer versions of CMake, starting at the version pinned in `.circleci/ci-oldest-reqs.txt`.
- Update the oldest supported version of CMake.

.circleci/config.yml
```diff
@@ -289,7 +289,7 @@ jobs:
 
   linux-python-36-oldest:
     docker:
-      - image: glotzerlab/ci:2020.10-py36
+      - image: glotzerlab/ci:2021.11-gcc7_py36
```
bdice: Revert:
```suggestion
      - image: glotzerlab/ci:2020.10-py36
```

bdice:
Ah. I see now that @tommy-waltmann already did the two last steps. The earliest version we support now is 3.12.0. That's great!
> - Using a local copy of the container, try running/re-running with progressively newer versions of CMake, starting at the version pinned in .circleci/ci-oldest-reqs.txt.
> - Update the oldest supported version of CMake.

doc/source/gettingstarted/installation.rst
```diff
@@ -35,16 +35,16 @@ The following are **required** for installing **freud**:
 - A C++14-compliant compiler
 - `Python <https://www.python.org/>`__ (>=3.6)
 - `NumPy <https://www.numpy.org/>`__ (>=1.14)
-- `Intel Threading Building Blocks <https://www.threadingbuildingblocks.org/>`__
+- `Intel Threading Building Blocks <https://www.threadingbuildingblocks.org/>`__ (>=2017.03.R2)
 - `Cython <https://cython.org/>`__ (>=0.29.14)
 - `scikit-build <https://scikit-build.readthedocs.io/>`__ (>=0.10.0)
-- `CMake <https://cmake.org/>`__ (>=3.6.3)
+- `CMake <https://cmake.org/>`__ (>=3.12.4)
```
bdice: We test CI with 3.12.0.
```suggestion
- `CMake <https://cmake.org/>`__ (>=3.12.0)
```

doc/source/gettingstarted/installation.rst
```diff
@@ -35,16 +35,16 @@ The following are **required** for installing **freud**:
...
 
 .. note::
 
     Depending on the generator you are using, you may require a newer version of CMake.
-    In particular, on Windows Visual Studio 2017 requires at least CMake 3.7.1, while Visual Studio 2019 requires CMake 3.14.
-    For more information on specific generators, see the `CMake generator documentation <https://cmake.org/cmake/help/git-stage/manual/cmake-generators.7.html>`__.
+    In particular, Visual Studio 2019 requires CMake 3.14. For more information on specific
```
bdice:
```suggestion
    In particular, Visual Studio 2019 requires CMake >= 3.14.
```

doc/source/gettingstarted/installation.rst
```diff
@@ -35,16 +35,16 @@ The following are **required** for installing **freud**:
...
+    generators, see the `CMake generator documentation <https://cmake.org/cmake/help/git-stage/manual/cmake-generators.7.html>`__.
```
bdice: We separate sentences by newlines in the rst docs.
```suggestion
    For more information on specific generators, see the `CMake generator documentation <https://cmake.org/cmake/help/git-stage/manual/cmake-generators.7.html>`__.
```

doc/source/gettingstarted/installation.rst
```diff
@@ -35,15 +35,15 @@ The following are **required** for installing **freud**:
...
 - `Cython <https://cython.org/>`__ (>=0.29.14)
 - `scikit-build <https://scikit-build.readthedocs.io/>`__ (>=0.10.0)
-- `CMake <https://cmake.org/>`__ (>=3.6.3)
+- `CMake <https://cmake.org/>`__ (>=3.12.0)
 
 .. note::
 
     Depending on the generator you are using, you may require a newer version of CMake.
-    In particular, on Windows Visual Studio 2017 requires at least CMake 3.7.1, while Visual Studio 2019 requires CMake 3.14.
+    In particular, Visual Studio 2019 requires CMake >= 3.14.
```
bdice:
```suggestion
    In particular, Visual Studio 2019 requires CMake >= 3.14.0.
```

bdice (approved):
I pushed a few changes. This looks good to me.

vyasr:
@tommy-waltmann @joaander I'm a little confused by this PR. I like having moving the CMake link commands into more local scopes for each extension module, and if the new version works that is fine. But the original version did use `target_link_libraries`, it just included a specific workaround for windows when TBB provided a FindTBB based solution rather than the newer config file approach. Did that problem go away?

**EDIT**
There's an error in my description above. The original workaround was for when TBB provided a TBB-config.cmake, not when we fell back to FindTBB.cmake. The FindTBB logic was previously correct, while the config logic was not and required this workaround.

tommy-waltmann:
It appears that issue is gone, yes. Although I don't understand why it was there in the first place. Our `FindTBB.cmake` creates the tbb target `TBB::tbb` with all the necessary properties set, so we shouldn't have needed to call `target_link_libraries` before with anything other than the `TBB::tbb` target. Perhaps the `FindTBB.cmake` was updated since that workaround was originally written?

PR #869, opened 2021-11-06
Refactor/util histogram axes

### LINKED ISSUE #826: util::Histogram<T>::Axes should be exposed as util::Axes
The code I'm writing for #820 and the code for `CorrelationFunction` both require two histograms: one for computing the values S(k) or C(r) and a second histogram used for normalizing the bin counts. In both cases, the same `Axes` object could be used twice.

However, the `Axes` object (which does not depend on the `util::Histogram<T>` or type `T`) is currently exposed only in the `util::Histogram<T>` namespace as `util::Histogram<T>::Axes`. It seems that this should be exposed as `util::Axes`. I accidentally re-used the `axes` instance with a different `T` and it worked fine, as expected.

Move `using Axes = std::vector<std::shared_pointer<Axis>>` outside of the `Histogram` class, to just below the `Axis` class.

I think `AxisIterator` can be removed -- it is not used in the code.

Then the `CorrelationFunction` can be refactored to use only one `Axes` instance for both histograms. (It is already possible to use the same `Axes` instance for both histograms, but this is not obvious because the type name includes the `Histogram<T>` namespace.)

andkerr:
Refactoring and code cleanup in Freud's c++ histogram utility.

## Description
This PR re-namespaces the `util::Histogram<T>::Axes` object in the histogram utility to `util::Axes`, which allows one Axes instance to be used for both histograms in `CorrelationFunction`.

I refactored the `CorrelationFunction` code to account for the namespace change, and also removed the `AxisIterator` object from Histogram.h that was going unused,

## Motivation and Context
Resolves: #826 

## How Has This Been Tested?
~~All but two tests from `test_diffraction_DiffractionPattern` pass, but I think this is because recent tweaks to those tests that incorporate diffraction normalization by N are not included in this branch yet. Would it be appropriate to merge `fix/diffraction-normalization` into this branch to make them pass?~~

~~Not all code has been refactored yet, I am working on this.~~

Existing tests are passing.

## Types of changes

- [x] Refactoring

## Checklist:

### DIFF
--- a/cpp/density/CorrelationFunction.cc
+++ b/cpp/density/CorrelationFunction.cc
@@ -37,9 +37,8 @@ CorrelationFunction<T>::CorrelationFunction(unsigned int bins, float r_max) : Bo
     m_histogram = util::Histogram<unsigned int>(axes);
     m_local_histograms = util::Histogram<unsigned int>::ThreadLocalHistogram(m_histogram);
 
-    typename util::Axes axes_rdf;
-    axes_rdf.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
-    m_correlation_function = util::Histogram<T>(axes_rdf);
+    axes.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
+    m_correlation_function = util::Histogram<T>(axes);
     m_local_correlation_function = CFThreadHistogram(m_correlation_function);
 }
 
--- a/cpp/density/CorrelationFunction.cc
+++ b/cpp/density/CorrelationFunction.cc
@@ -32,12 +32,12 @@ CorrelationFunction<T>::CorrelationFunction(unsigned int bins, float r_max) : Bo
     // We must construct two separate histograms, one for the counts and one
     // for the actual correlation function. The counts are used to normalize
     // the correlation function.
-    util::Histogram<unsigned int>::Axes axes;
+    util::Axes axes;
     axes.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
     m_histogram = util::Histogram<unsigned int>(axes);
     m_local_histograms = util::Histogram<unsigned int>::ThreadLocalHistogram(m_histogram);
 
-    typename util::Histogram<T>::Axes axes_rdf;
+    typename util::Axes axes_rdf;
     axes_rdf.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
     m_correlation_function = util::Histogram<T>(axes_rdf);
     m_local_correlation_function = CFThreadHistogram(m_correlation_function);
--- a/cpp/density/CorrelationFunction.cc
+++ b/cpp/density/CorrelationFunction.cc
@@ -31,13 +31,11 @@ CorrelationFunction<T>::CorrelationFunction(unsigned int bins, float r_max) : Bo
 
     // We must construct two separate histograms, one for the counts and one
     // for the actual correlation function. The counts are used to normalize
-    // the correlation function.
-    util::Axes axes;
-    axes.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
+    // the correlation function. The histograms can share the same set of axes.
+    const auto axes = util::Axes {std::make_shared<util::RegularAxis>(bins, 0, r_max)};
     m_histogram = util::Histogram<unsigned int>(axes);
     m_local_histograms = util::Histogram<unsigned int>::ThreadLocalHistogram(m_histogram);
 
-    axes.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
     m_correlation_function = util::Histogram<T>(axes);
     m_local_correlation_function = CFThreadHistogram(m_correlation_function);
 }
--- a/cpp/locality/BondHistogramCompute.h
+++ b/cpp/locality/BondHistogramCompute.h
@@ -122,7 +122,6 @@ protected:
         m_local_histograms; //!< Thread local bin counts for TBB parallelism
 
     using BondHistogram = util::Histogram<unsigned int>;
-    using BHAxes = util::Axes;
 };
 
 }; }; // namespace freud::locality
--- a/cpp/locality/BondHistogramCompute.h
+++ b/cpp/locality/BondHistogramCompute.h
@@ -122,7 +122,7 @@ protected:
         m_local_histograms; //!< Thread local bin counts for TBB parallelism
 
     using BondHistogram = util::Histogram<unsigned int>;
-    using BHAxes = typename BondHistogram::Axes;
+    using BHAxes = typename util::Axes;
 };
 
 }; }; // namespace freud::locality
--- a/cpp/locality/BondHistogramCompute.h
+++ b/cpp/locality/BondHistogramCompute.h
@@ -122,7 +122,7 @@ protected:
         m_local_histograms; //!< Thread local bin counts for TBB parallelism
 
     using BondHistogram = util::Histogram<unsigned int>;
-    using BHAxes = typename util::Axes;
+    using BHAxes = util::Axes;
 };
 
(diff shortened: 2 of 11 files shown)


cpp/density/CorrelationFunction.cc
```diff
@@ -32,14 +32,13 @@ CorrelationFunction<T>::CorrelationFunction(unsigned int bins, float r_max) : Bo
...
-    util::Histogram<unsigned int>::Axes axes;
+    util::Axes axes;
     axes.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
     m_histogram = util::Histogram<unsigned int>(axes);
     m_local_histograms = util::Histogram<unsigned int>::ThreadLocalHistogram(m_histogram);
 
-    typename util::Histogram<T>::Axes axes_rdf;
-    axes_rdf.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
-    m_correlation_function = util::Histogram<T>(axes_rdf);
+    axes.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
```
bdice: This line should be removed, because there is only one `axes` that is used twice. This second call to `push_back` will make `axes` 2D instead of 1D.
```suggestion
```

cpp/locality/BondHistogramCompute.h
```diff
@@ -122,7 +122,7 @@ class BondHistogramCompute
         m_local_histograms; //!< Thread local bin counts for TBB parallelism
 
     using BondHistogram = util::Histogram<unsigned int>;
-    using BHAxes = typename BondHistogram::Axes;
+    using BHAxes = typename util::Axes;
```
bdice: I don't think the `typename` is needed anymore, but please check to make sure this compiles.
```suggestion
    using BHAxes = util::Axes;
```

A slightly more useful refactoring could be to find-replace `BHAxes` with `util::Axes` and remove this definition here. The only reason we needed the type `BHAxes` was that this was a long name that depended on `util::Histogram<unsigned int>`, but the new name `util::Axes` is vastly shorter.

cpp/pmft/PMFTXYT.cc
```diff
@@ -49,7 +49,7 @@ PMFTXYT::PMFTXYT(float x_max, float y_max, unsigned int n_x, unsigned int n_y, u
     m_pcf_array.prepare({n_x, n_y, n_t});
 
     // Construct the Histogram object that will be used to keep track of counts of bond distances found.
-    BondHistogram::Axes axes;
+    util::Axes axes;
     axes.push_back(std::make_shared<util::RegularAxis>(n_x, -x_max, x_max));
     axes.push_back(std::make_shared<util::RegularAxis>(n_y, -y_max, y_max));
     axes.push_back(std::make_shared<util::RegularAxis>(n_t, 0, constants::TWO_PI));
```
bdice: If this initializer list works as I hope, it might be cleaner:
```suggestion
    auto const axes = util::Axes {std::make_shared<util::RegularAxis>(n_x, -x_max, x_max),
        std::make_shared<util::RegularAxis>(n_y, -y_max, y_max),
        std::make_shared<util::RegularAxis>(n_t, 0, constants::TWO_PI)};
```

This should work by using the `std::initializer_list<T>` constructor of a `std::vector<T>`:
https://github.com/glotzerlab/freud/blob/3dad11885dcfebfa79fe2575f4e8b556ab3caba2/cpp/util/Histogram.h#L277
https://en.cppreference.com/w/cpp/container/vector/vector

Key benefit: the vector is initialized with the correct size and elements, and thus can be declared `const`.

You might make the same change for other instances of this pattern everywhere that `Axes` is used.

cpp/util/Histogram.h
```diff
@@ -172,6 +172,8 @@ class RegularAxis : public Axis
     float m_inverse_bin_width; //!< Inverse of bin width
 };
 
+using Axes = std::vector<std::shared_ptr<Axis>>;
```
bdice, line 175: Excellent, this is the right move.

bdice (changes requested):
@andkerr Very nice job so far on this. I have some suggestions for expansion/improvement.

cpp/density/CorrelationFunction.cc
```diff
@@ -32,14 +32,13 @@ CorrelationFunction<T>::CorrelationFunction(unsigned int bins, float r_max) : Bo
     // We must construct two separate histograms, one for the counts and one
     // for the actual correlation function. The counts are used to normalize
     // the correlation function.
-    util::Histogram<unsigned int>::Axes axes;
+    util::Axes axes;
     axes.push_back(std::make_shared<util::RegularAxis>(bins, 0, r_max));
```
bdice: Same suggestion as I gave in [another comment](https://github.com/glotzerlab/freud/pull/869/files#r745254349) -- use the `std::initializer_list<T>` form of the vector constructor so that this can be `const`. It being `const` might have prevented the error in [this comment](https://github.com/glotzerlab/freud/pull/869/files#r744189073) with making a 2D `axes` where it was meant to be 1D. 😉
```suggestion
    auto const axes = util::Axes {std::make_shared<util::RegularAxis>(bins, 0, r_max)};
```

andkerr:
Just implemented these changes. Instances of `Axes` objects should now be populated via initializer lists, rather than multiple `push_back()` calls.

j-proc (commented):
This looks good to me and seems like you addressed all feedback from @bdice. Is the note "Not all code has been refactored yet, I am working on this" still valid in the description?

cpp/locality/BondHistogramCompute.h
```diff
@@ -122,7 +122,7 @@ class BondHistogramCompute
         m_local_histograms; //!< Thread local bin counts for TBB parallelism
 
     using BondHistogram = util::Histogram<unsigned int>;
-    using BHAxes = typename BondHistogram::Axes;
+    using BHAxes = util::Axes;
```
bdice: If all uses of `BHAxes` have been removed, this line can be deleted.
```suggestion
```

andkerr: Great, I just pushed this change. My bad for not catching this.

andkerr: What does it mean for a variable to be prefixed with `m_` by the way? I notice this happens a lot in some c++ files.

bdice: It is a naming convention used in freud to indicate member variables (as opposed to variables declared in the function or passed as arguments). I am not sure where the convention came from, it predates my work on the package. 😊

bdice: https://stackoverflow.com/questions/13018189/what-does-m-variable-prefix-mean

I agree with the suggestion in this SO post that this convention is a little antiquated or indicative of confusing design: “Your classes and functions should be small enough that you don't need [the prefix].”

joaander: This is a very common naming convention for C++ and C# projects. Both make "this->" implicit when accessing member variables so the `m_` prefix makes it easier to understand when you are referencing a member variable vs a local variable. Python does not have this problem because `self.` is always explicit.

@bdice I don't understand your comment. Are you suggesting we rewrite all of HOOMD and freud with smaller classes and methods so that there is no ambiguity? Or are you suggesting that we use a less "antiquated" language than C++ (which has been around since 1982)?

bdice: @joaander In some C++ projects I have looked at more recently, I have seen style guides suggesting to avoid conventions like `m_` prefixes in favor of small, concise classes that do not need such a distinction, or explicitly using `this->`. I have no complaint about the use of `m_` in freud or HOOMD and am not advocating for changing existing conventions or languages. I only meant that if I were to write a new library from scratch (and had no reason to prefer consistent style with HOOMD or freud), I might consider not using that convention.

bdice (approved):
Great work @andkerr!

Issue #821, opened 2021-08-27
Require r_min >= 0 at the C++ level.


bdice:
Currently the C++ code in freud does not ensure that `r_min` is non-negative (>= 0) in several classes. Values of `r_min < 0` should raise an error in the constructor like this:

https://github.com/glotzerlab/freud/blob/eb2a97bac302ed45236f26e0a7e6a0346071c579/cpp/density/RDF.cc#L17-L28

This applies to classes like `RDF` ~and `CorrelationFunction`~ and other classes that deal with radial distances that cannot be negative. This also affects `NeighborQuery` related classes, like `NeighborQueryPerPointIterator` (and related query constructors) and the method `NeighborList.filter_r`.

The structure factor calculations in #660 and #820 have `k_min` values where this same rule should apply (but that can be resolved in those PRs before merging them).

Fixing this will require tests that check for invalid values and ensure that an error is raised.

## Unwanted behavior:
```python
>>> freud.density.RDF(100, 10, r_min=-1)
freud.density.RDF(bins=100, r_max=10.0, r_min=-1.0
```

## Desired behavior:
```python
>>> freud.density.RDF(100, 10, r_min=-1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "density.pyx", line 615, in freud.density.RDF.__cinit__
ValueError: RDF requires r_min to be non-negative.
```

SYEDZIAULISLAM:
@bdice what was the exact issue over here? as i seen that **r_min** are not defined in the class like **"CorrelationFunction"** and many class does not have **r-min** value.

bdice:
@SYEDZIAULISLAM Good catch, the `CorrelationFunction` class doesn't support `r_min`. It's just defined with `r_min=0`.
https://github.com/glotzerlab/freud/blob/eb2a97bac302ed45236f26e0a7e6a0346071c579/cpp/density/CorrelationFunction.cc#L36

This issue definitely applies to `RDF` and the `NeighborQuery` classes mentioned above, which do take `r_min` parameters. There may not be any other classes that this affects. Part of resolving this issue will be to verify that there are no other exceptions.

SyedZiaul:
@bdice could you check whether #825 solve this issue .

vyasr:
#825 is only a partial resolution of this issue, see [#825 (comment)](https://github.com/glotzerlab/freud/pull/825#issuecomment-913825792).

AlainKadar:
[...]

PR #825, opened 2021-08-29
added r_min and r_max to the RDF & NeighborList classes

SyedZiaul:
## Description
This PR adds error messages in `RDF` and `NeighborList`  for negative `r_min` and `r_max` values, as well as cases where `r_max <= r_min`.

## Motivation and Context

Partly resolves: #821 

## How Has This Been Tested?
Added tests that ensure errors are raised as expected when negative values are provided, or when `r_max < r_min`

## Types of changes

- [x] New feature (non-breaking change which adds or improves functionality)

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] I have added tests that cover my changes (if relevant).
- [x] All new and existing tests passed.
- [ ] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst).
- [ ] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

PR #871, opened 2021-11-09
Add radial distance checks

andkerr:
This PR completes work started in #825, based on changes suggested in [this](https://github.com/glotzerlab/freud/pull/825#issuecomment-913825792) comment.

## Description
I add checks for appropriate radial distance arguments in the `RDF` and `NeighborList` class constructors, and unit tests that verify that invalid parameters raise appropriate error messages, rather than produce undefined behavior.

## Motivation and Context
Resolves Issue #821 and PR #825.

Will this PR require any changes to docs for the corresponding modules?

## How Has This Been Tested?
Existing code compiles. I am sorting out the unit tests (some of my checks for error messages do not pass right now).

## Screenshots
N/A

## Types of changes

## Checklist:

### DIFF
--- a/cpp/density/LocalDensity.cc
+++ b/cpp/density/LocalDensity.cc
@@ -20,7 +20,7 @@ LocalDensity::LocalDensity(float r_max, float diameter)
 
     if (diameter < 0)
     {
-        throw std::invalid_argument("LocalDensity requires diameter to be non-negative");
+        throw std::invalid_argument("LocalDensity requires diameter to be non-negative.");
     }
 }
 
--- a/cpp/density/LocalDensity.cc
+++ b/cpp/density/LocalDensity.cc
@@ -12,7 +12,17 @@ namespace freud { namespace density {
 
 LocalDensity::LocalDensity(float r_max, float diameter)
     : m_box(box::Box()), m_r_max(r_max), m_diameter(diameter)
-{}
+{
+    if (r_max <= 0)
+    {
+        throw std::invalid_argument("Local density calculations require that r_max must be positive");
+    }
+
+    if (diameter < 0)
+    {
+        throw std::invalid_argument("Local density calculations require that diameter must be non-negative");
+    }
+}
 
 void LocalDensity::compute(const freud::locality::NeighborQuery* neighbor_query,
                            const vec3<float>* query_points, unsigned int n_query_points,
--- a/cpp/density/LocalDensity.cc
+++ b/cpp/density/LocalDensity.cc
@@ -15,12 +15,12 @@ LocalDensity::LocalDensity(float r_max, float diameter)
 {
     if (r_max <= 0)
     {
-        throw std::invalid_argument("Local density calculations require that r_max must be positive");
+        throw std::invalid_argument("LocalDensity requires r_max to be positive.");
     }
 
     if (diameter < 0)
     {
-        throw std::invalid_argument("Local density calculations require that diameter must be non-negative");
+        throw std::invalid_argument("LocalDensity requires diameter to be non-negative");
     }
 }
 
--- a/cpp/locality/NeighborQuery.h
+++ b/cpp/locality/NeighborQuery.h
@@ -265,7 +265,16 @@ public:
         : NeighborPerPointIterator(query_point_idx), m_neighbor_query(neighbor_query),
           m_query_point(query_point), m_finished(false), m_r_max(r_max), m_r_min(r_min),
           m_exclude_ii(exclude_ii)
-    {}
+    {
+        if (r_max <= 0)
+        {
+            throw std::invalid_argument("Neighbor queries require that r_max must be positive.");
+        }
+        if (r_max <= r_min)
+        {
+            throw std::invalid_argument("Neighbor queries require that r_max must be greater than r_min.");
+        }
+    }
 
     //! Empty Destructor
     ~NeighborQueryPerPointIterator() override = default;
--- a/cpp/locality/NeighborQuery.h
+++ b/cpp/locality/NeighborQuery.h
@@ -268,11 +268,11 @@ public:
     {
         if (r_max <= 0)
         {
-            throw std::invalid_argument("Neighbor queries require that r_max must be positive.");
+            throw std::invalid_argument("NeighborQuery requires r_max to be positive.");
         }
         if (r_max <= r_min)
         {
-            throw std::invalid_argument("Neighbor queries require that r_max must be greater than r_min.");
+            throw std::invalid_argument("NeighborQuery requires that r_max must be greater than r_min.");
         }
     }
 
--- a/tests/test_density_LocalDensity.py
+++ b/tests/test_density_LocalDensity.py
@@ -100,4 +100,4 @@ class TestLocalDensity:
         with pytest.raises(ValueError):
             freud.density.LocalDensity(-1, 10)
         with pytest.raises(ValueError):
-            freud.density.LocalDensity(2, 0)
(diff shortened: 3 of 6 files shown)


cpp/density/LocalDensity.cc
```diff
@@ -12,7 +12,17 @@ namespace freud { namespace density {
 
 LocalDensity::LocalDensity(float r_max, float diameter)
     : m_box(box::Box()), m_r_max(r_max), m_diameter(diameter)
-{}
+{
+    if (r_max <= 0)
+    {
+        throw std::invalid_argument("Local density calculations require that r_max must be positive");
```
bdice: This suggestion matches the style of the existing error messages more closely. See #825 for other examples.
```suggestion
        throw std::invalid_argument("LocalDensity requires r_max to be positive.");
```

cpp/density/LocalDensity.cc
```diff
@@ -12,7 +12,17 @@ namespace freud { namespace density {
...
+    }
+
+    if (diameter < 0)
+    {
+        throw std::invalid_argument("Local density calculations require that diameter must be non-negative");
```
bdice:
```suggestion
        throw std::invalid_argument("LocalDensity requires diameter to be non-negative.");
```

cpp/locality/NeighborQuery.h
```diff
@@ -265,7 +265,16 @@ class NeighborQueryPerPointIterator : public NeighborPerPointIterator
         : NeighborPerPointIterator(query_point_idx), m_neighbor_query(neighbor_query),
           m_query_point(query_point), m_finished(false), m_r_max(r_max), m_r_min(r_min),
           m_exclude_ii(exclude_ii)
-    {}
+    {
+        if (r_max <= 0)
+        {
+            throw std::invalid_argument("Neighbor queries require that r_max must be positive.");
```
bdice:
```suggestion
            throw std::invalid_argument("NeighborQuery requires r_max to be positive.");
```

cpp/locality/NeighborQuery.h
```diff
@@ -265,7 +265,16 @@ class NeighborQueryPerPointIterator : public NeighborPerPointIterator
...
+        }
+        if (r_max <= r_min)
+        {
+            throw std::invalid_argument("Neighbor queries require that r_max must be greater than r_min.");
```
bdice:
```suggestion
            throw std::invalid_argument("NeighborQuery requires that r_max must be greater than r_min.");
```

tests/test_density_LocalDensity.py
```diff
@@ -94,3 +94,10 @@ def test_points_ne_query_points(self):
         ) / v_around
         correct_density = [cd0, cd1, 0]
         npt.assert_allclose(ld.density, correct_density, rtol=1e-4)
+
+    def test_invalid_radial_distances(self):
+        # make sure that invalid r_max and diameter arguments raise errrors
+        with pytest.raises(ValueError):
+            freud.density.LocalDensity(-1, 10)
+        with pytest.raises(ValueError):
+            freud.density.LocalDensity(2, 0)
```
bdice: Diameter is allowed to be zero. A negative value should raise an error.
```suggestion
            freud.density.LocalDensity(2, -1)
```

bdice (commented):
Thanks for working on this!

vyasr:
Thanks @andkerr!

andkerr:
@bdice, I'm running into a problem that I haven't seen before. It seems like extra validation I added to the `LocalDensity` and `NeighborPointQueryIterator` constructors are causing pytest to abort on a fatal error when it tries to run the test I wrote to check that the appropriate errors are raised for invalid parameters. It looks like the problem is specifically with the additional checks in `LocalDensity.cc`. When I comment them out, the test I added simply fails with no error. 

'm not sure where my mistake is, I've done my best to match the form of other calls to `std::invalid_argument`. This is the error message I get:

Screen Shot 2021-11-21 at 2 52 05 PM

bdice:
@andkerr It's because the constructor for `LocalDensity` is missing an `except +` in the Cython bindings: https://github.com/glotzerlab/freud/blob/94e8f2404c3b8f4976634e2ce42c7eaf474dcf88/freud/_density.pxd#L38

Compare to the RDF constructor, which has `except +`:
https://github.com/glotzerlab/freud/blob/94e8f2404c3b8f4976634e2ce42c7eaf474dcf88/freud/_density.pxd#L52

Please add `except +` to all the functions (probably all constructors and `compute` methods) that can raise an error. 👍

Cython docs: https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#error-return-values

andkerr:
Great, I added the `except +` flag, tests are passing now.

cpp/locality/NeighborQuery.h
```diff
@@ -265,7 +265,16 @@ class NeighborQueryPerPointIterator : public NeighborPerPointIterator
...
+    {
+        if (r_max <= 0)
+        {
+            throw std::invalid_argument("NeighborQuery requires r_max to be positive.");
+        }
+        if (r_max <= r_min)
+        {
+            throw std::invalid_argument("NeighborQuery requires that r_max must be greater than r_min.");
+        }
+    }
```
bdice, line 277: Can we add Python tests for this? It probably needs to be tested using concrete subclasses like AABBQuery and LinkCell. We need to be sure we get `except +` on all the methods that can construct a NeighborQuery -- that might take some careful thought.

andkerr: Sure thing, I'll take care of this.

andkerr:
I added checks for appropriate errors when invalid radius arguments are passed to `AABB Query` and `LinkCell`. Both classes already had the necessary `except +` flags in the Cython declarations. I had to pass some additional argument to the calls to `query()` (`num_neighbors`, `mode`) to prevent Freud from throwing a RuntimeError, and am not sure if this was the correct choice.

cpp/density/LocalDensity.cc
```diff
@@ -12,7 +12,17 @@ namespace freud { namespace density {
...
-{}
+{
+    if (r_max <= 0)
+    {
+        throw std::invalid_argument("LocalDensity requires r_max to be positive.");
+    }
+
+    if (diameter < 0)
+    {
+        throw std::invalid_argument("LocalDensity requires diameter to be non-negative");
```
bdice:
```suggestion
        throw std::invalid_argument("LocalDensity requires diameter to be non-negative.");
```

cpp/locality/NeighborQuery.h
bdice: @andkerr I changed the new tests a bit in ad4e6f4e9148796d7690fd749d85e812c2f2513d. It looks like you defined the tests in both AABBQuery and LinkCell classes, but the logic can be defined once in their shared parent class.

bdice (approved):
LGTM, I pushed some small tweaks. Automerging when tests pass.

tests/test_density_LocalDensity.py
```diff
@@ -94,3 +94,10 @@ def test_points_ne_query_points(self):
         ) / v_around
         correct_density = [cd0, cd1, 0]
         npt.assert_allclose(ld.density, correct_density, rtol=1e-4)
+
+    def test_invalid_radial_distances(self):
+        # make sure that invalid r_max and diameter arguments raise errrors
```
bdice:
```suggestion
        """Ensure that invalid r_max and diameter arguments raise errors."""
```

andkerr:
Nice!

PR #872, opened 2021-11-18
Fix StaticStructureFactorDebye not giving S(0)=N

### LINKED ISSUE #850: StaticStructureFactorDebye doesn't give S(0)=N
**Describe the bug**
StaticStructureFactorDebye doesn't give S(0)=N

**To Reproduce**
```
import freud
sf=freud.diffraction.StaticStructureFactorDebye(bins=100,k_max=10,k_min=0)
L = 10
N = 1000
box, points = freud.data.make_random_system(L, N)
system = freud.AABBQuery.from_system((box, points))
sf.compute(system)
print(sf.S_k[0],N,sf.bin_centers[0])
```

**Error output**
Gives following output:
`989.64526 1000 0.05000000`

 - OS: macOS 10.15.7
 - Version of Python 3.9.1
 - Version of freud 2.7

**Additional context**
The issue is in the bin centre position. This answer is consistent with ase implementation. However if we require k_min=0, we should expect S(k_min=0) to be equal to N as per documentation. @bdice suggests re-centering bins in k-space as s possible solution. Alternatively an additional bin can be added if k_min is zero that only handles k=0.

Tests should be made to ensure that this behaviour is consistent with docs. Additionally accumulation tests should also be made to confirm this works correctly.

DomFijan:
Static structure factor Debye method doesn't give `S(0) = N` as per design and documentation.

## Description
Change binning in StaticStructureFactor base class to center first bin at `k_min` and last bin at `k_max`, as per discussion in #850 .

## Motivation and Context
Resolves: #850 

## How Has This Been Tested?
Tests will have to be change because testing agains outside libraries will be broken. Tests for S(0)=N will have to be added for `StaticStructureFactorDebye`.

## Types of changes

## Checklist:


andkerr:
@DomFijan I was working through the math to center the k bins at the k_min and k_max values that get passed to the S(q) constructor, and I think I came up with a slightly different implementation that preserves the number of bins instead of the bin width:

```
const float delta = (k_max - k_min) / static_cast<float>(bins - 1);
const float target_k_max = k_max + 0.5 * delta;
const float target_k_min = k_min - 0.5 * delta;
const auto axes = util::Axes {std::make_shared<util::RegularAxis>(bins, target_k_min, target_k_max)};
```

Is there a strong reason to prefer one method of bin-centering over the other? I am curious.

edit: Here are the results when I follow your example, are these values correct?:
```
#
# create S(q) instance, call compute
#
>>> print(sf.S_k[0],N,sf.bin_centers[0])
1000.0 1000 -3.725290298461914e-09
```

DomFijan:
In consultation with @tommy-waltmann we decided that the issue is that there is no issue but rather that the docs need to be updated to reflect what is actually going on. Setting ``k_min``=0 does not mean that zeroth bin is centered at zero, but rather that the edge of the zeroth bin is at zero. This means that the only time we can get S(0)=N is when ``k_min`` is set to zero and when bin width goes to zero which can be effectively set by setting large number of ``bins``. I updated the docs to reflect this. I will also add some more tests to test this properly.

tests/test_diffraction_StaticStructureFactor.py
```diff
@@ -220,6 +220,30 @@ def test_repr(self):
...
         assert str(sf) == str(eval(repr(sf)))
 
+    def test_S_0_is_N(self):
+        L = 10
+        N = 1000
+        sf = self.build_structure_factor_object(bins=100, k_max=0.1)
+        box, points = freud.data.make_random_system(L, N)
+        system = freud.AABBQuery.from_system((box, points))
+        sf.compute(system)
+        assert np.isclose(sf.S_k[0], N)
```
bdice, line 248: This PR doesn't actually implement the proposal in issue #850. What we need is that the lowest bin _center_ is at `k_min` (its left edge would be in negative k-space) and the highest bin _center_ is at `k_max`.

> For comparison, the dynasor code puts it first bin center at 0 and its last bin center at `k_max`. That means the first bin's left edge is slightly below 0 and the last bin's right edge is slightly above `k_max`. This seems appropriate enough for our Debye code, since it only uses the bin centers.
https://github.com/glotzerlab/freud/issues/850#issuecomment-944799533

This test passes for `StaticStructureFactorDebye`, but not for the reason it should. The changes in b9a36120bc13ad31230f691c79fb66c39f66f582 make the `k_max` so small that `sf.bin_centers[0] == 0.0005`. That's close enough to zero that the Debye method gives `sf.S_k[0] == 999.99896`, which in turn passes the test `assert np.isclose(999.99896, 1000)`. If you change `k_max` back to 10, you get `sf.bin_centers[0] == 0.05` and `sf.S_k[0] == 989.6409`, which fails the test: `assert np.isclose(989.6409, 1000)`. That value of `k_max = 10` was deliberately chosen, and we should keep it there. There are other sets of parameters that can work, but `bins=100, k_max = 10` is a known good set and `bins=100, k_max=0.1` causes the test to pass but for the wrong reasons.

bdice (dismissed):

bdice:
> Setting `k_min`=0 does not mean that zeroth bin is centered at zero, but rather that the edge of the zeroth bin is at zero.

That's the problem that #850 aimed to resolve. The desired behavior is that the lowest bin _should_ be centered at `k_min` (not with its lower edge at `k_min`).

> This means that the only time we can get S(0)=N is when `k_min` is set to zero and when bin width goes to zero which can be effectively set by setting large number of `bins`.

Since the Debye method isn't truly a histogram, but an evaluation of the function at each bin center, we don't have to treat it like a histogram. All RDFs and correlation functions and the S(k) direct method have physical meaning to the bin edges, because they define where contributions from a given `\vec{r}` or `\vec{k}` belong. The bin edges are not physically meaningful for the Debye method. The bin centers are the relevant values, since they define where the function is evaluated.

bdice:
I pushed a few commits that implement the logic I had in mind for resolving #850. I reverted the changes to `k_max` in the tests and added logic to the C++ that mirrors my expectations for the class. I updated the docs and tests to match.

@tommy-waltmann @DomFijan What do you think? Does this change in conventions make sense to you?

tommy-waltmann:
The change in conventions is fine. I think the Debye method would have been okay whether we put the first k-space point at exactly 0, or if it was offset by half a bin width. Given that the changes are already there now for putting the first k-space point at 0, we should go with that.

I would like to add that if the Debye method isn't truly a histogram calculation, then our docs, api, internal classes, etc. should reflect that, or it will be confusing conceptually. For example, I would recommend changing the `bins` argument in the Debye constructor from `bins` to something like `num_points` that doesn't make people immediately think of a histogram. Any other properties of that class that use histogram language should also be changed. Likewise, our docs for that class should not use histogram terminology or or really talk about histograms at all. Internally, if we could not use the `Histogram` and `Axes` classes to implement the Debye method, that would be ideal, although I recognize that there may be some ease of parallel implementation related to using those classes, because the infrastructure for it is already there.

bdice:
@tommy-waltmann I understand what you're going for. However, I am not sure if the goal of conceptual purity is worth re-implementing freud's core functionality and internal/external APIs with a different name. It would be a pretty big undertaking. I would prefer to see this PR merged as-is. I think it's okay to use `bins` (and a few other histogram-like conventions) for the sake of API uniformity with the rest of freud.

I will leave this decision to you, @DomFijan, and others.

tommy-waltmann:
I don't think it would be a huge undertaking, honestly. I would be fine with just making changes in python APIs and docs so users aren't conceptually confused. We would only have to change the name of two python properties, a constructor argument in python, and re-word a paragraph in the documentation.

bdice:
@tommy-waltmann For scoping purposes, could we do that in a separate PR?

tommy-waltmann:
I'd be fine with that. I'll make an issue and then give what we have here a full review.

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -15,9 +15,42 @@
 
 namespace freud { namespace diffraction {
 
+namespace {
```
tommy-waltmann, line 18: Why the empty namespace?

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -15,9 +15,42 @@
...
+    }
+    if (k_min < 0)
+    {
+        throw std::invalid_argument("StaticStructureFactorDebye requires k_min to be non-negative.");
+    }
+    if (k_max <= k_min)
+    {
+        throw std::invalid_argument(
+            "StaticStructureFactorDebye requires that k_max must be greater than k_min.");
+    }
```
tommy-waltmann, line 52: I realize this validation logic cannot go in the super constructor anymore due to shifting needed in `StaticStructureFactorDebye`, but the code itself is the same and now exists in multiple places. To remove code duplication, let's put this is a helper method/function.

tests/test_diffraction_StaticStructureFactor.py
```diff
@@ -132,7 +144,7 @@ def test_attribute_access(self):
         assert np.isclose(sf.k_min, k_min)
         if hasattr(sf, "num_sampled_k_points"):
             assert sf.num_sampled_k_points == num_sampled_k_points
-        npt.assert_allclose(sf.bounds, (k_min, k_max))
```
tommy-waltmann, line 135: reason for this change?

tests/test_diffraction_StaticStructureFactor.py
```diff
@@ -204,8 +221,9 @@ def test_attribute_shapes(self):
             bins, k_max, k_min, num_sampled_k_points
         )
         assert sf.bin_centers.shape == (bins,)
-        assert sf.bin_edges.shape == (bins + 1,)
-        npt.assert_allclose(sf.bounds, (k_min, k_max))
+        if not self.DEBYE:
+            assert sf.bin_edges.shape == (bins + 1,)
+        npt.assert_allclose(sf.bounds, (k_min, k_max), rtol=1e-5, atol=1e-5)
```
tommy-waltmann, line 226: Why add the `rtol` and `atol` arguments?

tests/test_diffraction_StaticStructureFactor.py
```diff
@@ -281,7 +318,7 @@ def test_debye_validation(self):
         system = freud.locality.NeighborQuery.from_system((box, points))
         sf.compute(system)
         Q, S = self._validate_debye_method(system, bins, k_max, k_min)
-        npt.assert_allclose(sf.bin_centers, Q)
+        npt.assert_allclose(sf.bin_centers, Q, rtol=1e-5, atol=1e-5)
```
tommy-waltmann, line 321: Same comment as above.

tommy-waltmann (changes requested):
I had a few things I wanted to mention, but nothing major.

cpp/diffraction/StaticStructureFactorDebye.cc
bdice: > Unnamed namespaces are typically used when you have a lot of content that you want to ensure stays local to a given file [...]

https://www.learncpp.com/cpp-tutorial/unnamed-and-inline-namespaces/

These helper functions shouldn't be a part of the `freud::diffraction` public namespace. We only want to use them in this translation unit. Note: anonymous namespaces should never be used in headers (`*.h`), only implementation files (`*.cc`).

C++ Core Guidelines has advice on these topics as well:
- [SF.22: Use an unnamed (anonymous) namespace for all internal/non-exported entities](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities)
- [SF.21: Don’t use an unnamed (anonymous) namespace in a header](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-unnamed)

cpp/diffraction/StaticStructureFactorDebye.cc
bdice: Defining validation logic in each class is consistent with the rest of freud. We do that so the error messages can be highly descriptive without needing to mess with string concatenation / string formatting through a generalized function.

If you still think this is necessary, I would defer this to a separate PR.

tests/test_diffraction_StaticStructureFactor.py
bdice: The Debye method has a small amount of floating point error in the reported `k_min, k_max` after performing the shifting and computing the bin centers (which effectively shifts back to the input values). `k_min` was reporting values around `3e-9` for me, instead of exactly zero. I don't think we can refactor the arithmetic to reduce the floating point error. It would probably take a large amount of refactoring to allow the Histogram to be provided bin centers instead of bin edges.

tests/test_diffraction_StaticStructureFactor.py
bdice: See response above.

tests/test_diffraction_StaticStructureFactor.py
bdice: See response on other comment thread.

tests/test_diffraction_StaticStructureFactor.py
tommy-waltmann: In that case we can leave it as is.

tommy-waltmann (approved):
This looks good to me! Codecov appears to be lying about the drop in coverage, we have tests that cover the things it says we haven't covered.

bdice:
@tommy-waltmann Thanks! FYI, code coverage has not worked correctly since the switch to scikit-build. We might want to disable it.

PR #937, opened 2022-04-07
Update Name of Debye Properties

### LINKED ISSUE #934: Improve naming of StaticStructureFactorDebye properties

The calculation done by the `StaticStructureFactorDebye` does not actually involve histogramming in any way, but still uses histogram style naming conventions for its properties because other methods of computing structure factors do involve histogramming. For example, `bins` is a constructor argument, `bin_centers` is a property, etc.

This is conceptually confusing and better naming conventions would make it easier to understand. In addition, there are paragraphs in the documentation for `StaticStructureFactorDebye` which use histogram language to describe the classes properties and functionality.

Rename properties which have misleading names and reword documentation to remove histogram-style language.

**Additional context**
This issue is being made after some discussion on #872

tommy-waltmann:
## Description

this PR updates the names of some of the properties in the `StaticStructureFactorDebye` class. Since the calculation does not involve a histogram, those names are conceptually confusing and this PR updates the docs to remove histogram language as well.

This PR also does a significant re-work of the tests in `test_diffraction_StaticStructureFactor`, since the format in the previous version could not have handled changing property names well.

## Motivation and Context
There was discussion on changing the property names on #872 , and this PR implements those ideas

Resolves: #934 

## How Has This Been Tested?
Tests have been re-worked significantly to take advantage of pytest fixtures. No new tests have been added, but the tests have been re-worked to be more flexible and better parametrized.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,14 @@ The format is based on
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## vX.Y.Z -- XXXX-XX-XX
+
+### Added
+* (breaking) Some `freud.diffraction.StaticStructureFactorDebye` property names changed to be more descriptive.
+
+### Fixed
+* `freud.diffraction.StaticStructureFactorDebye` implementation now gives `S_k[0] = N`.
+
 ## v2.8.0 -- 2022-01-25
 
 ### Added
--- a/doc/source/reference/credits.rst
+++ b/doc/source/reference/credits.rst
@@ -326,7 +326,9 @@ Tommy Waltmann
 * Contributed code, design, and testing for ``StaticStructureFactorDirect`` class.
 * Refactor tests for ``StaticStructureFactor`` classes.
 * Improve CMake build system to use more modern style.
-* Remove CI build configurations from CircleCI which were already covered by CIBuildWheel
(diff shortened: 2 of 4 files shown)


joaander:
Will this go in a freud 3.0 release to reflect the breaking change in the API?

tommy-waltmann:
We had previously declared the `StaticStructureFactorDebye` class's API unstable, so it doesn't necessitate a major version bump, although it's definitely something to consider.

We should have some more discussion on going to a v3 because there are some other issues that have the 3.0 label, and I would want to address those before going to a version 3. If we choose to address those before a version 3 release, this PR needs to be merged into a `next` branch.

joaander:
OK, if the API is marked unstable its fine.

j-proc (approved):
Seems good to me with what Josh specified about the unstable API. Everything looks clean and the test values seem to be the same despite some reworking. Approving.

I'm curious if this resolves all the planned changes for this code an thus could be marked stable for the future? Apologies if that's here and I missed it.

tommy-waltmann:
That's a good question. we had previously labeled both the debye method and the direct method as unstable, and at this point we've discussed it enough that I'm okay with where it's at going forward. Is there a standard way for declaring or announcing that a previously unstable API is now considered stable? Maybe adding something like that to the changelog?

PR #943, opened 2022-04-14
Fix/noncubic diffraction

tommy-waltmann:
## Description

This PR disables the diffraction calculation for noncubic boxes

## Motivation and Context

The diffraction calculation doesn't give correct results if the box is not a cube, and we don't want users getting incorrect results and not realizing it. This is discussed on #940 and is a change that is overdue.

## How Has This Been Tested?

Tests have been added which assert that the proper exception is raised.

## Types of changes

## Checklist:

### DIFF
--- a/doc/source/reference/credits.rst
+++ b/doc/source/reference/credits.rst
@@ -329,6 +329,7 @@ Tommy Waltmann
 * Remove CI build configurations from CircleCI which were already covered by CIBuildWheel.
 * Change property names in ``StaticStructureFactorDebye`` class.
 * Reformat static structure factor tests.
+* ``DiffractionPattern`` now raises an error when used with non-cubic boxes.
 
 Maya Martirossyan
 
--- a/freud/box.pyx
+++ b/freud/box.pyx
@@ -595,8 +595,16 @@ cdef class Box:
     @property
     def cubic(self):
         """bool: Whether the box is a cube."""
-        return self.Lx == self.Ly and self.Ly == self.Lz and self.xy == self.yz\
-            and self.yz == self.xz and self.xz == 0
+    return (
+        not self.is2d
+        and np.allclose(
+            [self.Lx, self.Lx, self.Ly, self.Ly, self.Lz, self.Lz],
+            [self.Ly, self.Lz, self.Lx, self.Lz, self.Lx, self.Ly],
+            rtol=1e-5,
+            atol=1e-5,
+        )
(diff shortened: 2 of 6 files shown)


freud/box.pyx
```diff
@@ -592,6 +592,12 @@ cdef class Box:
 
         return np.array(l_contains_mask).astype(bool)
 
+    @property
+    def cubic(self):
+        """bool: Whether the box is a cube."""
+        return self.Lx == self.Ly and self.Ly == self.Lz and self.xy == self.yz\
+            and self.yz == self.xz and self.xz == 0
```
vyasr: You may want to use something like np.isclose instead of direct equality in case the boxes are coming from gsd or OVITO and have numerical imprecision introduced somewhere in the file reading/writing.

tommy-waltmann: yes, I meant to write myself a comment about this. Thanks for the reminder.

bdice: A few details to watch for:
- `allclose` is asymmetric, which has consequences:
  - 0 should be the first argument for the tilt factor check to ensure the `rtol` is applied to the potentially nonzero value (tilt factors)
  - it's probably best to check multiple directions for the box lengths, like ensuring that `Lx ~= Ly` and `Ly ~= Lx`. I used an array below to check both directions.
- box lengths are single-precision floats so NumPy needs a looser `atol`, like `1e-5` (instead of the default of `1e-8`) for the typical boxes we handle (order of 10^1 length units on each side)

```suggestion
    return (
        not self.is2d
        and np.allclose(
            [self.Lx, self.Lx, self.Ly, self.Ly, self.Lz, self.Lz],
            [self.Ly, self.Lz, self.Lx, self.Lz, self.Lx, self.Ly],
            rtol=1e-5,
            atol=1e-5,
        )
        and np.allclose(0, [self.xy, self.yz, self.xz], rtol=1e-5, atol=1e-5)
    )
```

bdice (approved):

doc/source/reference/credits.rst
```diff
@@ -329,6 +329,7 @@ Tommy Waltmann
 * Remove CI build configurations from CircleCI which were already covered by CIBuildWheel.
 * Change property names in ``StaticStructureFactorDebye`` class.
 * Reformat static structure factor tests.
+* ``DiffractionPattern`` now raises an error when used with non-cubic boxes.
```
vyasr, line 332: It's sort of odd that the names here are not fully qualified like the ones in the Changelog, but I guess that's following the existing pattern so no worries.

PR #966, opened 2022-05-19
Document replicating per-site properties of a lattice

### LINKED ISSUE #963: Support arbitrary attributes associated with basis positions in UnitCell
## Description

I propose adding arbitrary attributes based on `UnitCell.basis_positions` such as orientations, types, etc. For an API something like `h5py`'s or `xarray`'s `attrs`, I think would work well. We would then return the attributes properly replicated on a `generate_system` call. Below is an example of the proposed API.

```python
uc = freud.data.UnitCell.bcc()
uc.lattice_attrs.orientation = np.array([[1.0, 0.0, 0.0, 0.0], [6.12e-17, 7.07e-01, 7.07e-01, 0.0]])
box, position, extras = uc.generate_system(10)
print(extras["orientation"])
# [[1.0, 0.0, 0.0, 0.0] * 10 ** 3, [6.12e-17, 7.07e-01, 7.07e-01, 0.0] * 10 **3]
```

## Alternatives

* Since #942, supporting extra parameters has been made much simpler, and we could leave that there.
* We could add an argument to `generate_systems` that does the same thing.
* We could also accept extra basis position relative parameters as keyword arguments (i.e. `**kwargs`).

b-butler:
## Description
Documents how to replicate properties like orientation, type, etc. when using `freud.data.Unitcell.generate_system`.

## Motivation and Context
Resolves: #963

## How Has This Been Tested?
Unnecessary

## Types of changes

## Checklist:

### DIFF
--- a/freud/data.py
+++ b/freud/data.py
@@ -80,8 +80,8 @@ class UnitCell:
                     N = np.product(n_repeats)
                 indices = np.repeat(np.arange(len(uc.basis_positions)), N)
 
-        Below is an example of expanding basis position properties such as type
-        to a replicated lattice.
+        Below is an example of expanding basis position properties such as (in
+        this case, types) to a replicated lattice.
 
         Example::
 
--- a/freud/data.py
+++ b/freud/data.py
@@ -87,8 +87,9 @@ class UnitCell:
 
             >>> uc = freud.data.UnitCell.bcc()
             >>> n_repeats = (10, 5, 4)
-            >>> system = uc.generate_system(N)
-            >>> indices = np.repeat(np.arange(len(uc.basis_positions)), np.product(N))
+            >>> system = uc.generate_system(n_repeats)
+            >>> N = np.product(n_repeats)
+            >>> indices = np.repeat(np.arange(len(uc.basis_positions)), N)
             >>> # An array of types for all points
             >>> types = np.array([0, 1])[indices]
             >>> len(types)
--- a/freud/data.py
+++ b/freud/data.py
@@ -70,21 +70,29 @@ class UnitCell:
             Positions are generated in the order of the instance's
             ``basis_positions``. The first :math:`N_{replica}` positions come
             from the first basis position, the next :math:`N_{replica}` the
-            second, etc. A NumPy one-liner that generates the list of indices is
-            ``np.repeat(np.arange(len(uc.basis_positions)), np.product(N))``.
+            second, etc. To generate the indices use the following expression.
+
+            .. code-block:: python
+
+                if isinstance(n_repeats, int):
+                    N = n_repeats ** dimension
+                else:
+                    N = np.product(n_repeats)
+                indices = np.repeat(np.arange(len(uc.basis_positions)), N)
 
         Below is an example of expanding basis position properties such as type
         to a replicated lattice.
 
-        .. code-block:: python
+        Example::
 
-            uc = freud.data.UnitCell.bcc()
-            N = (10, 5, 4)
-            system = uc.generate_system(N)
-            indices = np.repeat(
-                np.arange(len(uc.basis_positions)), np.product(N))
-            # An array of types for all points
-            types = np.array([0, 1])[indices]
+            >>> uc = freud.data.UnitCell.bcc()
+            >>> n_repeats = (10, 5, 4)
+            >>> system = uc.generate_system(N)
+            >>> indices = np.repeat(np.arange(len(uc.basis_positions)), np.product(N))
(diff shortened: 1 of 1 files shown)


freud/data.py
```diff
@@ -68,9 +68,24 @@ def generate_system(self, num_replicas=1, scale=1, sigma_noise=0, seed=None):
...
-            :math:`N_{replica}` the second, etc.
+            ``basis_positions``. The first :math:`N_{replica}` positions come
+            from the first basis position, the next :math:`N_{replica}` the
+            second, etc. A NumPy one-liner that generates the list of indices is
+            ``np.repeat(np.arange(len(uc.basis_positions)), np.product(N))``.
+
+        Below is an example of expanding basis position properties such as type
+        to a replicated lattice.
+
+        .. code-block:: python
```
bdice: I'd write this as an example that will be recognized by the docs and will be doctested. Example of example writing: https://github.com/glotzerlab/freud/blob/3daeba88bb8e6e11973694a78f6eb652978b7f16/freud/box.pyx#L397-L406

freud/data.py
```diff
@@ -68,9 +68,24 @@ def generate_system(self, num_replicas=1, scale=1, sigma_noise=0, seed=None):
 
         Note:
             Positions are generated in the order of the instance's
-            ``basis_positions``. The first :math:`N_{replica}`
-            positions come from the first basis position, the next
-            :math:`N_{replica}` the second, etc.
+            ``basis_positions``. The first :math:`N_{replica}` positions come
+            from the first basis position, the next :math:`N_{replica}` the
+            second, etc. A NumPy one-liner that generates the list of indices is
+            ``np.repeat(np.arange(len(uc.basis_positions)), np.product(N))``.
```
bdice: It's not very clear how `N` is defined. It's close to `num_replicas` but if `num_replicas` is a scalar, you may need `num_replicas**3` in 3D and `num_replicas**2` in 2D. That kind of wrecks the "one liner" appeal but it is much more explicit about what is meant.

freud/data.py
```diff
@@ -68,9 +68,24 @@ def generate_system(self, num_replicas=1, scale=1, sigma_noise=0, seed=None):
...
+
+        Below is an example of expanding basis position properties such as type
+        to a replicated lattice.
+
+        .. code-block:: python
+
+            uc = freud.data.UnitCell.bcc()
+            N = (10, 5, 4)
```
bdice: I would call this `num_replicas` to align with the function parameter.

freud/data.py
b-butler: I forgot freud did doc testing.

freud/data.py
b-butler: Yeah, I forgot about the scalar case.

vyasr:
@b-butler FYI you can run the doctests locally (they're in `tests/test_doctests.py`) to make sure that the changes you're applying the example leave the sample code in a valid state.

freud/data.py
```diff
@@ -68,9 +68,33 @@ def generate_system(self, num_replicas=1, scale=1, sigma_noise=0, seed=None):
 
         Note:
             Positions are generated in the order of the instance's
-            ``basis_positions``. The first :math:`N_{replica}`
-            positions come from the first basis position, the next
-            :math:`N_{replica}` the second, etc.
+            ``basis_positions``. The first :math:`N_{replica}` positions come
+            from the first basis position, the next :math:`N_{replica}` the
```
vyasr, line 72: Since you're updating this note, it might be helpful to say explicitly that the behavior is analogous to `np.repeat` rather than `np.tile`.

freud/data.py
```diff
@@ -68,9 +68,33 @@ def generate_system(self, num_replicas=1, scale=1, sigma_noise=0, seed=None):
...
+            .. code-block:: python
+
+                if isinstance(n_repeats, int):
+                    N = n_repeats ** dimension
+                else:
+                    N = np.product(n_repeats)
+                indices = np.repeat(np.arange(len(uc.basis_positions)), N)
+
+        Below is an example of expanding basis position properties such as type
+        to a replicated lattice.
```
vyasr:
```suggestion
        Below is an example of expanding basis position properties such as (in
        this case, types) to a replicated lattice.
```

vyasr (approved):
Some minor suggestions but LGTM if @bdice is happy.

tommy-waltmann:
@bdice do you have any comments on this?

bdice (approved):
Looks good to me. I pushed one small fix to wording.

PR #975, opened 2022-06-06
Test/pytest mark parametrize

### LINKED ISSUE #814: Introduce `@pytest.mark.parametrize` to unit tests
Many of Freud's unit tests currently implement `for` loops to test various combinations of input parameters. For example, test_diffraction_DiffractionPattern.py currently includes:
```
for grid_size in (256, 1024):
    for output_size in (255, 256, 1023, 1024):
          for zoom in (1, 2.5, 4.123):
```
in the test for diffraction peaks in a simple cubic system.

With `@pytest.mark.parametrize`, the code above can refactor to
```
@pytest.mark.parametrize("grid_size, output_size, zoom", [
    (grid_size, output_size, zoom)
    for grid_size in (256, 1024)
    for output_size in (255, 256, 1023, 1024)
    for zoom in (1, 2.5, 4.123)
])
```
and be placed immediately before the simple cubic system test function. When run with pytest,
the printout will show a test for each combination of parameters, and whether is passes or not. This modification may also make tests easier to read.

This issue proposes similar changes to all of Freud's unit tests, replacing nested `for` loops with the appropriate `@pytest.mark.parameterize`expression.

Care must be taken to avoid changing tests such that they always pass.

kodyt:
## Description
Most of the unit tests do not utilize `@pytest.mark.parametrize`. These changes refactor many of the tests to use the parameterisation.

## Motivation and Context

Resolves: #814 

## How Has This Been Tested?
This is the test

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have added tests that cover my changes (if relevant).
- [X] All new and existing tests passed.
- [X] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst).


doc/source/reference/credits.rst
```diff
@@ -360,6 +360,10 @@ Dylan Marx
 
 * Contributed documentation for ``NeighborQuery`` class.
 
+Kody Takada
+
+* Contributed refactored unit tests
```
tommy-waltmann:
```suggestion
* Contributed refactored unit tests.
```

tests/test_data.py
```diff
@@ -33,55 +33,58 @@ def test_fcc(self):
             points, [[0, 0, -0.5], [0, -0.5, 0], [-0.5, 0, 0], [-0.5, -0.5, -0.5]]
         )
 
-    def test_scale(self):
+    @pytest.mark.parametrize("scale", [scale for scale in [0.5, 2]])
```
tommy-waltmann:
```suggestion
    @pytest.mark.parametrize("scale", [0.5, 2])
```

tests/test_data.py
```diff
@@ -33,55 +33,58 @@ def test_fcc(self):
...
+        box, points = freud.data.UnitCell.fcc().generate_system(scale=scale)
+        assert box == freud.box.Box.cube(scale)
+        npt.assert_array_equal(
+            points,
+            scale
+            * np.array([[0, 0, -0.5], [0, -0.5, 0], [-0.5, 0, 0], [-0.5, -0.5, -0.5]]),
+        )
+
+    @pytest.mark.parametrize(
+        "num_replicas", [num_replicas for num_replicas in range(1, 10)]
```
tommy-waltmann:
```suggestion
        "num_replicas", list(range(1, 10))]
```

tests/test_data.py
```diff
@@ -33,55 +33,58 @@ def test_fcc(self):
...
+        test_points = (test_points - (num_replicas * 0.5)).reshape(-1, 3)
+
+        npt.assert_allclose(
+            sort_rounded_xyz_array(points),
+            sort_rounded_xyz_array(box.wrap(test_points)),
+        )
+
+    @pytest.mark.parametrize(
+        "num_replicas",
+        [num_replicas for num_replicas in (0, 2.5, -1, [2, 2, 0], [2, 2, 2], "abc")],
```
tommy-waltmann:
```suggestion
        [0, 2.5, -1, [2, 2, 0], [2, 2, 2], "abc"],
```

tests/test_data.py
tommy-waltmann: Instead of doing a list comprehension (https://www.geeksforgeeks.org/python-list-comprehension/) every time to create the list, it is fine to just define the list with its elements in a more typical fashion.

tommy-waltmann: This comment applies to all such places in this PR.

tests/test_order_Hexatic.py
```diff
@@ -71,7 +71,10 @@ def test_compute(self):
 
             npt.assert_allclose(hop.particle_order[0], 1.0 + 0.0j, atol=1e-1)
 
-    def test_weighted_random(self):
+    k = ("k", [k for k in range(0, 12)])
+
+    @pytest.mark.parametrize(*k)
+    def test_weighted_random(self, k):
```
tommy-waltmann, line 75: Since `test_weighted_random` and `test_normalization` aren't related tests, I would like to see the k-values used written out for each test individually in the `pytest.mark.parametrize` decorator.

tommy-waltmann (dismissed):
This is a great start @kodyt, our unit tests are looking a lot cleaner. I have a few comments, and one of them applies to most of the unit tests which have been updated.

In addition, I think there are some more tests which could use parametrization, see:

1. `test_ql`, `test_ql_weighted`, and `test_wl` in `test_environment_LocalDescriptors`
2. `NeighborQueryTest::test_query_validate_points` in `test_locality_NeighborQuery` may be able to be split into two tests and have each test parametrized
3. `TestNeighborQueryAABB::test_r_guess_scale` in `test_locality_NeighborQuery`
4. `test_voronoi_neighbors_wrapped` in `test_locality_Voronoi`
5. `TestRotationalAutocorrelationReference` in `test_order_RotationalAutocorrelation`
6. `test_identical_environments` in `test_order_SolidLiquid`
7. `test_multiple_l` in `test_order_Steinhardt` mayb e parametrizable, but I'll leave that up to you since the parametrization may be complicated

tests/test_data.py
kodyt: @tommy-waltmann I have updated all of the tests that you mentioned, but for `test_wl` in `test_environment_LocalDescriptors.py` becomes much slower when parameterised and ran. Is it worth refactoring this test? I included the parameterised version in the latest commit

tommy-waltmann: My intent was that all the tests in `test_environment_LocalDescriptors` be parametrized on the `struct_func`, not on the `L` value.

kodyt: My bad that would make more sense, I have fixed it now.

tests/test_data.py
```diff
@@ -127,14 +128,15 @@ def test_seed(self):
...
-        for N in (0, 1, 10, 100, 1000):
-            for is2D in (True, False):
-                box, points = freud.data.make_random_system(
-                    box_size=10, num_points=N, is2D=is2D
-                )
-                assert points.shape == (N, 3)
-                assert box.is2D == is2D
+    @pytest.mark.parametrize(
+        "N, is2D", [(N, is2D) for N in (0, 1, 10, 100, 1000) for is2D in (True, False)]
+    )
```
bdice: Separate these parametrizations since they are orthogonal:
```suggestion
    @pytest.mark.parametrize("N", (0, 1, 10, 100, 1000))
    @pytest.mark.parametrize("is2D", (True, False))
```

tests/test_density_CorrelationFunction.py
```diff
@@ -455,13 +456,12 @@ def value_func(_r):
         for nq, neighbors in test_set:
```
bdice, line 465: In this case, parametrizing over `rv` will require the test to re-generate the test set and compute the `supposed_correlation` multiple times. I think that might be a nontrivial computation and would potentially increase test time. It may be possible to use a fixture that is cached to avoid recomputing that data?

tests/test_density_GaussianDensity.py
```diff
@@ -78,61 +78,64 @@ def test_change_box_dimension(self):
         with pytest.raises(ValueError):
             gd.compute((test_box, test_points))
 
-    def test_sum_2d(self):
+    num_points = ("num_points", [1, 10, 100])
+
+    @pytest.mark.parametrize(*num_points)
+    def test_sum_2d(self, num_points):
```
bdice, line 82: It's a little strange to shadow the previously declared `num_points` with the function parameter `num_points`. I would try to choose distinct names here to avoid confusion.

tests/test_density_RDF.py
```diff
@@ -13,25 +13,23 @@
 
 
 class TestRDF:
-    def test_generateR(self):
+    @pytest.mark.parametrize("r_min", [0, 0.05, 0.1, 1.0, 3.0])
+    def test_generateR(self, r_min):
```
bdice: It might make sense to rename this test to `test_generate_bins`.
```suggestion
    def test_generate_bins(self, r_min):
```

tests/test_density_RDF.py
```diff
@@ -74,44 +72,44 @@ def test_invalid_rdf(self):
...
+        box, points = freud.data.make_random_system(box_size, num_points)
+        # This test is slow, and since it's a validation of the underlying
+        # algorithm and not the API we don't need to test all possible
+        # inputs, so we only test the fastest one (AABBQuery).
+        nq = freud.locality.AABBQuery(box, points)
+        neighbors = {"mode": "ball", "r_max": r_max, "exclude_ii": True}
+
+        rdf = freud.density.RDF(bins, r_max, r_min)
+
+        if r_min != 3.0:
```
bdice: Why do we set `reset=False` explicitly for values besides 3.0? This is a special case of the parameterization but we don't have a comment explaining why this special case is used. This needs a comment -- or perhaps separated into another test, if it's covering something a case of the RDF behavior that is sufficiently unique.

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -88,68 +88,75 @@ def test_attribute_shapes(self):
...
+        "grid_size, output_size, view_orientation",
+        [
+            (grid_size, output_size, view_orientation)
+            for grid_size in (255, 256)
+            for output_size in (255, 256)
+            for view_orientation in rowan.random.rand(10)
+        ],
+    )
+
+    @pytest.mark.parametrize(*vals)
```
bdice: `vals` is only used a couple of times. I'd rather copy-paste the parameterizations and keep them close to the test definitions.

```suggestion
    @pytest.mark.parametrize("grid_size", (255, 256))
    @pytest.mark.parametrize("output_size", (255, 256))
    @pytest.mark.parametrize("view_orientation", rowan.random.rand(10))
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -88,68 +88,75 @@ def test_attribute_shapes(self):
...
+        diff = dp.diffraction
+        max_index = np.unravel_index(np.argmax(diff), diff.shape)
+        center_index = (output_size // 2, output_size // 2)
+        assert max_index == center_index
+
+        # The value at k=0 should be N_points because of
+        # normalization by the number of points
+        npt.assert_allclose(dp.diffraction[center_index], len(positions))
+
+    @pytest.mark.parametrize(*vals)
```
bdice:
```suggestion
    @pytest.mark.parametrize("grid_size", (255, 256))
    @pytest.mark.parametrize("output_size", (255, 256))
    @pytest.mark.parametrize("view_orientation", rowan.random.rand(10))
```

tests/test_diffraction_DiffractionPattern.py
bdice: Better yet, we could define `test_center_unordered` and `test_center_ordered` together by parametrizing the system. Then you don't even need to copy-paste the parametrizations because they're only defined once. 😉
```python
    @pytest.mark.parametrize("box, positions", (
        freud.data.make_random_system(box_size=10, num_points=1000),
        freud.data.UnitCell.bcc().generate_system(10),
    ))

File: tests/test_diffraction_DiffractionPattern.py [hunk 13]
```diff
@@ -159,76 +166,83 @@ def test_repr(self):
...
+
+    @pytest.mark.parametrize(
+        "grid_size, output_size, zoom",
+        [
+            (grid_size, output_size, zoom)
+            for grid_size in (256, 1024)
+            for output_size in (255, 256, 1023, 1024)
+            for zoom in (1, 2.5, 4.123)
+        ],
+    )
```
bdice:
```suggestion
    @pytest.mark.parametrize("grid_size", (256, 1024))
    @pytest.mark.parametrize("output_size", (255, 256, 1023, 1024))
    @pytest.mark.parametrize("zoom", (1, 2.5, 4.123))
```

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -159,76 +166,83 @@ def test_repr(self):
...
         length = 1
         box, positions = freud.data.UnitCell.sc().generate_system(
             num_replicas=16, scale=length, sigma_noise=0.1 * length
         )
         # Pick a non-integer value for zoom, to ensure that peaks besides k=0
         # are not perfectly aligned on pixels.
-        dp = freud.diffraction.DiffractionPattern(grid_size=512)
-        dp.compute((box, positions), zoom=4.123)
+        dp = freud.diffraction.DiffractionPattern(
+            grid_size=grid_size, output_size=output_size
```
bdice: One of the cases that is no longer explicitly covered here is the _default_ behavior of `output_size` when no value is provided. I don't recall exactly how that works but I would like to ensure that case is still tested. I remember specifically designing this test to ensure that case of default `output_size` was covered.

tests/test_environment_LocalDescriptors.py
```diff
@@ -199,7 +199,17 @@ def test_repr(self):
         comp = freud.environment.LocalDescriptors(8, True)
         assert str(comp) == str(eval(repr(comp)))
 
-    def test_ql(self):
+    struct_func = [
+        "struct_func",
```
bdice: A better name than `struct_func` might be `unit_cell`.

tests/test_locality_NeighborQuery.py
```diff
@@ -39,20 +39,23 @@ def build_query_object(cls, box, ref_points, r_max=None):
...
+            np.zeros(0, dtype=np.float32),
+            np.zeros(shape=(0, 3), dtype=np.float32),
+        ],
+    )
+    def test_query_validate_points(self, empty_points):
         L = 10  # Box Dimensions
         r_max = 2.01  # Cutoff radius
         box = freud.box.Box.cube(L)
 
         # It's not allowed to have an empty NeighborQuery
```
bdice: Only part of this test is parametrized. But the rest of the test does not depend on the values of `empty_points` and is now being tested redundantly. Either separate the parametrized portion of the test into a new test named `test_query_empty_points`, or revert the changes and keep the inner `for` loop rather than parametrizing it. I vote for breaking up the test, because this is really testing multiple kinds of behavior with a single test name.

bdice (commented):
Nice work! I had a handful of suggestions that may be helpful for writing cleaner / more concise tests. I only looked over about half of the PR. I am not planning to make another review pass unless specifically requested, but I was curious to see how this looked and I support this refactoring.

tests/test_locality_NeighborQuery.py
tommy-waltmann: I also vote for 2 tests, one which tests a set of valid inputs, and another test which tests a set of invalid inputs.

tests/test_density_RDF.py
kodyt: I believe the reason for `reset` being set to `False` explicitly is just to test that `compute` will work when `reset` is called as an argument, and the 3.0 value was chosen arbitrarily. I added a comment to explain this.

tests/test_density_GaussianDensity.py
```diff
@@ -78,61 +78,62 @@ def test_change_box_dimension(self):
...
         sigma = 2
         box_size = width
         gd = freud.density.GaussianDensity(width, r_max, sigma)
-        for num_points in [1, 10, 100]:
-            box, points = freud.data.make_random_system(box_size, num_points, is2D=True)
-            gd.compute(system=(box, points))
-            # This has discretization error as well as single-precision error
-            assert np.isclose(np.sum(gd.density), num_points, rtol=1e-4)
+        box, points = freud.data.make_random_system(box_size, num_points, is2D=True)
+        gd.compute(system=(box, points))
```
tommy-waltmann, line 90: I've noticed that since parametrizing these tests we are actually no longer testing that a `GaussianDensity` can call `compute` multiple times with different data. Can a small test be added which ensures this works?

tests/test_diffraction_DiffractionPattern.py
```diff
@@ -88,68 +88,47 @@ def test_attribute_shapes(self):
...
+        (
+            freud.data.make_random_system(box_size=10, num_points=1000),
+            freud.data.UnitCell.bcc().generate_system(10),
+        ),
+    )
+    @pytest.mark.parametrize("grid_size", [255, 256])
+    @pytest.mark.parametrize("output_size", [255, 256])
+    @pytest.mark.parametrize(
+        "view_orientation",
+        [view_orientation for view_orientation in rowan.random.rand(10)],
```
tommy-waltmann:
```suggestion
        rowan.random.rand(10),
```

tests/test_locality_NeighborQuery.py
tommy-waltmann: This comment still needs to be addressed.

tests/test_locality_NeighborQuery.py
```diff
@@ -512,39 +516,43 @@ def test_corner_2d(self):
...
-                    "Wrong nlist length for N = {}, " "num_neighbors = {}, length = {}"
-                ).format(N, k, len(nlist))
-                nlist_array = nlist[:]
-                for i in range(N):
-                    assert ([i, i] == nlist_array).all(axis=1).any()
+        positions = box.wrap(L / 2 * np.random.rand(N, 3))
+        """ks = [1, 5]
+        if N > 10:
+            ks.extend([10, 50])
+        for k in ks:"""
```
tommy-waltmann: Let's remove this.
```suggestion
```

tests/test_order_SolidLiquid.py
```diff
@@ -35,21 +35,27 @@ def test_nlist(self):
 
         npt.assert_array_equal(nlist[:], comp.nlist[:])
 
-    def test_identical_environments(self):
+    comp_default = freud.order.SolidLiquid(6, q_threshold=0.7, solid_threshold=6)
+    comp_no_norm = freud.order.SolidLiquid(
+        6, q_threshold=0.3, solid_threshold=6, normalize_q=False
+    )
```
tommy-waltmann: Let's put these inside object construction calls inside the pytest.mark.parametrize block. I don't want random global variables floating around the class.

tests/test_order_SolidLiquid.py
```diff
@@ -35,21 +35,27 @@ def test_nlist(self):
...
+
+    @pytest.mark.parametrize(
+        "comp, query_args",
+        [
+            (comp, query_args)
+            for comp in (comp_default, comp_no_norm)
+            for query_args in (dict(r_max=2.0), dict(num_neighbors=12))
+        ],
+    )
+    def test_identical_environments(self, comp, query_args):
```
tommy-waltmann, line 51: This test is another case like `GaussianDensity` where by adding the parametrization we have implicitly lost test coverage for calling compute multiple times in a row. Please add a small test which covers this case.

tommy-waltmann (changes requested):
This PR looks nicer every time I review. I have a few more comments and suggestions and then this should be ready to go.

tests/test_density_CorrelationFunction.py
```diff
@@ -445,6 +442,19 @@ def value_func(_r):
...
+    def test_points_ne_query_points_real(self, rv, correlation_calc):
+        r_max = 10.0
+        bins = 100
+        dr = r_max / bins
+        box_size = r_max * 5
+        box = freud.box.Box.square(box_size)
+
+        ocf = freud.density.CorrelationFunction(bins, r_max)
+
+        query_points, query_values, supposed_correlation = correlation_calc
```
bdice: There are a few places where variables have names like "supposed" and "correct" -- I tend to call the expectation values "expected" and the value computed by the test "actual." This is in line with pytest's usual nomenclature (search the docs and you'll see examples of "actual, expected"). See also: https://blog.8-p.info/en/2020/02/02/expected-actual/

As a reminder related to that naming convention, [tests with `allclose` or relative tolerances are asymmetric](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html) between `expected` and `actual`. Always call `numpy.allclose(actual, expected)`.

tests/test_locality_NeighborQuery.py
```diff
@@ -39,34 +39,34 @@ def build_query_object(cls, box, ref_points, r_max=None):
...
+        r_max = 2.01  # Cutoff radius
+        box = freud.box.Box.cube(L)
 
+        # It's not allowed to have an empty NeighborQuery
         # It's not allowed to have one point as a 1D array
-        with pytest.raises(ValueError):
-            points = np.zeros(shape=(3), dtype=np.float32)
-            self.build_query_object(box, points, r_max)
-
         # It's not allowed to have an array without shape (N, 3)
```
bdice: These code comments should be in the parametrize call, next to the parameters that they describe.

tests/test_locality_NeighborQuery.py
```diff
@@ -375,87 +375,83 @@ def test_exclude_ii(self):
 
         assert ij1 == ij2
 
-    def test_exhaustive_search(self):
+    i = ("i", list(range(10)))
+
+    @pytest.mark.parametrize(*i)
+    def test_exhaustive_search(self, i):
```
bdice: Seeing a variable `i` like this is a red flag. The convention of names like `i` is to represent a single integer, and usually only defined in the scope of a loop. This `i` is a list of integers, and any code in this class that uses the variable `i` will implicitly shadow the `i` defined here, which isn't great.

Upon further investigation, I see that `i` is only used to enumerate the test's random seeds. I suggest something like:
```suggestion
    @pytest.mark.parametrize("seed", range(1, 10))
    def test_exhaustive_search(self, seed):
```
and then remove `seed = 0` and replace `seed=seed + i` with `seed=seed` below.

tests/test_locality_NeighborQuery.py
```diff
@@ -375,87 +375,83 @@ def test_exclude_ii(self):
...
+        result = list(nq.query(points, dict(mode="ball", r_max=r_max, exclude_ii=True)))
+        ijs = {(x[1], x[0]) for x in result}
+        counts = Counter([x[1] for x in result])
+        counts_list = [counts[j] for j in range(N)]
+
+        try:
+            assert exhaustive_ijs == ijs
+        except AssertionError:
+            print(f"Failed neighbors, random seed: {seed} (i={i})")
+            raise
```
bdice: I'm pretty sure this `try/except AssertionError` is a relic of code that existed before converting to pytest. Pytest will give us all the debugging info we need (test name and parameters) so we can make this much simpler.
```suggestion
        assert exhaustive_ijs == ijs
```

tests/test_locality_NeighborQuery.py
```diff
@@ -375,87 +375,83 @@ def test_exclude_ii(self):
...
+
+        try:
+            assert exhaustive_counts_list == counts_list
+        except AssertionError:
+            print(f"Failed neighbor counts, random seed: {seed} (i={i})")
+            raise
```
bdice:
```suggestion
        assert exhaustive_counts_list == counts_list
```

tests/test_locality_NeighborQuery.py
```diff
@@ -375,87 +375,83 @@ def test_exclude_ii(self):
...
+
+    @pytest.mark.parametrize(*i)
+    def test_exhaustive_search_asymmetric(self, i):
```
bdice: As above, parametrize the seeds instead of `i`.

tests/test_locality_NeighborQuery.py
```diff
@@ -375,87 +375,83 @@ def test_exclude_ii(self):
...
-            except AssertionError:
-                print(f"Failed neighbor counts, random seed: {seed} (i={i})")
-                raise
+        nq = self.build_query_object(box, points2, r_max)
+        result = list(nq.query(points, dict(mode="ball", r_max=r_max)))
+        ijs = {(x[0], x[1]) for x in result}
+        counts = Counter([x[0] for x in result])
+        counts_list = [counts[j] for j in range(N)]
+
+        try:
```
bdice: Remove `try/except` as above. You can probably remove a lot of cases like this where the code says `except AssertionError`. The output from pytest should be sufficient for debugging and we don't need to carry the extra baggage of fancy errors when most of that info will be in the pytest report.

tests/test_density_RDF.py
bdice: Testing `reset` behavior should be done in a separate test function.

tests/test_locality_NeighborQuery.py
```diff
@@ -512,39 +508,40 @@ def test_corner_2d(self):
...
+        positions = box.wrap(L / 2 * np.random.rand(N, 3))
+
+        nq = self.build_query_object(box, positions, L / 10)
+
+        nlist = nq.query(
+            positions, dict(num_neighbors=k, exclude_ii=True)
+        ).toNeighborList()
+        assert len(nlist) == k * N, (
+            "Wrong nlist length for N = {}," "num_neighbors = {}, length = {}"
+        ).format(N, k, len(nlist))
```
bdice: This falls into the same category of assertion error messages that are no longer needed when using `parametrize`. Same for a few other assertions in this test function and perhaps elsewhere.

```suggestion
        assert len(nlist) == k * N
```

tests/test_order_Hexatic.py
```diff
@@ -36,20 +36,20 @@ def test_compute_random(self):
         hop.compute((box, points))
         npt.assert_allclose(np.mean(hop.particle_order), 0.0 + 0.0j, atol=1e-1)
 
-    def test_compute(self):
+    @pytest.mark.parametrize("i", range(6))
+    def test_compute(self, i):
```
bdice: ⚠️ This is not a parameterized test. This is a single test that constructs a single set of points using a `for` loop. All the points are intended to be in a single array. Please revert the changes in this test function.

tests/test_order_Hexatic.py
```diff
@@ -88,34 +89,32 @@ def test_weighted_random(self):
...
-            )
-            psi_k_weighted, _ = np.histogram(
-                voro.nlist.query_point_indices,
-                bins=len(points),
-                weights=voro.nlist.weights * np.exp(thetas * k * 1.0j),
-            )
-            psi_k_weighted /= total_weights
-
-            npt.assert_allclose(hop.particle_order, psi_k_weighted, atol=1e-5)
+        # Ensure that \psi'_k is between 0 and 1
```
bdice, line 82: The tests for `\psi'_0` and `\psi'_1` just above this comment could be separated into their own test functions rather than repeating those checks for all values of `k` (`k` is not used in those tests).

tests/test_order_Steinhardt.py
```diff
@@ -71,23 +71,29 @@ def test_qlmi(self, sph_l):
...
-            npt.assert_allclose(comp.particle_order, [1, 0, 1], atol=1e-5)
+        comp = freud.order.Steinhardt(odd_l)
+        comp.compute((box, positions), neighbors={"num_neighbors": 2})
+        npt.assert_allclose(comp.particle_order, [1, 0, 1], atol=1e-5)
+
+    @pytest.mark.parametrize("even_l", range(0, 20, 2))
+    def test_l_axis_aligned_even(self, even_l):
+        # This test has three points along the z-axis. By construction,
+        # all three points should
+        # have perfect order for even l.
```
bdice:
```suggestion
        # all three points should have perfect order for even l.
```

tests/test_order_Steinhardt.py
```diff
@@ -278,7 +282,8 @@ def test_compute_twice_norm(self):
 
         npt.assert_array_almost_equal(first_result, second_result)
 
-    def test_rotational_invariance(self):
+    @pytest.mark.parametrize("i", range(10))
+    def test_rotational_invariance(self, i):
```
bdice: Let's use the name `seed` instead of `i`. It isn't clear how `i` is used, but parametrizing over multiple seeds is easier to understand.
```suggestion
    @pytest.mark.parametrize("seed", range(10))
    def test_rotational_invariance(self, seed):
```

bdice (changes requested):
I have a handful of comments that I think should be addressed before merging. Some tests were incorrectly parametrized, others could benefit from simplification or clearer names.

tests/test_locality_NeighborQuery.py
```diff
@@ -39,33 +39,33 @@ def build_query_object(cls, box, ref_points, r_max=None):
...
             self.build_query_object(box, points, r_max)
 
-        # It's not allowed to have an array without shape (N, 3)
-        with pytest.raises(ValueError):
-            points = np.zeros(shape=(1, 2), dtype=np.float32)
-            self.build_query_object(box, points, r_max)
-        with pytest.raises(ValueError):
-            points = np.zeros(shape=(1, 4), dtype=np.float32)
-            self.build_query_object(box, points, r_max)
+    def test_query_valid_points(self):
```
bdice: This test could be parametrized on input data, like the invalid cases.

tests/test_locality_NeighborQuery.py
```diff
@@ -512,39 +486,36 @@ def test_corner_2d(self):
         result = list(nq.query(positions[[0]], dict(mode="nearest", num_neighbors=3)))
         assert get_point_neighbors(result, 0) == {0, 1, 2}
 
-    def test_random_system_query(self):
+    inputs = {10: [1, 5], 100: [1, 5, 10, 50], 500: [1, 5, 10, 50]}
+
+    @pytest.mark.parametrize(
+        "N, k", [(N, k) for N, value in inputs.items() for k in value]
+    )
```
bdice:
```suggestion
    @pytest.mark.parametrize(
        "N, k", [(N, k) for N in (10, 100, 500) for k in (1, 5, 10, 50) if k < N]
    )
```

bdice (approved):

tests/test_order_SolidLiquid.py
```diff
@@ -35,21 +35,39 @@ def test_nlist(self):
...
+        npt.assert_array_equal(comp.num_connections, 12)
 
-        for comp in (comp_default, comp_no_norm):
-            for query_args in (dict(r_max=2.0), dict(num_neighbors=12)):
-                comp.compute((box, positions), neighbors=query_args)
-                assert comp.largest_cluster_size == len(positions)
-                assert len(comp.cluster_sizes) == 1
-                assert comp.cluster_sizes[0] == len(positions)
-                npt.assert_array_equal(comp.num_connections, 12)
+    def test_identical_environments_compute(self):
```
tommy-waltmann:
```suggestion
    def test_multiple_compute(self):
```

tommy-waltmann (changes requested):
This looks great! I just have one name for a test I'd like to see changed to be more in line with what its testing,. Then make sure this branch is up to date with master, and this PR will be ready to merge.

bdice:
Something unexpected just happened in this branch. There are now 200 commits and diffs from unrelated PRs?

tommy-waltmann:
I think a rebase was done here instead of a merge to get the newest commits from master onto this branch. I don't know why the changes from other PRs are showing in the diff, as they are also on master.

tommy-waltmann:
Alright, I think that should resolve it.

tommy-waltmann (approved):
LGTM!

PR #980, opened 2022-06-27
Improve the documentation of Environment module

### LINKED ISSUE #977: Cite Erin Teich's "identity crisis" paper for environment matching
I believe `EnvironmentMotifMatch`, the ~artist~ algorithm formerly known as `MatchEnv`, was first developed and implemented for [Erin's 2019 "identity crisis"](https://www.nature.com/articles/s41467-018-07977-2) paper. If that's the case, we should add a note in the docs that users should cite that paper if they use that method. If that's not the case, there is probably something else we should advise users to cite.

I could certainly be wrong about this, so maybe those who know the history of the code better than me (@vyasr , @bdice , @joaander , @harperic ) can offer some input.

Charlottez112:
## Description
Include discussion from #979 and #278, and add citation to Erin's paper (resolves #977).

I deleted some of the old docs because when I first tried using `Environment`, I found them confusing. I also added some warnings to warn users of the potential pitfalls based on our recent discussion. 

## Types of changes

## Checklist:

### DIFF
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -678,7 +678,7 @@ cdef class EnvironmentCluster(_MatchEnv):
 
 cdef class EnvironmentMotifMatch(_MatchEnv):
     r"""Find matches between local arrangements of a set of points and
-    a provided motif :cite:`Teich2019`.
+    a provided motif, as done in :cite:`Teich2019`.
     """  # noqa: E501
 
     cdef freud._environment.EnvironmentMotifMatch * thisptr
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -715,8 +715,8 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None). Defines the environment of the query
-                particles.
+                Defines the environment of the query particles.
+                (Default value: None). 
             registration (bool, optional):
                 If True, first use brute force registration to orient one set
                 of environment vectors with respect to the other set such that
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -550,7 +550,7 @@ cdef class EnvironmentCluster(_MatchEnv):
         for every pair of matched points between the sets, the distance between
         the matched pair is less than :code:`threshold`.
 
-        When :code:`registration=False`, environments are not rotated prior to comparison 
+        When :code:`registration=False`, environments are not rotated prior to comparison
         between them. Which pairs of vectors are compared depends on the order in
         which the vectors of the environment are traversed.
 
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -535,14 +535,15 @@ cdef class EnvironmentCluster(_MatchEnv):
         Using a distance cutoff for :code:`'env_neighbors'` would
         lead to situations where the :code:`'cluster_environments'`
         contain different number of particles.
-        ..warning::
+
+        .. warning::
 
             All vectors of :code:`'cluster_environments'` are defined with
-            respect to the origin. O vectors are only used to pad the cluster
+            respect to the origin. Zero vectors are only used to pad the cluster
             vectors so that they have the same shape. In a future version of
-            freud, 0-padding will be removed.
+            freud, zero-padding will be removed.
 
-        ..warning::
+        .. warning::
             Comparison between two sets of environments is only made when both
             environments contain same number of particles.
 
@@ -574,17 +575,16 @@ cdef class EnvironmentCluster(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None). Defines the particle neighborhoods to be
-                compared.
+                (Default value: None). Defines the particle neighborhoods for
+                comparing environments.
             env_neighbors (:class:`freud.locality.NeighborList` or dict, optional):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
-                `query arguments
+                `query arguments.
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                This argument is used to define the neighbors of the
-                environment that motifs are registered against. If ``None``,
-                the value provided for ``neighbors`` will be used. (Default
-                value: None).
+                This argument is used to define the environments of each particle.
+                If ``None``, the value provided for ``neighbors`` will be used.
+                (Default value: None).
             registration (bool, optional):
                 If True, first use brute force registration to orient one set
                 of environment vectors with respect to the other set such that
@@ -688,9 +688,10 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
         r"""Determine clusters of particles that match the motif provided by
         motif.
 
-        ..warning::
+        .. warning::
             Comparison between two sets of environments is only made
             when both environments contain same number of particles.
+
         Args:
             system:
                 Any object that is a valid argument to
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -562,7 +562,7 @@ cdef class EnvironmentCluster(_MatchEnv):
 
         .. warning::
 
-            Comparison between two sets of environments is only made when both
+            Comparisons between two environments are only made when both
             environments contain the same number of particles.
 
         Example::
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -544,22 +544,32 @@ cdef class EnvironmentCluster(_MatchEnv):
 
         Two environments are compared using `point-set registration
         <https://en.wikipedia.org/wiki/Point-set_registration)>`_.
-        Specifically, we use the `Kabsch algorithm
+
+        The thresholding criterion we apply in order to determine if the two point sets
+        match is quite conservative: the two point sets match if and only if,
+        for every pair of matched points between the sets, the distance between
+        the matched pair is less than :code:`threshold`.
+
+        When :code:`registration=False`, environments are not rotated prior to comparison 
+        between them. Which pairs of vectors are compared depends on the order in
+        which the vectors of the environment are traversed.
+
+        When :code:`registration=True`, we use the `Kabsch algorithm
         <https://en.wikipedia.org/wiki/Kabsch_algorithm>`_ to find the optimal
         rotation matrix mapping one environment onto another. The Kabsch
         algorithm assumes a one-to-one correspondence between points in the two
         environments. However, we typically do not know which point in one environment
         corresponds to a particular point in the other environment. The brute force
         solution is to try all possible correspondences, but the resulting
-        combinatorial explosion makes this problem infeasible, so we use a tolerance
-        :code:`threshold` to terminate the search when we have found a permutation
+        combinatorial explosion makes this problem infeasible, so we use the thresholding
+        criterion above to terminate the search when we have found a permutation
         of points that results in a sufficiently low RMSD.
 
         .. note::
 
             Using a distance cutoff for :code:`env_neighbors` could
             lead to situations where the :code:`cluster_environments`
-            contain different numbers of particles. In this case, the
+            contain different numbers of neighbors. In this case, the
             environments which have a number of neighbors less than
             the environment with the maximum number of neighbors
             :math:`k_{max}` will have their entry in :code:`cluster_environments`
@@ -579,9 +589,9 @@ cdef class EnvironmentCluster(_MatchEnv):
         .. warning::
 
             Comparisons between two environments are only made when both
-            environments contain the same number of particles.
-            However, this is not currently enforced, so no warning will be given
-            at runtime if such mismatched environments are provided.
+            environments contain the same number of neighbors.
+            However, no warning will be given at runtime if mismatched
+            environments are provided for comparison.
 
         Example::
 
@@ -718,11 +728,10 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
     a provided motif, as done in :cite:`Teich2019`.
 
     A particle's environment can only match the motif if it contains the
-    same number of particles as the motif. Any environment with a
-    different number of particles than the motif will always fail to match
+    same number of neighbors as the motif. Any environment with a
+    different number of neighbors than the motif will always fail to match
     the motif. See :class:`freud.environment.EnvironmentCluster.compute` for
-    the algorithm we use.
-
+    the matching criterion.
     """  # noqa: E501
 
     cdef freud._environment.EnvironmentMotifMatch * thisptr
@@ -742,9 +751,9 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
         .. warning::
 
             Comparisons between two environments are only made when both
-            environments contain the same number of particles.
-            However, this is not currently enforced, so no warning will be given
-            at runtime if such mismatched environments are provided.
+            environments contain the same number of neighbors.
+            However, no warning will be given at runtime if mismatched
+            environments are provided for comparison.
 
         Args:
             system:
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -536,9 +536,9 @@ cdef class EnvironmentCluster(_MatchEnv):
         neighbors as defined by :code:`env_neighbors`.
         For example, :code:`env_neighbors= {'num_neighbors': 8}` means that every
         particle's local environment is defined by its 8 nearest neighbors.
-        Then, each particle's environment is compared to the environments of a
-        selected list of other particles in the system, controlled by the
-        :code:`neighbors` parameter. For example, :code:`neighbors={'r_max': 3.0}`
+        Then, each particle's environment is compared to the environments of
+        particles that satisfy a different cutoff parameter :code:`neighbors`.
+        For example, :code:`neighbors={'r_max': 3.0}`
         means that the environment of each particle will be compared to the
         environment of every particle within a distance of 3.0.
 
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -550,7 +550,7 @@ cdef class EnvironmentCluster(_MatchEnv):
         the environment with the maximum number of neighbors
         :math:`k_{max}` will have their entry in `cluster_environments`
         padded with zero vectors. For example, a cluster environment
-        with :math:`m < k`neighbors, will have :math:`k - m` zero
+        with :math:`m < k` neighbors, will have :math:`k - m` zero
         vectors at the end of its entry in `cluster_environments`.
 
         .. warning::
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -706,8 +706,9 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
         r"""Determine which particles' local environment matches given environment motif.
 
         .. warning::
-            Comparison between two sets of environments is only made
-            when both environments contain the same number of particles.
+
+            Comparisons between two environments are only made when both
+            environments contain the same number of particles.
 
         Args:
             system:
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -582,8 +582,8 @@ cdef class EnvironmentCluster(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None). Defines the particle neighborhoods for
-                comparing environments.
+                Defines the particle neighborhoods for comparing
+                environments. (Default value: None). 
             env_neighbors (:class:`freud.locality.NeighborList` or dict, optional):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -556,9 +556,9 @@ cdef class EnvironmentCluster(_MatchEnv):
         .. warning::
 
             All vectors of :code:`'cluster_environments'` are defined with
-            respect to the origin. Zero vectors are only used to pad the cluster
-            vectors so that they have the same shape. In a future version of
-            freud, zero-padding will be removed.
+            respect to the query particle. Zero vectors are only used to pad
+            the cluster vectors so that they have the same shape.
+            In a future version of freud, zero-padding will be removed.
 
         .. warning::
 
@@ -592,16 +592,16 @@ cdef class EnvironmentCluster(_MatchEnv):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
                 Defines the particle neighborhoods for comparing
-                environments. (Default value: None).
+                environments (Default value: None).
             env_neighbors (:class:`freud.locality.NeighborList` or dict, optional):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
-                `query arguments.
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
+                `query arguments
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
                 This argument is used to define the environments of each particle.
-                If ``None``, the value provided for ``neighbors`` will be used.
+                If ``None``, the value provided for ``neighbors`` will be used
                 (Default value: None).
             registration (bool, optional):
                 If True, first use brute force registration to orient one set
@@ -708,7 +708,7 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
 
         .. warning::
             Comparison between two sets of environments is only made
-            when both environments contain same number of particles.
+            when both environments contain the same number of particles.
 
         Args:
             system:
@@ -725,8 +725,8 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                Defines the environment of the query particles.
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
+                Defines the environment of the query particles
                 (Default value: None).
             registration (bool, optional):
                 If True, first use brute force registration to orient one set
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -593,7 +593,7 @@ cdef class EnvironmentCluster(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
-                Defines the particle neighborhoods for comparing
+                Defines the neighbors used for comparing different particle
                 environments (Default value: None).
             env_neighbors (:class:`freud.locality.NeighborList` or dict, optional):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -703,8 +703,7 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
 
     def compute(self, system, motif, threshold, neighbors=None,
                 registration=False):
-        r"""Determine clusters of particles that match the motif provided by
-        motif.
+        r"""Determine which particles' local environment matches given environment motif.
 
         .. warning::
             Comparison between two sets of environments is only made
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -512,7 +512,7 @@ cdef class _MatchEnv(_PairCompute):
 
 cdef class EnvironmentCluster(_MatchEnv):
     r"""Clusters particles according to whether their local environments match
-    or not, according to various shape matching metrics.
+    or not, according to various shape matching metrics :cite:`Teich2019`.
     """
 
     cdef freud._environment.EnvironmentCluster * thisptr
@@ -532,11 +532,19 @@ cdef class EnvironmentCluster(_MatchEnv):
                 global_search=False):
         r"""Determine clusters of particles with matching environments.
 
-        In general, it is recommended to specify a number of neighbors rather
-        than just a distance cutoff as part of your neighbor querying when
-        performing this computation. Using a distance cutoff alone could easily
-        lead to situations where a point doesn't match a cluster because a
-        required neighbor is just outside the cutoff.
+        Using a distance cutoff for :code:`'env_neighbors'` would
+        lead to situations where the :code:`'cluster_environments'`
+        contain different number of particles.
+        ..warning::
+
+            All vectors of :code:`'cluster_environments'` are defined with
+            respect to the origin. O vectors are only used to pad the cluster
+            vectors so that they have the same shape. In a future version of
+            freud, 0-padding will be removed.
+
+        ..warning::
+            Comparison between two sets of environments is only made when both
+            environments contain same number of particles.
 
         Example::
 
@@ -566,7 +574,8 @@ cdef class EnvironmentCluster(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None).
+                (Default value: None). Defines the particle neighborhoods to be
+                compared.
             env_neighbors (:class:`freud.locality.NeighborList` or dict, optional):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
@@ -583,8 +592,10 @@ cdef class EnvironmentCluster(_MatchEnv):
                 option incurs a significant performance penalty.
                 (Default value = :code:`False`)
             global_search (bool, optional):
-                 If True, do an exhaustive search wherein the environments of
+                If True, do an exhaustive search wherein the environments of
                 every single pair of particles in the simulation are compared.
+                Thus it is equivalent to including all particles in the system
+                in :code:`'neighbors'`.
                 If False, only compare the environments of neighboring
                 particles. Enabling this option incurs a significant
                 performance penalty. (Default value = :code:`False`)
@@ -659,15 +670,8 @@ cdef class EnvironmentCluster(_MatchEnv):
 
 
 cdef class EnvironmentMotifMatch(_MatchEnv):
-    r"""Find matches between local arrangements of a set of points and a provided motif.
-
-    In general, it is recommended to specify a number of neighbors rather than
-    just a distance cutoff as part of your neighbor querying when performing
-    this computation since it can otherwise be very sensitive. Specifically, it
-    is highly recommended that you choose a number of neighbors query that
-    requests at least as many neighbors as the size of the motif you intend to
-    test against. Otherwise, you will struggle to match the motif. However,
-    this is not currently enforced.
+    r"""Find matches between local arrangements of a set of points and
+    a provided motif :cite:`Teich2019`.
     """  # noqa: E501
 
     cdef freud._environment.EnvironmentMotifMatch * thisptr
@@ -684,6 +688,9 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
         r"""Determine clusters of particles that match the motif provided by
         motif.
 
+        ..warning::
+            Comparison between two sets of environments is only made
+            when both environments contain same number of particles.
         Args:
             system:
                 Any object that is a valid argument to
@@ -700,7 +707,8 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None).
+                (Default value: None). Defines the environment of the query
+                particles.
             registration (bool, optional):
                 If True, first use brute force registration to orient one set
                 of environment vectors with respect to the other set such that
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -611,7 +611,7 @@ cdef class EnvironmentCluster(_MatchEnv):
                 (Default value = :code:`False`)
             global_search (bool, optional):
                 If True, do an exhaustive search wherein the environments of
-                every single pair of particles in the simulation are compared.
+                every single pair of particles are compared.
                 Thus it is equivalent to including all particles in the system
                 in :code:`'neighbors'`.
                 If False, only compare the environments of neighboring
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -513,6 +513,17 @@ cdef class _MatchEnv(_PairCompute):
 cdef class EnvironmentCluster(_MatchEnv):
     r"""Clusters particles according to whether their local environments match
     or not, using various shape matching metrics defined in :cite:`Teich2019`.
+
+    Note that two sets of neighbor lists are used. :code:`'env_neighbors'`
+    defines the query particles' environments. :code:`'neighbors'`
+    defines the neighborhood where query particles' environments are compared
+    to find the common clusters.
+
+    For example, with :code:`'env_neighbors = {num_neighbors: 12}'` and
+    :code:`'neighbors = {'num_neighbors': 15}'`, the :code:`compute` function
+    would construct an environment that contains 12 neighbors for a query
+    particle, and compare it with the environments of the particle's 15 nearest
+    neighbors.
     """
 
     cdef freud._environment.EnvironmentCluster * thisptr
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -570,9 +570,10 @@ cdef class EnvironmentCluster(_MatchEnv):
 
         .. warning::
 
-            All vectors of :code:`cluster_environments` are defined with
-            respect to the query particle. Zero vectors are only used to pad
-            the cluster vectors so that they have the same shape.
+            All vectors of :code:`cluster_environments` and :code:`point_environments`
+            are defined with respect to the query particle.
+            Zero vectors are only used to pad the cluster vectors so that they
+            have the same shape.
             In a future version of freud, zero-padding will be removed.
 
         .. warning::
@@ -648,6 +649,14 @@ cdef class EnvironmentCluster(_MatchEnv):
         if env_neighbors is None:
             env_neighbors = neighbors
         env_nlist, env_qargs = self._resolve_neighbors(env_neighbors)
+
+        if global_search:
+            warnings.warn(
+                "The global search option is deprecated and will be removed in "
+                "version 3.0. If you want this behavior, use a NeighborList "
+                "composed of all pairs of particles in the system as the `neighbors`.",
+                FutureWarning
+            )
 
         self.thisptr.compute(
             nq.get_ptr(), nlist.get_ptr(), dereference(qargs.thisptr),
@@ -727,7 +736,8 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
 
     def compute(self, system, motif, threshold, neighbors=None,
                 registration=False):
-        r"""Determine which particles' local environment matches given environment motif.
+        r"""Determine which particles have local environments
+            matching the given environment motif.
 
         .. warning::
 
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -521,8 +521,8 @@ cdef class EnvironmentCluster(_MatchEnv):
 
     For example, with :code:`'env_neighbors = {num_neighbors: 12}'` and
     :code:`'neighbors = {'num_neighbors': 15}'`, the :code:`compute` function
-    would construct an environment that contains 12 neighbors for a query
-    particle, and compare it with the environments of the particle's 15 nearest
+    constructs an environment that contains 12 neighbors for each query
+    particle, and compares it with the environments of the particle's 15 nearest
     neighbors.
     """
 
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -600,7 +600,7 @@ cdef class EnvironmentCluster(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
-                This argument is used to define the environments of each particle.
+                Defines the neighbors used as the environment of each particle.
                 If ``None``, the value provided for ``neighbors`` will be used
                 (Default value: None).
             registration (bool, optional):
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -532,9 +532,15 @@ cdef class EnvironmentCluster(_MatchEnv):
                 global_search=False):
         r"""Determine clusters of particles with matching environments.
 
-        Using a distance cutoff for :code:`'env_neighbors'` would
+        Using a distance cutoff for :code:`'env_neighbors'` could
         lead to situations where the :code:`'cluster_environments'`
-        contain different number of particles.
+        contain different numbers of particles. In this case, the
+        environments which have a number of neighbors less than
+        the environment with the maximum number of neighbors
+        :math:`k_{max}` will have their entry in `cluster_environments`
+        padded with zero vectors. For example, a cluster environment
+        with :math:`m < k`neighbors, will have :math:`k - m` zero
+        vectors at the end of its entry in `cluster_environments`.
 
         .. warning::
 
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -514,10 +514,10 @@ cdef class EnvironmentCluster(_MatchEnv):
     r"""Clusters particles according to whether their local environments match
     or not, using various shape matching metrics defined in :cite:`Teich2019`.
 
-    Note that two sets of neighbor lists are used. :code:`'env_neighbors'`
-    defines the query particles' environments. :code:`'neighbors'`
+    Note that two sets of neighbor lists are used: :code:`'env_neighbors'`
+    defines the query particles' environments, and :code:`'neighbors'`
     defines the neighborhood where query particles' environments are compared
-    to find the common clusters.
+    to determine the clusters.
 
     For example, with :code:`'env_neighbors = {num_neighbors: 12}'` and
     :code:`'neighbors = {'num_neighbors': 15}'`, the :code:`compute` function
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -550,7 +550,7 @@ cdef class EnvironmentCluster(_MatchEnv):
             freud, zero-padding will be removed.
 
         .. warning::
-        
+
             Comparison between two sets of environments is only made when both
             environments contain the same number of particles.
 
@@ -583,7 +583,7 @@ cdef class EnvironmentCluster(_MatchEnv):
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
                 Defines the particle neighborhoods for comparing
-                environments. (Default value: None). 
+                environments. (Default value: None).
             env_neighbors (:class:`freud.locality.NeighborList` or dict, optional):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
@@ -716,7 +716,7 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
                 Defines the environment of the query particles.
-                (Default value: None). 
+                (Default value: None).
             registration (bool, optional):
                 If True, first use brute force registration to orient one set
                 of environment vectors with respect to the other set such that
--- a/freud/environment.pyx
+++ b/freud/environment.pyx
@@ -532,26 +532,41 @@ cdef class EnvironmentCluster(_MatchEnv):
                 global_search=False):
         r"""Determine clusters of particles with matching environments.
 
-        Note that two sets of neighbor lists are used: :code:`env_neighbors`
-        defines the query particles' environments, and :code:`neighbors`
-        defines the neighborhood where query particles' environments are compared
-        to determine the clusters.
+        An environment is defined by the bond vectors between a particle and its
+        neighbors as defined by :code:`env_neighbors`.
+        For example, :code:`env_neighbors= {'num_neighbors': 8}` means that every
+        particle's local environment is defined by its 8 nearest neighbors.
+        Then, each particle's environment is compared to the environments of a
+        selected list of other particles in the system, controlled by the
+        :code:`neighbors` parameter. For example, :code:`neighbors={'r_max': 3.0}`
+        means that the environment of each particle will be compared to the
+        environment of every particle within a distance of 3.0.
 
-        For example, with :code:`env_neighbors = {'num_neighbors': 12}` and
-        :code:`neighbors = {'num_neighbors': 15}`, the :code:`compute` function
-        constructs an environment that contains 12 neighbors for each query
-        particle, and compares it with the environments of the particle's 15
-        nearest neighbors.
+        Two environments are compared using `point-set registration
+        <https://en.wikipedia.org/wiki/Point-set_registration)>`_.
+        Specifically, we use the `Kabsch algorithm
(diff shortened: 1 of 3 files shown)


freud/environment.pyx
```diff
@@ -512,7 +512,7 @@ cdef class _MatchEnv(_PairCompute):
 
 cdef class EnvironmentCluster(_MatchEnv):
     r"""Clusters particles according to whether their local environments match
-    or not, according to various shape matching metrics.
+    or not, according to various shape matching metrics :cite:`Teich2019`.
```
tommy-waltmann:
```suggestion
    or not, using various shape matching metrics defined in :cite:`Teich2019`.
```

freud/environment.pyx
```diff
@@ -532,11 +532,20 @@ cdef class EnvironmentCluster(_MatchEnv):
...
+        .. warning::
+
+            All vectors of :code:`'cluster_environments'` are defined with
+            respect to the origin. Zero vectors are only used to pad the cluster
+            vectors so that they have the same shape. In a future version of
+            freud, zero-padding will be removed.
+
+        .. warning::
+            Comparison between two sets of environments is only made when both
+            environments contain same number of particles.
```
tommy-waltmann:
```suggestion
        .. warning::
        
            Comparison between two sets of environments is only made when both
            environments contain the same number of particles.
```

freud/environment.pyx
```diff
@@ -566,25 +575,27 @@ cdef class EnvironmentCluster(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None).
+                (Default value: None). Defines the particle neighborhoods for
+                comparing environments.
```
tommy-waltmann:
```suggestion
                Defines the particle neighborhoods for comparing
                environments. (Default value: None). 
```

freud/environment.pyx
```diff
@@ -566,25 +575,27 @@ cdef class EnvironmentCluster(_MatchEnv):
...
             env_neighbors (:class:`freud.locality.NeighborList` or dict, optional):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
-                `query arguments
+                `query arguments.
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
```
tommy-waltmann:
```suggestion
                `query arguments
                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
```

freud/environment.pyx
```diff
@@ -700,7 +708,8 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
                 <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None).
+                (Default value: None). Defines the environment of the query
+                particles.
```
tommy-waltmann:
```suggestion
                Defines the environment of the query particles.
                (Default value: None). 
```

freud/environment.pyx
```diff
@@ -659,15 +670,8 @@ cdef class EnvironmentCluster(_MatchEnv):
...
-
-    In general, it is recommended to specify a number of neighbors rather than
-    just a distance cutoff as part of your neighbor querying when performing
-    this computation since it can otherwise be very sensitive. Specifically, it
-    is highly recommended that you choose a number of neighbors query that
-    requests at least as many neighbors as the size of the motif you intend to
-    test against. Otherwise, you will struggle to match the motif. However,
-    this is not currently enforced.
+    r"""Find matches between local arrangements of a set of points and
+    a provided motif :cite:`Teich2019`.
```
tommy-waltmann:
```suggestion
    r"""Find matches between local arrangements of a set of points and
    a provided motif, as done in :cite:`Teich2019`.
```

freud/environment.pyx
```diff
@@ -532,11 +532,20 @@ cdef class EnvironmentCluster(_MatchEnv):
...
-        lead to situations where a point doesn't match a cluster because a
-        required neighbor is just outside the cutoff.
+        Using a distance cutoff for :code:`'env_neighbors'` would
+        lead to situations where the :code:`'cluster_environments'`
+        contain different number of particles.
+
+        .. warning::
+
+            All vectors of :code:`'cluster_environments'` are defined with
+            respect to the origin. Zero vectors are only used to pad the cluster
```
tommy-waltmann: If I understand this calculation right, these vectors are defined relative to the particle at the center of the motif, not relative to the origin.

freud/environment.pyx
```diff
@@ -532,11 +532,20 @@ cdef class EnvironmentCluster(_MatchEnv):
...
         r"""Determine clusters of particles with matching environments.
 
-        In general, it is recommended to specify a number of neighbors rather
-        than just a distance cutoff as part of your neighbor querying when
-        performing this computation. Using a distance cutoff alone could easily
-        lead to situations where a point doesn't match a cluster because a
-        required neighbor is just outside the cutoff.
+        Using a distance cutoff for :code:`'env_neighbors'` would
+        lead to situations where the :code:`'cluster_environments'`
+        contain different number of particles.
```
tommy-waltmann:
```suggestion
        Using a distance cutoff for :code:`'env_neighbors'` could
        lead to situations where the :code:`'cluster_environments'`
        contain different numbers of particles. In this case, the
        environments which have a number of neighbors less than
        the environment with the maximum number of neighbors
        :math:`k_{max}` will have their entry in `cluster_environments`
        padded with zero vectors. For example, a cluster environment
        with :math:`m < k`neighbors, will have :math:`k - m` zero
        vectors at the end of its entry in `cluster_environments`.
```

tommy-waltmann (changes requested):
Thanks for this update @Charlottez112 , this is a good start. I've left some suggestions I'd like addressed, and I would also like to see a conceptual explanation in `EnvironmentClusters` of why two definitions of neighbors may be used to do the calculation.

freud/environment.pyx
Charlottez112: The center of the motif is moved to the origin I think. Otherwise the zero vectors would mean particles at the origin. See in this [comment](https://github.com/glotzerlab/freud/pull/979#issuecomment-1164555080) where Erin replied to me

freud/environment.pyx
Charlottez112: May I ask why you want to delete it? Without the period it looks kinda weird. We can also move this sentence so that it's always at the end.
image

freud/environment.pyx
tommy-waltmann: In that case, I think it would be more appropriate to say something along the lines of "All vectors of `cluster_environments` are defined relative to the average position of the particles in the cluster.".

freud/environment.pyx
tommy-waltmann: I'm not asking to remove the period, I'm asking to move it outside the hyperlink. I think this will keep the period in the rendered documentation, but keep it from becoming highlighted purple once the link is clicked.

Charlottez112:
I just noticed that there are also a bunch of zeros added to the `point_environments` computed for the query particles. I'm not sure if these zeros are necessary. I guess they're also just paddings? Maybe we should also warn users about it

freud/environment.pyx
Charlottez112: Yeah that makes sense. I see why this can be confusing now. But I'll just check with @erteich again to make sure.

freud/environment.pyx
```diff
@@ -684,6 +706,10 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
         r"""Determine clusters of particles that match the motif provided by
         motif.
```
tommy-waltmann:
```suggestion
        r"""Determine which points' local environment matches the given environment motif.
```

freud/environment.pyx
```diff
@@ -684,6 +706,10 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
...
 
+        .. warning::
+            Comparison between two sets of environments is only made
+            when both environments contain the same number of particles.
+
```
tommy-waltmann:
```suggestion
        .. warning::
        
            Comparison between two environments is only made when
            both environments contain the same number of particles.

```

freud/environment.pyx
```diff
@@ -699,7 +725,8 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
+                Defines the environment of the query particles
```
tommy-waltmann, line 774:
```suggestion
                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
                Defines the environment of the query points.
```

freud/environment.pyx
```diff
@@ -512,7 +512,18 @@ cdef class _MatchEnv(_PairCompute):
 
 cdef class EnvironmentCluster(_MatchEnv):
     r"""Clusters particles according to whether their local environments match
```
tommy-waltmann, line 514:
```suggestion
    r"""Clusters points according to whether their local environments match
```

freud/environment.pyx
```diff
@@ -512,7 +512,18 @@ cdef class _MatchEnv(_PairCompute):
...
+    Note that two sets of neighbor lists are used. :code:`'env_neighbors'`
+    defines the query particles' environments. :code:`'neighbors'`
+    defines the neighborhood where query particles' environments are compared
+    to find the common clusters.
+
+    For example, with :code:`'env_neighbors = {num_neighbors: 12}'` and
+    :code:`'neighbors = {'num_neighbors': 15}'`, the :code:`compute` function
+    would construct an environment that contains 12 neighbors for a query
+    particle, and compare it with the environments of the particle's 15 nearest
+    neighbors.
```
tommy-waltmann:
```suggestion
    Note:
    
        Two sets of neighbor lists are used: :code:`'env_neighbors'`
        defines the query points' environments, and :code:`'neighbors'`
        defines the edges (neighbors) on the point graph where query
        point environments are compared to determine the clusters.

        For example, with :code:`'env_neighbors = {num_neighbors: 12}'` and
        :code:`'neighbors = {'num_neighbors': 15}'`, the :code:`compute` function
        constructs an environment that contains 12 neighbors for each query
        point, and finds clusters by traversing the point graph where there is an edge
        between each point and its 15 nearest neighbors.
```

There may be better words to express the difference between the two sets of neighbors. What we need to make clear here is that `env_neighbors` defines the local environment, and `neighbors` defines how the graph is traversed to determine the clusters. If anyone has any other suggestions, feel free to jump in.

freud/environment.pyx
```diff
@@ -532,11 +543,27 @@ cdef class EnvironmentCluster(_MatchEnv):
                 global_search=False):
         r"""Determine clusters of particles with matching environments.
```
tommy-waltmann, line 533:
```suggestion
        r"""Determine clusters of points with matching environments.
```

freud/environment.pyx
```diff
@@ -532,11 +543,27 @@ cdef class EnvironmentCluster(_MatchEnv):
...
 
-        In general, it is recommended to specify a number of neighbors rather
-        than just a distance cutoff as part of your neighbor querying when
-        performing this computation. Using a distance cutoff alone could easily
-        lead to situations where a point doesn't match a cluster because a
-        required neighbor is just outside the cutoff.
+        Using a distance cutoff for :code:`'env_neighbors'` could
+        lead to situations where the :code:`'cluster_environments'`
+        contain different numbers of particles. In this case, the
```
tommy-waltmann:
```suggestion
        contain different numbers of points. In this case, the
```

freud/environment.pyx
```diff
@@ -532,11 +543,27 @@ cdef class EnvironmentCluster(_MatchEnv):
...
+        the environment with the maximum number of neighbors
+        :math:`k_{max}` will have their entry in `cluster_environments`
+        padded with zero vectors. For example, a cluster environment
+        with :math:`m < k`neighbors, will have :math:`k - m` zero
+        vectors at the end of its entry in `cluster_environments`.
+
+        .. warning::
+
+            All vectors of :code:`'cluster_environments'` are defined with
+            respect to the query particle. Zero vectors are only used to pad
```
tommy-waltmann:
```suggestion
            respect to the query point. Zero vectors are only used to pad
```

freud/environment.pyx
```diff
@@ -532,11 +543,27 @@ cdef class EnvironmentCluster(_MatchEnv):
...
+            the cluster vectors so that they have the same shape.
+            In a future version of freud, zero-padding will be removed.
+
+        .. warning::
+
+            Comparison between two sets of environments is only made when both
+            environments contain the same number of particles.
```
tommy-waltmann:
```suggestion
            environments contain the same number of points.
```

freud/environment.pyx
```diff
@@ -565,26 +592,28 @@ cdef class EnvironmentCluster(_MatchEnv):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None).
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
+                Defines the particle neighborhoods for comparing
```
tommy-waltmann:
```suggestion
                Defines the point neighborhoods for comparing
```

freud/environment.pyx
```diff
@@ -565,26 +592,28 @@ cdef class EnvironmentCluster(_MatchEnv):
...
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                This argument is used to define the neighbors of the
-                environment that motifs are registered against. If ``None``,
-                the value provided for ``neighbors`` will be used. (Default
-                value: None).
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
+                This argument is used to define the environments of each particle.
```
tommy-waltmann:
```suggestion
                This argument is used to define the environment of each point.
```

freud/environment.pyx
```diff
@@ -565,26 +592,28 @@ cdef class EnvironmentCluster(_MatchEnv):
...
             registration (bool, optional):
                 If True, first use brute force registration to orient one set
                 of environment vectors with respect to the other set such that
                 it minimizes the RMSD between the two sets. Enabling this
                 option incurs a significant performance penalty.
                 (Default value = :code:`False`)
             global_search (bool, optional):
-                 If True, do an exhaustive search wherein the environments of
+                If True, do an exhaustive search wherein the environments of
                 every single pair of particles in the simulation are compared.
```
tommy-waltmann:
```suggestion
                every single pair of points are compared.
```

tommy-waltmann (changes requested):
I added more suggestions. Some of them were related to changes made in the PR already, and some were related to problems with the way the documentation was originally written. For example, the word particles was used a lot, and as a general philosophy in freud, we like to use the word points instead because it is more general.

There was also one fairly large suggestion related to clarifying the difference between `env_neighbors` and `neighbors` which I would like some feedback on. I tried to use graph language to help clarify the difference, but I'm not sure if it will make sense to most readers or not.

freud/environment.pyx
Charlottez112: I think using "graph" makes it more complex, because we don't mention graph anywhere else. Do you want to use graph language because of the algorithm?

Charlottez112:
> I added more suggestions. Some of them were related to changes made in the PR already, and some were related to problems with the way the documentation was originally written. For example, the word particles was used a lot, and as a general philosophy in freud, we like to use the word points instead because it is more general.

I think using `particle` is fine right now since it's actually used commonly in various modules. Like for Steinhardt order parameters we access per particle quantity using `particle_order`, same for `particle_msd` etc. If this is something you guys want to do for version 3, replacing all "particles" with "points", we can definitely do that here too.

freud/environment.pyx
tommy-waltmann: Yes, basically the algorithm is graph traversal while removing edges between points with different environments and then doing clustering on the resulting graph. This suggestion can be disregarded though if that's too confusing

tommy-waltmann:
> I think using `particle` is fine right now since it's actually used commonly in various modules.

I've marked the comments involving particle vs point as resolved. Some of them had grammatical suggestions as well, which I'll go back through and make again.

freud/environment.pyx
tommy-waltmann:
```suggestion
            Comparisons between two environments are only made when both
            environments contain the same number of particles.
```

freud/environment.pyx
```diff
@@ -565,26 +592,28 @@ cdef class EnvironmentCluster(_MatchEnv):
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                (Default value: None).
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
+                Defines the particle neighborhoods for comparing
+                environments (Default value: None).
```
tommy-waltmann:
```suggestion
                Defines the neighbors used for comparing different particle
                environments (Default value: None).
```

freud/environment.pyx
```diff
@@ -565,26 +592,28 @@ cdef class EnvironmentCluster(_MatchEnv):
...
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                This argument is used to define the neighbors of the
-                environment that motifs are registered against. If ``None``,
-                the value provided for ``neighbors`` will be used. (Default
-                value: None).
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
+                This argument is used to define the environments of each particle.
+                If ``None``, the value provided for ``neighbors`` will be used
```
tommy-waltmann:
```suggestion
                Defines the neighbors used as the environment of each particle.
                If ``None``, the value provided for ``neighbors`` will be used
```

freud/environment.pyx
tommy-waltmann:
```suggestion
                every single pair of particles are compared.
```

freud/environment.pyx
tommy-waltmann:
```suggestion
        r"""Determine which particles' local environment matches given environment motif.
```

tommy-waltmann (changes requested):
Just grammatical things, we should be good to go after this.

freud/environment.pyx
```diff
@@ -512,7 +512,18 @@ cdef class _MatchEnv(_PairCompute):
 
 cdef class EnvironmentCluster(_MatchEnv):
     r"""Clusters particles according to whether their local environments match
-    or not, according to various shape matching metrics.
+    or not, using various shape matching metrics defined in :cite:`Teich2019`.
+
+    Note that two sets of neighbor lists are used. :code:`'env_neighbors'`
+    defines the query particles' environments. :code:`'neighbors'`
+    defines the neighborhood where query particles' environments are compared
+    to find the common clusters.
```
tommy-waltmann:
```suggestion
    Note that two sets of neighbor lists are used: :code:`'env_neighbors'`
    defines the query particles' environments, and :code:`'neighbors'`
    defines the neighborhood where query particles' environments are compared
    to determine the clusters.
```

freud/environment.pyx
tommy-waltmann:
```suggestion
    For example, with :code:`'env_neighbors = {num_neighbors: 12}'` and
    :code:`'neighbors = {'num_neighbors': 15}'`, the :code:`compute` function
    constructs an environment that contains 12 neighbors for each query
    particle, and compares it with the environments of the particle's 15 nearest
    neighbors.
```

tommy-waltmann (changes requested):

freud/environment.pyx
```diff
@@ -532,11 +543,27 @@ cdef class EnvironmentCluster(_MatchEnv):
...
-        lead to situations where a point doesn't match a cluster because a
-        required neighbor is just outside the cutoff.
+        Using a distance cutoff for :code:`'env_neighbors'` could
+        lead to situations where the :code:`'cluster_environments'`
+        contain different numbers of particles. In this case, the
+        environments which have a number of neighbors less than
+        the environment with the maximum number of neighbors
+        :math:`k_{max}` will have their entry in `cluster_environments`
+        padded with zero vectors. For example, a cluster environment
+        with :math:`m < k`neighbors, will have :math:`k - m` zero
```
vyasr:
```suggestion
        with :math:`m < k` neighbors, will have :math:`k - m` zero
```

freud/environment.pyx
```diff
@@ -512,7 +512,18 @@ cdef class _MatchEnv(_PairCompute):
...
+    Note that two sets of neighbor lists are used: :code:`'env_neighbors'`
+    defines the query particles' environments, and :code:`'neighbors'`
+    defines the neighborhood where query particles' environments are compared
+    to determine the clusters.
+
+    For example, with :code:`'env_neighbors = {num_neighbors: 12}'` and
+    :code:`'neighbors = {'num_neighbors': 15}'`, the :code:`compute` function
+    constructs an environment that contains 12 neighbors for each query
+    particle, and compares it with the environments of the particle's 15 nearest
+    neighbors.
```
vyasr: This should probably go into the docstring of compute rather than the constructor since it's discussing parameters for that function.

freud/environment.pyx
```diff
@@ -512,7 +512,18 @@ cdef class _MatchEnv(_PairCompute):
...
-    or not, according to various shape matching metrics.
+    or not, using various shape matching metrics defined in :cite:`Teich2019`.
+
+    Note that two sets of neighbor lists are used: :code:`'env_neighbors'`
+    defines the query particles' environments, and :code:`'neighbors'`
+    defines the neighborhood where query particles' environments are compared
+    to determine the clusters.
+
+    For example, with :code:`'env_neighbors = {num_neighbors: 12}'` and
+    :code:`'neighbors = {'num_neighbors': 15}'`, the :code:`compute` function
```
vyasr:
```suggestion
    For example, with :code:`env_neighbors = {num_neighbors: 12}` and
    :code:`neighbors = {'num_neighbors': 15}`, the :code:`compute` function
```

freud/environment.pyx
```diff
@@ -512,7 +512,18 @@ cdef class _MatchEnv(_PairCompute):
 
 cdef class EnvironmentCluster(_MatchEnv):
     r"""Clusters particles according to whether their local environments match
-    or not, according to various shape matching metrics.
+    or not, using various shape matching metrics defined in :cite:`Teich2019`.
+
+    Note that two sets of neighbor lists are used: :code:`'env_neighbors'`
+    defines the query particles' environments, and :code:`'neighbors'`
```
vyasr:
```suggestion
    Note that two sets of neighbor lists are used: :code:`env_neighbors`
    defines the query particles' environments, and :code:`neighbors`
```

freud/environment.pyx
```diff
@@ -565,26 +592,28 @@ cdef class EnvironmentCluster(_MatchEnv):
...
                 it minimizes the RMSD between the two sets. Enabling this
                 option incurs a significant performance penalty.
                 (Default value = :code:`False`)
             global_search (bool, optional):
-                 If True, do an exhaustive search wherein the environments of
-                every single pair of particles in the simulation are compared.
+                If True, do an exhaustive search wherein the environments of
+                every single pair of particles are compared.
+                Thus it is equivalent to including all particles in the system
+                in :code:`'neighbors'`.
```
vyasr, line 644: If `global_search=True`, are the `neighbors` ignored?

freud/environment.pyx
```diff
@@ -532,11 +543,27 @@ cdef class EnvironmentCluster(_MatchEnv):
                 global_search=False):
         r"""Determine clusters of particles with matching environments.
 
-        In general, it is recommended to specify a number of neighbors rather
-        than just a distance cutoff as part of your neighbor querying when
-        performing this computation. Using a distance cutoff alone could easily
-        lead to situations where a point doesn't match a cluster because a
-        required neighbor is just outside the cutoff.
+        Using a distance cutoff for :code:`'env_neighbors'` could
```
vyasr: This entire comment feels like it belongs in a `.. note::` block rather than just free-floating. This docstring could benefit from adding a short description of _how_ clusters are matched. That might be a more useful way of presenting the discussion of `neighbors` vs `env_neighbors` currently in the class docstring. Here's a suggested starting point:

> An environment is defined by the bond vectors between a particle and its neighbors as defined by `env_neighbors`. For example, `env_neighbors={'num_neighbors': 8}` means that every particle's local neighborhood is defined by its eight nearest neighbors. Then, each particle's environment is compared to the environment of the other particles in the system. Which pairs of environments are compared is controlled by the `neighbors` parameter. For example, `neighbors={'r_max': 3.0}` means that the environment of each particle will be compared to the environment of every particle within a distance of 3.0.
>
> Two environments are compared using [point-set registration](https://en.wikipedia.org/wiki/Point-set_registration). Specifically, we use the [Kabsch algorithm](https://en.wikipedia.org/wiki/Kabsch_algorithm) to find the optimal rotation matrix mapping one environment onto another. However, the Kabsch algorithm assumes a one-to-one correspondence between points in each environment, whereas we typically do not know which point in one environment corresponds to a particular point in another environment. The brute force solution here is to try all possible correspondences, but the resulting combinatorial explosion makes this problem infeasible, so we use a tolerance threshold to terminate the search when we have found a permutation of points that results in a sufficiently low RMSD.

Then you can put the "Using a distance cutoff..." part into a note.

vyasr (changes requested):

freud/environment.pyx
```diff
@@ -681,8 +703,12 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
...
     def compute(self, system, motif, threshold, neighbors=None,
                 registration=False):
-        r"""Determine clusters of particles that match the motif provided by
-        motif.
+        r"""Determine which particles' local environment matches given environment motif.
+
+        .. warning::
+
+            Comparisons between two environments are only made when both
+            environments contain the same number of particles.
```
vyasr: I think we can be more explicit here.
```suggestion
            A particle's environment can only match the motif if it contains the
            same number of particles as the motif. Any environment with a
            different number of particles than the motif will _always_ fail to match
            the motif.
```

vyasr: Based on my recommendation in #979, I think we probably also want to mention that no warning will be given at runtime if a user provides such a mismatched environment (at least for now).

freud/environment.pyx
Charlottez112: I think so. I haven't checked the code though. I believe that's why in #979 @erteich also suggested removing the global search flag. see the last few lines of this [comment](https://github.com/glotzerlab/freud/pull/979#issuecomment-1163219389)

vyasr: Ah yes good call. Can you make another issue for removing `global_search` and also add that to the 3.0 milestone (since that's also a breaking change)?

freud/environment.pyx
Charlottez112: I added that no warning will be given, and used what you suggested here as the explanation of how  `EnvironmentMotifMatch` works, which I think is a better fit

freud/environment.pyx
Charlottez112: This is great. I added this to `EnvironmentCluster.compute`

freud/environment.pyx
```diff
@@ -565,26 +609,28 @@ cdef class EnvironmentCluster(_MatchEnv):
...
                 it minimizes the RMSD between the two sets. Enabling this
                 option incurs a significant performance penalty.
                 (Default value = :code:`False`)
             global_search (bool, optional):
-                 If True, do an exhaustive search wherein the environments of
-                every single pair of particles in the simulation are compared.
+                If True, do an exhaustive search wherein the environments of
+                every single pair of particles are compared.
+                Thus it is equivalent to including all particles in the system
+                in :code:`'neighbors'`.
```
vyasr, line 644: Can you add something like
```
if global_search:
    warnings.warn(
        "The global search option is deprecated and will be removed in version 3.0. If you want this behavior, use a NeighborList composed of all pairs of particles in the system as the `neighbors`.", FutureWarning)
```
in the code?

freud/environment.pyx
```diff
@@ -681,8 +727,14 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
 
     def compute(self, system, motif, threshold, neighbors=None,
                 registration=False):
-        r"""Determine clusters of particles that match the motif provided by
-        motif.
+        r"""Determine which particles' local environment matches given environment motif.
```
vyasr:
```suggestion
        r"""Determine which particles have local environments matching the given motif.
```

vyasr (approved):
I have a couple of small suggestions, but otherwise this looks good to go!

freud/environment.pyx
Charlottez112: do we need to add a test for this?

freud/environment.pyx
Charlottez112: I forgot to check with you. It seems like `point_environments` are also zero-padded. I added a few words in the documentation to warn user about it, but we haven't discussed whether we'll remove those zeros as well

freud/environment.pyx
erteich: Hi @Charlottez112 and @tommy-waltmann , cluster_environments is a list of environments that were found via clustering, meaning that each environment in cluster_environments corresponds to the environment of a single "cluster" of particles with similar environments (ie neighborhoods). The way the code works is that cluster_environments[i]  is actually the average environment taken over all the environments of particles in cluster i. So if cluster 0 consists of 4 particles (indexed k1-k4) which have similar environments defined by their 6 nearest neighbors, then cluster_environments[0] is a set of 6 vectors, where each vector is the average taken over the corresponding matching vector in the environments of particles k1-k4. As an example, cluster_environments[0][0] will be a vector, calculated as the following average: (k1_environment[0]+k2_environment[0]+k3_environment[0]+k4_environment[0])/4. Each vector in every particle's environment is defined _with respect to the particle_, which is what enables these averages to take place across different particle environments. Put another way, one could imagine placing each particle at the origin when defining the vectors that make up its environment. So I would say "cluster_environments[i] contains the average environment of the cluster of particles that is indexed by i. Each vector in cluster_environments[i] is defined with respect to an imaginary particle sitting at the origin." Does this make sense and seem okay? Feel free to change the wording, especially of the word "imaginary." But I can't think of a clearer way of saying it- presumably the user will understand how environments are generally defined throughout the code, and so saying that this variable contains "average environments" won't be too confusing?

The averaging process can be seen in the code, if you are curious or if looking at the code itself would be clearer than my explanation: line 626-631 of [environment.pyx](https://github.com/glotzerlab/freud/blob/master/freud/environment.pyx) shows that cluster_environments returns the set of environments produced by getClusterEnvironments() in [MatchEnv.h](https://github.com/glotzerlab/freud/blob/master/cpp/environment/MatchEnv.h) (defined lines 283-286). That function just returns the object m_cluster_environments, which is populated in lines 656-660 of [MatchEnv.cc](https://github.com/glotzerlab/freud/blob/master/cpp/environment/MatchEnv.cc) from the object cluster_env, which is computed in lines 632-639 of MatchEnv.cc via the method getAvgEnv (defined lines 192-238). Wow now that I actually summarized it I'm not sure if this is clearer than my explanation... but just in case you want to see how it is working under the hood, that is what's happening!

freud/environment.pyx
erteich: Just as another point of clarification, @Charlottez112 , point_environments is the set of all environments of every single particle in the system, rather than the set of average environments of clusters of particles (which is what cluster_environments is). I think that whatever is being done for cluster_environments in terms of zero-padding should also be done for point_environments, since cluster_environments is just built via averages of point_environments.

freud/environment.pyx
```diff
@@ -532,11 +532,56 @@ cdef class EnvironmentCluster(_MatchEnv):
...
+        Specifically, we use the `Kabsch algorithm
+        <https://en.wikipedia.org/wiki/Kabsch_algorithm>`_ to find the optimal
+        rotation matrix mapping one environment onto another. The Kabsch
+        algorithm assumes a one-to-one correspondence between points in the two
+        environments. However, we typically do not know which point in one environment
+        corresponds to a particular point in the other environment. The brute force
+        solution is to try all possible correspondences, but the resulting
+        combinatorial explosion makes this problem infeasible, so we use a tolerance
+        :code:`threshold` to terminate the search when we have found a permutation
+        of points that results in a sufficiently low RMSD.
```
erteich, line 566: I think this is a wonderful explanation, thank you! I would say as a final sentence here: 

The thresholding criterion we apply in order to determine if the two point sets have sufficiently low RMSD (i.e. "match") is quite conservative: the two point sets match if and only if, for every pair of matched points between the sets, the distance between the matched pair is less than :code:`threshold`.

That way, there is no ambiguity about how matching works here. Otherwise the user might think that we match according to RMSD lower than the threshold, which is not actually precisely what is happening.

freud/environment.pyx
```diff
@@ -532,11 +532,56 @@ cdef class EnvironmentCluster(_MatchEnv):
...
+
+        .. note::
+
+            Using a distance cutoff for :code:`env_neighbors` could
+            lead to situations where the :code:`cluster_environments`
+            contain different numbers of particles. In this case, the
```
erteich: Instead of "Using a distance cutoff for :code:`env_neighbors` could lead to situations where the :code:`cluster_environments` contain different numbers of particles", I would say:

"Using a distance cutoff for :code:`env_neighbors` could lead to situations where the :code:`cluster_environments` contain different numbers of vectors."

Or instead of vectors above, you could write "neighbors." Either choice removes ambiguity about whether we are referring to the number of particles in each cluster, or the number of particles that define environments.

freud/environment.pyx
```diff
@@ -659,15 +714,15 @@ cdef class EnvironmentCluster(_MatchEnv):
 
 
 cdef class EnvironmentMotifMatch(_MatchEnv):
-    r"""Find matches between local arrangements of a set of points and a provided motif.
+    r"""Find matches between local arrangements of a set of points and
+    a provided motif, as done in :cite:`Teich2019`.
+
+    A particle's environment can only match the motif if it contains the
+    same number of particles as the motif. Any environment with a
+    different number of particles than the motif will always fail to match
```
erteich: Instead of "A particle's environment can only match the motif if it contains the same number of particles as the motif. Any environment with a different number of particles than the motif will always fail to match the motif" I would say:

"A particle's environment can only match the motif if its number of neighbors equals the number of vectors in the motif. Any environment with a different number of neighbors than the number of vectors in the motif will always fail to match the motif."

I think this is clearer, and drives home the idea that the motif that the user provides and might want to search for within the system can be any set of points or vectors. It does not necessarily have to be a neighborhood that is already calculated via cluster_environments, although maybe that will be the most common use case.

freud/environment.pyx
```diff
@@ -681,8 +736,15 @@ cdef class EnvironmentMotifMatch(_MatchEnv):
...
-        r"""Determine clusters of particles that match the motif provided by
-        motif.
+        r"""Determine which particles have local environments
+            matching the given environment motif.
+
+        .. warning::
+
+            Comparisons between two environments are only made when both
+            environments contain the same number of particles.
+            However, this is not currently enforced, so no warning will be given
```
erteich: Now I think it is confusing to say "However, this is not currently enforced" here. That makes it sound like it's not currently enforced that comparisons between two environments are only made when both environments contain the same number of particles. But that is enforced. I would just delete that part of the sentence and only say:

"No warning will be given at runtime if mismatched environments are provided for comparison to the motif."

freud/environment.pyx
Charlottez112: Thank you @erteich! I'll include the definition of cluster_environment in the documentation as well. This is crystal clear. I guess in my original change what was confusing, or what I think @tommy-waltmann was confused about, was that the "origin" could mean the origin of the system. But your example is great, and it definitely helps to explain how the cluster_environments are computed.

freud/environment.pyx
vyasr: I agree. The expected invariant should be that every point environment corresponding to a specific cluster environment should have the same number of vectors. In other words, if `cluster_idx[i] == j`, then `point_environments[i].shape == cluster_environments[j].shape`.

freud/environment.pyx
vyasr: Ah that is a good point. We perform brute force registration between pairs of environments, but the threshold is applied to the vector distance between each pair of vectors in the environments, not to the RMSD.

freud/environment.pyx
vyasr: Based on your previous comment, I would like us to be consistent in describing an environment as either a set of vectors or a set of neighbors. Using consistent terminology will help avoid confusion. I am inclined to use vectors to avoid overloading the word neighbors any further, but I could be convinced either way. I would propose to describe a particle's environment as "the set of vectors between the particle and its `env_neighbors`".

freud/environment.pyx
Charlottez112: I was looking at the code, and it seems like if `registration=False`, only the thresholding criterion @erteich mentioned above is used? i.e., we brute force all pairs of vectors of two environments and compute the squared distance. No RMSD is computed for two sets of environments.
The Kabsch algorithm is only used if `registration=True`, so RMSD is only computed and minimized to find the rotation if `registration=True`?

freud/environment.pyx
erteich: agreed! thanks for putting it so succinctly @vyasr !

freud/environment.pyx
erteich: That is fine with me! I think using "vectors" is appropriate, and the most accurate.

freud/environment.pyx
erteich: @Charlottez112 that was a great catch, and you're mostly totally right. The thresholding criterion is used in both cases (registration = True and registration = False), as can be seen in line 313 (registration=True) and line 329 (registration=False) of MatchEnv.cc. But indeed, you're right that the Kabsch algorithm and the brute force method of comparing random pairs of vectors in order to minimize RMSD only happens if registration=True. So I would instead maybe say the following to be even more clear:

```suggestion
        Two environments are compared using `point-set registration
        <https://en.wikipedia.org/wiki/Point-set_registration)>`_.
        Specifically, when :code:`registration=True`, we use the `Kabsch algorithm
        <https://en.wikipedia.org/wiki/Kabsch_algorithm>`_ to find the optimal
        rotation matrix mapping one environment onto another. The Kabsch
        algorithm assumes a one-to-one correspondence between points in the two
        environments. However, we typically do not know which point in one environment
        corresponds to a particular point in the other environment. The brute force
        solution is to try all possible correspondences, but the resulting
        combinatorial explosion makes this problem infeasible, so we use a tolerance
        :code:`threshold` to terminate the search when we have found a permutation
        of points that results in a sufficiently low RMSD.
        The thresholding criterion we apply in order to determine if the two point sets 
        have sufficiently low RMSD (i.e. `match`) is quite conservative: the two point 
        sets match if and only if, for every pair of matched points between the sets, the 
        distance between the matched pair is less than :code:threshold. When 
        :code:`registration=False`, environments are not rotated prior to comparison 
        between them. Instead, all pairs of points are exhaustively compared and the 
        same thresholding criterion is used: two point sets match if and only if, for every pair 
        of matched points between the sets, the distance between the pair is less than 
        :code:threshold.
```

freud/environment.pyx
vyasr: That would be good! Something using `pytest.warns` would be helpful. And perhaps include something like `assert freud.__version__ < '3.0.0'` so that we make sure to remove this in 3.0

freud/environment.pyx
vyasr: You're right, the RMSD computation only happens if `registration=True`. I forgot that code path even existed. @erteich what's the value in _not_ performing registration? That code path leads to somewhat arbitrary results since it depends on the order in which the elements of the environment are traversed, which isn't great. Is it just useful for quick testing because it's faster to run without registration?

You mentioned the possibility of different similarity functions (plugging in a different `isSimilar` implementation). With vs without registration actually seems like that's exactly what it does. Maybe we should refactor the code to treat those as two different similarity functions, and in 3.0 allow users to specify which similarity function to use instead? That feels like a more flexible API.

erteich: Hi @vyasr , hopefully the comment I just made (I actually made it last night but forgot to "submit" it, so I think it is only visible to you now?) somewhat addresses your question. You're right that it depends on the order in which the elements of the environment are traversed- it is certainly not the most ideal way to do this. Like the method using registration though, the positive side is that the thresholding is such that the match found (local minimum) satisfies a certain similarity criterion, regardless of whether it is a true global minimum or not. The downside is that there may be cases in which local minima that meet the similarity criterion exist but are not found, although my intuition is that that would be rare for most use cases. But certainly there must be much better ways of doing this- I would be interested in coding one up if it is quick-ish to do.

Regarding different similarity functions, I'm into the idea of refactoring the code! That seems cool to me, since the two methods are using fairly different algorithms, as you said. There are definitely use cases in which users would not want to use registration- for example, if users wanted to distinguish between two different crystal grains, they would be interested in environments that are the same without rotation. One set of similar environments would be one crystal grain (oriented in one direction) and the other set of similar environments would be another crystal grain (oriented in another direction).

freud/environment.pyx
```diff
@@ -532,11 +532,56 @@ cdef class EnvironmentCluster(_MatchEnv):
...
-        performing this computation. Using a distance cutoff alone could easily
-        lead to situations where a point doesn't match a cluster because a
-        required neighbor is just outside the cutoff.
+        An environment is defined by the bond vectors between a particle and its
+        neighbors as defined by :code:`env_neighbors`.
+        For example, :code:`env_neighbors= {'num_neighbors': 8}` means that every
+        particle's local environment is defined by its 8 nearest neighbors.
+        Then, each particle's environment is compared to the environments of a
+        selected list of other particles in the system, controlled by the
+        :code:`neighbors` parameter. For example, :code:`neighbors={'r_max': 3.0}`
```
cbkerr: I was confused about what was doing the selecting. Suggested fix to put what controls it first:
```suggestion
        Then, each particle's environment is compared to each of the environments of nearby
        particles that satisfy a different cutoff parameter :code:`neighbors`.
        For example, :code:`neighbors={'r_max': 3.0}`
```

freud/environment.pyx
```diff
@@ -512,7 +512,7 @@ cdef class _MatchEnv(_PairCompute):
 
 cdef class EnvironmentCluster(_MatchEnv):
     r"""Clusters particles according to whether their local environments match
-    or not, according to various shape matching metrics.
+    or not, using various shape matching metrics defined in :cite:`Teich2019`.
```
cbkerr, line 515: Redundant words?

```suggestion
    using various shape matching metrics defined in :cite:`Teich2019`.
```

freud/environment.pyx
```diff
@@ -532,11 +532,56 @@ cdef class EnvironmentCluster(_MatchEnv):
...
-        required neighbor is just outside the cutoff.
+        An environment is defined by the bond vectors between a particle and its
+        neighbors as defined by :code:`env_neighbors`.
+        For example, :code:`env_neighbors= {'num_neighbors': 8}` means that every
+        particle's local environment is defined by its 8 nearest neighbors.
+        Then, each particle's environment is compared to the environments of a
+        selected list of other particles in the system, controlled by the
+        :code:`neighbors` parameter. For example, :code:`neighbors={'r_max': 3.0}`
+        means that the environment of each particle will be compared to the
+        environment of every particle within a distance of 3.0.
```
cbkerr, line 543:
```suggestion
        means that the environment of each particle will be compared to each of the
        environments of every particle within a distance of 3.0.
```

freud/environment.pyx
```diff
@@ -532,11 +532,56 @@ cdef class EnvironmentCluster(_MatchEnv):
...
+        rotation matrix mapping one environment onto another. The Kabsch
+        algorithm assumes a one-to-one correspondence between points in the two
+        environments. However, we typically do not know which point in one environment
+        corresponds to a particular point in the other environment. The brute force
+        solution is to try all possible correspondences, but the resulting
+        combinatorial explosion makes this problem infeasible, so we use a tolerance
+        :code:`threshold` to terminate the search when we have found a permutation
+        of points that results in a sufficiently low RMSD.
+
+        .. note::
```
cbkerr, line 568: I think this note can be another paragraph as part of the explanation

freud/environment.pyx
```diff
@@ -565,26 +610,28 @@ cdef class EnvironmentCluster(_MatchEnv):
...
                 Either a :class:`NeighborList <freud.locality.NeighborList>` of
                 neighbor pairs to use in the calculation, or a dictionary of
                 `query arguments
-                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_
-                This argument is used to define the neighbors of the
-                environment that motifs are registered against. If ``None``,
-                the value provided for ``neighbors`` will be used. (Default
-                value: None).
+                <https://freud.readthedocs.io/en/stable/topics/querying.html>`_.
+                Defines the neighbors used as the environment of each particle.
```
cbkerr, line 631: Brilliantly clear :)

freud/environment.pyx
```diff
@@ -532,11 +532,56 @@ cdef class EnvironmentCluster(_MatchEnv):
...
+            Zero vectors are only used to pad the cluster vectors so that they
+            have the same shape.
+            In a future version of freud, zero-padding will be removed.
+
+        .. warning::
+
+            Comparisons between two environments are only made when both
+            environments contain the same number of particles.
+            However, this is not currently enforced, so no warning will be given
+            at runtime if such mismatched environments are provided.
```
cbkerr: other instance of Erin's suggestion:
```suggestion
            However, no runtime warning will be given
            if such mismatched environments are provided.
```

freud/environment.pyx
```diff
@@ -532,11 +532,56 @@ cdef class EnvironmentCluster(_MatchEnv):
                 global_search=False):
         r"""Determine clusters of particles with matching environments.
 
-        In general, it is recommended to specify a number of neighbors rather
-        than just a distance cutoff as part of your neighbor querying when
-        performing this computation. Using a distance cutoff alone could easily
-        lead to situations where a point doesn't match a cluster because a
-        required neighbor is just outside the cutoff.
+        An environment is defined by the bond vectors between a particle and its
```
cbkerr, line 535: This extra-specific term also jumped out at me. I like vectors as discussed later: https://github.com/glotzerlab/freud/pull/980/files#r915013462

```suggestion
        An environment is defined by the vectors between a particle and its
```

freud/environment.pyx
cbkerr: The term "bond vectors" threw me off in the first paragraph. I also prefer simply "vectors".

cbkerr (approved):
A big bump in clarity, from the perspective of a user new to environment matching
Thank you all!

I made a few small suggestions to improve clarity further.

freud/environment.pyx
Charlottez112: Adding "each of" may lead to more confusion, because it could mean that we compute multiple environments for one particle, and we compare every single environment of one neighbor particle, which I don't think is what's happening.

freud/environment.pyx
Charlottez112: See the other comment about adding "each of"

freud/environment.pyx
Charlottez112: > You mentioned the possibility of different similarity functions (plugging in a different `isSimilar` implementation). With vs without registration actually seems like that's exactly what it does. Maybe we should refactor the code to treat those as two different similarity functions, and in 3.0 allow users to specify which similarity function to use instead? That feels like a more flexible API.

This probably needs further discussion, and is out of the scope of this PR. I can create another issue if that's better.

freud/environment.pyx
Charlottez112: Could you explain why? I think "bond vector" is actually clearer, and it's consistent with the usage of "bond" in other places of the `Environment` module.

freud/environment.pyx
Charlottez112: After reading the whole thing I again I think "neighbor" is better. A major reason is that we use `env_neighbors`, and also we don't really talk about vectors until we explain why we add zero vectors. Even though we use the word neighbor a lot, I think it is pretty clear what we refer to

freud/environment.pyx
vyasr: Agreed, definitely out of scope for this PR. I'm just using these discussions as a way to chart a path forward, but we definitely don't need to address all the discussions here.

freud/environment.pyx
vyasr: My concern with using neighbors is that for someone looking at this for the first time they will easily get lost between the discussion of `env_neighbors` vs `neighbors`. IMO the documentation for both `EnvironmentCluster` and `EnvironmentMotifMatch` would be clearest if we used the word "neighbors" exceedingly sparingly:
- Environments are defined by the vectors between a particle and its nearest neighbors
- A motif is a specific environment that we expect to observe in a system.
- When clustering:
    - The nearest neighbors used to compute each particle's environment are selected based on the `env_neighbors` argument 
    - The environment of each particle is compared to the environment of its nearest neighbors, where the set of nearest neighbors is determined based on the `neighbors` argument.
- When performing motif matching:
    -  The environment of each particle is computed using the `neighbors` argument.

While writing the above, I realized that there is another significant source of confusion. The `neighbors` argument for `EnvironmentMotifMatch` is really equivalent to the `env_neighbors` argument for `EnvironmentCluster`, right?

Charlottez112: > While writing the above, I realized that there is another significant source of confusion. The `neighbors` argument for `EnvironmentMotifMatch` is really equivalent to the `env_neighbors` argument for `EnvironmentCluster`, right?

Yes, and I agree that our use of `env_neighbor` and `neighbor` can be confusing to new users, especially because of the case you mentioned here.
But because we use `neighbors` as keywords and in our examples we use `neighbors` instead of `vectors`, I feel like `neighbors` is more consistent, or at least we will have to use them interchangeably.
Since we added the example distinguishing between `env_neighbors` and `neighbors`, and extra doc in their definition to explain the different, I think it is pretty clear.

Also `neighbors` actually makes more sense once one understands how it works, because they are indeed neighbors, just used differently for this computation. Vectors are implied, and used only to specify the location of the neighbors.

freud/environment.pyx
Charlottez112: >  You're right that it depends on the order in which the elements of the environment are traversed- it is certainly not the most ideal way to do this. 

I'm still confused about @vyasr's question on how the order of traversing the vectors affects the comparison. Since we compare every single pair of vectors of 2 environments, all possible correspondence is checked right? And all of them needs to satisfy the threshold. It's just if `registration = True`, the query environment or the motif will be rotated before comparison. So if we use the same threshold for `registration = True` and `registration = False`, 2 environments are less likely to be considered match in the latter case?

freud/environment.pyx
vyasr: Rather than debating this amongst ourselves, I think the best approach here is to solicit opinions from people who aren't familiar with the code. @Charlottez112 could you ask a couple of Glotzer lab members who don't currently use environment matching to read the docs and explain what they think the docs mean? That will be much more useful than our opinions since at this point everyone on this thread already has a solid understanding of the code.

freud/environment.pyx
vyasr: All possible correspondences are _checked_, but look at the code [here](https://github.com/glotzerlab/freud/blob/master/cpp/environment/MatchEnv.cc#L329). It simply checks whether two vectors are close enough, then tries to insert. That means that for a given vector `i` in the first environment, the first vector `j` that is close enough will be set as a match, even if there is another _closer_ vectors `k` that happens to be checked after it. We could modify that logic if we wanted to, although that's out of scope for this PR. I just wanted to point out the issue.

When `registration=True` you already have a set of pairs as determined by the optimal rotation, so you don't have this problem. Of course, it's not really _that_ different in the long run. The arbitrariness is just pushed to the order of permutations used in the brute force search, whereas for `registration=False` it's the order in which pairs are checked.

freud/environment.pyx
cbkerr: I added it for parallelism because you can't directly compare a particle's environment to a _list_ of particle environments. I meant for the "each of" to indicate the iteration that happens when comparing the environment to each element in the list of neighbor particle environments.

freud/environment.pyx
cbkerr: The word "bond" doesn't add any new information.

"Vector connecting two points" is a common phrase.
Adding the word "bond" makes users wonder if there is something special happening between those particles.

Additionally, freud uses the word "bond" to mean "neighboring particles", so the term is redundant when "its neighbors" appears in the description.

freud/environment.pyx
cbkerr: > it could mean that we compute multiple environments for one particle, and we compare every single environment of one neighbor particle, which I don't think is what's happening.

Actually I do see the confusion here.
To resolve my confusion that led me to add "each of", we would need to rewrite the whole sentence.

I think this half of the change makes the paragraph more clear:

```suggestion
        Then, each particle's environment is compared to the environments of
        particles that satisfy a different cutoff parameter :code:`neighbors`.
        For example, :code:`neighbors={'r_max': 3.0}`
```

freud/environment.pyx
cbkerr: I think this works without "each of" because "environment" is singular.

freud/environment.pyx
Charlottez112: > All possible correspondences are _checked_, but look at the code [here](https://github.com/glotzerlab/freud/blob/master/cpp/environment/MatchEnv.cc#L329). It simply checks whether two vectors are close enough, then tries to insert. That means that for a given vector `i` in the first environment, the first vector `j` that is close enough will be set as a match, even if there is another _closer_ vectors `k` that happens to be checked after it. We could modify that logic if we wanted to, although that's out of scope for this PR. I just wanted to point out the issue.

I see what you mean now. Since the problem is that once `threshold` is satisfied by the pair of vectors, the `vec_map` is no longer updated for this pair of vectors, would it be easy to fix?
Also we don't ever mention why we use graphs. Is it common practice for point set registration?

freud/environment.pyx
Charlottez112: > Rather than debating this amongst ourselves, I think the best approach here is to solicit opinions from people who aren't familiar with the code. @Charlottez112 could you ask a couple of Glotzer lab members who don't currently use environment matching to read the docs and explain what they think the docs mean?

I can do that. I think I'll just ask people to take a look at the paragraph where we explain the difference between `env_neighbors` and `neighbors` at group meeting or something. I feel like it's more efficient that way, rather than requesting people's review on this already long PR.

Anyway, my point is that we cannot just simply replace the "neighbors" with "vectors". I tried that and it's not as clear and consistent. Since my prelim report and slides are due in less than 2 weeks, I don't think I'll get to this before that.

tommy-waltmann (changes requested):
The documentation looks good at this point, but someone appears to have committed an older version of the `freud-examples` submodule. Those changes will need to be reverted before this PR can be merged.

tommy-waltmann (approved):
LGTM

freud/environment.pyx
erteich: I agree with @vyasr that soliciting people's opinions will be the way to go- we can also change the "neighbors" argument in EnvironmentMatchMotif to "env_neighbors" if that would be more consistent, because you're completely right that they serve the same function. For historical reasons those terms weren't consistent but perhaps it would be better to make them consistent now that we are clarifying things. Perhaps then the use of "neighbors" (or "env_neighbors") wouldn't throw people off with respect to the neighbors vs vectors debate. I think either term would honestly probably be fine as long as we are clear in our explanation of terms. Describing motifs as sets of "vectors" in EnvironmentMotifMatch appealed to me because the motif that the user provides and might want to search for within the system can be any set of vectors, not necessarily an environment "neighborhood" already found via EnvironmentCluster. But again, I could really go either way.

freud/environment.pyx
erteich: @Charlottez112 , I'm not sure what you mean by using graphs- which part of the code are you referring to? 

Regarding the easy fix, I do think there are much more elegant and globally minimal ways of solving this matching problem. Because of the thresholding, my method will never give a "false" match (as defined by the threshold criterion) between environments, but there are edge cases where it could fail to find the match, for the reasons Vyas was saying. What I could have done originally, for example, was implement a solution to the [assignment problem](https://en.wikipedia.org/wiki/Assignment_problem) like the [Hungarian algorithm](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linear_sum_assignment.html#scipy.optimize.linear_sum_assignment). One could then check if each matched vector pair that minimizes the cost function, the sum of distances between all matched vector pairs, satisfies the threshold criterion. I have no idea about the timing of this type of solution but that would be more ideal than what I implemented, probably.

erteich (approved):

erteich (approved):

erteich (approved):

freud/environment.pyx
vyasr: I am also not sure what graphs are being referred to. In general, though, point set registration is a very hard problem with a wide range of different methods for solution. I don't recall if any of them are graph-based, but there are lots of interesting alternatives including e.g. stochastic solutions that involve things like fitting smoothed distributions of the data.

vyasr:
I've opened a couple of issues with additional suggestions for the 3.0 API.

PR #987, opened 2022-07-06
Fix Debye Structure Factor Implementation in 2D

### LINKED ISSUE #985: 2D Debye Structure Factor incorrectly uses 3D formula
@DomFijan @tommy-waltmann I'm not sure if PR #936 is correctly analyzing the physics of the 2D problem. I threw an error for 2D calculations for a reason, and should have left a comment about why. It requires a derivation which I have finally done and shown below, with a reference.

Copying a derivation of the Debye formula from https://github.com/wojdyr/debyer/blob/master/doc/index.rst and adapting it for 2D:

> Let's re-derive it once more. We start with the amplitude of the scattered wave. If we ignore non-elastic scattering and assume that each photon is scattered only once, the amplitude is a simple sum over all atoms:
> $$\Psi(\boldsymbol{Q}) = \sum_i \psi_i =\sum_i f_i \exp(-i \boldsymbol{Q} \cdot \boldsymbol{r}_i)$$
> (*i* has two unrelated meanings here, index and the imaginary unit, hopefully it's not too confusing).
> The intensity of the scattered wave is given as
>
> $$I(\boldsymbol{Q}) = \left| \Psi (\boldsymbol{Q}) \right|^2 = \Psi(\boldsymbol{Q}) \Psi^{*}(\boldsymbol{Q})$$
>
> where the complex conjugate
>
> $$\Psi^{*}(\boldsymbol{Q}) = \sum_i \psi^{*}_i = \sum_i f_i^{*} \exp(i \boldsymbol{Q} \cdot \boldsymbol{r}_i)$$
>
> Therefore, with the assumption that atomic form factors are real and $f^{*}=f$,
>
> $$I(\boldsymbol{Q}) = \sum_i \sum_j \psi_i \psi^{*}_j = \sum_i \sum_j f_i f_j \exp \left[ -i \boldsymbol{Q} \cdot \left(\boldsymbol{r}_i-\boldsymbol{r}_j \right) \right]$$
>
> The Debye scattering equation gives spherically averaged intensity. The averaging is similar to calculating the surface area of sphere. It is more elegant to use double integral, but knowing the circumference formula we can take a shortcut and use one integral:
> $$A_{sphere} = \int_0^\pi 2\pi r\sin\theta \, r \mathrm{d} \theta = 2\pi r^2 [ - \cos \theta ]_0^\pi = 4 \pi r^2$$
>
> Now, let $\gamma$ be the angle between $Q$ and $r_{ij}$.
>
> $$I(\boldsymbol{Q}) = \sum_i \sum_j f_i f_j \exp \left( -i Q r_{ij} \cos \gamma \right)$$
>
> Averaging the exponential part gives
>
> $$\left\langle \exp( -iQr_{ij} \cos \gamma ) \right\rangle = \frac{1}{4\pi r_{ij}^2} \int_0^\pi \exp\left( -iQr_{ij} \cos\gamma
    \right) ~2\pi r_{ij}^2 \sin\gamma ~\mathrm{d}\gamma$$
>
> $$= \frac{1}{2} \int_0^\pi \exp\left( -iQr_{ij} \cos\gamma \right) \sin\gamma ~\mathrm{d}\gamma$$
>
> $$= \frac{1}{2} \left[ \frac{\exp\left( -iQr_{ij} \cos\gamma \right)} {iQr_{ij}} \right]_0^\pi$$
>
> $$= \frac{\exp(iQr_{ij})-\exp(-iQr_{ij})}{2iQr_{ij}}$$
>
> $$= \frac{\sin(Qr_{ij})}{Qr_{ij}}$$
>
> which proves the Debye formula.

Because 2D uses polar coordinates and not spherical coordinates, the $\sin\gamma$ term goes away. This fundamentally changes the integral.

For 2D,

$$A_{circle} = \int_0^{2 \pi} r~\mathrm{d} \theta = 2\pi r$$

Averaging the exponential part gives

$$\left\langle \exp( -iQr_{ij} \cos \gamma ) \right\rangle = \frac{1}{2\pi r_{ij}} \int_0^\pi \exp\left( -iQr_{ij} \cos\gamma
    \right) ~2 r_{ij} ~\mathrm{d}\gamma$$

$$= \frac{1}{\pi} \int_0^\pi \exp\left( -iQr_{ij} \cos\gamma \right) ~\mathrm{d}\gamma$$

WolframAlpha says [this integrates to a Bessel function](https://www.wolframalpha.com/input?i=integrate+%281%2Fpi%29*e%5E%28-i*k*x*cos%28gamma%29%29+from+gamma%3D0+to+gamma%3Dpi) $J_0(Qr)$.

I double-checked the 3D case, and [WolframAlpha agrees that it reduces to sinc(Qr)](https://www.wolframalpha.com/input?i=integrate+1%2F2+e%5E%28-i*k*x*cos%28gamma%29%29*sin%28gamma%29+from+gamma%3D0+to+gamma%3Dpi).

Here is a reference (found after doing the derivation myself) that agrees with my assessment that this should be a Bessel function $J_0(Qr)$.

- The Debye Scattering Formula in $n$ Dimensions, Thomas Wieder. J. Math. Comput. Sci. 2 (2012), No. 4, 1086-1090. http://www.scik.org/index.php/jmcs/article/viewFile/263/120

image

tommy-waltmann:
## Description

This PR fixes the bug which was introduced in #936 which does not account for the change in geometry that occurs when going from 3D to 2D. The fix uses a cylindrical bessel function instead of a sinc function.

## Motivation and Context
Resolves: #985 

## How Has This Been Tested?

No new tests have been added yet, but they will. I am still looking for a library to compare to which will validate this calculation. ASE was used to validate the 3D case, but they can't handle 2D systems.

## Types of changes

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
                                                "MinSizeRel" "RelWithDebInfo")
 endif()
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # This setting is required when building shared libraries from object libraries.
--- a/cpp/diffraction/StaticStructureFactorDebye.cc
+++ b/cpp/diffraction/StaticStructureFactorDebye.cc
@@ -87,7 +87,7 @@ void StaticStructureFactorDebye::accumulate(const freud::locality::NeighborQuery
                     // other library's implementation is unique only for complex
                     // numbers, otherwise it just tries to call
                     // std::cyl_bessel_j.
-                    S_k += bessel::cyl_j0(std::complex<double>(k * distance));
+                    S_k += std::real(bessel::cyl_j0(std::complex<double>(k * distance)));
 #else
                     S_k += std::cyl_bessel_j(0, k * distance);
 #endif
--- a/cpp/diffraction/StaticStructureFactorDebye.cc
+++ b/cpp/diffraction/StaticStructureFactorDebye.cc
@@ -4,6 +4,7 @@
 #include <cmath>
 #include <limits>
 #include <stdexcept>
+#include <bessel-library/bessel-library.hpp>
 
 #include "NeighborQuery.h"
 #include "StaticStructureFactorDebye.h"
@@ -80,7 +81,16 @@ void StaticStructureFactorDebye::accumulate(const freud::locality::NeighborQuery
             {
                 if (box.is2D())
                 {
+                    #ifdef __clang__
+                    // clang doesn't support the special math functions, use
+                    // other library instead. The cast is needed because the
+                    // other library's implementation is unique only for complex
+                    // numbers, otherwise it just tries to call
+                    // std::cyl_bessel_j.
+                    S_k += bessel::cyl_j0(std::complex<double>(k * distance));
+                    #else
                     S_k += std::cyl_bessel_j(0, k * distance);
+                    #endif
                 }
                 else
                 {
--- a/cpp/diffraction/StaticStructureFactorDebye.cc
+++ b/cpp/diffraction/StaticStructureFactorDebye.cc
@@ -3,7 +3,6 @@
 
 #ifdef __clang__
 #include <bessel-library.hpp>
-#else
 #endif
 #include <cmath>
 #include <limits>
@@ -84,15 +83,20 @@ void StaticStructureFactorDebye::accumulate(const freud::locality::NeighborQuery
             {
                 if (box.is2D())
                 {
+                    // floating point precision errors can cause k to be
+                    // slightly negative, and make evaluating the cylindrical
+                    // bessel function impossible.
+                    auto abs_k = std::abs(k);
+
 #ifdef __clang__
-                    // clang doesn't support the special math functions, use
-                    // other library instead. The cast is needed because the
-                    // other library's implementation is unique only for complex
-                    // numbers, otherwise it just tries to call
-                    // std::cyl_bessel_j.
-                    S_k += std::real(bessel::cyl_j0(std::complex<double>(k * distance)));
+                    // clang doesn't support the special math functions in
+                    // C++17, so we use another library instead. The cast is
+                    // needed because the other library's implementation is
+                    // unique only for complex numbers, otherwise it just tries
+                    // to call std::cyl_bessel_j.
+                    S_k += std::real(bessel::cyl_j0(std::complex<double>(abs_k * distance)));
 #else
-                    S_k += std::cyl_bessel_j(0, k * distance);
+                    S_k += std::cyl_bessel_j(0, abs_k * distance);
 #endif
                 }
                 else
--- a/cpp/diffraction/StaticStructureFactorDebye.cc
+++ b/cpp/diffraction/StaticStructureFactorDebye.cc
@@ -1,10 +1,10 @@
 // Copyright (c) 2010-2020 The Regents of the University of Michigan
 // This file is from the freud project, released under the BSD 3-Clause License.
 
+#include <bessel-library/bessel-library.hpp>
 #include <cmath>
 #include <limits>
 #include <stdexcept>
-#include <bessel-library/bessel-library.hpp>
 
 #include "NeighborQuery.h"
 #include "StaticStructureFactorDebye.h"
@@ -81,16 +81,16 @@ void StaticStructureFactorDebye::accumulate(const freud::locality::NeighborQuery
             {
                 if (box.is2D())
                 {
-                    #ifdef __clang__
+#ifdef __clang__
                     // clang doesn't support the special math functions, use
                     // other library instead. The cast is needed because the
                     // other library's implementation is unique only for complex
                     // numbers, otherwise it just tries to call
                     // std::cyl_bessel_j.
                     S_k += bessel::cyl_j0(std::complex<double>(k * distance));
-                    #else
+#else
                     S_k += std::cyl_bessel_j(0, k * distance);
-                    #endif
+#endif
                 }
                 else
                 {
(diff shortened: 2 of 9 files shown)


tommy-waltmann:
Compiler support for new C++17 math functions seems a little suspect: trying to compile freud with C++17 using clang++13 does not resolve the advanced math function `std::cyl_bessel_j`

vyasr:
https://en.cppreference.com/w/cpp/compiler_support/17#C.2B.2B17_library_features

Apple clang does not yet support "Mathematical Special Functions"

joaander:
> https://en.cppreference.com/w/cpp/compiler_support/17#C.2B.2B17_library_features
> 
> Apple clang does not yet support "Mathematical Special Functions"

Well, that is unfortunate. We could conditionally error on unsupported platforms. Or find another embeddable implementation of the function. `boost` is not an option - we spent a lot of work removing boost from freud years ago. It would be a huge cost to introduce as a dependency again, especially now with the pypi wheel builds.

vyasr:
I agree that adding Boost as a dependency is not an option. However, the Boost license is compatible with the BSD-3 license, so what if we just copy in the necessary files for the one Bessel function that we need and just make a note to replace it with the `std` function once we have coverage across all platforms/compilers that we care about? All the relevant code is in [this file](https://github.com/boostorg/math/blob/develop/include/boost/math/special_functions/bessel.hpp) or in other `bessel*` files that it includes, and it shouldn't require more than minor modifications to get rid of the dependency on Boost policies etc. Bessel functions aren't terribly difficult to implement, so I wouldn't hate this solution as a stopgap.

tommy-waltmann:
That sounds like the best option at this point, I wasn't sure of the legality of copying files from other projects which is why I didn't suggest it.

tommy-waltmann:
I have played around with this, and there is a significant cascade of header files we would need to copy from the boost math library over into freud just to get `cyl_bessel_j`, to the point where we have to include a significant amount of the math library in our own source code.

Someone else can try this and maybe they can figure out how to use less header files and get all the required functions and definitions.

vyasr:
I forgot how intertwined the various components of Boost are. It looks like you've found a good alternative though. I'm fine with moving forward with something like that (haven't looked at the source yet) as a short-term solution until Apple Clang gets updated.

Lina492375qW1188 (approved):

DomFijan:
What happens when you have a quasi 2D system? Are the results the same if you use 2d and 3d formula?

tommy-waltmann:
I have not tried this myself, but my guess would be no they are not the same. I don't think they should be the same either.

DomFijan:
Users that simulate soft matter system will be interested in obtaining the results for quasi 2D systems that they approximate with 2D simulations. In their case (is this the majority of Freud users?) they would want to obtain a quasi 2D result with the 3D structure factor formula from their 2D simulation? The true 2D case is useful in some very niche settings where hypothetical 2D systems are studied (such as lattice models?). If this is the case we should make this very clear in docs or maybe even have some extra arguments for specifically computing correct s(q) for quasi2D systems.

tommy-waltmann:
We can add a note in the docs saying that for quasi-2D systems, use points in the XY plane with a 3D box.

tommy-waltmann:
Would you like to do that @DomFijan ?

joaander (approved):
Looks great, thanks!

cpp/diffraction/StaticStructureFactorDebye.cc
```diff
@@ -78,7 +81,28 @@ void StaticStructureFactorDebye::accumulate(const freud::locality::NeighborQuery
             double S_k = 0.0;
             for (const auto& distance : distances)
             {
-                S_k += util::sinc(k * distance);
+                if (box.is2D())
+                {
+                    // floating point precision errors can cause k to be
+                    // slightly negative, and make evaluating the cylindrical
+                    // bessel function impossible.
+                    auto abs_k = std::abs(k);
```
bdice, line 89: Should this be max(0, k)?

extern/bessel-library.hpp
```diff
@@ -0,0 +1,105 @@
+/*
+ * Taken from https://gtihub.com/jodesarro/bessel-library.git, distributed under
```
bdice, line 2: “github” has a typo

freud/diffraction.pyx
```diff
@@ -96,6 +96,18 @@ cdef class StaticStructureFactorDebye(_StaticStructureFactor):
     <https://en.wikipedia.org/wiki/Structure_factor>`__. For a full derivation
     see :cite:`Farrow2009`. Note that the definition requires :math:`S(0) = N`.
 
+    .. note::
+
+        For 2D systems freud uses the bessel function :math:`J_0` instead of the
+        :math:`\text{sinc}` function in the equation above. See this `link
+        <http://www.scik.org/index.php/jmcs/article/viewFile/263/120>`__
```
bdice, line 103: This should be a proper citation.

freud/diffraction.pyx
```diff
@@ -96,6 +96,18 @@ cdef class StaticStructureFactorDebye(_StaticStructureFactor):
     <https://en.wikipedia.org/wiki/Structure_factor>`__. For a full derivation
     see :cite:`Farrow2009`. Note that the definition requires :math:`S(0) = N`.
 
+    .. note::
+
+        For 2D systems freud uses the bessel function :math:`J_0` instead of the
```
bdice, line 101: “Bessel” should be capitalized.

CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
                                                "MinSizeRel" "RelWithDebInfo")
 endif()
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
bdice, line 21: FYI, we may need a newer manylinux version tag to support C++17. This may cause problems at the next release time.

bdice (commented):
A few notes — thanks @tommy-waltmann for this!

tommy-waltmann:
There are quite a few comments here, some will need addressing before the next release. Would you like to address them in followup PR?

PR #997, opened 2022-08-11
Dependabot file path fix

tommy-waltmann:
## Description

This PR fixes an issue with file paths dependabot uses to check for dependencies.

## Motivation and Context

On #951 , the path to environment files used by CI was changed and dependabot wasn't reconfigured to search the new paths to update those dependencies automatically.

## How Has This Been Tested?

Dependabot doesn't explicitly say how it searches file paths, so we can't know for sure if this is working correctly before it is merged, all we can do is follow other examples, such as what is done by [gsd](https://github.com/glotzerlab/gsd/blob/trunk-patch/.github/dependabot.yml).

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation improvement (updates to user guides, docstrings, or developer docs)

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.
- [ ] I have updated the documentation (if relevant).
- [ ] I have added tests that cover my changes (if relevant).
- [ ] All new and existing tests passed.
- [ ] I have updated the [credits](https://github.com/glotzerlab/freud/blob/master/doc/source/reference/credits.rst).
- [ ] I have updated the [Changelog](https://github.com/glotzerlab/freud/blob/master/ChangeLog.md).

PR #1007, opened 2022-08-12
Re-arrange Some Requirements Files

tommy-waltmann:
## Description

This PR changes the directory of the oldest requirements file so dependabot will not try to bump the versions anymore.

## Motivation and Context

On #997 , we attempted a fix so dependabot would find our testing requirements files and resume bumping them monthly again. This attempt made it so dependabot also tried to bump the oldest testing requirements file as well, which is undesirable.

## How Has This Been Tested?

Similar to #997 , we cannot test that dependabot doesn't sense the oldest requirements file until the PR is merged, we can only follow other examples. We can, however, observe that the python 3.6 tests pass, meaning that the files path adjustments were done correctly.

## Types of changes

## Checklist:


bdice (approved):
Approving this -- but I'd suggest adopting the structure we _know_ works for signac, rather than making up a new structure that is untested...

joaander:
> Approving this -- but I'd suggest adopting the structure we _know_ works for signac, rather than making up a new structure that is untested...

Agreed. I suggested that Tommy use a separate directory. However, I was thinking of a separate directory *outside* the `env-files` directory (specified as the root in the dependabot config) so that the behavior would not change subject future changes in the way dependabot scans.

I worry that this PR as is won't change anything. *signac*'s config demonstrates that dependabot performs some amount of recursive searching. It is possible that it will search inside `oldest`, though it may not (the behavior is not documented).

tommy-waltmann:
Okay, I can re-arrange some things. I created a different structure previously because I was unaware at the time about the troubles with dependabot search paths.

tommy-waltmann:
Going to wait until this round of dependabot PRs gets merged before moving further with this. Its creating merge conflicts that I only want to go about resolving once after all the dependency versions are bumped.

bdice:
Sorry, bumped the close button. I was trying to comment.

If you update/move the requirements files first, dependabot will close or recreate the corresponding PRs for new versions. You won’t have to fix merge conflicts.

tommy-waltmann:
This appears to be working right as of now. I find it awkward, however, that the conda environment files are not in the same directories as their corresponding pip requirements files due to the new structure that we need for dependabot to work right. I don't know if it would violate common practice or not to put conda env files in a `requirements/` directory.

.github/dependabot.yml
```diff
@@ -1,10 +1,5 @@
 version: 2
 updates:
-  - package-ecosystem: pip
```
bdice, line 3: This PR disables dependabot for pip packages entirely. That seems wrong.

bdice (changes requested):

.github/dependabot.yml
tommy-waltmann: Yes, that is wrong.

bdice (approved):

joaander (commented):
I have no comments that I have not already provided in other threads.

PR #1010, opened 2022-08-25
YAML style issue templates

tommy-waltmann:
## Description

This PR updates the issue templates to the new issue forms. I have also added a form for tasks, so we now have bug reports, feature requests, and tasks.

## Motivation and Context

Github created a more appealing way of allowing users to file issues.

## How Has This Been Tested?

We cannot test this until it gets merged. I followed much of the style used in glotzerlab/hoomd-blue#1384 to be safe.

## Types of changes

## Checklist:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
+name: Bug Report
+description: Report a bug to help us improve.
```
vyasr: We should add default labels as well. freud should be better about labeling issues in general if possible. This one should be 'bug'.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+      label: freud Version
+      description: What version of freud are you using?
+      placeholder: v2.10.0
+    validations:
+      required: true
+  - type: input
+    attributes:
+      label: Python Version
+      description: What version of python are you using?
+      placeholder: v3.8.2
```
vyasr, line 36: Given how many people are running on supercomputing clusters with arbitrarily slow version updating, I wouldn't include a default here.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+    attributes:
+      label: Error Output
+      description: Please indicate the output from the above script and what you expect the output to be.
+    validations:
+      required: true
+  - type: input
+    attributes:
+      label: freud Version
+      description: What version of freud are you using?
+      placeholder: v2.10.0
```
vyasr: We should update [the bumpversion config in setup.cfg](https://github.com/glotzerlab/freud/blob/master/setup.cfg#L25) to also update this file.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+      label: Python Version
+      description: What version of python are you using?
+      placeholder: v3.8.2
+  - type: dropdown
+    attributes:
+      label: System Platform
+      description: What platform are you using?
+      options:
+        - MacOS
+        - Linux
```
vyasr, line 43: Windows too.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+  - type: dropdown
+    attributes:
+      label: Installation method
+      description: How did you install freud?
+      options:
+        - Download from PyPI
+        - Download from conda-forge
+        - built from source
```
vyasr, line 53: Containers are another option that I assume is reasonable common given the containers that @joaander maintains. Is there a way to include an "other" option where users can provide more info on alternate methods of installation? That would save us from having to maintain an exhaustive list here, just a sufficiently comprehensive collection of common methods.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+  - type: dropdown
+    attributes:
+      label: Developer
+      description: Who should fix this?
+      options:
+        - I can!
+        - I can, but I'll need some help and guidance.
+        - Someone else should fix this.
```
vyasr: Can we reword this to something that sounds friendlier and less demanding?

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,29 @@
+name: Feature Request
+description: Suggest a new feature for freud
+body:
+  - type: textarea
+    attributes:
+      label: Problem
```
vyasr: Calling this "Problem" sounds odd, but given the context and that the next section is the "Description" I'm not sure what alternative to suggest. We can move forward as is, just making this comment in case someone else comes up with a better alternative.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,29 @@
...
+        label: Alternatives
+        description: Describe any alternatives you have considered.
+  - type: dropdown
+    attributes:
+      label: Developer
+      description: Who should implement the new feature?
+      options:
+        - I can!
+        - I can, but I'll need some help/guidance.
+        - Someone else will need to implement this feature.
```
vyasr: Same as in the bug report, reword this.

vyasr (changes requested):
Do you think it's worth waiting for the next HOOMD minor release to see whether there are any issues to iron out, or do you think we'll get around to another freud release first?

tommy-waltmann:
We don't have to wait until the next hoomd minor release to see how the issue forms look. AFAIK, this is all done through github on the main/master/trunk-minor branch. @vyasr you should be able to try and file an issue in hoomd and see how the new forms look right now.

.github/ISSUE_TEMPLATE/bug_report.yml
tommy-waltmann: AFAIK, the placeholder isn't a default option, its something that is shown in faded text to show users how to format their response. Try filing a hoomd issue so you can see how these look.

.github/ISSUE_TEMPLATE/bug_report.yml
tommy-waltmann: I can add a container option, but I think those 4 options form a comprehensive set. Can you think of another way a user could install/use freud that doesn't fall under those categories?

I can add these options as placeholders in a text area if we need to allow more flexibility.

.github/ISSUE_TEMPLATE/feature_request.yml
tommy-waltmann: I mostly tried to copy the formatting from the previous issue templates, where the first section was something like "What is the problem this feature is trying to solve?". Feel free to suggest another way of wording the title.

.github/ISSUE_TEMPLATE/bug_report.yml
tommy-waltmann: I thought this sounded friendlier than the typical "me/someone else" options, do others think this sounds demanding?

.github/ISSUE_TEMPLATE/bug_report.yml
joaander:
```suggestion
description: Report a bug in freud.
```

.github/ISSUE_TEMPLATE/bug_report.yml
joaander: Yes, the placeholder disappears when the user clicks in the text box.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+  - type: textarea
+    attributes:
+      label: Description
+      description: A clear and concise description of the bug.
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Steps to Reproduce
+      description: Please provide a *minimal* script to reproduce the bug.
```
joaander, line 14: Consider `render: python` here so the user automatically enters text in a code block.

.github/ISSUE_TEMPLATE/bug_report.yml
joaander: The real question is: If a user installed freud via some other method, would we even want to support it? Could we, if the binary build is out of our control?

Yes, add the `glotzerlab-software` container to the list.

.github/ISSUE_TEMPLATE/bug_report.yml
joaander: The language I used in HOOMD is:
```
        - I will submit a pull request that fixes this problem.
        - Would someone else please fix this?
```

.github/ISSUE_TEMPLATE/feature_request.yml
joaander: I use "Description" in HOOMD.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,29 @@
+name: Feature Request
+description: Suggest a new feature for freud
+body:
+  - type: textarea
+    attributes:
+      label: Problem
+      description: Is your feature request related to a problem? Please describe.
```
joaander: "problem" is mentioned again here. For HOOMD, I use the language "What new capability would you like in HOOMD-blue?"

.github/ISSUE_TEMPLATE/feature_request.yml
joaander: In HOOMD, I use:
```
        - I will submit a pull request that implements this feature.
        - Would someone else please implement this?
```

joaander (changes requested):
Also add `.github/ISSUE_TEMPLATE/config.yml`. In it:
- Choose whether to allow blank issues or not.
- Populate the `contact_links:` section.

joaander:
> We don't have to wait until the next hoomd minor release to see how the issue forms look. AFAIK, this is all done through github on the main/master/trunk-minor branch. @vyasr you should be able to try and file an issue in hoomd and see how the new forms look right now.

Unfortunately, I changed the default branch on hoomd to `trunk-minor`, so the forms disappeared. They will come back after the next minor release. You can look at GSD as an example: https://github.com/glotzerlab/gsd/issues/new/choose

vyasr:
> Unfortunately, I changed the default branch on hoomd to `trunk-minor`, so the forms disappeared. They will come back after the next minor release. You can look at GSD as an example: https://github.com/glotzerlab/gsd/issues/new/choose

Right, I thought that with the choice of default branch we wouldn't see it. Thanks for pointing out that GSD has it, though.

.github/ISSUE_TEMPLATE/bug_report.yml
vyasr: Thanks for that, I didn't realize how the placeholder works. I see it now.

.github/ISSUE_TEMPLATE/bug_report.yml
vyasr: I'm fine with not adding anything other than containers. The containers (specifically `glotzerlab-software`) should be there.

.github/ISSUE_TEMPLATE/bug_report.yml
vyasr: I think that sounds nicer. Saying that someone else *should* fix it is what sounds demanding to me. It's a command instead of a request.

.github/ISSUE_TEMPLATE/feature_request.yml
vyasr: There is also a "Description" section below, though. Otherwise I would also have suggested "Description" here.

As I said, I don't have a great alternative to suggest so if neither of you can come up with one it's fine with me if to move forward as is.

tommy-waltmann: I may just reformat the first few sections here, we don't have to strictly follow the wording/section layout that was used before.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,62 @@
...
+        - built from source
+  - type: dropdown
+    attributes:
+      label: Developer
+      description: Who should fix this?
+      options:
+        - I will submit a pull request that fixes this problem.
+        - Would someone else please fix this?
+      validations:
+        required: true
```
joaander:
```suggestion
```
On second though, let's remove this section. Users that plan to implement a fix can mention so in the comments.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,29 @@
...
+        description: What additional information is useful to understand this request?
+  - type: dropdown
+    attributes:
+      label: Developer
+      description: Who should implement the new feature?
+      options:
+        - I will submit a pull request that implements this feature.
+        - Would someone else please implement this?
+    validations:
+      required: true
```
joaander:
```suggestion
```

joaander (approved):
Thanks! You can merge this as is or apply my suggested edits first.

.github/ISSUE_TEMPLATE/bug_report.yml
tommy-waltmann: How about we just remove the required validation?

bdice (approved):

bdice:
I think this is ready for merge after @vyasr approves (currently requested changes).

vyasr (approved):
LGTM

Issue #991, opened 2022-07-20
More options to color Voronoi diagram by


dnusha1994:
**Is your feature request related to a problem? Please describe.**
**Using Freud to do Voronoi tesselation of lipid bilayer phosphates**

**Describe the solution you'd like**
**It would be useful to have the option to colour the Voronoi polygons by their area as well as by the number of sides they have. Is there any way this could be achieved?**

**Describe alternatives you've considered**

**Additional context**

tommy-waltmann:
Yes, that would be a simple and useful addition.

The ideal solution to achieve this would involve changing the argument `color_by_sides` for the `Voronoi.plot` method. It would need to be generalized to something like `color_by`, which instead of taking a bool, would take one of a set of acceptable strings (i.e. `color_by='sides'`, or `color_by='volume'`, etc.).

This proposed solution would be API breaking, though, so it would have to be pushed to a v3 release. A non-API breaking way to handle this would be to add an extra argument named something like `color_by_volume=bool`, which is clunkier, but would not force a major version bump. @bdice @vyasr thoughts?

In the meantime, a matplotlib script to achieve the desired color scheme can always be written by hand.

bdice:
I tend to prefer custom matplotlib for user-defined visualizations. However, I would support the plan to make a PR with API-breaking changes in 3.0 if someone wants to make those changes. @dnusha1994 Would you like to contribute that feature?

When I designed other plotting APIs in freud, I tended towards forwarding `*args, **kwargs`, which allows passing `c=some_array, cmap="viridis"` or similar. That's similar to how [pandas `DataFrame.plot`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.html) approaches the problem -- just let that logic be handled by matplotlib. However, that only works well for plotting points (with scatter). The patch collections used for Voronoi need more complicated logic to set the colors. See: https://github.com/glotzerlab/freud/blob/f0e81994b2912dbca3e335c1b8f178b4fb9675b9/freud/plot.py#L442-L472

vyasr:
I agree that we shouldn't overengineer plotting code in freud; the plotting available here should be pretty minimal, just enough for users to validate and get a feel for their analysis. This particular change doesn't seem too heavyweight, though, and I'm fine with adding it. I think the API-breaking change to `color_by` is probably the way to go since in the interim a sample matplotlib script should suffice to get users through the interim period.

tommy-waltmann:
addressed in #1012

PR #1012, opened 2022-09-22
Color Voronoi Diagrams by Area

tommy-waltmann:
## Description

This option adds the ability to color voronoi plots by the area of the Voronoi polytopes. It involves an API change from the option `color_by_sides=True/False` to `color_by='sides'/''area'/None`.

## Motivation and Context
Resolves: #991 

## How Has This Been Tested?
I have tested this locally with a variety of arguments and input values. See screenshots section below.

## Screenshots

With `voro.plot()`
voronoi_color_by_None

With `voro.plot(color_by='sides')`
voronoi_color_by_sides

With `voro.plot(color_by='area')`
voronoi_color_by_area

## Types of changes

## Checklist:


tommy-waltmann:
NOTE: The changes in this PR will require an update to the examples repository.

doc/source/gettingstarted/migration.rst
```diff
@@ -23,3 +23,6 @@ Overview of API Changes
     * - Normalize small system RDFs to 1.
       - ``freud.density.RDF(..., normalize=True)``
       - ``freud.density.RDF(..., normalization_mode='finite_size')``
+    * - Color voronoi diagram by area of the cells.
+      - ``N/A``
+      - ``voro.plot(..., color_by='area')``
```
b-butler: The breaking change is in setting `color_by_sides=False`. That should be what is shown here.

freud/plot.py
```diff
@@ -408,21 +408,22 @@ def density_plot(density, box, ax=None):
...
-        polytopes (:class:`numpy.ndarray`):
-            Array containing Voronoi polytope vertices.
         ax (:class:`matplotlib.axes.Axes`): Axes object to plot.
             If :code:`None`, make a new axes and figure object.
             (Default value = :code:`None`).
-        color_by_sides (bool):
-            If :code:`True`, color cells by the number of sides.
-            If :code:`False`, random colors are used for each cell.
-            (Default value = :code:`True`).
+        color_by (bool):
```
b-butler:
```suggestion
        color_by (str):
```

freud/plot.py
```diff
@@ -440,19 +441,25 @@ def voronoi_plot(box, polytopes, ax=None, color_by_sides=True, cmap=None):
...
         colors = np.random.RandomState().permutation(np.arange(len(patches)))
         num_colors = np.unique(colors).size
+    else:
+        raise RuntimeError(f"Invalid color_by option {color_by}.")
 
     # Ensure we have enough colors to uniquely identify the cells
-    if cmap is None:
-        if color_by_sides and num_colors <= 10:
+    continuous_colorby_options = ["area"]
+    if cmap is None and color_by not in continuous_colorby_options:
```
b-butler, line 461:
```suggestion
    if cmap is None and num_colors is not None:
```
This is equivalent and simpler.

freud/plot.py
```diff
@@ -485,11 +493,12 @@ def voronoi_plot(box, polytopes, ax=None, color_by_sides=True, cmap=None):
     ax.set_aspect("equal", "datalim")
 
     # Add colorbar for number of sides
-    if color_by_sides:
+    color_by_labels = dict(sides="Number of sides", area="Polytope Area")
```
b-butler, line 496:
```suggestion
```

freud/plot.py
```diff
@@ -485,11 +493,12 @@ def voronoi_plot(box, polytopes, ax=None, color_by_sides=True, cmap=None):
...
+    if color_by is not None:
         ax_divider = make_axes_locatable(ax)
         cax = ax_divider.append_axes("right", size="7%", pad="10%")
         cb = Colorbar(cax, patch_collection)
-        cb.set_label("Number of sides")
+        cb.set_label(color_by_labels[color_by])
```
b-butler, line 501:
```suggestion
        cb.set_label(
            "Number of sides" if color_by == "sides" else "Polytope Area")
```

b-butler (changes requested):
Left some comments to address.

freud/plot.py
tommy-waltmann: I wrote this in a way that would allow us to more easily add options for the `color_by` argument. When we want to add a new option, we can just add it to the proper list/dictionary. The same reasoning applies to the next 2 comments you left.

b-butler: Sure, but is that something we are really planning for? Being flexible is great but here it just adds code (in my opinion). There is also no guarantee that the next option, if there ever is one, would be continuous. I wouldn't try to block the PR for this though, so if you want to do it this way, then use `{"area"}` since a set is really what you want.

tommy-waltmann: If the next option is not continuous, then we would just not add it to the list named `continuous_colorby_options`, which intuitively makes sense. If the next option was continuous, then we would only add an element to a list and everything would just work properly. I chose to write it this way so it will be more easily extended for more options. Also, defining the list allows me an extra variable to explain what's going on, which is naturally more self-documenting than the suggested change.

The same logic holds for the other suggested changes below this one.

b-butler (approved):
:+1:

PR #1013, opened 2022-09-27
RDF Normalization Mode Argument

tommy-waltmann:
## Description

This PR changes the `normalize` argument to `freud.density.RDF` to be more descriptive, and clarifies what the default value is.

## Motivation and Context

There was a lot of discussion about how the RDF should be normalized, what the options should be, and how the API should look in the issues linked below.

Resolves: #635 
Related to #396

## How Has This Been Tested?

I have added a test for both of the valid arguments, and a test case for the invalid arguments.

## Types of changes

## Checklist:

### DIFF
--- a/cpp/density/RDF.h
+++ b/cpp/density/RDF.h
@@ -16,8 +16,16 @@ namespace freud { namespace density {
 class RDF : public locality::BondHistogramCompute
 {
 public:
+
+    enum NormalizationMode
+    {
+        infer,
+        finite_size
+    };
+
     //! Constructor
-    RDF(unsigned int bins, float r_max, float r_min = 0, bool normalize = false);
+    RDF(unsigned int bins, float r_max, float r_min = 0,
+            NormalizationMode normalization_mode = NormalizationMode::infer);
 
     //! Destructor
     ~RDF() override = default;
@@ -51,7 +59,7 @@ public:
     }
 
 private:
-    bool m_normalize;                //!< Whether to enforce that the RDF should tend to 1 (instead of
+    NormalizationMode m_norm_mode;   //!< Whether to enforce that the RDF should tend to 1 (instead of
                                      //!< num_query_points/num_points).
     util::ManagedArray<float> m_pcf; //!< The computed pair correlation function.
     util::ManagedArray<float>
--- a/cpp/density/RDF.h
+++ b/cpp/density/RDF.h
@@ -16,7 +16,6 @@ namespace freud { namespace density {
 class RDF : public locality::BondHistogramCompute
 {
 public:
-
     enum NormalizationMode
     {
         infer,
@@ -25,7 +24,7 @@ public:
 
     //! Constructor
     RDF(unsigned int bins, float r_max, float r_min = 0,
-            NormalizationMode normalization_mode = NormalizationMode::infer);
+        NormalizationMode normalization_mode = NormalizationMode::infer);
 
     //! Destructor
     ~RDF() override = default;
--- a/cpp/density/RDF.h
+++ b/cpp/density/RDF.h
@@ -16,15 +16,16 @@ namespace freud { namespace density {
 class RDF : public locality::BondHistogramCompute
 {
 public:
+    //! Enum for each normalization mode
     enum NormalizationMode
     {
-        infer,
+        exact,
         finite_size
     };
 
     //! Constructor
     RDF(unsigned int bins, float r_max, float r_min = 0,
-        NormalizationMode normalization_mode = NormalizationMode::infer);
+        NormalizationMode normalization_mode = NormalizationMode::exact);
 
     //! Destructor
     ~RDF() override = default;
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -78,7 +78,7 @@ release = "2.11.0"
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = 'en'
+language = "en"
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -78,7 +78,7 @@ release = "2.11.0"
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
--- a/doc/source/gettingstarted/migration.rst
+++ b/doc/source/gettingstarted/migration.rst
@@ -1,0 +1,25 @@
+.. _migration:
+
+============================
+Migration to freud Version 3
+============================
+
+Version 3 of the freud library introduces a few breaking changes to make the API
+more intuitive and facilitate future development. This guide explains how to
+alter scripts which use freud v2 APIs so they can be used with freud version 3.
+
+Overview of API Changes
+=======================
+
+.. list-table::
+    :header-rows: 1
+
+    * - Goal
+      - v2 API
+      - v3 API
+    * - Use default RDF normalization.
+      - ``freud.density.RDF(..., normalize=False)``
+      - ``freud.density.RDF(..., normalization_mode='infer')``
+    * - Normalize small system RDFs to 1.
+      - ``freud.density.RDF(..., normalize=True)``
+      - ``freud.density.RDF(..., normalization_mode='finite_size')``
--- a/doc/source/gettingstarted/migration.rst
+++ b/doc/source/gettingstarted/migration.rst
@@ -6,7 +6,7 @@ Migration to freud Version 3
 
 Version 3 of the freud library introduces a few breaking changes to make the API
 more intuitive and facilitate future development. This guide explains how to
-alter scripts which use freud v2 APIs so they can be used with freud version 3.
+alter scripts which use freud v2 APIs so they can be used with freud v3.
 
 Overview of API Changes
 =======================
@@ -19,7 +19,7 @@ Overview of API Changes
       - v3 API
     * - Use default RDF normalization.
       - ``freud.density.RDF(..., normalize=False)``
-      - ``freud.density.RDF(..., normalization_mode='infer')``
+      - ``freud.density.RDF(..., normalization_mode='exact')``
     * - Normalize small system RDFs to 1.
       - ``freud.density.RDF(..., normalize=True)``
       - ``freud.density.RDF(..., normalization_mode='finite_size')``
--- a/freud/density.pyx
+++ b/freud/density.pyx
@@ -582,11 +582,15 @@ cdef class RDF(_SpatialHistogram1D):
     systems of 100 particles the RDF will differ by 1%).
 
     .. note::
-        The ``normalization_mode`` argument should not be used if
-        :code:`query_points` is provided as a different set of points, or if
-        unusual query arguments are provided to :meth:`~.compute`, specifically
-        if :code:`exclude_ii` is set to :code:`False`. This normalization is
-        not meaningful in such cases and will simply convolute the data.
+        For correct normalization behavior when using
+        ``normalization_mode='infer'``, let the set of points be either: 1) the
+        same as the set of query points or 2) completely disjoint from the set
+        of query points.
+
+    .. note::
+        For correct normalization behavior when using
+        ``normalization_mode='finite_size'``, do not allow particles to be their
+        own neighbor.
 
     .. note::
         **2D:** :class:`freud.density.RDF` properly handles 2D boxes.
--- a/freud/density.pyx
+++ b/freud/density.pyx
@@ -611,7 +611,7 @@ cdef class RDF(_SpatialHistogram1D):
             ``finite_size``, adds an extra rescaling factor of
             :math:`\frac{N_{query\_points}}{N_{query\_ponts} - 1}` so the RDF
             values will tend to 1 at large :math:`r` for small systems (Default
-            value = :code:`'infer'`).
+            value = :code:`'exact'`).
     """
     cdef freud._density.RDF * thisptr
 
--- a/freud/density.pyx
+++ b/freud/density.pyx
@@ -585,12 +585,12 @@ cdef class RDF(_SpatialHistogram1D):
         For correct normalization behavior when using
         ``normalization_mode='exact'``, let the set of points be either: 1) the
         same as the set of query points or 2) completely disjoint from the set
-        of query points.
+        of query points (points shouldn't contain any particles in query points).
 
     .. note::
         For correct normalization behavior when using
         ``normalization_mode='finite_size'``, do not allow particles to be their
-        own neighbor.
+        own neighbor (:code:`exclude_ii` must be set to :code:`False`).
 
     .. note::
         **2D:** :class:`freud.density.RDF` properly handles 2D boxes.
--- a/freud/density.pyx
+++ b/freud/density.pyx
@@ -610,10 +610,11 @@ cdef class RDF(_SpatialHistogram1D):
     cdef freud._density.RDF * thisptr
 
     def __cinit__(self, unsigned int bins, float r_max, float r_min=0,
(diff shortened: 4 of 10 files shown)


DomFijan:
I can take a look at this on Friday.

freud/density.pyx
```diff
@@ -576,10 +576,17 @@ cdef class RDF(_SpatialHistogram1D):
...
+    rescale the results and force the long range behavior to 1. Note that this
+    option will have little to no effect on larger systems (for example, for
+    systems of 100 particles the RDF will differ by 1%).
+
+    .. note::
+        The ``normalization_mode`` argument should not be used if
+        :code:`query_points` is provided as a different set of points, or if
+        unusual query arguments are provided to :meth:`~.compute`, specifically
+        if :code:`exclude_ii` is set to :code:`False`. This normalization is
+        not meaningful in such cases and will simply convolute the data.
```
DomFijan: Could you raise an Error in compute if you explicitly check for these two conditions? As far as I can see this is not checked for.

freud/density.pyx
```diff
@@ -576,10 +576,17 @@ cdef class RDF(_SpatialHistogram1D):
...
-    particles the RDF will differ by 1%).
+    deviation is noticeable, the ``normalization_mode`` argument may be used to
+    rescale the results and force the long range behavior to 1. Note that this
+    option will have little to no effect on larger systems (for example, for
+    systems of 100 particles the RDF will differ by 1%).
+
+    .. note::
+        The ``normalization_mode`` argument should not be used if
+        :code:`query_points` is provided as a different set of points, or if
+        unusual query arguments are provided to :meth:`~.compute`, specifically
```
DomFijan: unusual? You already explicitly mention which argument is problematic. Are there any others?

DomFijan (commented):
This looks good. I just two questions related to docstrings.

freud/density.pyx
tommy-waltmann: To be honest, I am not sure exact what is meant by "unusual" in this docstring because I didn't write it originally. I chose not to change the wording because of that. Perhaps @vyasr or @bdice can comment on the original intent of using the word "unusual"?

freud/density.pyx
tommy-waltmann: For a typical use case of the RDF module, `compute` will be called repeatedly on many frames in a trajectory in order to accumulate a smooth RDF curve. Checking the conditions in the code would create significant, often unnecessary overhead, so we choose to document these cases instead of checking them explicitly for the performance benefit.

freud/density.pyx
DomFijan: Next sentence also states that if such unusual arguments are provided the data will be "convoluted"? Does that mean it will be wrong? I'd avoid using word convoluted here. In this specific context convolution can also be interpreted as a mathematical operation...

vyasr:
I've updated the PR description to say that this is related to #396 but not that it resolves it. There is additional work to be done to resolve that issue.

freud/density.pyx
```diff
@@ -593,27 +600,23 @@ cdef class RDF(_SpatialHistogram1D):
...
-            not meaningful in such cases and will simply convolute the data.
-
+        normalization_mode (str, optional):
+            There are two valid string inputs for this argument. The first
+            option, ``infer``, handles the normalization as shown mathematically
+            at the beginning of this class's docstring. The other option,
+            ``finite_size``, adds an extra rescaling factor of
+            :math:`\frac{N_{query\_points}}{N_{query\_ponts} - 1}` so the RDF
+            values will tend to 1 at large :math:`r` for small systems (Default
+            value = :code:`'infer'`).
```
vyasr: We actually need to do a little more work for `infer`. The current behavior of `infer` is what we want if `query_points is points` (or equivalently if the user doesn't specify `query_points`). In the case where `query_points != points`, though, I believe that we should instead be normalizing by `N_query_points * N_points`.

freud/density.pyx
vyasr: I agree, this is poor wording. Let me try to explain it better.

If you follow through the derivation of the `n`-point correlation function for a system of `N` particles, you'll find that it has a prefactor of the form `N!/(N-n)!` For a 2-point correlation, this is simply `N * (N-1)`. There's an additional factor of `N^2` that pops up (IIRC from the volume integral, but it's been a long time so don't quote me on that) that results in a prefactor that looks like `(N-1)/N`. For an ideal gas, all the other terms in the correlation function are unity, so `g(r) -> (N-1)/N = 1 - 1/N`. In the thermodynamic limit of `N->\infty`, `1-1/N -> 1`.

The inconsistency that led to the addition of the `normalize` flag arises because when most packages compute the pair correlation function they end up just dividing by `N` in a place where freud chooses to be exact and divide by `N-1`. For large systems the difference is negligible, but it is noticeable when working with small systems (aka far from the thermodynamic limit). All that `normalize` does is multiply the result by `N/N-1` to flip between freud's convention and the convention used by other tools.

Unfortunately, the normalization by values like `N` is only applicable when working with a single set of particles. The actual values get more complicated when you want to compute, for example, the RDF between two different subsets of points in the system that represent two different types of particles. In that situation, the math demands a different normalization. Ultimately the normalization constant boils down to something proportional to the number of pairs of points considered for the RDF. Other tools like VMD or LAMMPS have access to the whole system, so when a user requests the RDF computation between two subsets of the system, those tools can explicitly compute the number of pairs by looking at how many distinct pairs exist between the two sets of particles. freud cannot do this because the only information freud has is the two sets of input positions; it doesn't know if some of those actually represent the same particle (and we don't want to try to infer that based on overlapping positions, that's out of scope for freud). 

The point of this warning is basically to state that the `N->N-1` transition is _only_ valid in a case where the RDF is computed based on a cutoff distance (`r_max`) where `exclude_ii=True` and `query_points=points`. If `exclude_ii=False`, then the division is already by `N` rather than `N-1` because we're including a particle's interaction with itself (this is where the `N-1` comes from in the normal case of `exclude_ii=True`) so the additional normalization would produce the wrong result. If `query_points != points`, then any simple rescaling will be wrong unless the two sets of points are completely disjoint (and freud can't know that), so the `normalize` behavior is pretty useless.

The purpose of switching from `normalize` to `normalize_mode` is to be able to handle some of the simplest cases where we know the normalization, even if it has a form different from just `N/N-1`. The obvious cases that we can handle are:
- `finite_size`: Apply the finite size correction of `N/N-1` so that even for small system sizes `g(r)->1` as `r->\infty`. This option will give a user incorrect results if they specify `exclude_ii=False` and will result in 
`g(r)->1+1/N`.
- `infer`: If `query_points == points`, apply the finite size correction, otherwise don't. This option will give incorrect results if, for instance, `query_points != points` but there are some shared points between the two sets.
- `none`: Don't apply any normalization at all. This mode doesn't currently exist, but would essentially not do any division based on `N` and leave that up to the user. This mode is what advanced users will need if they use `query_points` that are distinct from but have a nonempty intersection with `points`.

Hopefully that explanation clears up any confusion here.

tommy-waltmann:
> I've updated the PR description to say that this is related to #396 but not that it resolves it. There is additional work to be done to resolve that issue.

Good point, I'm not currently planning on implementing the new solution on this PR, so it wouldn't resolve that issue.

freud/density.pyx
tommy-waltmann: Is your comment here suggesting that we implement a new normalization solution in this PR? You've previously stated (see your last comment in #635) that we may not want to do that in order to move things along a bit quicker.

freud/density.pyx
tommy-waltmann: Thanks for the detalied explanation! I'd like to respond to one of your bullet points at the bottom because I think the information is wrong.

The infer mode does not, in any case, apply the finite size correction. The user must specify that they want the finite size correction using the `normalization_mode='finite_size'`, otherwise it will not happen. This was the previous behavior as well when `normalization=False`.

vyasr: You're right, my original post was a little bit imprecise. What I was calling "infer" isn't exactly the current behavior. I had intended to use the current behavior when `query_points == points` and the finite size correct when `query_points == points`. That's why I called it "infer"; it infers the appropriate behavior in at least the two simplest cases.

I think we have two options:
1. Implement the "infer" behavior that I described above. It would be fairly easy, but it is nonzero work.
2. Just have the default remain exactly the same as the current behavior `normalization=False` but name the mode something (`exact`? not sure) to more precisely characterize what it's doing: it's the correct calculation when the RDF is computed for a single set of points against itself.

The more I think about it, the more inclined I am to choose option 2. RDF normalization is an esoteric enough topic that I would prefer not to have anything implicit.

freud/density.pyx
vyasr: [I clarified exactly what I intended](https://github.com/glotzerlab/freud/pull/1013/files/82d9c27ccb0a5d7430151359a9ecd710b303ef53#r989267661).

freud/density.pyx
tommy-waltmann: I think leaving the current behavior is fine, and calling it `exact` better reflects what its actually doing.

freud/density.pyx
```diff
@@ -593,27 +604,23 @@ cdef class RDF(_SpatialHistogram1D):
...
-            not meaningful in such cases and will simply convolute the data.
-
+        normalization_mode (str, optional):
+            There are two valid string inputs for this argument. The first
+            option, ``exact``, handles the normalization as shown mathematically
+            at the beginning of this class's docstring. The other option,
+            ``finite_size``, adds an extra rescaling factor of
+            :math:`\frac{N_{query\_points}}{N_{query\_ponts} - 1}` so the RDF
+            values will tend to 1 at large :math:`r` for small systems (Default
+            value = :code:`'infer'`).
```
DomFijan:
```suggestion
            value = :code:`'exact'`).
```

DomFijan: I think this was supposed to be exact now if I understand this correctly?

freud/density.pyx
```diff
@@ -576,10 +576,21 @@ cdef class RDF(_SpatialHistogram1D):
...
+    .. note::
+        For correct normalization behavior when using
+        ``normalization_mode='exact'``, let the set of points be either: 1) the
+        same as the set of query points or 2) completely disjoint from the set
+        of query points.
+
+    .. note::
+        For correct normalization behavior when using
+        ``normalization_mode='finite_size'``, do not allow particles to be their
+        own neighbor.
```
DomFijan:
```suggestion
        own neighbor (:code:`exclude_ii` must be set to :code:`False`).
```

DomFijan: I feel like it's valuable to have exact options to which this refers be spelled out. Helps the docs be more informative IMO.

freud/density.pyx
```diff
@@ -576,10 +576,21 @@ cdef class RDF(_SpatialHistogram1D):
...
+    deviation is noticeable, the ``normalization_mode`` argument may be used to
+    rescale the results and force the long range behavior to 1. Note that this
+    option will have little to no effect on larger systems (for example, for
+    systems of 100 particles the RDF will differ by 1%).
+
+    .. note::
+        For correct normalization behavior when using
+        ``normalization_mode='exact'``, let the set of points be either: 1) the
+        same as the set of query points or 2) completely disjoint from the set
+        of query points.
```
DomFijan:
```suggestion
        of query points (points shouldn't contain any particles in query points).
```

DomFijan: I feel like being overly descriptive helps users when reading docs, especially ones less familiar with set theory.

DomFijan (changes requested):
I think there is a typo in one of the changes to docstrings (please double check that). And I have some improvements suggested. I can't see anything else wrong.

freud/density.pyx
tommy-waltmann: Good idea. I left it that way because one could give a custom neighborlist and in that case there are no query arguments, but having the exact answer for the much more common case will help.

DomFijan (approved):

cpp/density/RDF.h
```diff
@@ -16,8 +16,16 @@ namespace freud { namespace density {
 class RDF : public locality::BondHistogramCompute
 {
 public:
+    //! Enum for each normalization mode
+    enum NormalizationMode
```
vyasr, line 20: It's up to you whether to do it in this PR or to fix all instances of this in a follow-up PR, but I would recommend moving all uses of `enum` in freud to `enum class`. Scoped enums are definitely preferable (although they require very slightly more hoops to jump through with Cython until we can update to Cython 3.0) but given that all our enums are unscoped right now I don't mind adding this one like this for now then changing them all later. Up to you.

doc/source/conf.py
```diff
@@ -78,7 +78,7 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
```
vyasr, line 81: Seems unrelated?

doc/source/gettingstarted/migration.rst
```diff
@@ -0,0 +1,25 @@
+.. _migration:
+
+============================
+Migration to freud Version 3
+============================
+
+Version 3 of the freud library introduces a few breaking changes to make the API
+more intuitive and facilitate future development. This guide explains how to
+alter scripts which use freud v2 APIs so they can be used with freud version 3.
```
vyasr: For consistency with the earlier v2
```suggestion
alter scripts which use freud v2 APIs so they can be used with freud v3.
```

doc/source/gettingstarted/migration.rst
```diff
@@ -0,0 +1,25 @@
...
+
+.. list-table::
+    :header-rows: 1
+
+    * - Goal
+      - v2 API
+      - v3 API
+    * - Use default RDF normalization.
+      - ``freud.density.RDF(..., normalize=False)``
+      - ``freud.density.RDF(..., normalization_mode='infer')``
```
vyasr:
```suggestion
      - ``freud.density.RDF(..., normalization_mode='exact')``
```

freud/density.pyx
```diff
@@ -576,10 +576,21 @@ cdef class RDF(_SpatialHistogram1D):
...
+    .. note::
+        For correct normalization behavior when using
+        ``normalization_mode='exact'``, let the set of points be either: 1) the
+        same as the set of query points or 2) completely disjoint from the set
+        of query points (points shouldn't contain any particles in query points).
+
+    .. note::
+        For correct normalization behavior when using
+        ``normalization_mode='finite_size'``, do not allow particles to be their
+        own neighbor (:code:`exclude_ii` must be set to :code:`False`).
```
vyasr: The same restrictions for `exact` also apply to `finite_size`.

vyasr:
```suggestion
        For correct normalization behavior when using
        ``normalization_mode='finite_size'`` the ``points`` _must_ be the
        same as the ``query_points`` and ``exclude_ii`` must be
        set to ``False``.
```

vyasr (changes requested):
It would be nice to actually validate the long-distance behaviors of the RDF for a small system of particles as well, but getting that test to behave nicely statistically will probably be hard since the whole point is to test a small system where statistics are bad, so if you have a hard time writing a robust test it's fine to skip it.

cpp/density/RDF.h
tommy-waltmann: I agree with the enum classes suggestion, but I'll wait until the cython 3.0 release to take a PR out to handle it. If you'd like to make an issue so we can track it, I would appreciate it. In the meantime, this particular enum is wrapped by the RDF class, so it doesn't leak outside `freud::density::RDF`.

doc/source/conf.py
tommy-waltmann: Newer versions of sphinx generate a warning when the language isn't explicitly chosen to be english. I can remove this if you want, but I thought I'd sneak it in here since it seems too small to be its own PR.

cpp/density/RDF.h
vyasr: Done: https://github.com/glotzerlab/freud/issues/1025.

vyasr (approved):

Issue #984, opened 2022-07-02
Remove global_search flag of EnvironmentCluster


Charlottez112:
Remove the `global_search` flag of `EnvironmentCluster.compute`, since it is equivalent to using the entire system as the neighbor list for `neighbors`.
If `global_search = True`, neighbor list passed to `neighbors` is ignored.
See related PRs: #979 #980

tommy-waltmann:
closed via #1039

PR #1039, opened 2022-11-21
Remove global search flag

tommy-waltmann:
## Description

This PR removes the global_search flag in `EnvironmentCluster`, in favor of having users just give a neighborlist where every particle is a neighbor of every other particle.

## Motivation and Context

This PR makes the API more intuitive and less confusing. 

Resolves: #984 

## How Has This Been Tested?

I have converted a previous test that used the `global_search=True` option.

## Types of changes

## Checklist:


vyasr:
@tommy-waltmann I didn't remember to mention this in earlier PRs, but one thing that might be helpful to maintain continuously on the next branch is some sort of migration guide. We can expand on it when it's actually time for the 3.0 release, but it would be helpful during each PR to maintain some minimal notes for how users should adapt to the change. In the case of removing `global_search`, it can be emulated by constructing a `NeighborList` of all pairs of particles as `neighbors`.

tommy-waltmann:
> @tommy-waltmann I didn't remember to mention this in earlier PRs, but one thing that might be helpful to maintain continuously on the next branch is some sort of migration guide. We can expand on it when it's actually time for the 3.0 release, but it would be helpful during each PR to maintain some minimal notes for how users should adapt to the change. In the case of removing `global_search`, it can be emulated by constructing a `NeighborList` of all pairs of particles as `neighbors`.

@vyasr That already exists, it's in `doc/source/gettingstarted/migration.rst`.

vyasr:
Right sorry I forgot that we added that in #1013. In that case the action item would be to update migration.rst for this change as well (but I also see that this PR is a draft so no rush, I just did a batch review this weekend).

doc/source/gettingstarted/migration.rst
```diff
@@ -32,3 +32,6 @@ Overview of API Changes
     * - Create a custom neighborlist from arrays.
       - ``freud.locality.NeighborList.from_arrays(..., distances=...)``
       - ``freud.locality.NeighborList.from_arrays(..., vectors=...)``
+    * - Compare environments of all particle pairs with ``EnvironmentCluster``.
+      - ``env_cluster.compute(..., global_search=True)``
+      - ``env_cluster.compute(..., neighbors=all_pairs_nlist, ...)``
```
vyasr, line 37: Maybe worth showing how to construct `all_pairs_nlist`. Depends on #1051's new feature.

tests/test_environment_MatchEnv.py
```diff
@@ -204,6 +204,24 @@ def test_cluster_hardr(self):
             err_msg="BCC Cluster Environment fail",
         )
 
+    def _make_global_neighborlist(self, sys):
+        """Get neighborlist where all particles are neighbors."""
+        box, points = sys
```
vyasr: This function signature is strange. If it's intended to be called with an arbitrary system-like object, we should be calling `from_system` and then accessing the attributes. If it's always called with a `box, points` tuple (which seems to be the case) it should just accept them as separate parameters. The current approach communicates mixed messages about the intent.

vyasr (approved):
Couple of minor comments but otherwise LGTM.

doc/source/gettingstarted/migration.rst
tommy-waltmann: Good suggestion. However, I'd like to wait until we have an official release out with the changes from #1051 on it before adding the utility method in the example.

tests/test_environment_MatchEnv.py
tommy-waltmann: I just wrote this quick-and-dirty without thinking too much about the function signature since its just a helper method in a unit test, but I can update it to be clearer.

PR #1042, opened 2022-11-30
NEW feature: intermediate scattering function

### LINKED ISSUE #1040: A Request for Intermediate Scattering function
### Description

The intermediate scattering function is defined as the Fourier transform of the Van Hove function:
image
Instead of Fourier transform, these functions can also be directly computed from the atomic trajectories:
image
`Fs` and `Fd` are self and distinct parts. 

Is there any plan to support this function? Otherwise, I can write one following structure factor code and MSD part. 

### Proposed Solution
```
isf = freud.Scattering.Intermedidate(k_space, ...)
# points = (N_frames, N_particles, 3)
# points = (N_frames, M_particles, 3)
isf.compute((box, points)).query(query_points)
isf.self_part
isf.distinct_part
```
### Additional Context

Reference: 
https://www.lehigh.edu/imi/teched/AtModel/Lecture_11_Micoulaut_Atomistics_Glass_Course.pdf

### Developer

Would someone else please implement this?

Roy-Kid:
## Description
Please assist me in improving the code @tommy-waltmann 
To calculate time-dependent intermediate scattering function. Here are the refs:
1. https://en.wikipedia.org/wiki/Dynamic_structure_factor
2. https://www.lehigh.edu/imi/teched/AtModel/Lecture_11_Micoulaut_Atomistics_Glass_Course.pdf

## Motivation and Context
Here we add a class derived from `StaticStructureFactorDirect` to calculate the time-dependent intermediate scattering function. 
Resolves: #1040

## How Has This Been Tested?
This code has not been tested before complete

## Types of changes

- [ x] New feature (non-breaking change which adds or improves functionality)

## Checklist:

- [ x] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/master/CONTRIBUTING.md) document.
- [x ] My code follows the code style of this project.


tommy-waltmann:
Hi @Roy-Kid, thanks for getting such a quick start! This is great and I can tell this took significant effort on your part to write. Before I get into more detailed comments, I want to talk about some conceptual things related to the intermediate scattering function.

The first thing is that the intermediate scattering function is not a static structure factor, it is dynamic in that it needs data from multiple different times in order to compute. At the same time, it shares the same method of sampling vectors in k-space with the `StaticStructureFactorDirect` class. The intermediate scattering function can also be thought of as a structure factor because it is the autocorrelation of the direct structure factor. With those ideas in mind, I want to alter the inheritance hierarchy among structure factor classes so the intermediate scattering function can fit in nicely. I will implement those changes since they alter the structure of previously developed code somewhat.

The second thing is about whether or not we should be able to accumulate the intermediate scattering function. In the MSD code, the calculation can be accumulated over more points because each point's MSD can be computed independent of the other points. In the code for the other structure factors, the calculation can be accumulated over many frames because each frame can be computed independent of the other frames, and the histogram freud computes internally can have more counts added to it. The intermediate scattering function, by contrast to the previous two calculations, needs all positions at all times to compute, which means accumulation conceptually does not make sense.

Going forward, I will jump in and alter the inheritance hierarchy of the structure factor classes, remove the ability to accumulate from the class, and make some other small changes to the API. Then, if you wish, you can jump back in and make some of the more detailed changes like updating the documentation, adding more tests, validating the calculation on your use case, etc.

Once again, thanks for taking the initiative with this feature, I'm sure other people will find use for it as well!

Charlottez112:
> Once again, thanks for taking the initiative with this feature, I'm sure other people will find use for it as well!

Yes I will. Thanks guys!

tommy-waltmann:
One more conceptual thing that I forgot to mention before: There should be no need for use of a neighbor list in this calculation. Since the sums are over all particles, we already know that all particles are neighbors of each other. `NeighborQuery`/`NeighborList` objects are for defining/finding neighbors when it isn't already obvious which particle pairs the calculations need to consider.

It will likely, in fact, slow down the `IntermediateScattering` implementation to use a `NeighborQuery` because the `NeighborQuery` will build a complex data structure and find neighbors according to search parameters, but it will just end in all particle pairs being neighbors, and it would be much faster to do a nested loop over all particle indices.

Roy-Kid:
>  Going forward, I will jump in and alter the inheritance hierarchy of the structure factor classes, remove the ability to accumulate from the class, and make some other small changes to the API. 

Thanks for all your correction! I have some misunderstandings about the intermediate scattering function. Now it seems that it is not a simple combination of MSD and static scattering function (even if some methods can be reused). I'll leave the code to you next. Thank you for your adjustment in the code architecture.

tommy-waltmann:
Hi, @Roy-Kid at this point I have finished restructuring the code so that the calculation will fit correctly into freud's structure.. Everything will compile, but the implementation is wrong. Are you willing to give the implementation another try? It should only involve changing the `compute` function on the C++ side of the code.

Roy-Kid:
> Hi, @Roy-Kid at this point I have finished restructuring the code so that the calculation will fit correctly into freud's structure.. Everything will compile, but the implementation is wrong. Are you willing to give the implementation another try? It should only involve changing the `compute` function on the C++ side of the code.

I'm so sorry for replying late. I'm busy with my mid-term assessment and PhD application, so I don't have time to develop the code these days. If you have time, could you please continue to finish the work, or wait for me until the next year February? 

Sorry again for the delay...

Roy-Kid:
Hi, @tommy-waltmann. I try to implement the code of `compute`, and it looks like done. But I get a fatal error at `getSelfFunction` of IntermediateScattering.cc line 121. I am confused with those `reduce` functions, so I don't figure out how to fix this problem. Could you please help me with this?

tommy-waltmann:
@Roy-Kid the issue was in access out-of-bounds memory in two spots, I've made some changes and there are no more crashes. The code should be in a place where you can continue with the implementation.

Issue #372, opened 2019-08-08
Implement Solid-angle Nearest Neighbors (SANN)


bdice:
The Solid-angle Nearest Neighbors method of determining particle neighbors [1] has been used in several recent papers. We would like to implement this feature in freud so it can be compared to other analysis methods. @csadorf expressed interest in helping with this.

1. https://aip.scitation.org/doi/full/10.1063/1.4729313

**Describe the solution you'd like**
I would build this from the `next` branch, which will become freud 2.0. It should be possible to build this method by using the other methods `query` and `queryBall` and increasing the number of neighbors until finding a solution of the iterative problem described in [1] above. See the two expressions below:

![image](https://user-images.githubusercontent.com/3943761/62735579-7ad0a480-b9f9-11e9-8d18-ef39e9936c51.png)
![image](https://user-images.githubusercontent.com/3943761/62735599-82904900-b9f9-11e9-855d-de26670c4b5e.png)

In particular, I think `queryBall` might be a lot faster than `query`, so if you're choosing one method to "query and expand," I would try using `queryBall` method internally and expand the search radius until the computed solid angle is >= 4pi.

Steps to implement:
- add a "mode" for SANN to the QueryArgs object:
https://github.com/glotzerlab/freud/blob/b25c6e2455dd9e6d412f5eadd1f8f91384a91262/freud/locality.pyx#L78-L85
https://github.com/glotzerlab/freud/blob/b25c6e2455dd9e6d412f5eadd1f8f91384a91262/freud/_locality.pxd#L21-L23
https://github.com/glotzerlab/freud/blob/b25c6e2455dd9e6d412f5eadd1f8f91384a91262/cpp/locality/NeighborQuery.h#L29-L49

- add virtual method to NeighborQuery parent class
https://github.com/glotzerlab/freud/blob/b25c6e2455dd9e6d412f5eadd1f8f91384a91262/cpp/locality/NeighborQuery.h#L110-L113

- add method definition to AABBQuery and LinkCell classes

**Describe alternatives you've considered**
It might be possible to implement this in `NeighborQuery` directly, instead of implementing it twice in `AABBQuery` and `LinkCell`, but I'm not sure. @vyasr would have more insight on that.

**Additional context**
Papers citing the SANN method: https://aip.scitation.org/doi/citedby/10.1063/1.4729313

vyasr:
[...]

PR #1051, opened 2022-12-15
NeighborList Filters and SANN

tommy-waltmann:
## Description

This PR adds a new concept in freud, the NeighborList filter, and the SANN neighbor finding method implemented as a neighborlist filter. The SANN method finds neighbors based on solid angle occupied by each neighbor up to a total of 4pi. For more information, see https://aip.scitation.org/doi/10.1063/1.4729313

I have also added a method for sorting a `NeighborList` by either distance or point_index to the python API. It was needed for the SANN method implementation and I think it has general utility as well.

## Motivation and Context

This is a useful way of finding neighbors, which was not previously available in freud.

## How Has This Been Tested?

Tests have been added in `test_locality_Filter.py`. I would like to see some validation of the calculation on a real system before this PR can be merged.

## Types of changes

## Checklist:


DomFijan:
I will either look at this by the end of this week or if not it'll have to wait until January. (More likely January)
@tommy-waltmann Do we still want to do several more tests for couple of different crystal structures? Only FCC is currently tested and only agains python implementation.

tommy-waltmann:
> I will either look at this by the end of this week or if not it'll have to wait until January. (More likely January) @tommy-waltmann Do we still want to do several more tests for couple of different crystal structures? Only FCC is currently tested and only agains python implementation.

Its fine if this waits until January, I'm not in a rush to get it merged. I don't think we need more official unit tests, but I would like to see this gives the same results from the paper in the PR description.

vyasr:
See #372 just in case you weren't aware of the history here ;) I've reopened it so that it can be linked to and closed by this PR for historical context.

tommy-waltmann:
I've tested the freud implementation against simon's notebook from #372 and the sets of neighbors computed are the same for a 10,000 particle, random system with box length 10.

Freud_neighbors

Simon_neighbors

cpp/locality/FilterSANN.cc
```diff
@@ -0,0 +1,72 @@
...
+    using BondVector = tbb::enumerable_thread_specific<std::vector<NeighborBond>>;
+    BondVector filtered_bonds;
+
+    // parallelize over query_point_index
+    util::forLoopWrapper(0, sorted_nlist.getNumQueryPoints(), [&](size_t begin, size_t end) {
+        // grab thread-local vector
+        BondVector::reference local_bonds(filtered_bonds.local());
+        for (auto i = begin; i < end; i++)
+        {
+            unsigned int m = 3;
```
DomFijan: We used name "m" here for consistency with the algorithm in the paper. Would it make sense to use a different variable name that's more descriptive? I'm not sure what our usual policies are on this matter.

cpp/locality/FilterSANN.cc
```diff
@@ -0,0 +1,72 @@
...
+        }
+    });
+
+    // combine thread-local NeighborBond vectors into a single vector
+    tbb::flattened2d<BondVector> flat_filtered_bonds = tbb::flatten2d(filtered_bonds);
+    std::vector<NeighborBond> sann_bonds(flat_filtered_bonds.begin(), flat_filtered_bonds.end());
+
+    // sort final bonds array by distance
+    tbb::parallel_sort(sann_bonds.begin(), sann_bonds.end(), compareNeighborDistance);
+    // tbb::parallel_sort(sann_bonds.begin(), sann_bonds.end(), compareNeighborBond);
```
DomFijan:
```suggestion
```
This line is commented out?

freud/locality.pyx
```diff
@@ -1272,3 +1286,94 @@ cdef class Voronoi(_Compute):
...
+    arguments. Then, based on the arrangement of particles, their shapes, and
+    other criteria determined by the derived class, some of the neighbors are
+    removed from the unfiltered :class:`.NeighborList`.
+
+    The compute method of each :class:`.Filter` class will take a system object
+    along with a neighbors dictionary specifying query arguments. The
+    ``neighbors`` dictionary along with the system object will be used to build
+    the unfiltered neighborlist, which will then be filtered according to the
+    filter class. After the calculation, the filtered neighborlist will be
+    available as the property ``filtered_nlist``
```
DomFijan:
```suggestion
    available as the property ``filtered_nlist`` .
```

DomFijan (changes requested):
I don't see any immediate issues. Some possible minor improvements are in the comments. Since you tested against Simons notebook successfully, do we think there should be more tests? Would a test against this notebook be a good addition to tests as well?

cpp/locality/FilterSANN.cc
tommy-waltmann: I'm fine with keeping the name `m`, if further clarification of the variable name is necessary, a comment can be added to the line where it is defined.

tommy-waltmann:
As far as using Simon's implementation for more testing, we already have a python implementation of SANN in the tests, so I don't see a need to add another one. Anyone can feel free to add it if they want to.

tommy-waltmann:
I think we could use one more test comparing results between implementations for a random system.

DomFijan (approved):
Looks good. I don't have any more suggestions or issues with this.

vyasr:
I won't have time to look too closely at this, but the design of the filters looks solid. However, I am a bit curious about the potential for user error. IIUC this filtering approach to computing the SANN requires that the query arguments result in a sufficient number of neighbors to satisfy the SANN criteria, correct? Of course, plenty of other methods in freud can also give nonsense results when not provided sensible query arguments (e.g. RDFs with no neighbors), but the failure mode for SANN seems a bit different since it wouldn't just provide physically meaningless results but would actually not even satisfy the core criterion of the algorithm (a total solid angle of 4 pi). Am I understanding that correctly? If so, could we add a means to warn the user in such a scenario?

tommy-waltmann:
@vyasr That's right, if the user doesn't give a large enough r_max, for example, there may be particles without enough neighbors to fill out the 4*pi solid angle. I think it wouldn't be too hard to check that and send some warning to terminal output.

Perhaps a note in the documentation about this possibility would also be appropriate.

tommy-waltmann:
Another thought I've had on this: we could make the default behavior (i.e. neighbors=None in the `compute` method) be to build an unfiltered neighborlist where all particles are neighbors of all other particles, resulting in the full SANN calculation happening by default.

tommy-waltmann:
> Another thought I've had on this: we could make the default behavior (i.e. neighbors=None in the `compute` method) be to build an unfiltered neighborlist where all particles are neighbors of all other particles, resulting in the full SANN calculation happening by default.

In light of what's going on in #1039 , it may be useful to just directly expose a factory method in the Neighborlist class which takes a system and set of query points and returns a Neighborlist where every pair of points i, j (i =/= j) are neighbors. Then we could either call that method internally here if `neighbors=None`, or ask users to call it directly in their scripts.

joaander:
> > Another thought I've had on this: we could make the default behavior (i.e. neighbors=None in the `compute` method) be to build an unfiltered neighborlist where all particles are neighbors of all other particles, resulting in the full SANN calculation happening by default.
> 
> In light of what's going on in #1039 , it may be useful to just directly expose a factory method in the Neighborlist class which takes a system and set of query points and returns a Neighborlist where every pair of points i, j (i =/= j) are neighbors. Then we could either call that method internally here if `neighbors=None`, or ask users to call it directly in their scripts.

That's a nice idea to add some amount of correctness at the cost of performance. However, note that it still only guarantees correctness in a limited set of circumstances (when the solution involves only neighbor pairs within the minimum image). For example, in small (or dilute) boxes, one of the SANN neighbors might easily be aa non-minumum image of any particle in the box (including itself).

tommy-waltmann:
> > > Another thought I've had on this: we could make the default behavior (i.e. neighbors=None in the `compute` method) be to build an unfiltered neighborlist where all particles are neighbors of all other particles, resulting in the full SANN calculation happening by default.
> > 
> > 
> > In light of what's going on in #1039 , it may be useful to just directly expose a factory method in the Neighborlist class which takes a system and set of query points and returns a Neighborlist where every pair of points i, j (i =/= j) are neighbors. Then we could either call that method internally here if `neighbors=None`, or ask users to call it directly in their scripts.
> 
> That's a nice idea to add some amount of correctness at the cost of performance. However, note that it still only guarantees correctness in a limited set of circumstances (when the solution involves only neighbor pairs within the minimum image). For example, in small (or dilute) boxes, one of the SANN neighbors might easily be aa non-minumum image of any particle in the box (including itself).

I think the vast majority of use cases will not require that we consider neighbors that are more than a single image convention away. In those cases, freud has utilities like `replicate_system` to allow users to work with cases like that. I've recently added some warnings that will alert users to situations where those utilities may be necessary, as requested by @vyasr's first comment.

vyasr:
Originally the implementation that Simon, Bradley and I considered was based on iteratively performing ball queries until a sufficient number of neighbors was found to complete the sphere. I'm on board with the current approach of filtering, which circumvents some of the implementation complexity associated with that iterative procedure by pushing the burden of having an effective starting point onto the user, but I think we probably should help out users a bit more than we are now. In that vein, a few points:
- I don't think the `all_pairs` implementation is necessary, but since you've already added it I don't mind it being present either.
- I'm happy to see that the warning was added, but I am not sure if it is strict enough. Should the SANN calculation even be considered valid if not enough neighbors are present? I'm inclined to include an additional parameter `allow_incomplete_shell = False` that would control whether insufficient neighbors is considered a warning or an error, and default to error. Admittedly I am not the most familiar with SANN, so I don't know whether a warning or an error is a more appropriate default (or whether it is important enough to make such a change).
- What I think would be the most helpful is to add two new examples to [our documentation](https://freud.readthedocs.io/en/latest/gettingstarted/examples.html):
  1. If a user is seeing the warning about too few neighbors, what is the best way to fix it? Manually increasing the thresholds and crossing your fingers is one option, but would be crucial to show IMO is an example of a Python script that iteratively runs ball queries until the threshold is large enough, then filters the final resulting NeighborList. I suspect that many users will end up doing something like this anyway, so we may as well codify a best practice here (along with strong warnings that users should make good informed choices for the initial `r_min`).
  2. In the case where there simply aren't enough neighbors in the system, show how `replicate_system` could be used. This case seems like it shouldn't occur too often in realistically sized systems, so I don't think it's critical.

joaander:
>   2. In the case where there simply aren't enough neighbors in the system, show how `replicate_system` could be used. This case seems like it shouldn't occur too often in realistically sized systems, so I don't think it's critical.

Indeed. Users should use replication in small N, high density systems. For low density systems of any N, users should not use SANN at all. I presented that possibility above as a pathological case to demonstrate that is is not feasible to find a complete SANN list for *any* arbitrary system the user might provide, even with massive numbers of system replicates and using all pairs.

One could probably formulate a rule of thumb for the search radius needed in terms of the mean free path.

tommy-waltmann:
I am on board with adding an extra argument as @vyasr suggested, and I think an error should be the default behavior. If we are going to have an iterative procedure for finding a good choice of query parameters, it will be necessary to have the ability to generate an error in `FilterSANN` to know when to try a different set of parameters.

I think the best iterative procedure would be a binary search with either `r_max` or `num_neighbors`, calling `FilterSANN` on each iteration and seeing if it fails or not. Using knowledge of the system, a user could probably make a good guess as to what the initial upper and lower bounds of the binary search should be so the iterative method converges quicker.

tommy-waltmann:
I have added the code to raise an error if there are unfilled neighbor shells. I cannot, however, figure out how to get the error message to format properly when the exception is raised. Here is an example of what the user sees right now when the exception is raised:
```
test_locality_Filter.py::TestSANN::test_SANN_simple Fatal Python error: Aborted

Current thread 0x0000000112f59e00 (most recent call first):
  File "/Users/tomwalt/code/freud/tests/test_locality_Filter.py", line 137 in test_SANN_simple
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/python.py", line 195 in pytest_pyfunc_call
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/python.py", line 1789 in runtest
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/runner.py", line 167 in pytest_runtest_call
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/runner.py", line 260 in <lambda>
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/runner.py", line 339 in from_call
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/runner.py", line 259 in call_runtest_hook
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/runner.py", line 220 in call_and_report
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/runner.py", line 131 in runtestprotocol
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/runner.py", line 112 in pytest_runtest_protocol
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/main.py", line 349 in pytest_runtestloop
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/main.py", line 324 in _main
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/main.py", line 270 in wrap_session
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/main.py", line 317 in pytest_cmdline_main
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/config/__init__.py", line 167 in main
  File "/Users/tomwalt/miniconda3/lib/python3.9/site-packages/_pytest/config/__init__.py", line 190 in console_main
  File "/Users/tomwalt/miniconda3/bin/pytest", line 10 in <module>
Abort trap: 6
(base) tomwalt@cheme-blyleven ~/code/freud/tests $
```
Does anyone know how to get exceptions from C++ to format nicely in python output?

joaander:
pybind11 translates c++ exceptions to Python ones :)

vyasr:
Cython also translates C++ exceptions to Python ones :) All that you should need is the `except + ` after the `cdef extern` of the `compute` method, which I see that you do have, so I'm not sure what's going on. Let me try to repro locally.

vyasr:
@tommy-waltmann I cannot reproduce that traceback locally. If I set `allow_incomplete_shells=False` I get a Python exception as expected. I would try wiping your build directory, it is possible that in an earlier iteration of the code you did not have the `except +` in the pxd file. Cython is not great at regenerating certain parts of the code when all that changes is a declaration in a pxd, so perhaps your build was a bit stale for some reason and it didn't pick up on the exception handling.

tommy-waltmann:
Do @b-butler or anyone else have more comments on this PR?

b-butler:
Given the large discussion is another opinion warranted you think @tommy-waltmann?

tommy-waltmann:
> Given the large discussion is another opinion warranted you think @tommy-waltmann?

Just making sure everyone who is tagged for review gets a chance to provide their opinion.

Issue #981, opened 2022-06-28
Remove zero vectors from the clusters computed by EnvironmentCluster


Charlottez112:
Right now the clusters found by `EnvironmentCluster` are padded with 0 vectors if their sizes (number of component particles) are smaller than the largest cluster. This can be confusing, and it's not necessary.  Based on recent discussions in #979, it seems like a better approach would be returning a list of numpy arrays, rather than a 2D numpy array with zeros.

`point_environments` computed by `EnvironmentCluster` are also padded with zeros. We will need to remove zero vectors from those too. (discussed in #980)

tommy-waltmann:
Closed via #1064

PR #1064, opened 2023-01-27
Return ragged arrays in Environment Matching Code

tommy-waltmann:
## Description

This PR makes the `point_environments` property of the `EnvironmentCluster` and `EnvironmentMotifMatch`, as well as the `cluster_environments` property of `EnvironmentCluster`, return a list if numpy arrays of different sizes.

## Motivation and Context

This is a more intuitive way to return the data to the user, and the padded zero vectors in the previous implementation were confusing.

Resolves: #981 

## How Has This Been Tested?

Tests have been added to ensure raggedness of the properties which have had their behavior changed.

## Types of changes

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -12,10 +12,10 @@ and this project adheres to
 
 ### Changed
 * The `normalize` argument to `freud.density.RDF` is now `normalization_mode`.
+* Rmove zero-padding from arrays in `freud.environment.EnvironmentCluster` and `freud.environment.EnvironmentMotifMatch` and replace with ragged lists of NumPy arrays.
 
 ### Removed
 * The `global_search` flag in `freud.environment.EnvironmentCluster`.
-* Zero padding from properties of `freud.environment.EnvironmentCluster` and `freud.environment.EnvironmentMotifMatch`.
 
 ## v2.12.1 -- 2022-12-05
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,7 @@ and this project adheres to
 
 ### Removed
 * The `global_search` flag in `freud.environment.EnvironmentCluster`.
+* Zero padding from properties of `freud.environment.EnvironmentCluster` and `freud.environment.EnvironmentMotifMatch`.
 
 ## v2.12.1 -- 2022-12-05
 
--- a/cpp/environment/MatchEnv.cc
+++ b/cpp/environment/MatchEnv.cc
@@ -544,9 +544,6 @@ void EnvironmentCluster::compute(const freud::locality::NeighborQuery* nq,
         ;
     }
 
-    // reallocate the m_point_environments array
-    m_point_environments.prepare({Np, dj.m_max_num_neigh});
-
     size_t bond(0);
     // loop through points
     for (unsigned int i = 0; i < Np; i++)
@@ -640,11 +637,12 @@ unsigned int EnvironmentCluster::populateEnv(EnvDisjointSet dj)
                 label_ind = label_map[c];
             }
 
+            m_point_environments.push_back(std::vector<vec3<float>>());
             // label this particle in m_env_index
             m_env_index[particle_ind] = label_ind;
             for (unsigned int m = 0; m < part_vecs.size(); m++)
             {
-                m_point_environments(particle_ind, m) = part_vecs[m];
+                m_point_environments[particle_ind].push_back(part_vecs[m]);
             }
             particle_ind++;
         }
@@ -690,9 +688,6 @@ void EnvironmentMotifMatch::compute(const freud::locality::NeighborQuery* nq,
     auto* end = begin + counts.size();
     auto max_num_neigh = *std::max_element(begin, end);
     dj.m_max_num_neigh = max_num_neigh;
-
-    // reallocate the m_point_environments array
-    m_point_environments.prepare({Np, max_num_neigh});
 
     // create the environment characterized by motif. Index it as 0.
     // set the IGNORE flag to true, since this is not an environment we have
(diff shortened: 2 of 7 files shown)


tommy-waltmann:
In addition to the changes here, I think the restriction that the disjoint set data structure (`EnvDisjointSet`) have a max number of neighbors in any given environment can be lifted now. I didn't write that code though, so I'm not quite sure, @vyasr any thoughts?

vyasr:
Hmm yeah I think that sounds right. The only reason that was needed was to properly size the output arrays and to control iteration bounds, but with a ragged array you can iterate properly over each subarray (environment) so it should be safe to stop tracking that. Let me know if you run into any issues with removing it, but I think it should be fine.

cpp/environment/MatchEnv.cc
```diff
@@ -228,10 +235,10 @@ std::vector<vec3<float>> EnvDisjointSet::getAvgEnv(const unsigned int m)
 
     // loop through the vectors in env now, dividing by the total number
     // of contributing particle environments to make an average
-    for (unsigned int n = 0; n < m_max_num_neigh; n++)
+    for (auto& vec : env)
     {
-        vec3<float> normed = env[n] / static_cast<float>(N);
-        env[n] = normed;
+        vec3<float> normed = vec / static_cast<float>(N);
+        vec = normed;
```
b-butler: Is there a reason not to do this
```suggestion
        vec = vec / static_cast<float>(N);
```

b-butler (approved):
I am not overly familiar with the environment matching code, but the changes look good.

cpp/environment/MatchEnv.cc
```diff
@@ -228,10 +235,9 @@ std::vector<vec3<float>> EnvDisjointSet::getAvgEnv(const unsigned int m)
 
     // loop through the vectors in env now, dividing by the total number
     // of contributing particle environments to make an average
-    for (unsigned int n = 0; n < m_max_num_neigh; n++)
+    for (auto& vec : env)
     {
-        vec3<float> normed = env[n] / static_cast<float>(N);
-        env[n] = normed;
+        vec = vec / static_cast<float>(N);
```
bdice: Use vector operator overloads.

https://github.com/glotzerlab/freud/blob/dea4859079218a565918cc98d38143354571e01b/cpp/util/VectorMath.h#L237

```suggestion
        vec /= static_cast<float>(N);
```

cpp/environment/MatchEnv.cc
```diff
@@ -614,9 +611,11 @@ unsigned int EnvironmentCluster::populateEnv(EnvDisjointSet dj)
 
             // label this particle in m_env_index
             m_env_index[particle_ind] = label_ind;
-            for (unsigned int m = 0; m < part_vecs.size(); m++)
+
+            m_point_environments.emplace_back();
+            for (auto& part_vec : part_vecs)
```
```suggestion
            for (auto const& part_vec : part_vecs)
```

cpp/environment/MatchEnv.cc
```diff
@@ -714,9 +701,10 @@ void EnvironmentMotifMatch::compute(const freud::locality::NeighborQuery* nq,
         // grab the set of vectors that define this individual environment
         std::vector<vec3<float>> part_vecs = dj.getIndividualEnv(dummy);
 
-        for (unsigned int m = 0; m < part_vecs.size(); m++)
+        m_point_environments.emplace_back();
+        for (auto& part_vec : part_vecs)
```
```suggestion
        for (auto const& part_vec : part_vecs)
```

doc/source/reference/credits.rst
```diff
@@ -338,6 +338,7 @@ Tommy Waltmann
 * Extending plotting options for the ``Voronoi`` module.
 * Work with Bradley Dice on adding neighbor vectors to ``freud.locality.NeighborList``.
 * Remove `global_search` flag in ``freud.environment.EnvironmentCluster``.
+* Remove zero-pading from arrays in ``freud.environment.EnvironmentCluster`` and ``freud.environment.EnvironmentMotifMatch``.
```
bdice:
```suggestion
* Remove zero-padding from arrays in ``freud.environment.EnvironmentCluster`` and ``freud.environment.EnvironmentMotifMatch`` and replace with ragged lists of NumPy arrays.
```

tests/test_environment_MatchEnv.py
```diff
@@ -11,6 +11,13 @@
 matplotlib.use("agg")
 
 
+def assert_ragged_array(arr):
+    """Assert the given array is a list of numpy arrays."""
+    assert type(arr) == list
```
bdice: Always prefer isinstance over raw type checks unless you’re in a very specific circumstance (signac uses raw type checks for synced collections for performance reasons in one particular case).
```suggestion
    assert isinstance(arr, list)
```

tests/test_environment_MatchEnv.py
```diff
@@ -11,6 +11,13 @@
...
+    for a in arr:
+        assert type(a) == np.ndarray
```
bdice:
```suggestion
        assert isinstance(a, np.ndarray)
```

bdice (commented):
A few small suggestions.

ChangeLog.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to
 
 ### Removed
 * The `global_search` flag in `freud.environment.EnvironmentCluster`.
+* Zero padding from properties of `freud.environment.EnvironmentCluster` and `freud.environment.EnvironmentMotifMatch`.
```
bdice: This should be a “Changed” message. It’s not an API removal, it is a change in the API return convention and type.

See my proposed wording for the credits line, for another perspective.

cpp/environment/MatchEnv.cc
tommy-waltmann: AFAIK, it is best practice (and established freud convention) to use `const auto` rather than `auto const`. I suspect in this context it will not make a difference, but I would rather keep the established convention.

bdice: East const vs. west const is a matter of style. I don't know which is more common in freud's codebase. I tend to fall in the east const (`auto const`) camp.

https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/
https://hackingcpp.com/cpp/design/east_vs_west_const.html

tommy-waltmann: I'm definitely more sympathetic to east const after reading those articles, but after checking the codebase the convention is overwhelmingly west const, so I'll stick with that.

PR #1074, opened 2023-02-03
order.nematic API change

### LINKED ISSUE #945: Document use of nematic order parameter from vector inputs
While freud implements a general solution for the nematic order parameter, new users are confused at how to specify the inputs. They want to give a list of vectors representing orientations.

Here is an example question on the mailing list: https://groups.google.com/g/freud-users/c/Fyklr5GxFvg/m/gdW5y0ovBAAJ

**Areas that need improvement**
1. Add a code snippet to the [API documentation](https://freud.readthedocs.io/en/stable/modules/order.html#freud.order.Nematic) that uses rowan to convert a list of vectors to the form freud needs. Perhaps something like this that Bradley wrote in the mailing list question linked above:
```python
import freud
import rowan

# Generate input vectors
random_quaternions = rowan.random.rand(1000)
reference_vector = [1, 0, 0]
input_vectors = rowan.rotate(random_quaternions, reference_vector)

# Compute quaternions mapping from the reference vector to the input vectors
orientations = rowan.vector_vector_rotation(reference_vector, input_vectors)

# Determine the nematic order -- this should be small because the inputs
# point in random directions.
nop = freud.order.Nematic(reference_vector)
nop.compute(orientations)
print(f"Nematic order value: {nop.order}")
```
2. Add a simpler example to the the [example notebook](https://freud.readthedocs.io/en/stable/gettingstarted/examples/module_intros/order.Nematic.html)

Tagging @DomFijan because he mentioned wanting to address this (and smectic order) in the developer's meeting today

DomFijan:
Proposal for change of nematic API change to accept 3D orientational vectors instead of quaternions and molecular director.

## Motivation and Context
Many new users are confused by the current nematic API. The source of this confusion is two fold: 
- The example (not covered by this PR - TODO) is often interpreted in such a way that molecular director is a free parameter that should be optimized (where in fact is not).
- The compute method accepts quaternions which are 4D arrays that represent rotation of a reference orientation (which is the molecular director defined in the constructor). However orientations are represented with 3D vectors, thus non-HOOMD users and first time users of nematic order are confused regarding why they have to change to the 4D representation and how this should be done.

An additional, but minor issue is that the compute method converts the quaternions to orientations to calculate the nematic order tensor anyways, which wastes computational time for non HOOMD users - they have to convert their representation to 4D and then compute re-converts it to a 3D orientation vector. 

Resolves: #945 .

The proposed solution is for order.nematic is to get rid of molecular director in the constructor and for the compute method to accept 3D orientation vectors instead of quaternions. This will provide a clean API which requires users to use quantities they would expect (and which are used in research papers as well) instead of having to learn how and why this can also be done using quaternions. An example for conversion of HOOMDs quaternion representation to orientation is provided in the docstring. The users in this case would use rowan to do this, whereas this was done in C++ previously. The overhead of this should be a slight slowdown. However, overhead of using 3D orientation vectors should be improved by a good margin since the computation of quaternions from orientation vectors and then back to orientation vectors is completely avoided. 

## How Has This Been Tested?
Existing tests have been modified to conform to new API.

## Types of changes

## Checklist:


AlainKadar (approved):

freud/order.pyx
```diff
@@ -162,20 +162,23 @@ cdef class Cubatic(_Compute):
...
     r"""Compute the nematic order parameter for a system of particles.
 
-    Args:
-        u (:math:`\left(3 \right)` :class:`numpy.ndarray`):
-            The nematic director of a single particle in the reference state
-            (without any rotation applied).
+    In some cases orientations are given as quaternions representing a rotation
+    from the starting reference state (orientation) - such as HOOMD-blue. To
+    convert from quaternions to orientations in 3D one should apply quaternions
+    on the initial reference orientation of the particle.
```
```suggestion
    Note:
        In some cases, such as HOOMD-blue simulations, orientations are represented as
        quaternions which rotate a particle from an initial reference orientation. To
        convert from the quaternion representation to the orientation vectors taken by *freud*,
        one should rotate the reference orientation by the particle's orientation quaternion. An example
        of this process using the *rowan* library is shown below
```

freud/order.pyx
```diff
@@ -162,20 +162,23 @@ cdef class Cubatic(_Compute):
...
+
+    Example::
+
+        >>> import rowan
+        >>> import numpy as np
+        >>> quats = np.array([[1, 0, 0, 0]] * 100)
+        >>> starting_orientation = np.array([1, 1, 0])
+        >>> orientations = rowan.rotate(quats,starting_orientation)
```
tommy-waltmann: Let's add two more lines to this example showing how orientations are given to the `Nematic` class's compute method.

tests/test_order_Nematic.py
```diff
@@ -29,6 +24,8 @@ def test_perfect(self):
         with pytest.raises(AttributeError):
             op_parallel.nematic_tensor
 
+        print(orientations)
+        print(orientations.shape)
```
tommy-waltmann:
```suggestion
```

freud/order.pyx
```diff
@@ -185,23 +188,24 @@ cdef class Nematic(_Compute):
...
-            freud.order.Nematic(u=[...])
+            freud.order.Nematic()
+            >>> print(nematic.order)
+            1.0
 
         Args:
-            orientations (:math:`\left(N_{particles}, 4 \right)` :class:`numpy.ndarray`):
-                Orientations to calculate the order parameter.
+            orientations (:math:`\left(N_{particles}, 3 \right)` :class:`numpy.ndarray`):
+                Orientation vectors for which to calculate the order parameter.
```
tommy-waltmann, line 208: Do the orientation vector inputs need to be unit length? If so, we should explicitly normalize them at some point in the calculation.

tommy-waltmann (changes requested):
Nice work on this so far @DomFijan ! Remember to give yourself some credit and update the credits file, make a changelog entry, and update the migration guide too.

freud/order.pyx
joaander: *rowan*, not *coxeter*.

freud/order.pyx
DomFijan: Is it ok if I do that in python using numpy, or should I do it in cpp?

tommy-waltmann: Would probably be best to do it in cpp, [line 51 of Nematic.cc](https://github.com/glotzerlab/freud/pull/1074/files#diff-e568d7b4eeba7cd5955620f05ca042b12159940641771f5f89f12efacca032bf) would be a good place.

DomFijan: I'm looking at the old code now. Previously we assumed the quaternions were unit quaternions and we did no explicit checks on quaternion normality.  I'm assuming that this was done to improve the speed of the code. Do you think we can instead write in the docs that we require users to provide normalized orientations?  The `u` (molecular director passed to the nematic constructor in the old code) was normalized in the constructor though.

DomFijan:
The pre-commit fails on an unrelated file `tests/test_locality_Filter.py:63:9: F841 local variable 'sol_dist' is assigned to but never used
        sol_dist = sorted_dist[mask]` . Should I go ahead and fix this in this pr?

freud/order.pyx
tommy-waltmann: Tools like `rowan` and `hoomd` always output unit quaternions, so it wasn't as much of a concern when quaternions were the input. For vector inputs, they could come from any arbitrary source, so I think explicit normalization is the best option. It takes very little time to do and computing this order parameter is already very computationally cheap.

tommy-waltmann:
> The pre-commit fails on an unrelated file `tests/test_locality_Filter.py:63:9: F841 local variable 'sol_dist' is assigned to but never used sol_dist = sorted_dist[mask]` . Should I go ahead and fix this in this pr?

Might as well just do it, yeah.

freud/order.pyx
DomFijan: I normalized the orientations in c++ for now. We can easily revert if we want to.

ChangeLog.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to
 * The `normalize` argument to `freud.density.RDF` is now `normalization_mode`.
 * The ``neighbors`` argument to ``env_neighbors`` for ``EnvironmentMotifMatch`` class.
 * The ``neighbors`` argument to ``cluster_neighbors`` for ``EnvironmentCluster`` class.
+* The `freud.order.Nematic` method no longer requires `u` ("nematic director") and uses orientation vectors instead of quaternions in the `compute` method.
```
```suggestion
* `freud.order.Nematic` uses orientation vectors instead of quaternions and a nematic director.

```

cpp/order/Nematic.cc
```diff
@@ -55,9 +47,8 @@ void Nematic::compute(quat<float>* orientations, unsigned int n)
     util::forLoopWrapper(0, n, [&](size_t begin, size_t end) {
         for (size_t i = begin; i < end; ++i)
         {
-            // get the director of the particle
-            quat<float> q = orientations[i];
-            vec3<float> u_i = rotate(q, m_u);
+            // get the orientation of the particle and normalize it
+            vec3<float> u_i = orientations[i] / std::sqrt(dot(orientations[i], orientations[i]));
```
tommy-waltmann:
```suggestion
            // get the orientation of the particle and normalize it
            auto u_i = orientations[i];
            u_i = u_i / std::sqrt(dot(u_i, u_i));
```
A compiler may be smart enough to optimize what was previously written to what I've suggested, but we might as well slightly change the way the code is written to give the compiler a little help.

doc/source/gettingstarted/migration.rst
```diff
@@ -41,3 +41,6 @@ Overview of API Changes
     * - Use ``EnvironmentCluster`` with a set of neighbors that define the cluster
       - ``freud.environment.EnvironmentCluster().compute(..., neighbors=...)``
       - ``freud.environment.EnvironmentCluster().compute(..., cluster_neighbors=...)``
+    * - Nematic API change: Nematic constructor no longer requires "nematic director" (molecular director) and compute method uses vector orientations (3D vectors) instead of quaternions (4D vectors).
+      - ``freud.order.Nematic(u=).compute(orientations = 4D_quaternions)``
+      - ``freud.order.Nematic().compute(orientations = 3D_orientation_vectors)``
```
tommy-waltmann:
```suggestion
    * - Compute nematic order parameter.
      - ``freud.order.Nematic(u=director).compute(orientations=quaternions)``
      - ``freud.order.Nematic().compute(orientations=orientation_vectors)``
```

tests/test_order_Nematic.py
```diff
@@ -1,23 +1,18 @@
...
 import pytest
-import rowan
 
 import freud
 
 
 class TestNematicOrder:
     def test_perfect(self):
-        """Test perfectly aligned systems with different molecular axes"""
+        """Test perfectly aligned systems with different alignment"""
```
tommy-waltmann:
```suggestion
        """Test different perfectly aligned systems."""
```

tommy-waltmann (changes requested):
Great work so far on this @DomFijan . I have some smaller comments and suggestions and one other point I want to put up for discussion:

I've gone back and forth a bit with myself about whether we should change the name of the input to `Nematic`'s `compute` method from `orientations` to something else because the input is different now. I think calling it `orientations` still makes some sense, but it feels awkward to change the input without changing the name of the input.

Perhaps `orientation_vectors` would be better to clarify that the inputs are vectors now, not quaternions. What do other people think about this?

joaander:
> Perhaps `orientation_vectors` would be better to clarify that the inputs are vectors now, not quaternions. What do other people think about this?

Check some publications and/or textbooks and see which name they give to this vector.

DomFijan:
I did a limited analysis on wording used for orientation vectors. It seems that in most cases papers just use "orientation" when talking about orientation vectors. [This](https://pubs-rsc-org.proxy.lib.umich.edu/en/content/articlehtml/2008/sm/b715764a) paper also used "orientation vector" once in one of the definitions but just "orientation" in others. [This](https://pubs-rsc-org.proxy.lib.umich.edu/en/content/articlehtml/2008/sm/b715764a) paper used phrase "orientation axis" once, but in the context of molecular systems.  [This](https://aip.scitation.org/doi/full/10.1063/1.3589961?mimetype=application%2Fpdf&crawler=true) paper has a very formal overview of how nematic order parameter is derived and always uses term "orientation". I think leaving "orientations" would make most sense, but I wouldn't be opposed to using "orientation_vectors" either, it seems that they usually mean the same thing anyways.

freud/order.pyx
```diff
@@ -185,23 +196,27 @@ cdef class Nematic(_Compute):
...
         """   # noqa: E501
+        if orientations.shape[1] == 4:
+            raise ValueError('Since freud v3.0, Nematic.compute() API takes'
+                             'orientation vectors instead of quaternions.')
```
```suggestion
            raise ValueError('In freud versions >=3.0.0, Nematic.compute() takes'
                             '3d orientation vectors instead of 4d quaternions.')
```

tommy-waltmann (changes requested):
One last change to wording, then this will be ready for merge. Thanks so much for your work on this @DomFijan !

tommy-waltmann (approved):
LGTM!

PR #1096, opened 2023-04-11
Compute Class NeighborList Lifetime

### LINKED ISSUE #1095: Voronoi nlist seems to deallocate if voronoi object is garbage collected
### Description

When computing a neighbor list using `freud.locality.Voronoi`, if the Voronoi object is garbage collected but the neighbor list is not, accessing the properties of the neighbor list results in a segmentation fault.

This happens for multiple Python versions as well, though I report the one used for the minimal working example below.

I can fix this, but it may be some time before I get to it, since there is a simple work around of just storing the Voronoi object.

### Steps to Reproduce

```python
import freud
import numpy as np

N_replica = 6
system = freud.data.UnitCell.sc().generate_system(N_replica)

def get_voronoi(system):
    voronoi = freud.locality.Voronoi()
    voronoi.compute(system)
    return voronoi.nlist

nlist = get_voronoi(system)
nlist.neighbor_counts
```

### Error Output

```shell
No error output just a segmentation fault.
```

### freud Version

v2.12.1

### Python Version

v3.10.6

### System Platform

Linux

### Installation method

built from source

### Developer

None

tommy-waltmann:
## Description

This PR creates copies of the neighborlists generated internally by many of freud's compute classes. On the backend, these neighborlists do not live past the lifetime of the compute classes that create them and we were using shallow neighborlist copies to expose them in python. Once the compute object was garbage collected, the pointers held by the neighborlist copies were invalid

## Motivation and Context

The original reason for this design decision was to prevent excessive copying of potentially large neighborlists to keep the code more performant and less memory intensive. However, this results in non-intuitive behavior from the perspective of the user.

Resolves: #1095 

## How Has This Been Tested?

I have added a test for all affected classes.

## Types of changes

## Checklist:


b-butler (approved):
Looks good to me! Thanks for the fix @tommy-waltmann.

tommy-waltmann:
Merging this will need to be downstream of #1097 getting merged.

tommy-waltmann:
This PR may also benefit from adding a section in the Using Freud Efficiently part of the documentation to prevent users from excessive neighborlist copies.

tommy-waltmann:
I decided on a different implementation which doesn't create any new copies. It instead gives the neighborlist a reference to the compute which generated it, preventing the compute (and underlying C++ neighborlist data) from being prematurely garbage collected.

b-butler (approved):
Thanks for the work @tommy-waltmann and nice solution.

PR #1184, opened 2023-12-04
Zero rotation nematic

AlainKadar:
## Description
Raise a warning when a zero vector is in the orientation list passed to `freud.order.Nematic.compute`

## Motivation and Context
Without this change, if user passes zero vector to `compute`, they may not realize why the director is incorrect.

## How Has This Been Tested?
Test was added to ensure that a warning is raised when the zero vector is passed.

## Types of changes

## Checklist:


freud/order.pyx
```diff
@@ -213,6 +213,10 @@ cdef class Nematic(_Compute):
         orientations = freud.util._convert_array(
             orientations, shape=(None, 3))
 
+        if len(np.where(~orientations.any(axis=1))[0])!=0:
+            warnings.warn('You supplied a zero vector in the orientations'
+                          'array, which causes unusual behavior.',
```
```suggestion
            warnings.warn('The zero vector was detected in the orientations array, which causes unusual behavior.',
```
To be clear, what exactly is the output when the zero vector is given as the input? I'd like to change the message to be a bit more explicit about what the user should expect to see.

ChangeLog.md
```diff
@@ -14,7 +14,7 @@ and this project adheres to
 * The `normalize` argument to `freud.density.RDF` is now `normalization_mode`.
 * The ``neighbors`` argument to ``env_neighbors`` for ``EnvironmentMotifMatch`` class.
 * The ``neighbors`` argument to ``cluster_neighbors`` for ``EnvironmentCluster`` class.
-* `freud.order.Nematic` uses orientation vectors instead of quaternions and a nematic director.
+* `freud.order.Nematic` uses orientation vectors instead of quaternions and a nematic director and raises warning when zero vector passed.
```
```suggestion
* `freud.order.Nematic` uses orientation vectors instead of quaternions and a nematic director.
* `freud.order.Nematic` raises a warning when the zero is vector passed.
```

tommy-waltmann (changes requested):

freud/order.pyx
AlainKadar: `op.order` is zero; Entries of `op.nematic_tensor` are `NaN`; `op.director` points in the wrong direction. However `op.particle_tensors` are correct. I agree a more explicit error message would be better but I don't think fitting in all these details is necessary.

tommy-waltmann: Let's use the following error message then:

'Including zero vector in the orientations array may lead to undefined behavior.'

tommy-waltmann (approved):
Looks good, great work on this @AlainKadar !

PR #1225, opened 2024-02-27
Feat/cnv

b-butler:
## Description
Adds a new feature, the continuous coordination number featured in chapter 5 of my dissertation.
The feature relaxes the Voronoi neighbor coordination number to the reals via one of many functions.
The PR implements the family of powers, logarithms, and exponentiation.

## Motivation and Context
Just a new potentially useful feature.

## How Has This Been Tested?
I wrote tests that look at basic behavior as well as exact correspondence to known values of this for perfect crystals. 2D is also tested.

## Types of changes

## Checklist:

- [x] I have updated the [credits](https://github.com/glotzerlab/freud/blob/main/doc/source/reference/credits.rst).


freud/density.pyx
```diff
@@ -732,3 +733,126 @@ cdef class RDF(_SpatialHistogram1D):
             return freud.plot._ax_to_bytes(self.plot())
         except (AttributeError, ImportError):
             return None
+
+
+cdef class ContinuousCoordination(_PairCompute):
+    r"""Computes the continuous local coordination number.
+
+    The :class:`ContinuousCoordination` impliments extensions of the Voronoi
```
tommy-waltmann:
```suggestion
    The :class:`ContinuousCoordination` class implements extensions of the Voronoi
```

freud/density.pyx
```diff
@@ -732,3 +733,126 @@ cdef class RDF(_SpatialHistogram1D):
...
+    discrete coordination number to the real numbers. These are motivated by the
+    dissertation <cite>. The formulas for the various implementations are:
```
tommy-waltmann: I think it would be best for now to not mention the dissertation until it is publicly citable. We should create an issue to cite the dissertation once its available for citation. Do you have any idea how long that will take?

freud/density.pyx
```diff
@@ -732,3 +733,126 @@ cdef class RDF(_SpatialHistogram1D):
...
+
+    Power
+
+    .. eq::
+
+        CN_p = N^{2.0 - m} \sum_{i=1}^{k}
+        {\left[\left(\frac{V_i}{V}\right)^{m}\right]}^{-1}
+
```
tommy-waltmann: These equations aren't rendering in the docs. I don't know exactly why, but we typically use a `.. math::` block.

freud/density.pyx
```diff
@@ -732,3 +733,126 @@ cdef class RDF(_SpatialHistogram1D):
...
+    and whose height is half the distance vector, and :math:`V` is the
+    volume/area of the Voronoi polytope.
+
+    Args:
+        powers (list[float]): The powers to compute the continuous coordination
+            number for.
+        compute_log (`bool`, optional): Whether to compute the log continuous
+            coordination number.
+        compute_exp (`bool`, optional): Whether to compute the exp continuous
+            coordination number.
```
tommy-waltmann: We should make every argument to the constructor optional. There may be a case where a user just wants to compute the log and/or exponential, not the powers. With that in mind, what do you believe the best set of default coordination numbers to compute are?

freud/density.pyx
```diff
@@ -732,3 +733,126 @@ cdef class RDF(_SpatialHistogram1D):
...
+        cdef bool is2D
+
+        neighbor_query = freud.locality.NeighborQuery.from_system(system)
+        is2D = neighbor_query.box.is2D
+
+        if voronoi is None:
+            voronoi = freud.locality.Voronoi()
+            voronoi.compute(system)
+        cpp_voronoi, cpp_nlist = (voronoi, voronoi.nlist)
+        self.thisptr.compute(cpp_voronoi.thisptr, cpp_nlist.get_ptr(), is2D)
```
tommy-waltmann: I'd like to see this compute take the voronoi object as the only argument. The neighborlist and number of dimensions can be accessed from the voronoi object in c++.

freud/density.pyx
```diff
@@ -732,3 +733,126 @@ cdef class RDF(_SpatialHistogram1D):
...
+    cdef freud._density.ContinuousCoordination* thisptr
+
+    def __cinit__(self, powers, compute_log=False, compute_exp=False):
+        self.thisptr = new freud._density.ContinuousCoordination(
+            powers, compute_log, compute_exp)
+
+    def __dealloc__(self):
+        del self.thisptr
+
+    def compute(self, system, voronoi=None):
```
tommy-waltmann: The system argument should be optional here. In cases where a user supplies an already computed voronoi object, a box/points is not necessary to do the calculation. In the code, we should check that in cases where both the system and voronoi are `None`, an error is thrown and mention in the docs that the system will not be used if a voronoi object is given.

freud/density.pyx
```diff
@@ -732,3 +733,126 @@ cdef class RDF(_SpatialHistogram1D):
...
+        return freud.util.make_managed_numpy_array(
+            &self.thisptr.getCoordination(),
+            freud.util.arr_type_t.FLOAT)
+
+    @property
+    def powers(self):
+        """list[float]: The powers to compute the continuous coordination number.
+
+        Changes to this property are not reflected when computing coordination
+        numbers.
```
tommy-waltmann: This property is read-only right now, so I think this comment is moot.

cpp/density/ContinuousCoordination.h
```diff
@@ -0,0 +1,65 @@
...
+
+    //! Get a reference to the last computed number of neighbors
+    const util::ManagedArray<float>& getCoordination() const
+    {
+        return m_coordination;
+    }
+
+private:
+    //! Get the number of coordination numbers to compute.
+    unsigned int getNumberOfCoordinations();
```
tommy-waltmann: I think this could be a useful property to expose in the python API.

freud/density.pyx
```diff
@@ -732,3 +733,126 @@ cdef class RDF(_SpatialHistogram1D):
...
+            voronoi = freud.locality.Voronoi()
+            voronoi.compute(system)
+        cpp_voronoi, cpp_nlist = (voronoi, voronoi.nlist)
+        self.thisptr.compute(cpp_voronoi.thisptr, cpp_nlist.get_ptr(), is2D)
+        return self
+
+    @_Compute._computed_property
+    def coordination(self):
+        """(:math:`(N_{points}, N_{coord}`) :class:`numpy.ndarray`: \
+                coordination of points per query point."""
```
tommy-waltmann: Let's make a note, either here or in the class docstring, about the order coordinations are returned in.

cpp/density/CMakeLists.txt
```diff
@@ -12,3 +14,8 @@ add_library(
   SphereVoxelization.cc)
 
 target_link_libraries(_density PUBLIC TBB::tbb)
+
+# We treat the extern folder as a SYSTEM library to avoid getting any diagnostic
+# information from it. In particular, this avoids clang-tidy throwing errors due
+# to any issues in external code.
+target_include_directories(_density SYSTEM PUBLIC ${PROJECT_SOURCE_DIR}/extern/)
```
tommy-waltmann: Which submodule does this compute use?

tommy-waltmann (changes requested):
I have quite a few comments. It looks like the actual implementation is in good shape, but the API should be tweaked a bit and the docs could use more explanation, which I've written in comments.

Also, I think this compute more naturally fits in the order module rather than the density module, since the coordinations it computes are generally used as order parameters.

cpp/density/CMakeLists.txt
b-butler: voro++ must be included here otherwise in compiling the `_density` target CMake fails to be able to find voro++ headers.

freud/density.pyx
b-butler: Not necessarily. Operations like `compute.powers.append(4)` would be completely syntactically correct and work... just not as intended by a user. We could mitigate this by manual converting to a tuple to prevent these shenanigans.

freud/density.pyx
b-butler: We would need to make `is2D` a compute argument then. That is fine, but perhaps a bit out of touch with other freud computes, but I can do that. We need to decide what happens if a system and is2D is passed.

freud/density.pyx
b-butler: I am not sure what the _correct_ default is. A user can easily do what you are saying by `ContinuousCoordination([], True)` for instance. I can set the default value to `()` which would be as unopinionated as possible.

freud/density.pyx
tommy-waltmann: The voronoi object has a box, which can be used in the c++ code when we need to know the number of dimensions.

freud/density.pyx
tommy-waltmann: That works for me. I am asking you for the most reasonable set of defaults because you have the most experience with this compute and probably know which coordinations are more/less useful.

freud/density.pyx
tommy-waltmann: For other cases like this, we have ignored that fact that a user may try to do that. It hasn't led to issues that I'm aware of.

b-butler: So do you want me to remove the comment or just not bother with changing it to a tuple?

freud/density.pyx
b-butler: Not really.

tommy-waltmann: Okay. I'll make an issue for it and we can add it at a later date.

doc/source/modules/density.rst
```diff
@@ -7,6 +7,7 @@ Density Module
 .. autosummary::
     :nosignatures:
 
+    freud.density.ContinuousCoordination
```
tommy-waltmann: This should be moved to the order module so it can show up in the right place in the docs.

doc/source/reference/credits.rst
```diff
@@ -296,6 +296,7 @@ Yezhi Jin
 
 Brandon Butler
 
+* Added the ``freud.density.ContinuousCoordination`` compute.
```
tommy-waltmann:
```suggestion
* Added the ``freud.order.ContinuousCoordination`` compute.
```

freud/order.pyx
```diff
@@ -1071,3 +1071,141 @@ cdef class RotationalAutocorrelation(_Compute):
...
+
+
+cdef class ContinuousCoordination(_PairCompute):
+    r"""Computes the continuous local coordination number.
+
+    The :class:`ContinuousCoordination` class implements extensions of the Voronoi
+    discrete coordination number to the real numbers. The formulas for the
+    various implementations are:
+
+    Power
```
```suggestion
    Power:
```

freud/order.pyx
```diff
@@ -1071,3 +1071,141 @@ cdef class RotationalAutocorrelation(_Compute):
...
+
+    .. math::
+
+        CN_p = N^{2.0 - m} \sum_{i=1}^{k}
+        {\left[\left(\frac{V_i}{V}\right)^{m}\right]}^{-1}
+
+    Log
```
```suggestion
    Log:
```

freud/order.pyx
```diff
@@ -1071,3 +1071,141 @@ cdef class RotationalAutocorrelation(_Compute):
...
+
+    .. math::
+
+        CN_{log} = \frac{-1}{\log{N}} \sum_{i=1}^{k}\log{\left(\frac{V_i}{V}\right)}
+
+    Exp
```
```suggestion
    Exponential:
```

freud/order.pyx
```diff
@@ -1071,3 +1071,141 @@ cdef class RotationalAutocorrelation(_Compute):
...
+
+    .. math::
+
+        CN_{exp} = \sum_{i=1}^{k}\exp{\left(\frac{V_i}{V} - \frac{1}{N} \right)}
+
+    where :math:`k` is the number of neighbors a particle has, :math:`V_i` is
+    the volume of the pyramid (or area of the triangle) whose base is the
```
```suggestion
    the volume of the pyramid (or area of the triangle in 2D) whose base is the
```

freud/order.pyx
```diff
@@ -1071,3 +1071,141 @@ cdef class RotationalAutocorrelation(_Compute):
...
+        less correlation.
+
+    Args:
+        powers (list[float], optional): The powers to compute the continuous
+            coordination number for. The default value indicates only compute
+            for power 2.0.
+            (Default value: None)
+        compute_log (`bool`, optional): Whether to compute the log continuous
+            coordination number.
+            (Default value: True)
```
```suggestion
            (Default value: :code:`True`)
```

freud/order.pyx
```diff
@@ -1071,3 +1071,141 @@ cdef class RotationalAutocorrelation(_Compute):
...
+        compute_exp (`bool`, optional): Whether to compute the exp continuous
+            coordination number.
+            (Default value: True)
```
```suggestion
            (Default value: :code:`True`)
```

freud/order.pyx
```diff
@@ -1071,3 +1071,141 @@ cdef class RotationalAutocorrelation(_Compute):
...
+            >>> box, points = freud.data.make_random_system(10, 100, seed=0)
+            >>> # Compute ContinuousCoordination
+            >>> coord = freud.order.ContinuousCoordination([2, 4], True)
+            >>> coord.compute(system=(box, points))
+            freud.order.ContinuousCoordination(...)
+
+        Args:
+            system (optional):
+                Any object that is a valid argument to
+                :class:`freud.locality.NeighborQuery.from_system`.
```
tommy-waltmann, line 1151:
```suggestion
                :class:`freud.locality.NeighborQuery.from_system`. (Default value: None).
```

freud/order.pyx
```diff
@@ -1071,3 +1071,141 @@ cdef class RotationalAutocorrelation(_Compute):
...
+        if voronoi is None:
+            voronoi = freud.locality.Voronoi()
+            voronoi.compute(system)
+        cpp_voronoi = voronoi
+        self.thisptr.compute(cpp_voronoi.thisptr)
+        return self
+
+    @_Compute._computed_property
+    def coordination(self):
+        """(:math:`(N_{points}, N_{coord}`) :class:`numpy.ndarray`: \
```
tommy-waltmann, line 1173:
```suggestion
        """(:math:`(N_{points}, N_{coord}`) :class:`numpy.ndarray`): \
```

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -212,6 +212,39 @@ void loopOverNeighbors(const NeighborQuery* neighbor_query, const vec3<float>* q
     }
 }
 
+//! Wrapper iterating looping over NeighborList.
+/*! This function is designed for computations that must performs some sort of
```
tommy-waltmann:
```suggestion
/*! This function is designed for computations that must perform some sort of
```

cpp/locality/NeighborComputeFunctional.h
```diff
@@ -212,6 +212,39 @@ void loopOverNeighbors(const NeighborQuery* neighbor_query, const vec3<float>* q
...
+ *  pre- or post-processing on a per-point basis. The provide compute function
```
tommy-waltmann:
```suggestion
 *  pre- or post-processing on a per-point basis. The provided compute function
```

tests/test_order_ContinuousCoordination.py
```diff
@@ -0,0 +1,68 @@
...
+        self.compute_log = True
+        self.compute_exp = True
+        self.coord = freud.order.ContinuousCoordination(
+            self.powers, self.compute_log, self.compute_exp
+        )
+        self.voronoi = freud.locality.Voronoi()
+
+    def compute(self, system):
+        self.voronoi.compute(system)
+        self.coord.compute(system, self.voronoi)
```
tommy-waltmann, line 31: Let's add test cases where we call the compute method with just a voronoi object, just a voronoi object which has not already called its own compute method, and neither a voronoi or system.

tests/test_order_ContinuousCoordination.py
```diff
@@ -0,0 +1,68 @@
...
+        self.box, self.pos = freud.data.make_random_system(
+            box_size, num_points, seed=123
+        )
+        # 0.0 is the same as just counting neighbor and can be used for testing
+        self.powers = [0.0, 2.0, 4.0, 8.0]
+        self.compute_log = True
+        self.compute_exp = True
+        self.coord = freud.order.ContinuousCoordination(
+            self.powers, self.compute_log, self.compute_exp
+        )
```
tommy-waltmann, line 26: Let's add test where we give the constructor many different combinations of coordinations and make it sure it computes the right number of coordinations.

tommy-waltmann (changes requested):
This is getting much better! I have mostly small comments this time about the docstrings and I've requested to more completely test the API.

freud/density.pyx
tommy-waltmann: Everything can be left as-is.

ChangeLog.md
```diff
@@ -4,6 +4,11 @@ The format is based on
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v.3.1.0 -- 20xx-xx-xx
+
+### Added
+* New continuous coordination number compute `freud.density.ContinuousCoordination`.
```
b-butler:
```suggestion
* New continuous coordination number compute `freud.order.ContinuousCoordination`.
```

freud/order.pyx
```diff
@@ -1071,3 +1071,154 @@ cdef class RotationalAutocorrelation(_Compute):
...
+
+    Exponential:
+
+    .. math::
+
+        CN_{exp} = \sum_{i=1}^{k}\exp{\left(\frac{V_i}{V} - \frac{1}{N} \right)}
+
+    where :math:`k` is the number of neighbors a particle has, :math:`V_i` is
+    the volume of the pyramid (or area of the triangle in 2D) whose base is the
+    Voronoi polytope facet between the central particle and neighobor :math:`i`
```
```suggestion
    Voronoi polytope facet between the central particle and neighbor :math:`i`
```

tommy-waltmann (approved):
I found one last typo, then this can be merged.

PR #1234, opened 2024-03-07
Add class method for Box class which create a box from Lx,Ly,Lz and angles

DomFijan:
Implements classmethods that create a box class from ~~lattice vector matrix and~~ L1, L2, L3 and angles, as well as methods that give L1, L2, L3 and angles from box class.

## Description
Crystals are usually defined via their lattice vectors and wykoff sites or lengths of lattice vectors and angles; while freud and hoomd use cell parameter matrix. In this PR I implement class method for creation of box ~~from lattice vectors or~~ from lattice vector lengths and angles. I've also implemented a method that gives lengths of box vectors and angles from box object.

## Motivation and Context
Lattice vectors or lattice vector lengths and angles are used to define real crystals and having a method that works with this makes creating crystals easier.

## How Has This Been Tested?
Tests were added for new methods for box testing.

## Types of changes

## Checklist:

- [x] I have updated the [credits](https://github.com/glotzerlab/freud/blob/main/doc/source/reference/credits.rst).


janbridley:
I personally am a proponent of using math.sqrt and math.acos, as for single values it is significantly faster than the numpy alternatives. The difference is on the order of μs, but free performance is free imo.

DomFijan:
> I personally am a proponent of using math.sqrt and math.acos, as for single values it is significantly faster than the numpy alternatives. The difference is on the order of μs, but free performance is free imo.

It seems numpy sqrt is used throughout freud from what I can tell. If we were to start using math functions we should change all occurrences. I agree it makes no sense to leave performance on the table, but realistically the differences will be so small they are inconsequential. Perhaps this would be a good first issue to tackle for someone new?

freud/box.pyx
```diff
@@ -921,6 +940,75 @@ cdef class Box:
...
+        Returns:
+            :class:`~.UnitCell`: A unit cell with the given lattice vectors.
+        """
+        lattice_matrix = np.asarray(lattice_vectors, dtype=np.float32)
+        v0 = lattice_matrix[0]
+        v1 = lattice_matrix[1]
+        v2 = lattice_matrix[2]
+        Lx = np.sqrt(np.dot(v0, v0))
+        a2x = np.dot(v0, v1) / Lx
+        Ly = np.sqrt(np.dot(v1, v1) - a2x * a2x)
```
janbridley:
```suggestion
        Ly = np.sqrt(np.dot(v1, v1) - a2x**2)
```

I think a square here better conveys the intent, as below we have a very similar function with two different variables `(a2x * a3x)`

tests/test_box_Box.py
```diff
@@ -485,6 +485,64 @@ def test_from_box(self):
...
+            np.pi / 2.25,
+            np.pi / 2.35,
+        )
+        box = freud.box.Box.from_box_lengths_and_angles(
+            *original_box_lengths_and_angles
+        )
+        lengths_and_angles_computed = box.to_box_lengths_and_angles()
+        assert np.allclose(lengths_and_angles_computed, original_box_lengths_and_angles)
+
+    def test_from_lattice_vectors_cubic_lattice(self):
```
janbridley: Should we parametrize these tests (cubic and non_cubic) as cuboids? The math is pretty simple and it is (arguably) simpler:

```python
    @pytest.mark.parametrize(
        "cuboid_lengths",
        [
            [1, 1, 1],
            [1, 2, 3],
            [1, 1, 0],
            [4, 2, 0],
        ],
    )
    def test_from_lattice_vectors_cuboids(self, cuboid_lengths):
        lattice_vectors = np.diag(cuboid_lengths)
        unit_cell = freud.box.Box.from_lattice_vectors(lattice_vectors)

        freud_solution = [*unit_cell.to_dict().values()]
        exact_solution = [*cuboid_lengths, 0, 0, 0, 2 + (cuboid_lengths[-1] != 0)]


        assert np.allclose(freud_solution, exact_solution, atol=1e-15)
```

janbridley:
> > I personally am a proponent of using math.sqrt and math.acos, as for single values it is significantly faster than the numpy alternatives. The difference is on the order of μs, but free performance is free imo.
> 
> It seems numpy sqrt is used throughout freud from what I can tell. If we were to start using math functions we should change all occurrences. I agree it makes no sense to leave performance on the table, but realistically the differences will be so small they are inconsequential. Perhaps this would be a good first issue to tackle for someone new?

Noted, and sounds good! I will raise an issue.

Edit: See #1236

tommy-waltmann:
> I personally am a proponent of using math.sqrt and math.acos, as for single values it is significantly faster than the numpy alternatives. The difference is on the order of μs, but free performance is free imo.

Keep in mind freud modules are compiled with Cython, I wouldn't trust any performance data that is done outside a cython-compiled python module for use in freud.

freud/box.pyx
```diff
@@ -689,6 +689,25 @@ cdef class Box:
                            [0, self.Ly, self.yz * self.Lz],
                            [0, 0, self.Lz]])
 
+    def to_box_lengths_and_angles(self):
+        r"""Return the box lengths and angles.
+
+        Returns:
+            tuple: The box vector lengths and angles in radians
```
tommy-waltmann:
```suggestion
            tuple: The box vector lengths and angles in radians :math:`(L_1, L_2, L_3, \alpha, \beta, \gamma)`.
```
Let's be more explicit about what exactly is being returned.

freud/box.pyx
```diff
@@ -921,6 +940,75 @@ cdef class Box:
                             "positional argument: L")
         return cls(Lx=L, Ly=L, Lz=0, xy=0, xz=0, yz=0, is2D=True)
 
+    @classmethod
+    def from_lattice_vectors(cls, lattice_vectors, dimensions: int = None):
```
tommy-waltmann: How is this different than `from_matrix`? `Box.from_matrix` makes a box from a matrix where the columns are the lattice vectors. This seems very similar to that.

freud/box.pyx
```diff
@@ -921,6 +940,75 @@ cdef class Box:
...
+    @classmethod
+    def from_box_lengths_and_angles(
+        cls,
+        L1: float,
+        L2: float,
+        L3: float,
+        alpha: float,
+        beta: float,
+        gamma: float,
+        dimensions: int = None,
```
tommy-waltmann: freud's documentation doesn't use types directly next to the parameter names in the function definitions.

freud/box.pyx
```diff
@@ -921,6 +940,75 @@ cdef class Box:
...
+    ):
+        r"""Construct a box from lengths and angles.
+
+        Args:
+            L1 (float): The length of the first lattice vector
+            L2 (float): The length of the second lattice vector
+            L3 (float): The length of the third lattice vector
+            alpha (float): The angle between the y and z axes in radians
+            beta (float): The angle between the x and z axes in radians
+            gamma (float): The angle between the x and y axes in radians
```
tommy-waltmann: I am confused by this description. The angle between axes is always 90 degrees.

freud/data.py
```diff
@@ -233,6 +233,56 @@ def hex(cls):
         fractions = np.array([[0, 0, 0], [0.5, 0.5, 0]])
         return cls([1, np.sqrt(3)], fractions)
 
+    @classmethod
+    def from_lattice_vectors(
```
tommy-waltmann: Since this method (and `from_box_lengths_and_angles`) just replaces two lines of user script, I don't think it's needed. I am open to discussion though.

tommy-waltmann (changes requested):
Great work on this!

I have some conceptual comments that I'd like to discuss before doing a more thorough review.

freud/box.pyx
DomFijan: Not sure how this got past me but with this in mind this opens a whole new debate. `from_matrix` in freud basically uses lattice vectors as input, but this is not stated explicitly. Instead a link to hoomd docs is given which further complicates the problem. In hoomd `from_matrix` wants upper triangular matrix that is already a box like object as defined in hoomd docs. What we should do is to have freuds `from_matrix` take upper triangular matrix as input, such that its functionality is equivalent to hoomd's and add `from_lattice_vectors` which has same function as current `from_matrix`?

freud/data.py
DomFijan: It's just a convenience function. I love these. If you strongly feel they are not needed we can remove them. Up to you.

freud/box.pyx
DomFijan: You are right, this should state angle between lattice vectors 1,2,3 etc.

freud/data.py
tommy-waltmann: My personal logic is that if we have a convenience function for making `UnitCell`'s in the `data` module for `Box.from_lattice_vectors` and `Box.from_box_lengths_and_angles`, then we should also have a convenience function for every factory method in the `Box` class (like `cube`, `from_box`, `from_matrix`, `square`, etc.). There are quite a few of those and it may get cumbersome to have so many convenience functions hanging around.

If you like having that convenience function, add one to your python scripts. It only takes two lines to write it, after all :)

freud/box.pyx
tommy-waltmann: I think it would be best to add to freud's documentation and explicitly state that the columns of the input to `from_matrix` are the lattice vectors of the box. Any other improvements to the documentation of `from_matrix` to make it less confusing would also be welcome.

Looking at the diff of the code between `from_lattice_vectors` and `from_matrix`, they do the exact same thing so I don't see a need to have both factory methods hanging around.

DomFijan: Is the functionality of freud's `from_matrix` equivalent to hoomd's `from_matrix`?

freud/data.py
janbridley: While I see what you're saying, there are a significant number of users/use cases for this specific functionality: making the translation between freud boxes and data from crystallography files more transparent is a good thing in my opinion. I think adding this particular convenience function provides enough to users that we can probably do without the rest mentioned above. If this truly doesn't fit though, I can live without it.

freud/box.pyx
tommy-waltmann: It appears that hoomd's `from_matrix` requires an upper triangular matrix, whereas freud's is more general and does not.

freud/data.py
tommy-waltmann: What data does a crystallography file typically contain? If its a common use case to make a freud `UnitCell` from crystallography file data, then it makes sense to have a convenience function for it. I personally don't work a lot with those, so I'm not entirely sure what would be most helpful for others who work with it.

I think this topic is worth further offline discussion as well.

janbridley: CIF files (the most common file type, at least for computational/experimental researchers) have unit cell information encoded as a/b/c/α/β/γ. For DFT peeps, VASP and ABINIT provide lattice vectors so those would work without this function.

[Crystallography Open Database](http://www.crystallography.net/cod/)
```cif
_cell_angle_alpha                90.000
_cell_angle_beta                 102.104(8)
_cell_angle_gamma                90.000
_cell_formula_units_Z            4
_cell_length_a                   7.9661(7)
_cell_length_b                   9.205(1)
_cell_length_c                   7.3198(8)
_cell_volume                     524.8(9)
```

[AFlow](https://www.aflowlib.org):
```cif
_cell_length_a  5.4030000000
_cell_length_b  3.4330000000
_cell_length_c  5.0790000000
_cell_angle_alpha  90.0000000000
_cell_angle_beta  132.3200000000
_cell_angle_gamma  90.0000000000
```

DomFijan:
I think it makes sense to remove the class methods for UnitCell for now and to completely remove from_lattice_vectors for now. We can discuss further how this should be approached and implement more substantial changes in a future PR. I will update PR description accordingly.

freud/box.pyx
```diff
@@ -922,6 +942,35 @@ cdef class Box:
...
+        """
+        a1 = np.array([L1, 0, 0])
+        a2 = np.array([L2 * np.cos(gamma), L2 * np.sin(gamma), 0])
+        a3x = np.cos(beta)
+        a3y = (np.cos(alpha) - np.cos(beta) * np.cos(gamma)) / np.sin(gamma)
+        a3z = np.sqrt(1 - a3x**2 - a3y**2)
+        a3 = np.array([L3 * a3x, L3 * a3y, L3 * a3z])
+        if dimensions is None:
+            dimensions = 2 if L3 == 0 else 3
+        return cls.from_lattice_vectors(np.array([a1, a2, a3]), dimensions=dimensions)
```
janbridley:
```suggestion
        return cls.from_matrix(np.array([a1, a2, a3]).T, dimensions=dimensions)
```

from_lattice_vectors was removed

janbridley (approved):
Looks good. I am a fan of parameterizing a few examples for each test, but this should be sufficient. Feel free to merge when ready.

freud/box.pyx
```diff
@@ -921,6 +942,41 @@ cdef class Box:
...
 
+    @classmethod
+    def from_box_lengths_and_angles(
+        cls, L1, L2, L3, alpha, beta, gamma, dimensions=None,
+    ):
+        r"""Construct a box from lengths and angles.
+
+        Args:
+            L1 (float):
+                The length of the first lattice vector
```
tommy-waltmann:
```suggestion
                The length of the first lattice vector.
```

freud/box.pyx
```diff
@@ -921,6 +942,41 @@ cdef class Box:
...
+            L2 (float):
+                The length of the second lattice vector
```
tommy-waltmann:
```suggestion
                The length of the second lattice vector.
```

freud/box.pyx
```diff
@@ -921,6 +942,41 @@ cdef class Box:
...
+            L3 (float):
+                The length of the third lattice vector
```
tommy-waltmann:
```suggestion
                The length of the third lattice vector.
```

freud/box.pyx
```diff
@@ -921,6 +942,41 @@ cdef class Box:
...
+            alpha (float):
+                The angle between second and third lattice vector in radians
```
tommy-waltmann:
```suggestion
                The angle between second and third lattice vector in radians.
```

freud/box.pyx
```diff
@@ -921,6 +942,41 @@ cdef class Box:
...
+            beta (float):
+                The angle between first and third lattice vector in radians
```
tommy-waltmann:
```suggestion
                The angle between first and third lattice vector in radians.
```

freud/box.pyx
```diff
@@ -921,6 +942,41 @@ cdef class Box:
...
+            gamma (float):
+                The angle between the first and second lattice vector in radians
```
tommy-waltmann:
```suggestion
                The angle between the first and second lattice vector in radians.
```

freud/box.pyx
```diff
@@ -921,6 +942,41 @@ cdef class Box:
...
+            dimensions (int):
+                The number of dimensions (Default value = :code:`None`)
```
tommy-waltmann:
```suggestion
                The number of dimensions (Default value = :code:`None`).
```

ChangeLog.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to
 
 ### Added
 * New continuous coordination number compute `freud.order.ContinuousCoordination`.
+* New class methods for construction of `freud.box.Box` from box lengths and angles.
+* New `freud.box.Box.to_box_lengths_and_angles` method for computing box vector lengths
+  and angles.
```
```suggestion
* New methods for conversion of box lengths and angles to/from `freud.box.Box`.
```

tests/test_box_Box.py
```diff
@@ -485,6 +485,28 @@ def test_from_box(self):
         box7 = freud.box.Box.from_matrix(box.to_matrix())
         assert np.isclose(box.to_matrix(), box7.to_matrix()).all()
 
+    def test_standard_orthogonal_box(self):
+        box = freud.box.Box.from_box((1, 2, 3, 0, 0, 0))
+        Lx, Ly, Lz, alpha, beta, gamma = box.to_box_lengths_and_angles()
+        npt.assert_allclose(
+            (Lx, Ly, Lz, alpha, beta, gamma), (1, 2, 3, np.pi / 2, np.pi / 2, np.pi / 2)
+        )
+
```
tommy-waltmann, line 494: I think a great test to add would be to start with 3 random lengths and 3 random angles, then call `from_box_lengths_and_angles` to get a box, then call `to_box_lengths_and_angles` on that box and assert that you recover the original random lengths and angles.

freud/box.pyx
tommy-waltmann, line 1: What are the ranges on the allowed angles (e.g. between 0-90 degrees)? What should happen if a user gives an angle which is outside the range? I am okay with a garbage in garbage out philosophy, but we should document what is and isn't considered garbage.

tommy-waltmann (changes requested):
This is getting really close!

Most of my comments are about small stuff like periods in the docstrings, but there is one simple but robust test of the code I would like to see added.

freud/box.pyx
tommy-waltmann: After thinking about it more, the angles should all be between 0-180 degrees. Let's add a small note to parameter descriptions in `from_box_lengths_and_angles` about this.

freud/box.pyx
```diff
@@ -921,6 +942,43 @@ cdef class Box:
                             "positional argument: L")
         return cls(Lx=L, Ly=L, Lz=0, xy=0, xz=0, yz=0, is2D=True)
 
+    @classmethod
+    def from_box_lengths_and_angles(
+        cls, L1, L2, L3, alpha, beta, gamma, dimensions=None,
+    ):
+        r"""Construct a box from lengths and angles.
+
+        All the angles provided must be between 0 and :math:`\pi`.
```
janbridley, line 951: This matches the [cif spec](https://www.iucr.org/__data/iucr/cifdic_html/1/cif_core.dic/Icell_angle_.html) - nice.

janbridley (approved):

tests/test_box_Box.py
```diff
@@ -485,6 +485,28 @@ def test_from_box(self):
...
+    def test_to_and_from_box_lengths_and_angles(self):
+        original_box_lengths_and_angles = (
+            1,
+            2,
+            3,
+            np.pi / 3,
+            np.pi / 2.25,
+            np.pi / 2.35,
```
tommy-waltmann: Each of the six numbers should be the output of a different call to `np.random.rand`. That way, each time the test is run it gets run with a different random value.

tommy-waltmann (changes requested):
One final thing, then we merge.

Great work on this, I think many people will find this useful!

tests/test_box_Box.py
tommy-waltmann: I am currently working on adding this.

freud/box.pyx
```diff
@@ -921,6 +942,43 @@ cdef class Box:
...
+                The number of dimensions (Default value = :code:`None`).
+
+        Returns:
+            :class:`freud.box.Box`: The resulting box object.
+        """
+        a1 = np.array([L1, 0, 0])
+        a2 = np.array([L2 * np.cos(gamma), L2 * np.sin(gamma), 0])
+        a3x = np.cos(beta)
+        a3y = (np.cos(alpha) - np.cos(beta) * np.cos(gamma)) / np.sin(gamma)
+        a3z = np.sqrt(1 - a3x**2 - a3y**2)
```
tommy-waltmann: This line appears to take the square root of a negative number at times, resulting in NaNs.

DomFijan: some combinations of angles are not possible-there is likely a constraint on their sum or something like that. Is this taken into account?

tommy-waltmann: Not in how I wrote the test case, which would explain the failures. The code should throw an error if the user gives inputs which violate the constraints.

ispivack:
All looks good to me, just one comment. If we want to just use freud for the EBT box, it would be good to have a convenience function converting to a unit vector + lengths description. Do we feel that this would fit into freud, or should this just be separately implemented in ebonds?

tommy-waltmann:
> All looks good to me, just one comment. If we want to just use freud for the EBT box, it would be good to have a convenience function converting to a unit vector + lengths description. Do we feel that this would fit into freud, or should this just be separately implemented in ebonds?

I would start by adding the functionality into `ebonds`. If it becomes cumbersome to have to do the conversion in ebonds constantly, then adding a method to freud to do that might be worth it.

Right now, the best way to get lengths and unit vectors would be to get the box vectors (they are the columns of `box.to_matrix()`) and get magnitude and direction from that.

ispivack (approved):

janbridley:
`np.testing.assert_allclose` sets `atol=0` by default. This will not work with floating point calculations, so I set atol to 1e-14. However, the tests are still failing due to mismatched `NaN`s.

tommy-waltmann:
> `np.testing.assert_allclose` sets `atol=0` by default. This will not work with floating point calculations, so I set atol to 1e-14. However, the tests are still failing due to mismatched `NaN`s.

It's fine to leave atol as 0.0. According to the [documentation](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) for `assert_allclose`: "It compares the difference between actual and desired to atol + rtol * abs(desired)", so having an rtol is enough to account for floating point error.

joaander:
> > `np.testing.assert_allclose` sets `atol=0` by default. This will not work with floating point calculations, so I set atol to 1e-14. However, the tests are still failing due to mismatched `NaN`s.
> 
> It's fine to leave atol as 0.0. According to the [documentation](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) for `assert_allclose`: "It compares the difference between actual and desired to atol + rtol * abs(desired)", so having an rtol is enough to account for floating point error.

A non-zero `atol` is needed when a value close to zero should be accepted as a desired (exactly) zero value. `atol + rtol * abs(desired) == 0.0` when `atol=0` and `desired=0`.

janbridley:
I have tested this method against box lengths and angles from 1,104 real CIF files and all cases pass. I can export the box params as a numpy array and save it as test data, or we can adjust the current test to be more realistic.

tommy-waltmann:
> I have tested this method against box lengths and angles from 1,104 real CIF files and all cases pass. I can export the box params as a numpy array and save it as test data, or we can adjust the current test to be more realistic.

The code should be written so invalid sets of angle inputs raise an error to the user instead of creating invalid boxes. This PR needs an update to the code, not just the tests, before it's ready to be merged.

Ideally, we would know the mathematical constraints on the box angles so we could check that the inputs are valid, and we should make an effort to figure out what those are. If we cannot get those relations, then we could move forward with an updated set of tests and a "garbage in, garbage out" note in the documentation, but again, I only think we should take the second option if we have already given the first option some effort and failed.

DomFijan:
I think I figured it out. The restriction is the following:
$1+2\cos(\alpha)\cos(\beta)\cos(\gamma) - \cos(\alpha)^2 - \cos(\beta)^2 - \cos(\gamma)^2>0$
This is part of the computation of volume of a Parallelepiped. This guarantees that the volume of the box defined is positive and must always be true. This statement should be equivalent to requiring the value under the sqrt in the from_box_lengths_and_angles function be positive (haven't done the maths to confirm that). 

The geometric interpretation of this is the violation of the cosine law.

tommy-waltmann (approved):
Great work! I know this will be useful for lots of peeps using freud.

PR #1256, opened 2024-06-03
Convert Box Module to Use nanobind

tommy-waltmann:
## Description

This PR changes the box module to use pybind11 instead of Cython. It also establishes a new pattern in the cmake code which other modules will be able to follow for when they are converted.

## Motivation and Context
Cython is old and causing us a lot of developer headaches, so we are converting freud to use pybind11 instead.

## How Has This Been Tested?
All old tests pass. Please build freud from source and test the box module manually for now.

## Types of changes

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/main/CONTRIBUTING.md) document.


cpp/box/Box.h
```diff
@@ -4,13 +4,16 @@
 #ifndef BOX_H
 #define BOX_H
 
-#include "utils.h"
 #include <algorithm>
 #include <complex>
+#include <iostream>
```
tommy-waltmann: I can remove this, its leftover from debugging.

cpp/box/Box.h
```diff
@@ -23,6 +26,9 @@ constexpr float TWO_PI = 2.0 * M_PI;
 
 namespace freud { namespace box {
 
+template<typename T>
+using pybind11_array = pybind11::array_t<T, pybind11::array::c_style | pybind11::array::forcecast>;
```
tommy-waltmann: Best place for this typedef? I was thinking in freud::util since it will likely be used by each freud module.

vyasr:
If you're considering switching away from Cython you might want to consider nanobind instead. [The benchmarks look quite good](https://nanobind.readthedocs.io/en/latest/benchmark.html) and the API is nearly identical.

joaander:
> If you're considering switching away from Cython you might want to consider nanobind instead. [The benchmarks look quite good](https://nanobind.readthedocs.io/en/latest/benchmark.html) and the API is nearly identical.

Thanks for sharing, I wasn't aware of nanobind. It looks perfect for what we do.

... and let the endless cycle of rewriting our codes continue ...

tommy-waltmann:
> If you're considering switching away from Cython you might want to consider nanobind instead. [The benchmarks look quite good](https://nanobind.readthedocs.io/en/latest/benchmark.html) and the API is nearly identical.

The ndarray class looks to be great for interoperatibility between python and C++, which will be perfect for freud!

cpp/CMakeLists.txt
```diff
@@ -6,16 +6,31 @@ if(WIN32)
...
+      PARENT_SCOPE)
+endif()
+
+add_subdirectory(box)
+
+# commented out for now, uncomment them as the conversion to pybind11 progresses
+# add_subdirectory(cluster) add_subdirectory(density)
+# add_subdirectory(diffraction) add_subdirectory(environment)
+# add_subdirectory(locality) add_subdirectory(order) add_subdirectory(parallel)
+# add_subdirectory(pmft) add_subdirectory(util)
```
DomFijan, line 31: are these the same for nanobind?

cpp/CMakeLists.txt
```diff
@@ -53,3 +68,4 @@ if(CMAKE_EXPORT_COMPILE_COMMANDS)
     COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/compile_commands.json
             ${PROJECT_SOURCE_DIR})
 endif()
+]]
```
DomFijan, line 71: should this be commented?

cpp/box/Box.h
```diff
@@ -93,12 +99,6 @@ class Box
         return !(*this == b);
     }
 
-    //! Set L, box lengths, inverses.  Box is also centered at zero.
-    void setL(const vec3<float>& L)
-    {
-        setL(L.x, L.y, L.z);
-    }
-
```
DomFijan, line 101: Why is this no longer needed?

cpp/CMakeLists.txt
tommy-waltmann: These lines are CMake code.

cpp/CMakeLists.txt
tommy-waltmann: I left some things in, but commented out for the purpose of not forgetting to include them in later PRs, but I could just go to the main branch to see what was there before. I'll just get rid of them.

cpp/box/Box.h
tommy-waltmann: It's not exported to python, nor was it used anywhere in the C++ code. The name clashes with a different method named `setL`, which nanobind cannot resolve.

DomFijan:
I was only able to go through a small portion of PR. Will give it another go soon.

cpp/box/Box.h
```diff
@@ -230,12 +230,15 @@ class Box
      *  \param Nvecs Number of vectors
      *  \param out The array in which to place the wrapped vectors.
```
janbridley:
```suggestion
     *  \param out_data The array in which to place the wrapped vectors.
```

DomFijan (commented):
This looks good. It seems that, besides changes on the python side, we only have to change the data types to the nanobind array on c++ side + some minor tweaks.

One thing that I don't understand is why there are several sets of cpp functions such as makefraction and makefractionalPython. I can see that the "python" functions are used for exporting to python, but why is there a need for the non "python" function? Couldn't the logic be condensed into a single function? it seems that the main difference between the two functions is that non-python functions operate on a single set of data while the python functions do that for the whole system. Could we have either a single function for both use cases or if that is not possible, better names would also help (ie calling them _single vs. whole system or something like that?)

tommy-waltmann:
> This looks good. It seems that, besides changes on the python side, we only have to change the data types to the nanobind array on c++ side + some minor tweaks.
> 
> One thing that I don't understand is why there are several sets of cpp functions such as makefraction and makefractionalPython. I can see that the "python" functions are used for exporting to python, but why is there a need for the non "python" function? Couldn't the logic be condensed into a single function? it seems that the main difference between the two functions is that non-python functions operate on a single set of data while the python functions do that for the whole system. Could we have either a single function for both use cases or if that is not possible, better names would also help (ie calling them _single vs. whole system or something like that?)

Generally speaking, there are 3 kinds of functions:

1) Function that operates on a single particle
2) Function that operates on the whole system and take raw pointers as inputs
3) Function that operates on the whole system and takes `nb::ndarray` as inputs, name ends with Python suffix

Type (1) exists to increase code readability and reduce code duplication (they are re-used sometimes in multiple places).

Type (2) is largely leftover from cython-era freud, but could provide a C++ interface for the box class if at some point we decide to provide that. If we don't care about preserving something that could be easier to turn into a C++ API later, we could remove this type of function.

Type (3) is so C++ and python can talk to each other with nanobind. When exporting functions, both pybind11 and nanobind cannot resolve two functions with the same name, so the suffix `Python` is added to the name.

DomFijan (approved):
In that case, I see no good reason to remove any of the 3 types of functions.

joaander:
One solution is to implement the first 2 methods in `Box.h` and move then 3rd into the nanobind module export file. This compartmentalizes the C++-only code in `libfreud.so` for future use by C++ libraries yet makes the Python API available to Python users.

The 3rd function in this case wouldn't be a member function. It could either be a static function that behaves like a member function (taking a `shared_ptr<Box>` as the first argument) - or implemented in a lambda within the nanobind `.def`.

tommy-waltmann:
> One solution is to implement the first 2 methods in `Box.h` and move then 3rd into the nanobind module export file. This compartmentalizes the C++-only code in `libfreud.so` for future use by C++ libraries yet makes the Python API available to Python users.
> 
> The 3rd function in this case wouldn't be a member function. It could either be a static function that behaves like a member function (taking a `shared_ptr<Box>` as the first argument) - or implemented in a lambda within the nanobind `.def`.

Other than cluttering `libfreud.so` with methods that we don't recommend using, what is the downside of the 3rd function type existing in `libfreud.so`?

I haven't separated out the nanobind export modules (i.e. _box.cpython...) from the C++ library (libfreud.so) on this PR. I do this in the parallel module branch (which is also ready to go once this PR is merged), so it may be easier to wait until the next PR to make those changes.

joaander:
> > One solution is to implement the first 2 methods in `Box.h` and move then 3rd into the nanobind module export file. This compartmentalizes the C++-only code in `libfreud.so` for future use by C++ libraries yet makes the Python API available to Python users.
> > The 3rd function in this case wouldn't be a member function. It could either be a static function that behaves like a member function (taking a `shared_ptr<Box>` as the first argument) - or implemented in a lambda within the nanobind `.def`.
> 
> Other than cluttering `libfreud.so` with methods that we don't recommend using, what is the downside of the 3rd function type existing in `libfreud.so`?
> 
> I haven't separated out the nanobind export modules (i.e. _box.cpython...) from the C++ library (libfreud.so) on this PR. I do this in the parallel module branch (which is also ready to go once this PR is merged), so it may be easier to wait until the next PR to make those changes.

Certainly. If we decide to take this approach, this is a rearrange we could do after the next PR.

Other than clutter, one possible downside I see is that `libfreud.so` may need to be linked to the `nanobind.so` shared object (I'm not sure yet, as I haven't tried this yet). We would have to figure out how to do this without `nanobind_add_module`. 

Even if the parts of nanobind needed by `libfreud.so` are header-only, this forces C++-only downstream users of freud to have a Python environment with nanobind installed for those headers.

It is really a question of how strongly do we want to support potential C++-only use-cases. If we think there is even a small chance users would like this in the future, it will be worth the small amount of extra effort during this rewrite. Refactoring the cluttered `libfreud.so` in the future will be a much larger project.

CMakeLists.txt
joaander, line 1: nanobind bumps the minimum to 3.15. Use
```
cmake_minimum_required(VERSION 3.15...3.30)
```
to set the minimum to 3.15 and accept new default policies introduced in 3.15 through 3.30. This way we can ensure that our CMake is as modern as possible.

CMakeLists.txt
```diff
@@ -36,6 +49,15 @@ if(TBB_FOUND)
...
 endif()
 
+# go find nanobind
+execute_process(
+  COMMAND ${Python_EXECUTABLE} "-m" "nanobind" "--cmake_dir"
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  OUTPUT_VARIABLE NB_DIR)
+list(APPEND CMAKE_PREFIX_PATH "${NB_DIR}")
+message(STATUS ${CMAKE_PREFIX_PATH})
+find_package(nanobind CONFIG REQUIRED)
```
joaander:
```suggestion
# go find nanobind
execute_process(
  COMMAND ${Python_EXECUTABLE} "-m" "nanobind" "--cmake_dir"
  OUTPUT_STRIP_TRAILING_WHITESPACE
  OUTPUT_VARIABLE nanobind_ROOT)
find_package(nanobind CONFIG REQUIRED)
```
The goal here is to find the **specific** nanobind associated with the already found Python. Use the project-specific root variable to accomplish this instead of the global `CMAKE_PREFIX_PATH`.

cpp/box/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+if(APPLE)
+  set_target_properties(_box PROPERTIES INSTALL_RPATH "@loader_path")
+else()
+  set_target_properties(_box PROPERTIES INSTALL_RPATH "\$ORIGIN")
+endif()
+
+if(_using_conda OR DEFINED ENV{CIBUILDWHEEL})
+  set_target_properties(_box PROPERTIES INSTALL_RPATH_USE_LINK_PATH True)
+endif()
```
joaander, line 19:
```suggestion
```
I don't think this is necessary. Please confirm.

tommy-waltmann:
> It is really a question of how strongly do we want to support potential C++-only use-cases. If we think there is even a small chance users would like this in the future, it will be worth the small amount of extra effort during this rewrite. Refactoring the cluttered `libfreud.so` in the future will be a much larger project.

Ok, then let's do this on the next PR.

cpp/CMakeLists.txt
```diff
@@ -6,16 +6,31 @@ if(WIN32)
...
+  set(_using_conda On)
+  set(_using_conda
+      On
+      PARENT_SCOPE)
+else()
+  set(_using_conda Off)
+  set(_using_conda
+      Off
+      PARENT_SCOPE)
+endif()
```
joaander, line 23:
```suggestion
```
I don't think this is necessary. I removed similar blocks from gsd, fresnel, and hoomd a long time ago.

cpp/box/Box.h
```diff
@@ -228,14 +228,17 @@ class Box
     /*! \param vecs Vectors of fractional coordinates between 0 and 1 within
      *         parallelepipedal box
      *  \param Nvecs Number of vectors
-     *  \param out The array in which to place the wrapped vectors.
+     *  \param out_data The array in which to place the wrapped vectors.
      */
-    void makeAbsolute(const vec3<float>* vecs, unsigned int Nvecs, vec3<float>* out) const
+    void makeAbsolutePython(nb_array<float, nb::shape<-1, 3>> vecs, unsigned int Nvecs,
+                            nb_array<float, nb::shape<-1, 3>> out) const
```
joaander: nanobind will silently convert arrays that are not the right type, shape, or ordering. This is fine for input arrays, but the output array must not be converted, or else we write output to temporary copy.

I see the Python side wraps output arrays with: `_convert_array(..., allow_copy=False)`. Could we instead use the [`noconvert`](https://nanobind.readthedocs.io/en/latest/ndarray.html#arrays-and-function-overloads) method that `nanobind` provides? One would hope that checking at the C++ level is faster - and that check must already be performed. It would also be less code for us to maintain in freud (no `_convert_array` method and no multiple calls to `_convert_array` in every method).

freud/box.py
```diff
@@ -204,16 +206,11 @@ def make_absolute(self, fractional_coordinates, out=None):
...
 
-        cdef const float[:, ::1] l_points = fractions
-        cdef unsigned int Np = l_points.shape[0]
-        cdef float[:, ::1] l_out = out
+        Np = fractions.shape[0]
 
-        self.thisptr.makeAbsolute(
-            <vec3[float]*> &l_points[0, 0], Np,
-            <vec3[float]*> &l_out[0, 0])
+        self._cpp_obj.makeAbsolute(fractions, Np, out)
```
joaander: Continuing the discussion about `_convert_array` - there is no need to access `Np = fractions.shape[0]` here (and in similar methods). That check can be made in C++. In most cases, I would hope that the input "array" provided by the user can be passed directly in to the C++ method.

To preserve full `API` compatibility with freud 2.0 in this specific method, the `flatten` code above would need to be wrapped in some sort of type check.
```
        flatten = fractions.ndim == 1
        fractions = np.atleast_2d(fractions)
```
I don't know what specific API use-case the `atleast_2d` achieves. I see that it is used extensively in the box module. Is that true in the rest of freud?

cpp/box/Box.h
```diff
@@ -337,12 +346,15 @@ class Box
      *  \param Nvecs Number of vectors
      *  \param out The array in which to place the wrapped vectors.
      */
-    void wrap(const vec3<float>* vecs, unsigned int Nvecs, vec3<float>* out) const
+    void wrapPython(nb_array<float, nb::shape<-1, 3>> vecs, unsigned int Nvecs,
```
joaander:
```suggestion
    void wrapPython(nb_array<float, nb::shape<-1, 3>> vecs,
```
This (and similar) methods should discover `N` from the shape of the array. The duplicated information could be inconsistent, leading to out of bounds array accesses or bugs where only part of an array is read.

joaander (commented):
Thanks! This looks like a clean port.

I have a number of recommendations to improve the code and a few points to discuss further.

cpp/box/CMakeLists.txt
joaander: Correction, I don't think it is necessary for conda. It may be needed for CIBUILDWHEEL.

cpp/box/Box.h
joaander: Also: When the output must be the same size as the input, this (and similar methods) must verify that is the case.

cpp/box/CMakeLists.txt
tommy-waltmann: This is still necessary when I build on linux (WSL2) in a conda environment. Without it, I get a runtime error saying module `_box` cannot find `libtbb.so`.

freud/box.py
tommy-waltmann: freud chooses to allow users to give and receive arrays containing only a single position as a 1d array instead of 2d if they want. Much of the pre/post processing via calls to `atleast_2d` (which adds a dimension to a 1d array so its 2d) and `squeeze` is to support this use case.

This happens throughout the box module, but there is only one other call to `atleast_2d` in freud, which is located in the `NeighborQuery.query` method.

CMakeLists.txt
```diff
@@ -36,6 +47,13 @@ if(TBB_FOUND)
     "[${TBB_LIBRARY}][${TBB_INCLUDE_DIR}]")
 endif()
 
+# go find nanobind
+execute_process(
+  COMMAND ${Python_EXECUTABLE} "-m" "nanobind" "--cmake_dir"
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  OUTPUT_VARIABLE nanobind_ROOT)
+find_package(nanobind CONFIG REQUIRED)
+
```
joaander, line 61:
```suggestion
if (nanobind_FOUND)
    find_package_message(nanobind "Found nanobind: ${nanobind_DIR} ${nanobind_VERSION}" "[${nanobind_DIR},${nanobind_VERSION}]")
endif()
```
I forgot to add this earlier. It prints a status message indicating which nanobind was found. This is immensely useful for debugging CMake builds.

Once https://github.com/wjakob/nanobind/pull/618 (or an alternative solution) is merged and a new nanobind release is pushed, then we can also add the version requirement to `find_package`:
```
find_package(nanobind 2.0 CONFIG REQUIRED)
```

cpp/box/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+# create the target and add the needed properties
+nanobind_add_module(_box SHARED module-box.cc export-box.h)
```
joaander:
```suggestion
nanobind_add_module(_box NB_SHARED module-box.cc export-box.h)
```
I think we want `NB_SHARED` here per the discussion in https://nanobind.readthedocs.io/en/latest/api_cmake.html#high-level-interface . This avoids duplicating the nanobind compiled code in every freud module.

cpp/box/export-box.h
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2010-2023 The Regents of the University of Michigan
+// This file is from the freud project, released under the BSD 3-Clause License.
+
+#ifndef EXPORT_BOX_H
+#define EXPORT_BOX_H
+
+#include "Box.h"
+
+#include <nanobind/ndarray.h>
+namespace nb = nanobind;
```
joaander: Usually we avoid aliasing namespaces in headers as files that `#include` the header are not expecting it.

This isn't critical here as this header has only one consumer: `module-box.cc`. That begs the question: Why not combine this code into `module-box.cc` and eliminate the need for the header?

cpp/box/export-box.h
```diff
@@ -0,0 +1,113 @@
...
+
+namespace freud { namespace box {
+
+template<typename T, typename shape = nb::shape<-1, 3>>
+using nb_array = nb::ndarray<T, shape, nb::device::cpu, nb::c_contig>;
+
+void makeAbsolutePython(Box box, nb_array<float, nb::shape<-1, 3>> vecs,
```
joaander: Avoid making extra copies of the box.
```suggestion
void makeAbsolutePython(const Box& box, nb_array<float, nb::shape<-1, 3>> vecs,
```
Apply this similarly to all methods.

Is the intention to pass `Box` around by copies instead of reference counting? This is OK as `Box` is a lightweight object. Larger classes should definitely be wrapped in `std::shared_ptr` to share reference counting with `nanobind`.

cpp/box/module-box.cc
```diff
@@ -0,0 +1,50 @@
...
+        .def("getPeriodicZ", &Box::getPeriodicZ)
+        .def("setPeriodic", &Box::setPeriodic)
+        .def("setPeriodicX", &Box::setPeriodicX)
+        .def("setPeriodicY", &Box::setPeriodicY)
+        .def("setPeriodicZ", &Box::setPeriodicZ)
+        .def("is2D", &Box::is2D)
+        .def("set2D", &Box::set2D)
+        .def("getVolume", &Box::getVolume)
+        .def("center", &centerPython)
+        .def("centerOfMass", &centerOfMassPython)
```
joaander:
```suggestion
        .def("centerOfMass", &python::centerOfMass)
```
It's a small thing, but I think the names might be cleaner if we used a namespace instead of a suffix. Could call it `&wrap::<methodName>` instead for something shorter than `python`.

joaander (commented):
Thanks @tommy-waltmann, this looks great and is exactly what I have been suggesting! This pattern will make a really clean separation between the C++ library and the Python bindings. If we applied this pattern in HOOMD, for example, we would no longer need all those `PYBIND11_EXPORT` macros.

cpp/box/export-box.h
tommy-waltmann: This code is intended to set the pattern for the rest of freud. Other modules will be significantly bigger and I thought it might be cumbersome to have so much wrapper code in the same file. Additionally, we may want to separate this translation unit into an `export-box.h` and an `export-box.cc` (or multiple translation units) to improve compile times with larger modules.

cpp/box/export-box.h
tommy-waltmann: I didn't think of this originally, you are right that we shouldn't make unnecessary copies of the box class. After reading the [documentation](https://nanobind.readthedocs.io/en/latest/porting.html#shared-pointers-and-holders), I think the `std::shared_ptr` approach will work best.

joaander (approved):
Thanks! This establishes a good pattern. We are ready to move on to the later modules.

PR #1260, opened 2024-06-24
Fix broken CI jobs on pybind11 branch.

joaander:
## Description

To the extent possible, fix the currently broken CI builds.

## Motivation and Context

It would be nice to fix failing CI checks one at a time instead of all at once.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -59,7 +59,10 @@ jobs:
 
       - name: Run Tests
         run: |
-          pytest tests/ -v
+          # Enable tests only for nanobind-converted modules
+          pytest tests/test_box_Box.py -v
+          pytest tests/test_parallel.py -v
+          # pytest tests/ -v
 
   tests_complete:
     name: All tests
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -26,6 +26,7 @@ jobs:
     name: Unit test [py${{ matrix.python }} ${{ matrix.os }}]
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest]
         python: ['39', '310', '311', '312']
--- a/.github/workflows/style_check.yml
+++ b/.github/workflows/style_check.yml
@@ -38,8 +38,7 @@ jobs:
           # uv pip install --no-build-isolation --no-deps -ve .
         env:
           CMAKE_ARGS: "-DCMAKE_EXPORT_COMPILE_COMMANDS=True"
-      - run: ls -lR
-      - name: Execute clang-tidy and cppcheck
-        run: |
-          pre-commit run cppcheck --all-files --hook-stage=manual -v
-          pre-commit run clang-tidy --all-files --hook-stage=manual -v
+      - name: cppcheck
+        run: pre-commit run cppcheck --all-files --hook-stage=manual -v
+      - name: clang-tidy
+        run: pre-commit run clang-tidy --all-files --hook-stage=manual -v
--- a/.github/workflows/style_check.yml
+++ b/.github/workflows/style_check.yml
@@ -38,5 +38,5 @@ jobs:
           CMAKE_ARGS: "-DCMAKE_EXPORT_COMPILE_COMMANDS=True"
       - name: Execute clang-tidy and cppcheck
         run: |
-          pre-commit run cppcheck --all-files --hook-stage=manual
-          pre-commit run clang-tidy --all-files --hook-stage=manual
+          pre-commit run cppcheck --all-files --hook-stage=manual -v
+          pre-commit run clang-tidy --all-files --hook-stage=manual -v
--- a/.github/workflows/style_check.yml
+++ b/.github/workflows/style_check.yml
@@ -36,6 +36,7 @@ jobs:
           uv pip install --no-build-isolation --no-deps -ve .
         env:
           CMAKE_ARGS: "-DCMAKE_EXPORT_COMPILE_COMMANDS=True"
+      - run: ls -lR
       - name: Execute clang-tidy and cppcheck
         run: |
           pre-commit run cppcheck --all-files --hook-stage=manual -v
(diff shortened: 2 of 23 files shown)


.github/workflows/run_tests.yml
```diff
@@ -26,6 +26,7 @@ jobs:
     name: Unit test [py${{ matrix.python }} ${{ matrix.os }}]
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
tommy-waltmann, line 29: Now that the PR is in working order, is this still needed?

.github/workflows/style_check.yml
```diff
@@ -5,36 +5,40 @@ concurrency:
...
 jobs:
   clang-tidy:
     name: Check C++ Code Style
-    runs-on: ubuntu-latest
-    container:
-      image: glotzerlab/ci:2021.11-clang10_py38
+    runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v4.1.6
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
```
tommy-waltmann, line 25: Can we just use the version here instead of the commit hash?

.github/workflows/style_check.yml
```diff
@@ -5,36 +5,40 @@ concurrency:
...
         with:
           submodules: true
-      - name: Install pre-commit
-        run: python3 -m pip install pre-commit
+      - name: Create Python Environment
+        uses: mamba-org/setup-micromamba@f8b8a1e23a26f60a44c853292711bacfd3eac822 # v1.9.0
```
tommy-waltmann, line 29: Same as above.

cpp/box/export_Box.h
```diff
@@ -13,32 +13,32 @@ namespace freud { namespace box { namespace wrap {
 template<typename T, typename shape = nanobind::shape<-1, 3>>
 using nb_array = nanobind::ndarray<T, shape, nanobind::device::cpu, nanobind::c_contig>;
 
-void makeAbsolute(std::shared_ptr<Box> box, nb_array<float, nanobind::shape<-1, 3>> vecs,
+void makeAbsolute(const std::shared_ptr<Box>& box, nb_array<float, nanobind::shape<-1, 3>> vecs,
```
tommy-waltmann, line 16: Passing shared pointers by const reference? That doesn't seem like the standard convention.

cpp/order/SolidLiquid.h
```diff
@@ -8,7 +8,7 @@
 #include <iterator>
 #include <vector>
 
-#include "Cluster.h"
+#include "../cluster/Cluster.h"
```
tommy-waltmann, line 11: I think this is unnecessary.

cpp/box/Box.h
```diff
@@ -367,7 +367,7 @@ class Box
      *  \param masses Optional array of masses, of length Nvecs
      *  \return Center of mass as a vec3<float>
      */
-    vec3<float> centerOfMass(vec3<float>* vecs, size_t Nvecs, const float* masses) const
+    vec3<float> centerOfMass(vec3<float>* vecs, size_t Nvecs, const float* masses = nullptr) const
```
tommy-waltmann, line 370: The masses passed in from python are guaranteed to be non-null.

tommy-waltmann (changes requested):
Thanks for this, it will be nice to have unit tests running again with only the converted modules.

cpp/box/Box.h
joaander: The masses coming from `ClusterProperties::compute` are not.
https://github.com/glotzerlab/freud/blob/ba5998eb4f6a51ba6930e200066abb09390ab059/cpp/cluster/ClusterProperties.cc#L60

This causes `clang-tidy` to fail:
https://github.com/glotzerlab/freud/actions/runs/9649688151/job/26613682932#step:6:57

cpp/order/SolidLiquid.h
joaander: `freud/cpp/cluster` is not in the include search path. `clang-tidy` cannot find the file without this change:
https://github.com/glotzerlab/freud/actions/runs/9649688151/job/26613682932#step:6:67

cpp/box/export_Box.h
joaander: Yes, it seems weird - but `clang-tidy` insists: https://github.com/glotzerlab/freud/actions/runs/9648239399/job/26608946198#step:6:151

It makes sense when you think about it. This function doesn't store the `shared_ptr` anywhere to hold on to the reference, so why not borrow the reference from the caller?
https://stackoverflow.com/questions/74329949/clang-tidy-and-boostintrusive-ptr

.github/workflows/style_check.yml
joaander: We *can*, but we **should not**: 
* https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash
* https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

.github/workflows/run_tests.yml
joaander: I think it is useful to keep. CI resources are plentiful. With the default `fail-fast=true`, a developer can solve only 1 error per commit-run iteration. With `fail-fast=false`, you can see all the different errors that occur in different matrix elements and make faster progress on solving them.

.github/workflows/style_check.yml
tommy-waltmann: Okay, but I have a couple of questions. Should we be doing this for every github action? How will we know to update the hash when a new version of the action is released?

cpp/order/SolidLiquid.h
tommy-waltmann: Okay, I suppose we can keep it for now and revert this change in a future PR converting the order module.

.github/workflows/style_check.yml
joaander: Yes, we should ideally use this for every GitHub action to ensure consistent behavior.
Dependabot will update both the hash and comment when new versions are released:
https://github.com/dependabot/dependabot-core/pull/5951

tommy-waltmann (approved):

PR #1263, opened 2024-06-25
Convert Locality Module to Use Nanobind

tommy-waltmann:
## Description

This PR converts the locality module to use nanobind. It also enables the data module because it's pure python, but uses `PeriodicBuffer` internally.

## Motivation and Context

We are removing Cython from freud, this is one of a series of PRs that accomplishes that goal.

## How Has This Been Tested?

Testing has been added in the CI for newly enabled modules.

## Types of changes

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/glotzerlab/freud/blob/main/CONTRIBUTING.md) document.


CMake/freud-macros.cmake
```diff
@@ -33,3 +33,17 @@ function(make_python_extension_module _target)
...
+
+# set the rpath for installing python extension modules
+function(target_set_install_rpath _target)
+  if(APPLE)
+    set_target_properties(${_target} PROPERTIES INSTALL_RPATH "@loader_path")
+  else()
+    set_target_properties(${_target} PROPERTIES INSTALL_RPATH "\$ORIGIN")
+  endif()
+
+  if(_using_conda OR DEFINED ENV{CIBUILDWHEEL})
```
joaander:
```suggestion
  if(_using_conda)
```
I modified the cibuildwheel action to set `INSTALL_RPATH_USE_LINK_PATH`.

cpp/locality/locality-files.cmake
```diff
@@ -0,0 +1,39 @@
...
+set(locality_sources
+    locality/AABBQuery.cc
+    locality/FilterRAD.cc
+    locality/FilterSANN.cc
+    locality/LinkCell.cc
+    locality/NeighborComputeFunctional.cc
+    locality/NeighborList.cc
+    locality/PeriodicBuffer.cc
+    locality/Voronoi.cc
+    # For now, compile voro++ object in directly.
```
joaander: Is there a problem with making this a permanent solution?
```suggestion
    # Compile voro++ object in directly.
```

cpp/util/ManagedArray.h
```diff
@@ -107,8 +94,8 @@ template<typename T> class ManagedArray
             // with a different data structure like std::vector, but it would
             // require writing additional gymnastics to ensure proper reference
             // management and should be carefully considered before any rewrite.
-            m_data = std::shared_ptr<std::shared_ptr<T>>(new std::shared_ptr<T>(
-                new T[size()], std::default_delete<T[]>())); // NOLINT(modernize-avoid-c-arrays)
+            m_data = std::shared_ptr<T>(new T[size()],
+                                        std::default_delete<T[]>()); // NOLINT(modernize-avoid-c-arrays)
```
joaander: In this rewrite, it is possible (and would make the code more clear) to use a std::vector to contain the data. We should remove this comment, replace `m_data` with a `std::vector<T>`. We will also need to replace `m_data.get()` with `m_data.data()`

With the change in data model, `prepare` no longer has the same semantics. We should remove it to ensure that all calls to it are changed. The contents of `prepare` should be moved to the constructor and all current uses of `prepare` should be replaced with `m_array = std::make_shared<MangedArray<T>>(shape)`.

cpp/util/ManagedArray.h
```diff
@@ -172,13 +157,13 @@ template<typename T> class ManagedArray
     //! Get the size of the current array.
     size_t size() const
     {
-        return *m_size;
+        return m_size;
     }
 
     //! Get the shape of the current array.
     std::vector<size_t> shape() const
```
joaander:
```suggestion
    const std::vector<size_t>& shape() const
```

cpp/util/ManagedArray.h
```diff
@@ -311,23 +296,23 @@ template<typename T> class ManagedArray
      */
     inline size_t getIndex(const std::vector<size_t>& indices) const
```
joaander, line 277: Forcing the caller to form a `std::vector` for every array element access is extreme. The amount of work to fix this may be for a future PR, but we should consider templating `ManagedArray` on the number of array dimensions and switching to `std::array` for `indices`. We could also do the same for `shape`.

cpp/util/export_ManagedArray.h
```diff
@@ -0,0 +1,63 @@
...
+{
+    ManagedArray<T>* self_cpp = nanobind::cast<ManagedArray<T>*>(self);
+    auto dims = self_cpp->shape();
+    auto ndim = dims.size();
+    auto data_ptr = self_cpp->get();
+    return nanobind::ndarray<nanobind::numpy, const T>((void*) data_ptr, ndim, &dims[0], self);
+}
+
+/* Need to alter array dimensions when returning an array of vec3*/
+template<typename T> nanobind::ndarray<nanobind::numpy, const T> toNumpyArrayVec3(nanobind::object self)
```
joaander: Could we use template specialization here to make `toNumpyArray` automatically do the right thing when given `vec3<T>`?

cpp/util/export_ManagedArray.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace wrap {
+
+template<typename T> nanobind::ndarray<nanobind::numpy, const T> toNumpyArray(nanobind::object self)
+{
+    ManagedArray<T>* self_cpp = nanobind::cast<ManagedArray<T>*>(self);
+    auto dims = self_cpp->shape();
+    auto ndim = dims.size();
+    auto data_ptr = self_cpp->get();
+    return nanobind::ndarray<nanobind::numpy, const T>((void*) data_ptr, ndim, &dims[0], self);
```
joaander: Nice! Glad to see this plan works with nanobind.

cpp/util/export_Vector.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace wrap {
+
+template<typename T> nanobind::ndarray<nanobind::numpy, const T> vectorToNumpyArray(nanobind::object self)
+{
+    std::vector<T>* self_cpp = nanobind::cast<std::vector<T>*>(self);
+    auto dims = {self_cpp->size()};
+    auto ndim = 1;
+    auto data_ptr = self_cpp->data();
+    return nanobind::ndarray<nanobind::numpy, const T>((void*) data_ptr, dims, self);
```
joaander: This does not work. `self` here is the `std::vector` itself. The vector is actually held by a `PeriodicBuffer` (or another class instance). We can either
1) Make a copy of the data here (removes zero-copy performance advantage).
or 
2) make this a template method that `PeriodicBuffer` (and other classes) can use so that the `self` object is an instance of the holder.

The problem with option 2 is that the user can now invalidate arrays:
```
obj.compute()
array = obj.result
array.sum() # valid
obj.compute() # resizes internal array
array.sum() # seg fault
```

If we need zero-copy performance we need some kind of holder to keep these copies alive. That could be `ManagedArray`, but it doesn't have to be. We should test if it works, but I think holding `std::vector` in a `std::shared_ptr` would fix the seg fault. This would require minimal modification to the classes that use `std::vector`.

joaander (commented):
Thanks! This looks like a good start. This first review covers only the data management side. I'll look at the locality classes in a future review.

cpp/locality/locality-files.cmake
tommy-waltmann: We can do it permanently. I think this comment was added a long time ago.

cpp/util/export_ManagedArray.h
tommy-waltmann: I'm not sure if that's doable, you may know more about how to do this than I do. We may have to extend this to handle cases like `ManagedArray<quat<float>>` later, so it's worth thinking about how to do it now.

cpp/util/export_Vector.h
tommy-waltmann: Yes, it seems I overlooked this detail. Ideally, the C++ classes could use a `shared_ptr` to hold the vector, but I'm not sure how we would make that work given that the current pattern is to export the holder class to python and give it a `toNumpyArray` method.

Another option is to force usage of `ManagedArray` as the only data container we allow for array data on the C++ side. This is a bit restrictive, ideally there would be a more flexible option.

cpp/util/ManagedArray.h
tommy-waltmann: The vector passed into this (and similar) methods of `ManagedArray` are typically throwaway objects (I don't know the technical term). Calling code usually looks like `ManagedArray({5, 4, 3})`, for example, if a `ManagedArray` is being constructed.

I would imagine that in the typical use case, the compiler will know how to optimize the extra operations away.

joaander: Yes, this is exactly the type of use-case `std::array` is designed for. We should use it.

I'm not fully up to date on the most modern C++ standards. From a quick search, it seems that what you describe *might* be possible in C++20 or 26
https://andreasfertig.blog/2021/08/cpp20-dynamic-allocations-at-compile-time/
https://stackoverflow.com/questions/77689295/can-you-convert-stdvector-to-stdarray-at-compile-time-without-making-the-vec

However, we target C++17 for conda-forge support and in any case would need to add the requisite `constexpr` and other keywords to make it possible for the compiler to optimize out the `std::vector`. 

There are other benefits to templating on the dimension as well - the compiler can further optimize (by unrolling) the index computations in `ManagedArray`.

cpp/util/export_ManagedArray.h
joaander: OK, I will push a commit that does this for Managed Array. Partial template specialization for functions is not allowed, so these methods will be moved to static methods in dummy structs.
https://stackoverflow.com/questions/5101516/why-can-function-templates-not-be-partially-specialized

joaander: See 161eb3c

tommy-waltmann:
There are some classes for histogramming in locality.py (`_SpatialHistogram` and `_SpatialHistogram1D`) that are unused by any compute in the locality module. They exist solely for the purpose of exposing common properties of classes that use binning to do their calculation. I choose to leave the conversion of those classes for a future PR because we have no way of testing them right now.

tommy-waltmann:
Please ignore the clang-tidy errors on this PR, they are related to uncoverted modules.

joaander (approved):
This looks great! Thanks! I like how the `locality.pyx` file chages considered a "rename" by git to `locality.py` with relatively minor changes - and no changes to the rest of the Python code.

PR #1267, opened 2024-07-02
Expose neighbors computed using local density via a neighbor list

DomFijan:
## Description
This pull request exposes neighbors computed using local density via a neighbor list. The local density computation currently averages the number of neighbors using the approach described in the documentation but does not expose the associated neighbor list. This enhancement computes and exposes the full weighted neighbor list based on the approach described in the local density documentation. The weighted neighbor list applies the same rules used to compute neighbors of particles, considering contributions based on distances as neighbor list weights.

## Motivation and Context
@rayasare requested this be exposed.

## How Has This Been Tested?
Tests and docs are TODO

## Types of changes

## Checklist:

### DIFF
--- a/cpp/density/LocalDensity.cc
+++ b/cpp/density/LocalDensity.cc
@@ -2,8 +2,8 @@
 // This file is from the freud project, released under the BSD 3-Clause License.
 
 #include "LocalDensity.h"
-#include "NeighborComputeFunctional.h"
 #include "NeighborBond.h"
+#include "NeighborComputeFunctional.h"
 #include <tbb/enumerable_thread_specific.h>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


joaander (commented):
Have you checked the memory utilization and performance of this code for typical use-cases of LocalDensity? (e.g. 1 million particle 2D systems with r_cut = ~50-100). I would think that the performance impact would be very noticeable and the memory usage could be massive.

Why can't the user interested in this feature compute a neighbor list via the normal means and add the weights computed by numpy to that?

DomFijan:
I haven't checked how this impacts memory/performance. If I added a flag that skips computation of the neighbor list, would this be an ok fix, or do you just think we should just drop this?

DomFijan:
In an offline discussion we have concluded that the same behavior can easily be achieved with current neighbor list infrastructure and that this code is not needed. Another issue was the fact that changes proposed impact the performance and memory for users using local density in its current form.

PR #1312, opened 2025-02-22
Add CIF file reader based on Parsnip

janbridley:
## Description

Many users wish to initialize crystals from CIF files, but encounter convention mismatches when converting between ASE and HOOMD scripts. Parsnip provides a lightweight alternative that has been tested against the Freud convention for accuracy.

~NOTE: parsnip is not yet on conda-forge, meaning the current environment file will not work. I can install with pip, but we could also just wait a few days for the conga-forge feedstock to be merged.~ 

## How Has This Been Tested?
New tests have been added in the style of other `freud/data.py` tests. There is additional testing on the Parsnip side, which I would be happy to expand on if we feel it necessary - I test the creation and initialization of data from a variety of CIF files, but I do not have tests for parsnip->HOOMD directly.

## Checklist:

### DIFF
--- a/.github/workflows/environments/environment.yaml
+++ b/.github/workflows/environments/environment.yaml
@@ -14,6 +14,7 @@ dependencies:
   - matplotlib
   - nanobind>=2.0
   - numpy
+  - parsnip-cif
   - pillow
   - pre-commit
   - pytest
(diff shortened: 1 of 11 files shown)


janbridley:
It looks like the build-wheels step fails because it does not copy my example file into the distribution - this is probably desired, as there is no reason to include that file otherwise. @joaander do you have a recommendation on the best path forward?

joaander (approved):
Thanks!

We will need to add `parsnip-cif` to the dependencies when the next release is built on conda-forge.

janbridley:
Noted - thanks!

PR #1318, opened 2025-03-11
Add `to_box_params` method for exporting boxes to GSD

janbridley:
## Description

Freud boxes are typically initialized from `Lx, Ly, Lz, xy, xz, yz` data, but there is no neat way to recover that data from the input. This method (which ~could~ should? be renamed for clarity) provides that functionality.

## Motivation and Context

Initializing GSD boxes can now be achieved in one function call:
```diff
frame = gsd.hoomd.Frame()
- frame.configuration.box = [freud_box.Lx, freud_box.Ly, ..., freud_box.yz]
+ frame.configuration.box = freud_box.to_box_params()
```

## How Has This Been Tested?
New tests have been added

## Checklist:

### DIFF
--- a/freud/box.py
+++ b/freud/box.py
@@ -648,7 +648,7 @@ class Box:
 
             >>> box = freud.box.Box.cube(L=10)
             >>> box.to_array()
-            array([10., 10., 10., 0., 0., 0.])
+            (10.0, 10.0, 10.0, 0.0, 0.0, 0.0)
 
         Returns:
             tuple:
--- a/freud/box.py
+++ b/freud/box.py
@@ -639,15 +639,15 @@ class Box:
             ]
         )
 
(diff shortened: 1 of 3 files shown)


freud/box.py
```diff
@@ -639,6 +639,24 @@ def to_matrix(self):
             ]
         )
 
+    def to_array(self):
```
DomFijan: The fact that the function is called "to_array" but returns a tuple is somewhat weird. Could we come up with a better name for the function? How about get_parameters or get_box_params or just box_parameters?

janbridley: I renamed to to_box_params, which mirrors the naming of to_dict and to_matrix but is more accurate for the new method

DomFijan (approved):

PR #1321, opened 2025-04-01
Fix PMFT plot and add cmap option

janbridley:
## Description, Motivation, and Context

PMFTXY.plot() called for the `self.PMFT`, `self.X`, and `self.Y` properties which were renamed to `self.pmft` and `self.bounds` in a recent PR. This changes fixes these calls, and adds an optional "cmap" parameter that passes through into the internal call to `matplotlib.imshow`.

## How Has This Been Tested?

Used with cmap="afmhot".

image

## Checklist:

### DIFF
--- a/freud/plot.py
+++ b/freud/plot.py
@@ -317,7 +317,7 @@ def histogram_plot(
     return ax
 
 
-def pmft_plot(pmft, ax=None):
+def pmft_plot(pmft, ax=None, cmap="viridis"):
     """Helper function to draw 2D PMFT diagram.
 
     Args:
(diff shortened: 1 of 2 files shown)


janbridley:
@joaander As mentioned in #1322, I pinned cmake<4.0.0 to get the builds working. However, we should probably address the underlying cause before merging this PR.

joaander:
You can revert the changes you made to workaround the failure to build tbb. #1320 (now merged) fixes the issue.

DomFijan:
So these issues didn't get raised because we don't have tests for plotting functions. Should we write some very simple test just to see if the code can be run?

There is a whole bigger issue re freud's plotting library which is hidden from the user, but most of it can be accessed through class_object.plot(). We might also consider, at some point, opening the plot submodule to the user but it might require us to consolidate the functions so that they use roughly the same style and have similar function variables and most importantly tests, but this is probably a problem for another day.

DomFijan (approved):

PR #1389, opened 2026-01-22
Add an optimized CellList implementation based on a linear contiguous buffer

janbridley:
## Description

Add a new cell list nearest neighbor search, which is significantly faster than the previous AABBQuery. Note that this is a fundamentally different architecture from LinkCell, which is extremely slow compared to both alternatives.

#### TODOs [DONE]

- [x] Fix on windows
- [x] Clean up and lint
- [x] Request review
- [x] Neighbor bonds are inverted occasionally?
- [x] Remove knn

### Architecture

This neighbor list is based on a spatially sorted linear memory region, with cells adjacent in the X direction contiguous in memory. We defer construction until the user attempts a query, allowing us to choose the optimal cell width for a given lookup. For `num_nearest` lookups, we estimate the cell width based on the density of the system, with an empirically determined scale factor for performance. Our choice of construction guarantees we generate every necessary ghost particle in a single layer of ghost cells, which is optimal for performance in `r_max` queries. For `num_nearest` queries, no such guarantee is possible so we do not support that mode. Correctly performing a knn query using this code could otherwise degenerate to all_pairs.

### Performance

For performance, I have two benchmarks -- one based on constructing a full neighbor list in python (`cq.toNeighborList()`) and a more representative test based on computing the RDF of a system with a single bin. This latter benchmark aims to test freud's internal use of lists in `NeighborComputeFunctional`, and is what the text outputs of the code measure performance for. Note that generating random systems and computing the RDF itself takes ~25-30% of the runtime of this benchmark, so the % change performance numbers are underestimates. 

Note in the benchmarks below, I test against [vesin](https://github.com/Luthaf/vesin), which is the fastest nearest neighbor library I've found. Note the results are not fully comparable, however, as they do not use the freud `toNeighborList` function that dominates the runtime.

#### OSX M1 Pro (Python 3.13)

Benchmarks for uniform random systems in random, lightly sheared boxes. r_cut=1.5 and rho=0.5.

```
============================================================
PERCENTAGE IMPROVEMENT: RDF (CellQuery vs AABBQuery)
============================================================

N = 1,000 particles:
  Serial:   +44.0% (AABB: 1.122ms -> Cell: 0.628ms)
  Parallel: +42.3% (AABB: 0.338ms -> Cell: 0.195ms)

N = 2,000 particles:
  Serial:   +49.1% (AABB: 2.305ms -> Cell: 1.174ms)
  Parallel: +45.0% (AABB: 0.685ms -> Cell: 0.377ms)

N = 4,000 particles:
  Serial:   +53.1% (AABB: 5.007ms -> Cell: 2.348ms)
  Parallel: +55.1% (AABB: 1.305ms -> Cell: 0.586ms)

N = 8,000 particles:
  Serial:   +53.7% (AABB: 11.093ms -> Cell: 5.133ms)
  Parallel: +48.7% (AABB: 2.672ms -> Cell: 1.371ms)

N = 16,000 particles:
  Serial:   +53.3% (AABB: 23.478ms -> Cell: 10.958ms)
  Parallel: +59.0% (AABB: 5.807ms -> Cell: 2.381ms)

N = 32,000 particles:
  Serial:   +56.5% (AABB: 43.443ms -> Cell: 18.909ms)
  Parallel: +64.2% (AABB: 11.455ms -> Cell: 4.097ms)

Average improvement across all particle counts:
  Serial:   +51.6%
  Parallel: +52.4%
============================================================
```
out

#### Purdue Anvil, `-n 8`

```
============================================================
PERCENTAGE IMPROVEMENT: RDF (CellQuery vs AABBQuery)
============================================================

N = 1,000 particles:
  Serial:   +52.6% (AABB: 1.936ms -> Cell: 0.917ms)
  Parallel: +41.2% (AABB: 0.418ms -> Cell: 0.246ms)

N = 2,000 particles:
  Serial:   +53.8% (AABB: 4.228ms -> Cell: 1.953ms)
  Parallel: +45.9% (AABB: 0.891ms -> Cell: 0.481ms)

N = 4,000 particles:
  Serial:   +55.6% (AABB: 8.794ms -> Cell: 3.903ms)
  Parallel: +48.5% (AABB: 1.751ms -> Cell: 0.901ms)

N = 8,000 particles:
  Serial:   +59.7% (AABB: 17.038ms -> Cell: 6.873ms)
  Parallel: +50.6% (AABB: 3.196ms -> Cell: 1.580ms)

N = 16,000 particles:
  Serial:   +61.1% (AABB: 36.066ms -> Cell: 14.044ms)
  Parallel: +60.6% (AABB: 8.019ms -> Cell: 3.158ms)

N = 32,000 particles:
  Serial:   +60.5% (AABB: 78.146ms -> Cell: 30.855ms)
  Parallel: +63.3% (AABB: 18.246ms -> Cell: 6.698ms)

Average improvement across all particle counts:
  Serial:   +57.2%
  Parallel: +51.7%
============================================================
```

### Comments

freud's `toNeighborList` is extremely slow for systems of reasonable size (<100k particles), mainly due to overhead in the tbb parallel loop. This is true more generally, with many parallel loops in freud incurring performance costs for small-ish systems. This is not a surprise, but does indicate an opportunity for more performance in the future -- although  I don't recall where I saw these figures, I have seen commentary that bs_thread_pool (which we use in SPATULA) has much lower overhead for a similar work-stealing paradigm. We do use more of TBB's machinery throughout freud, but it's worth considering.

Secondly, freud's lazy evaluation of neighbors makes evaluation of certain order parameters relatively inefficient. The issue is that we interleave the pair bond calculations (which have a fair amount of branching and indirection) with what are otherwise fairly dense calculations. This is most notable in fast order parameters like nematic and BOOD, but is true to a lesser extent for `environment` and `density` OPs as well.

Note that this cell list is optimized for uniform, dense systems which is a common pattern within the glotzerlab but perhaps not more generally. AABBQuery will be faster for spatially inhomogeneous data, although we avoid common problems with low-density simulations due to the linear layout of our memory. Because we never rebuild neighbor lists in freud, low-occupancy bins can be stored as efficiently as larger ones, and empty bins can be skipped in the spatial sort entirely. 

There is a wide variety of (reasonably) modern literature on neighbor list calculation. GROMACS advocates for a [blocked tree-based neighbor list](https://arxiv.org/pdf/1306.1737v1) similar to the current AABBQuery, with a few extra tweaks for SIMD between the particles themselves. I tested this as well, but the pattern does not fit freud's lazy evaluation well and the performance was not competitive for reasonable particle counts. There is also research on novel neighbor finding methods for (1) spatially inhomogeneous data ([SNN](https://github.com/nla-group/snn/tree/master), an approach I really like, but it degenerates to all pairs in the uniform case so not useful for crystals) and (2) kNN queries, which do not translate as efficiently to ball queries as the current code in the other direction.

## How Has This Been Tested?

Tests extending the existing `NeighborQueryTest` class have been implemented, with a variety of random systems also tested offline. 

## Checklist:


janbridley:
@joaander This PR is nearly ready for review, with Dom and I finishing up the final approach. Would you have some time to review this? You're quite familiar with spatial data code, so hopefully you can provide some insight (1) into how I've structured this, and (2) if there are any obvious performance degradations

freud/locality/CellIterator.h
```diff
@@ -0,0 +1,399 @@
...
+        {
+            throw std::runtime_error("Cell data is uninitialized.");
+        }
+
+        m_r_max_sq = r_max * r_max;
+        m_r_min_sq = r_min * r_min;
+
+        // Check if the query point is within the grid bounds. TODO: is this allowed?
+        unsigned int cell_idx_u;
+        if (!m_cell_query->getCellIdxSafe(m_query_point, cell_idx_u))
```
joaander, line 54: This will miss neighbors when the query point is just outside the cell bounds and is still within `r_cut` of points inside. I needed to handle that case in hoomd-rs. Can it occur in freud?

joaander (commented):
Thanks! It is great that you can improve performance so much with a rewrite of the cell list code.

freud/locality/CellIterator.h
janbridley: Can confirm this behavior -- good catch! I'm fixing that and will re-request review when it's sorted

janbridley: It looks like clamping the cell index is a ~simple~ actually rather involved fix, as it lets us use the existing iterator machinery to handle particles outside the cell. Performance will be worse for query points outside the grid, but I hope that is an uncommon case. I've documented that cell queries should not be used for kNN

joaander: Yes, I knew it would be an involved fix. I was hoping the answer was that freud didn't allow queries from outside the boundary.

janbridley: I think there is another bug as well, I’m testing random systems and in a few cases I miss neighbors that AABBQuery finds - I’ll let you know when it’s ready for re-review

janbridley:
Finalized plots that compare to other packages in the space, similar to the [Benchmarking NeighborQuery](https://freud.readthedocs.io/en/v2.13.2/gettingstarted/examples/examples/Benchmarking%20Neighbor%20Finding%20against%20scipy.html) example from version 2.0. As we'd hope, we outperform all other libraries I could find even with the fairly inefficient toNeighborList call. This isn't quite fair, as no other libraries provide a parallel implementation, but for end users this performance is what they'll get out of the box.

benchmark_locality_scipy

joaander:
> This isn't quite fair, as no other libraries provide a parallel implementation, but for end users this performance is what they'll get out of the box.

How much does the time change when you limit `freud` to 1 thread?

janbridley:
> > This isn't quite fair, as no other libraries provide a parallel implementation, but for end users this performance is what they'll get out of the box.
> 
> How much does the time change when you limit `freud` to 1 thread?

Slower than MDAnalysis and vesin in all cases and slower than scipy for some random systems. I will note, though, Freud's nlist is optimized for a lazy iterator pattern rather than just building an entire neighbor list at once. This significantly deteriorates performance when calling `toNeighborList` -- if we always built the entire list at the start, we could inline and optimize more aggressively. On the upside, though, it makes fast order parameters (q6, rdf, etc.) more efficient. I can't reproduce this at this point, but I had an early pure C version of this neighbor list that was about 25% faster than vesin (before I implemented the query iterator infrastructure). 
benchmark_locality_scipy

joaander (approved):
Thanks! This looks good.

Do you plan to address the remaining TODO comments before merging?

janbridley:
> Thanks! This looks good.
> 
> Do you plan to address the remaining TODO comments before merging?

I removed those TODOs -- the prefix scan is implemented in util for convenience, and the SoA ghost packet code is unlikely to yield a significant benefit, as most of the runtime is in the query

joaander:
The changes made by this PR are not documented in the the change log. Will need to correct that in a new PR.

PR #1399, opened 2026-03-05
Add type hints in Python

### LINKED ISSUE #1311: Add type hinting in python
### Description

Many freud classes and functions don't have type hints. Implementing them would help with using IDEs so that users don't have to reference docs all the time.

### Proposed Solution

Since we have recently made the port to nanobind, all python files are pure python files so using the full python feature set is possible and IDEs and language servers particularly benefit from type hits for code completion or easy docstring reading.
Type hints should be added such that all function returns and function variables are added according to the latest python standard.

### Additional Context

_No response_

DomFijan:
This PR adds missing type hinting for Python function returns.

## Description
Many Python functions have missing type hints.

## Motivation and Context
Resolves: #1311 

## How Has This Been Tested?
Locally run mypy succesfully.

## Checklist:


freud/box.py
```diff
@@ -75,15 +106,15 @@ def __init__(self, Lx, Ly, Lz=0, xy=0, xz=0, yz=0, is2D=None):
...
+    def L(self) -> npt.NDArray[np.floating]:
         r""":math:`\left(3, \right)` :class:`numpy.ndarray`: Get or set the
         box lengths along x, y, and z."""
         return np.asarray(
             [self._cpp_obj.getLx(), self._cpp_obj.getLy(), self._cpp_obj.getLz()]
         )
 
     @L.setter
-    def L(self, value):
+    def L(self, value: ScalarLike | Sequence[ScalarLike]) -> None:
```
janbridley, line 143: In cases like this where the valid sequences have a fixed length, I think it might be better to use `tuple[ScalarLike, ScalarLike, Scalarlike]`. A little more verbose, but gives you meaningful feedback if you pass an incorrectly typed input

freud/box.py
```diff
@@ -170,33 +201,35 @@ def __eq__(self, other):
...
     @property
-    def volume(self):
+    def volume(self) -> float:
         """float: The box volume (area in 2D)."""
         return self._cpp_obj.getVolume()
 
-    def make_absolute(self, fractional_coordinates, out=None):
+    def make_absolute(
+        self, fractional_coordinates: ArrayLike, out: ArrayLike | None = None
+    ) -> npt.NDArray[np.floating]:
```
janbridley, line 259: We defined FloatArray, could we use that as the return?

freud/box.py
```diff
@@ -249,7 +284,7 @@ def make_fractional(self, absolute_coordinates, out=None):
 
         return np.squeeze(out) if flatten else out
 
-    def get_images(self, vecs):
+    def get_images(self, vecs: ArrayLike) -> npt.NDArray[np.integer]:
```
janbridley, line 314: And intarray -- I think we can swap these all out with ruplacer

freud/data.py
```diff
@@ -35,19 +39,21 @@ class UnitCell:
             (Default value = ``[[0, 0, 0]]``).
     """
 
-    def __init__(self, box, basis_positions=None):
+    def __init__(
+        self, box: freud.box.Box | ArrayLike, basis_positions: ArrayLike | None = None
+    ) -> None:
         if basis_positions is None:
             basis_positions = [[0, 0, 0]]
```
janbridley, line 49: Because we automatically convert basis_positions if it is None, could we instead set the default to `basis_positions=((0,0,0),)`?

freud/diffraction.py
```diff
@@ -833,13 +880,15 @@ def k_vectors(self):
...
+    def __repr__(self) -> str:
         return (
             f"freud.diffraction.{type(self).__name__}(grid_size={self.grid_size}, "
             f"output_size={self.output_size})"
         )
 
-    def to_image(self, cmap="afmhot", vmin=None, vmax=None):
+    def to_image(
+        self, cmap: str = "afmhot", vmin: float | None = None, vmax: float | None = None
+    ) -> np.ndarray:
```
janbridley, line 897:
```suggestion
    ) -> npt.NDArray[np.uint8]:
```

Full shape annotation is [not supported](https://github.com/numpy/numpy/issues/16544) but it's nice to mark this as uint8 imo

freud/locality.py
```diff
@@ -27,13 +33,171 @@
...
+
+
+class _OvitoCollection(Protocol):
+    cell: _OvitoCell
+    particles: _OvitoParticles
+
+
+class _SystemWithBoxPoints(Protocol):
+    box: object
+    points: ArrayLike
```
janbridley: We could consider moving this to a freud/locality_typing.py or similar?

freud/locality.py
```diff
@@ -6,14 +6,20 @@
 locate points based on their proximity to other points.
 """
 
+from __future__ import annotations
+
 import inspect
+from collections.abc import Iterator, Mapping, Sequence
 from importlib.util import find_spec
+from typing import TYPE_CHECKING, Literal, Protocol, cast
```
janbridley: Even in Python we end up with reinterpret cast, it truly is inevitable

janbridley:
I'm generally happy with these changes, just a few consistency questions and a nitpick here or there. This will be quite useful to have, it's remarkable to see how far python typing has come.

As a quick heads-up: while the typing is correct at a hight level there are many cases where we don't formally satisfy all the requirements. I don't think this really matters but it does cause a lot of typing errors in LSPs:

image

janbridley:
Looks like we can fix some module-not-found errors (for the freud\._* pattern) using [nanobind stubgen](https://nanobind.readthedocs.io/en/latest/typing.html), if we're ok adding a bunch of .pyi files to the repo

DomFijan:
> Looks like we can fix some module-not-found errors (for the freud._* pattern) using [nanobind stubgen](https://nanobind.readthedocs.io/en/latest/typing.html), if we're ok adding a bunch of .pyi files to the repo

We should consider this for a future PR. Most of user facing functions are wrapped in python code so at first glance, I don't think users would benefit much from them. Did you have some concrete situation in mind where having them would be helpful?

janbridley (approved):
Looks great! This will be very nice to have

freud/data.py
DomFijan: We could but this would technically be an API breaking change.....

janbridley: Looks like (nested) tuples are [ArrayLike](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike), but it’s no problem either way

PR #1447, opened 2026-07-01
Default angular separation

### LINKED ISSUE #1241: Change default behavior of AngularSeparation
### Description

Either the default behavior of equiv_orientations in AngularSeparation must be updated to include the degenerate identity quaternion (-1,0,0,0) or the function behavior should be updated to not force the user to supply the degenerate negative quaternions.

### Motivation and Context

I believe there is some unintentional behavior in freud in AngularSeparation. AngularSeparation takes in a parameter called equiv_orientations, which allows you to give quaternions that map a shape to itself to account for shape symmetry.

However, if you don't give any value, this parameter defaults to (1,0,0,0), According to the documentation, you are meant to include the degenerate negative quaternions (-q for each q). If you explicitly give equiv_orientations = [(1,0,0,0),(-1,0,0,0)], the result is different. 

I'm not marking this as a bug, because the documentation correctly describes the behavior, but the default behavior in not physically relevant.

tfteague:
## Description
Added $(-1, 0, 0, 0)$ as a default equivalent quaternion to `AngularSeparationNeighbor`, `AngularSeparationGlobal`, and `LocalBondProjection`, in response to an open issue.

## Motivation and Context
The quaternions  $(1, 0, 0, 0)$ and $(-1, 0, 0, 0)$ both represent the identity, and so in all physical applications should both be used.

Resolves: #1241

## How Has This Been Tested?
 Added a test, `test_compute_defaults` for `AngularSeparationGlobal` in `test_environment_angular_separation.py` to ensure the default value provides the correct angular displacement.

## Checklist:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog] and this project adheres to
 
 * MSD with `mode=="direct"` is now slightly faster (#1414).
 * Nematic order parameter is now ~40x faster. (#1428)
+* Changed default equivalent orientations to include (-1,0,0,0) for `AngularSeparationNeighbor`, `AngularSeparationGlobal`, and `LocalBondProjection`(#1241)
 
 ## 3.5.0 -- 2025-10-08
 
--- a/doc/source/reference/credits.rst
+++ b/doc/source/reference/credits.rst
@@ -16,6 +16,7 @@ The following people contributed to the development of freud:
 * Bradley Dice, University of Michigan
 * Brandon Butler, University of Michigan
 * Bryan VanSaders, University of Michigan
+* Catherine La Riviere, University of Michigan
 * Carl Simon Adorf, University of Michigan
 * Carolyn Phillips, University of Michigan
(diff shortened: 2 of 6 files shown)


ChangeLog.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog] and this project adheres to
 
 * MSD with `mode=="direct"` is now slightly faster (#1414).
 * Nematic order parameter is now ~40x faster. (#1428)
+* Changed default equivalent orientations to include (-1,0,0,0) for `AngularSeparationNeighbor`, `AngularSeparationGlobal`, and `LocalBondProjection`(#1241)
```
janbridley, line 23:
```suggestion
* Changed default equivalent orientations to include (-1,0,0,0) for `AngularSeparationNeighbor`, `AngularSeparationGlobal`, and `LocalBondProjection`(#1447)
```

Our chanagelogs typically use the PR # , rather than the associated issue!

janbridley (approved):
Thanks!

